PR #14, opened 2018-03-09
Proper registration of a staticly available tracer

felixbarny:


apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
```diff
@@ -1,40 +1,121 @@
...
+     * So the solution is to provide a package-private method which the implementation can call.
+     * What's not so nice about this solution is that the implementation has to have a class in
+     * the same package as the api in order to be able to call this package private method.
+     * But as this is not a public method,
+     * the concrete registration mechanism can always be changed later without breaking compatibility.
+     * </p>
+     *
+     * @param tracer The tracer implementation to register.
+     */
+    static void register(Tracer tracer) {
```
alvarolobato: Why is this method static? You are inside a singleton, it should be safe not to have this static which in return will make this less testable.

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
```diff
@@ -1,40 +1,121 @@
...
+ * <p/>
+ */
 public class ElasticApm implements Tracer {
 
+    /**
+     * The singleton instance of the Tracer.
+     */
     private static final ElasticApm INSTANCE = new ElasticApm();
 
     private static Tracer tracer = NoopTracer.INSTANCE;
```
alvarolobato: I think this variable should not be static, you are inside a singleton. See below.

About thread safety, there are several parts where you access this variable in a non thread safe and some other ie. register/unregister where you have a`synchonized` block, that makes this class not thread safety.

Adding `synchronized` to all the methods will add too much contention here I think that the safest / most performant option is to make this variable volatile.

You could instead add a `ReadWrite` lock, but that seems overkill in this situation and make everything a lot more complex.

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
```diff
@@ -1,40 +1,121 @@
...
+     * Otherwise it would be part of the public API.
+     * </p>
+     * <p>
+     * Loading the concrete implementation from here is also not a good option,
+     * as it could not be easily set to a different (possibly mocked) implementation in tests.
+     * </p>
+     * <p>
+     * So the solution is to provide a package-private method which the implementation can call.
+     * What's not so nice about this solution is that the implementation has to have a class in
+     * the same package as the api in order to be able to call this package private method.
```
alvarolobato: NIT: api -> API

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
```diff
@@ -1,40 +1,121 @@
...
+     *
+     * @param tracer The tracer implementation to register.
+     */
+    static void register(Tracer tracer) {
+        synchronized (ElasticApm.class) {
+            ElasticApm.tracer = tracer;
+        }
+    }
+
+    // @VisibleForTesting
```
alvarolobato: Leftover?

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
```diff
@@ -1,40 +1,121 @@
...
+     * <p>
+     * If there is no current transaction, this method will return a noop transaction.
+     * </p>
+     *
+     * @return The currently running transaction, or a noop transaction (never <code>null</code>).
+     */
     @Override
     public Transaction currentTransaction() {
-        return tracer.currentTransaction();
+        Transaction transaction = tracer.currentTransaction();
```
alvarolobato: This is not thread safe, see above.

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
```diff
@@ -1,40 +1,121 @@
...
+     * the same package as the api in order to be able to call this package private method.
+     * But as this is not a public method,
+     * the concrete registration mechanism can always be changed later without breaking compatibility.
+     * </p>
+     *
+     * @param tracer The tracer implementation to register.
+     */
+    static void register(Tracer tracer) {
+        synchronized (ElasticApm.class) {
+            ElasticApm.tracer = tracer;
```
alvarolobato: Should we add some nullability checking here? and bail in case of a null parameter?

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
felixbarny: Then I'd have to make the `INSTANCE` variable package protected in order to be able to access the instance methods `register` and `unregister`, but that seems like a good option.

Sure, I could use volatile in order to make sure that I'm always reading the current value of `tracer` in each thread. But the worst case implication of an outdated read in this case is only that the `NoopTracer` is returned. I'd argue that is not really that important, as it only done once at startup time. Furthermore, ElasticAPM is initialized at the very start of the application. That means that it is unlikely that a Thread has ever read the tracer variable and thus being in a CPU cache.

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
felixbarny: Because the `tracer` variable is static atm. But we could definitely change that to being non static. I think in this case, it does not really affect testability, but it's always a good thing to minimize the usage of the `static` keyword.

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
felixbarny: Rather than throwing an exception, which is not always a good option within agents, we could just set it to `NoopTracer`. But on the only code path that leads to here `null`s are impossible.

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
felixbarny: Documents that this is method is for testing purposes without having a dependency on guava, where this annotation actually comes from.

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
felixbarny: see above

apm-agent-core/src/main/java/co/elastic/apm/impl/ElasticApmTracer.java
```diff
@@ -102,7 +103,15 @@ public static ElasticApmTracer get() {
         return instance;
     }
 
-    private static boolean isStarted() {
+    // @VisibleForTesting
```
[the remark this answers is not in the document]
felixbarny: see above

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
felixbarny: @danielmitterdorfer could you shed some light on wether a non-volatile read can lead to reading a corrupted value or if the worst case is only reading a slightly outdated value?

danielmitterdorfer: In the general case it can indeed happen that you read an inconsistent object state (to be clear: the object *reference* is always consistent (see [JLS §17.7](https://docs.oracle.com/javase/specs/jls/se9/html/jls-17.html#jls-17.7). In this special case, you are publishing this field via a static initializer which is always thread-safe (according to [JLS §12.4.2](https://docs.oracle.com/javase/specs/jls/se9/html/jls-12.html#jls-12.4.2)) so you should be fine here. However, IMHO you need to be careful with republication (i.e. assigning a different tracer instance to that field). It depends a bit on the tracer implementation whether you can do this in a thread-safe manner (e.g. if the tracer instance is immutable then even racy publishing is safe).

PR #18, opened 2018-03-20
Try to fix failing jenkins build

felixbarny:


felixbarny:
I have no clue why the Jenkins build fails. It works on my machine and on Travis. Something is wrong on Jenkins when it reads the keystore.jks file. The file seems to be corrupted ¯\_(ツ)_/¯

felixbarny:
jenkins retest this please

PR #24, opened 2018-03-21
Get real ip from proxy headers

felixbarny:


apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/servlet/ClientIpUtils.java
```diff
@@ -0,0 +1,53 @@
...
+public class ClientIpUtils {
+
+    /**
+     * This method returns the first IP from common HTTP header names used by reverse proxies.
+     * If there is no such header, it returns {@link HttpServletRequest#getRemoteAddr()}
+     *
+     * @param request The HTTP request.
+     * @return the
+     */
+    public static String getRealIp(HttpServletRequest request) {
```
jalvz: for reference, this is the apm-server implementation: https://github.com/elastic/apm-server/blob/12d38ced19a8a4a9555435e960bc47da4f824c6f/utility/ip.go#L12

felixbarny: As the apm server is doing this anyway, I'll close this PR.

felixbarny:
Turns out, the APM server only detects the real IP of the agent, not the IP of the actual user

alvarolobato:
Are you finally closing this? needs review?

Issue #6, opened 2018-03-08
Add configuration option for capturing form params


felixbarny:
In the first iteration, the `capture_body` option only works for `application/x-www-form-urlencoded` data. In future iterations, `multipart/form-data` or arbitrary plain-text data might be supported. 

Reading arbitrary text data, probably also multipart data, requires some extra effort as you can't just read the content as plain text from an `HttpServletRequest`. Also, you can't read the `InputStream` multiple times from the same request. That means that the agent must not call `javax.servlet.ServletRequest.getInputStream.read` for example, because it would break the monitored application. So the solution would involve wrapping the request and storing the values in a buffer as they are read by the underlying application.

PR #30, opened 2018-03-21
Implement capture_body config option

felixbarny:
Introduce Processor which processes events before they are reported
The processor is used to redact the request body based on the config


felixbarny:
closes #6

PR #41, opened 2018-03-26
WIP: Avoid creation of spy.log

felixbarny:
Use custom JdbcEventListenerFactory

P6spy 2.7.0 is not yet released so this is a WIP. See https://github.com/p6spy/p6spy/pull/437.

### DIFF
--- a/apm-agent-plugins/apm-jdbc-plugin/pom.xml
+++ b/apm-agent-plugins/apm-jdbc-plugin/pom.xml
@@ -15,7 +15,7 @@
         <dependency>
             <groupId>p6spy</groupId>
             <artifactId>p6spy</artifactId>
-            <version>3.6.0</version>
+            <version>3.7.0</version>
         </dependency>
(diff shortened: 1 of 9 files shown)


apm-agent-plugins/apm-jdbc-plugin/pom.xml
```diff
@@ -15,7 +15,7 @@
         <dependency>
             <groupId>p6spy</groupId>
             <artifactId>p6spy</artifactId>
-            <version>3.6.0</version>
+            <version>3.7.0-SNAPSHOT</version>
```
alvarolobato: Isn't this risky? You should use a real version. If a snapshot is a must, which is not recommended, at least resolve the snapshot and use the version with the timestamp so at least you have a reproducible build

felixbarny: That’s why it says wip in the title ;)

PR #44, opened 2018-03-28
Add support for Spring-MVC

felixbarny:
Sets the transaction name to ControllerName#methodName


felixbarny:
screen shot 2018-03-28 at 15 00 02

PR #48, opened 2018-04-04
Add ignore_urls configuration option

### LINKED ISSUE #5: Add configuration option to ignore certain URLs
Ignore URLs which start with a given prefix or use globs (maybe leveraging https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/util/AntPathMatcher.java). Or why not both?

Add benchmarks to get insight into the overhead in regards to latency, cpu and allocation rate.

felixbarny:
This option allows to configure a list of URL wildcard patterns
which should not be traced

closes #5

### DIFF
--- a/apm-agent-plugins/apm-web-plugin/src/main/java/co/elastic/apm/matcher/WildcardMatcher.java
+++ b/apm-agent-plugins/apm-web-plugin/src/main/java/co/elastic/apm/matcher/WildcardMatcher.java
@@ -1,0 +1,166 @@
+/*-
+ * #%L
+ * Elastic APM Java agent
+ * %%
+ * Copyright (C) 2018 the original author or authors
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


felixbarny:
Should the `WildcardMatcher` allow for ignoring the case of the matched string?
If yes, should that be the default behaviour?
If no, how should the String which is provided to `WildcardMatcher#valueOf(String wildcardString)` look like to indicate that it should be case-ignoring? Something like `(#i)foo*` (zsh-style Syntax).

This would be useful, for example, when using the `WildcardMatcher` to sanitize request parameters.

jalvz:
This is a very personal preference, but i'll go with the minimal scenario (not ignore case); and if users start asking for case ignore or more complicate things, evaluate it then.
More so as agents have their own independent release cycle...

apm-agent-plugins/apm-web-plugin/src/main/java/co/elastic/apm/matcher/WildcardMatcher.java
```diff
@@ -0,0 +1,165 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package co.elastic.apm.matcher;
+
+/**
```
jalvz: this is superb documentation! 💯

PR #50, opened 2018-04-06
Sanitize fields

### LINKED ISSUE #27: Filter HTTP headers and post params for sensitive information
see also https://www.elastic.co/guide/en/apm/agent/python/current/sanitizing-data.html

felixbarny:
closes #27


apm-agent-core/src/main/java/co/elastic/apm/sanitize/SanitizingCoreProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+        sanitizeContext(transaction.getContext());
+    }
+
+    @Override
+    public void processBeforeReport(ErrorCapture error) {
+        sanitizeContext(error.getContext());
+    }
+
+    private void sanitizeContext(Context context) {
+        sanitizeMap(context.getTags());
```
beniwohli: In the Python agent, we don't sanitize tags or custom context, as we assume that the user knows what they are putting in there.

apm-agent-core/src/main/java/co/elastic/apm/configuration/CoreConfiguration.java
```diff
@@ -104,6 +110,52 @@ public void assertValid(Double value) {
...
+            WildcardMatcher.valueOf("(?i)token"),
+            WildcardMatcher.valueOf("(?i)*key"),
+            WildcardMatcher.valueOf("(?i)*token"),
+            WildcardMatcher.valueOf("(?i)*session*"),
+            WildcardMatcher.valueOf("(?i)*credit*"),
+            WildcardMatcher.valueOf("(?i)*card*"),
+            // HTTP request header for basic auth, contains passwords
+            WildcardMatcher.valueOf("(?i)authorization"),
+            // HTTP request header cookies are already captured in Request.cookies
+            WildcardMatcher.valueOf("(?i)cookie"),
```
beniwohli: Does this mean that the `cookie` header will be set to `[REDACTED]` on every request? If that's the case, I'd rather remove it, as the value is close to zero (only thing it tells you that the request did in fact have a cookie header), and the potential for confusion is high.

In the Python agent, we handle the `cookie` header separately, by turning it into a map, sanitize the map, then turn it back into a string: https://github.com/elastic/apm-agent-python/blob/e74bd1a55b275e4d147699308c22785595c4636e/elasticapm/processors.py#L93-L96

felixbarny: It also tells the user that if they do want to collect these headers, they have to change the configuration. When I'm just dropping the header, they might not know what's going on.

But it probably makes sense to just hard-code to leave out the `Cooke` header as there is also `Request.cookie` and to leave the sanitization for `(?i)set-cookie` and `(?i)authorization` as-is.

Oh, and I also forgot to document that `(?i)` means case insensitive matching.

apm-agent-core/src/main/java/co/elastic/apm/sanitize/SanitizingCoreProcessor.java
felixbarny: good point

apm-agent-core/src/main/java/co/elastic/apm/configuration/CoreConfiguration.java
felixbarny: I'd like to avoid having to parse the cookie string to a map and back to a string as this creates a lot of garbage, which I'm trying to avoid in the Java agent.

PR #84, opened 2018-05-14
JDBC instrumentation

### LINKED ISSUE #35: Zero garbage JDBC tracing
P6spy allocates objects on each query:

https://github.com/p6spy/p6spy/blob/f9f852e895419a8adcda590618aec4ae47513cfe/src/main/java/com/p6spy/engine/wrapper/PreparedStatementWrapper.java#L59-L64

So investigate if p6spy could be easily modified to allow for object pooling or if we should rather implement our own JDBC wrapper. P6spy has an extensive test suite for all sorts of databases and application servers so it would be nice to leverage this.

felixbarny:
closes elastic/apm-agent-java#35

### DIFF
--- a/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/jdbc/StatementInstrumentation.java
+++ b/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/jdbc/StatementInstrumentation.java
@@ -1,0 +1,94 @@
+/*-
+ * #%L
+ * Elastic APM Java agent
+ * %%
+ * Copyright (C) 2018 the original author or authors
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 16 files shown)


apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/jdbc/StatementInstrumentation.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public void init(ElasticApmTracer tracer) {
+        StatementInstrumentation.tracer = tracer;
+        StatementInstrumentation.jdbcEventListener = new ApmJdbcEventListener(tracer);
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return not(isInterface())
+            .and(nameContains("Statement"))
```
tylerbenson: Just out of curiosity, why was this needed?  Is this a performance optimization to avoid the `isSubTypeOf` check?

felixbarny: See the comment in https://github.com/elastic/apm-agent-java/blob/master/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/servlet/ServletInstrumentation.java#L69
Basically it’s because the isSubType marcher can be pretty expensive. It causes a lot of type descriptions to be parsed by ByteBuddy. I stagemonitor, I cached TypeDescriptions to improve instrumentation performance (startup times). A user experienced an OOME because of an isSubType matcher. But I should re evaluate with a recent Byte Buddy version.

felixbarny:
TODO: more integration tests

PR #85, opened 2018-05-17
POC: distributed tracing

felixbarny:


axw:
I don't feel like I'm very well qualified for reviewing Java code any more, but here's a couple of high level comments:

- It would be nice to have the TraceContext parsing code in a separate repository, in a form usable by other applications. Just a nice-to-have, though.
- I think it would be a bit neater if the header decoding/encoding were separated from the core transaction API. i.e. rather than passing in a header string value, pass in the decoded struct. That way we could possibly accept/produce other header wire formats, too.

felixbarny:
Thx for the feedback!

1) The `TraceContext` class depends on classes which are specific to Elastic APM, like `Recyclable`, which would make this a bit difficult.
2) I did not decouple the header parsing yet, as this is only the internal API, which can be changed any time without the need for it being backwards compatible. So if we add support for other headers, I can just change it then. For now, I think it would only add complexity without providing benefits.

axw:
> I did not decouple the header parsing yet, as this is only the internal API, which can be changed any time without the need for it being backwards compatible. So if we add support for other headers, I can just change it then. For now, I think it would only add complexity without providing benefits.

Fair enough - I didn't realise this was an internal API. Agreed, only worth decoupling for a public API.

PR #97, opened 2018-06-04
docs: use shared attributes

axw:
Use the shared attributes provided
by elastic/docs for linking to other
product docs, forums, etc.


felixbarny:
Thanks Andrew :)

PR #110, opened 2018-06-27
Add logging configuration options

felixbarny:
* Add ability to log to a file
* Configure log level via config option

TODO: should we log to file or standard out by default?


docs/configuration.asciidoc
```diff
@@ -326,6 +326,56 @@ NOTE: All errors that are captured during a request by an ignored user agent are
...
+| Java System Properties      | Environment
+| `elastic.apm.logging.log_level` | `ELASTIC_APM_LOGGING.LOG_LEVEL`
+|============
+
+[float]
+[[config-logging-log-file]]
+==== `logging.log_file`
+
+Sets the path of the agent logs.
+The special value '_AGENT_HOME_' is a placeholder for the folder the elastic-apm-agent.jar is in.
```
alvarolobato: If you want to use _ you need tho scape them with \

apm-agent-core/src/main/java/co/elastic/apm/logging/LoggingConfiguration.java
```diff
@@ -19,18 +19,27 @@
...
 public class LoggingConfiguration extends ConfigurationOptionProvider {
 
-    private static final String LOGGING_CATEGORY = "Logging";
+    private static final String SYSTEM_OUT = "System.out";
     private static final String LOG_LEVEL_KEY = "logging.log_level";
+    private static final String LOG_FILE_KEY = "logging.log_file";
+    private static final String DEFAULT_LOG_FILE = SYSTEM_OUT;
+
+    private static final String LOGGING_CATEGORY = "Logging";
+    private static final String AGENT_HOME_PLACEHOLDER = "_AGENT_HOME_";
```
alvarolobato: This seems a bit awkward, to me, but I guess it is the best solution while still allowing relative paths from the application home.

apm-agent-core/src/main/java/co/elastic/apm/logging/LoggingConfiguration.java
```diff
@@ -45,9 +54,39 @@ public void onChange(ConfigurationOption<?> configurationOption, Level oldValue,
...
+        .key(LOG_FILE_KEY)
+        .configurationCategory(LOGGING_CATEGORY)
+        .description("Sets the path of the agent logs.\n" +
+            "The special value '_AGENT_HOME_' is a placeholder for the folder the elastic-apm-agent.jar is in.\n" +
+            "Example: '_AGENT_HOME_/logs/elastic-apm.log'\n" +
+            "\n" +
+            "When set to the special value 'System.out',\n" +
+            "the logs are sent to standard out.\n" +
+            "\n" +
+            "NOTE: When logging to a file,\n" +
```
alvarolobato: What, no file rotation?

apm-agent-core/src/main/java/co/elastic/apm/impl/ElasticApmTracer.java
```diff
@@ -119,10 +119,6 @@ public void onChange(ConfigurationOption<?> configurationOption, Double oldValue
         }
     }
 
-    public static ElasticApmTracerBuilder builder() {
-        return new ElasticApmTracerBuilder();
```
alvarolobato: Why are you moving away from having a singleton?

felixbarny: I just inlined the call to new ElasticApmTracerBuilder() and removed the static method on `ElasticApmTracer` here.

alvarolobato: Nevermind I may have missread the method :facepalm: I think I was thinking you always returned a single instance.

felixbarny: Actually, I previously did some changes around that. There is no singleton instance of `ElaticApmTracer` anymore. It gets instantiated in `premain` and all `ElasticApmInstrumentation` get the instance via a `init` method. This makes Unit testing with different settings and different instances of `ElaticApmTracer` way easier.

PR #114, opened 2018-06-29
Make sure the error context is always set

felixbarny:
Previously, most of the transaction context was set after the execution of a request.
When capturing an exception, the transaction context is copied over at it's current state.
That means when the user captures an exception, the the transaction context is not yet
populated and therefore, the Error context is empty.

This changes the initialization of the transaction context so that it is set
before the request starts.

That way, we can be sure that the Error context is always set


felixbarny:
Hey @beniwohli, could you review the comments in the commit message and in `ServletTransactionHelper`? I think the approach outlined in the comments should work, but I'd love to hear your opinion about it.

felixbarny:
Thanks for having a look :)

PR #123, opened 2018-07-10
Create spans from transaction

dmitry-saraykin:
This is convenient feature to use in reactive environment where it is not possible to use thread locals across the transaction lifecycle.

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Span.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Span.java
@@ -24,6 +24,7 @@ import co.elastic.apm.impl.stacktrace.Stacktrace;
 import co.elastic.apm.objectpool.Recyclable;
 
 import javax.annotation.Nullable;
+
 import java.util.ArrayList;
 import java.util.List;
 
@@ -50,21 +51,23 @@ public class Span extends AbstractSpan implements Recyclable, AutoCloseable {
      * The locally unique ID of the parent of the span.
      */
     @Deprecated
-    private SpanId parent = new SpanId();
+    private final SpanId parent = new SpanId();
     /**
      * Offset relative to the transaction's timestamp identifying the start of the span, in milliseconds
      * (Required)
      */
-    private double start;
+    private volatile double start;
     /**
      * Keyword of specific relevance in the service's domain (eg: 'db.postgresql.query', 'template.erb', etc)
      * (Required)
      */
     @Nullable
-    private String type;
+    private volatile String type;
+    
+    private volatile Span parentSpan;
 
     @Nullable
-    private Transaction transaction;
+    private volatile Transaction transaction;
 
     public Span start(ElasticApmTracer tracer, Transaction transaction, @Nullable Span parentSpan, long nanoTime, boolean dropped) {
         this.tracer = tracer;
@@ -84,9 +87,10 @@ public class Span extends AbstractSpan implements Recyclable, AutoCloseable {
             duration = nanoTime;
             timestamp = System.currentTimeMillis();
         }
+        this.parentSpan = parentSpan;
         return this;
     }
-
+    
     /**
      * The locally unique ID of the span.
      */
@@ -152,6 +156,7 @@ public class Span extends AbstractSpan implements Recyclable, AutoCloseable {
     }
 
     public void end(long nanoTime, boolean releaseActiveSpan) {
+        this.transaction.setCurrentSpan(this.parentSpan);
         if (isSampled()) {
             this.duration = (nanoTime - duration) / MS_IN_NANOS;
         }
@@ -160,6 +165,7 @@ public class Span extends AbstractSpan implements Recyclable, AutoCloseable {
         }
     }
 
+    @Override
     public void close() {
         end();
     }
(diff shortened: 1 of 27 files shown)


felixbarny:
Hey Dimitry and thanks for this awesome PR!

I'll have a closer look later this day. Could you explain why you changed all `Span`properties, like `start` and `type` to be volatile? Isn't a span only accessed from a single thread?

karmi:
Hi @dmitry-saraykin, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/apm-agent-java/pull/123.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

dmitry-saraykin:
Hi Felix,

In general  `Span`s and `Transaction`s can be used across multiple threads in case of multithreaded processing or reactive programming. Here is example of usage spans in multiple threads with spring webflux client
```
//Thread 1:
Span span = transaction.createSpan();
span.setType("ext");
span.name("Some micro-service call");
webClient.post().uri(...).retrieve().doOnSuccess(r-> {
    //This will happen in another thread:
    span.end();
});
```

felixbarny:
Jenkins test this, please.

apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Span.java
```diff
@@ -50,21 +51,23 @@
...
+    private volatile double start;
     /**
      * Keyword of specific relevance in the service's domain (eg: 'db.postgresql.query', 'template.erb', etc)
      * (Required)
      */
     @Nullable
-    private String type;
+    private volatile String type;
+    
+    private volatile Span parentSpan;
```
felixbarny: Add `@Nullable` annotation

apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Transaction.java
```diff
@@ -53,8 +56,12 @@
     /**
      * A mark captures the timing of a significant event during the lifetime of a transaction. Marks are organized into groups and can be set by the user or the agent.
      */
-    private final Map<String, Object> marks = new HashMap<>();
+    private final Map<String, Object> marks = new ConcurrentHashMap<>();
     private final SpanCount spanCount = new SpanCount();
+    /**
+     * Current span created for this transaction or null if there is no span at the moment. If there are nested spans this should contain the deepest.
+     */
+    private volatile Span currentSpan;
```
felixbarny: I'm not sure that approach works. For example, if you are starting two spans in parallel there could be a race condition so that one span is wrongly declared as the parent of another span when they should, in fact, be both direct children of the transaction. I'd rather remove the current span here and add `Span#createSpan` (or `Span.createChildSpan`?) as the only way of creating nested spans. I'd even consider removing `ElasticApmTracer#createSpan`, but I'm not certain about that yet.

apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Transaction.java
```diff
@@ -77,6 +84,10 @@
      * Transactions that are 'sampled' will include all available information. Transactions that are not sampled will not have 'spans' or 'context'. Defaults to true.
      */
     private boolean noop;
+    
+    public Transaction(CoreConfiguration coreConfiguration) {
```
felixbarny: Remove constructor and read the span limit in `#start` via `tracer.getConfig(CoreConfiguration.class).getTransactionMaxSpans()`

apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Transaction.java
```diff
@@ -209,6 +220,35 @@ public void setUser(String id, String email, String username) {
     public void end() {
         end(System.nanoTime(), true);
     }
+    
+    /**
+     * Creates span bound to the transaction
+     */
+    public Span createSpan() {
+        Span span;
+        span = tracer.createSpan();
```
felixbarny: tracer is nullable, so this needs a null check

apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Transaction.java
dmitry-saraykin: Agree. Also though about that.

dmitry-saraykin:
I've created `ElasticApm#startAsyncTransaction` method to create transactions which are not bound to the current thread and must be tracked manually.
Also added `Span#createSpan` method to create child spans.
Changed `Transaction#createSpan` to create always top level spans

felixbarny:
I like that approach of having a method for creating a non-thread bound transaction and span. I just think we should come up with a better name, as transactions are not inherently async when created with the `startAsyncTransaction` method. Some possibilities:

* `startNonActiveTransaction`, which would correspond to `co.elastic.apm.impl.ElasticApmTracer#activate`
* make non-active transactions the default and add a method for creating thread-bound transactions
* make non-active transactions the default and add a method for making the transaction active on the current thread, add another one to deactivate.
* make it only possible to create non-active transactions with the public API, possibly adding support for that later

felixbarny:
I'm working on some changes based on your PR

dmitry-saraykin:
- I think `startNonActiveTransaction` is not a good name because the user can think that the transaction is not going to be sent or tracked. A better name may be `startManualTransaction`. Or may be there should be some method which creates/reuses currently active transaction but tells not to `end` it automatically
- It should be possible to create transactions which are bounded to the thread as it will allow to use automatic instrumentation of all internal calls

Also found a bug with gzip. Currently agent uses zip instead of gizp and apm-server fails to parse such requests

felixbarny:
closing in favor of https://github.com/elastic/apm-agent-java/pull/133. @dmitry-saraykin, could you please review that PR if it covers your use case?

PR #147, opened 2018-08-01
Fix bug in intake v2 when log level is DEBUG

felixbarny:
When DEBUG logs are active the agent logs the JSON of the Transaction objects it serializes.
When serializing the transaction to a String, which is only done when it should be logged, it resets the JsonWriter which detaches the request body's OutputStream form the JsonWriter.

The solution is to buffer the serialized bytes when the payload should be logged. When the request is about to end, log the buffered bytes as a String and write the bytes to the underlying OutputStream.


felixbarny:
y u no test?
Jenkins, test this please!

simitt:
jenkins, retest this please.

felixbarny:
He won't listen!

PR #166, opened 2018-08-15
fix typo in intro doc, ist -> is

ravinaik1312:


felixbarny:
Thanks for spotting this!

PR #173, opened 2018-08-21
Remove class name check

eyalkoren:


eyalkoren:
@felixbarny no, please see the description and task breaking in [the related github issue](https://github.com/elastic/apm-dev/issues/295). I assumed we agreed we better provide worse performance than miss stuff, so as a first step I just removed these checks and planned that next step will include performance assessments. If you think the order of tasks should be reversed - no problem.

felixbarny:
I think it's worth to at least know how much worse the performance is when removing the name checks. Just to be sure we don't break/considerably slow down applications when doing that. Because it might even be better to not instrument everything but be fast than to instrument everything but to break applications.

felixbarny:
No, I don't. But I've encountered OOMEs in a different project when not pre-filtering before the `hasSuperType` check. That was with enabled type pool caching, however. As we don't have that enabled, we would probably just increase startup times and not break the application - probably.

eyalkoren:
See performance assessments in [the github issue](https://github.com/elastic/apm-dev/issues/295). If we can already agree on a specific implemenation, I will commit it within this pull request.

eyalkoren:
Due to [this defect](https://github.com/raphw/byte-buddy/issues/530) that @felixbarny encountered, we decided to currently not use method matching within type matching.
We will keep the current `nameContains` and see if it yields false negatives. The `Servlet` filter may be good enough after all since we would normally instrument the `service` method of some base/abstract `Servlet` implementation. For JDBC, all drivers tested so far were instrumented properly whilst relying on the `Connection` implementation containing the word "Connection". Usage of anonymous classes may cause misses, but not sure how likely these are for these use cases.

PR #175, opened 2018-08-23
WIP: integration into SLF4J's MDC for trace-log-correlation

felixbarny:


apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/AbstractSpan.java
```diff
@@ -138,12 +144,28 @@ public T activate() {
         final ElasticApmTracer tracer = this.tracer;
         previouslyActive = tracer.getActive();
         tracer.setActive(this);
+        List<SpanListener> spanListeners = tracer.getSpanListeners();
+        for (int i = 0; i < spanListeners.size(); i++) {
+            try {
+                spanListeners.get(i).onActivate(this);
+            } catch (Throwable t) {
+                logger.warn("Exception while calling {}#onActivate", spanListeners.get(i).getClass().getSimpleName(), t);
```
henrikno: You shouldn't catch-log all throwables willy-nilly. It will catch OutOfMemoryError and other serious VirtualMachineError that should bubble up.
Some programs re-throw if it's a subclass of VirtualMachineError | ThreadDeath | InterruptedException | LinkageError | ControlThrowable, which seems like a reasonable approach.

felixbarny: Thx for the suggestion. Ideally I would not catch Throwables at all, the problem is that `MethodHandle#invoke` throws a Throwable and that I need to handle that somehow.

apm-agent-core/src/main/java/co/elastic/apm/cache/WeakKeySoftValueLoadingCache.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        @Nullable
+        public V get(K key) {
+            if (valueReference == null) {
+                return null;
+            }
+            V value = this.valueReference.get();
+            if (value != null) {
+                return value;
```
eyalkoren: Logging the reclaimed `SoftReference`s could be useful (`INFO` level - shouldn't be often and if it is we want to know about it ASAP) so we can have an idea of how often they are GCed.
It can be even more useful if compared with the number of keys reclaimed in the underlying map but this may require extend `com.blogspot.mydailyjava.weaklockfree.WeakConcurrentMap.WithInlinedExpunction` and override `expungeStaleEntries`...

apm-agent-core/src/main/java/co/elastic/apm/impl/ElasticApmTracer.java
```diff
@@ -42,6 +42,7 @@
 
 import javax.annotation.Nullable;
 import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
```
eyalkoren: I don't think this import is used

apm-agent-core/src/test/java/org/slf4j/impl/StaticMDCBinder.java
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package org.slf4j.impl;
+
+import org.slf4j.helpers.BasicMDCAdapter;
+import org.slf4j.spi.MDCAdapter;
+
+public class StaticMDCBinder {
```
eyalkoren: I didn't find usages for this one

apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/AbstractSpan.java
```diff
@@ -138,12 +144,32 @@ public T activate() {
         final ElasticApmTracer tracer = this.tracer;
         previouslyActive = tracer.getActive();
         tracer.setActive(this);
+        List<SpanListener> spanListeners = tracer.getSpanListeners();
+        for (int i = 0; i < spanListeners.size(); i++) {
+            try {
+                spanListeners.get(i).onActivate(this);
+            } catch (Error e) {
```
eyalkoren: Isn't that the same as only catching Exceptions?

felixbarny: I would like to only catch `Exception`s but `MethodHandle#invoke` throws `Throwable`

apm-agent-core/src/test/java/org/slf4j/impl/StaticMDCBinder.java
felixbarny: right, this was just an experiment and should be removed

apm-agent-core/src/main/java/co/elastic/apm/impl/ElasticApmTracer.java
felixbarny: good catch

apm-agent-core/src/main/java/co/elastic/apm/cache/WeakKeySoftValueLoadingCache.java
```diff
@@ -0,0 +1,86 @@
...
+package co.elastic.apm.cache;
+
+import com.blogspot.mydailyjava.weaklockfree.WeakConcurrentMap;
+
+import javax.annotation.Nullable;
+import java.lang.ref.SoftReference;
+
+public class WeakKeySoftValueLoadingCache<K, V> {
+
+    private final WeakConcurrentMap<K, CacheValue<K, V>> cache = new WeakConcurrentMap.WithInlinedExpunction<>();
```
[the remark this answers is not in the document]
felixbarny: I'm not sure how a normal `WeakHashMap` behaves when used concurrently. For example, using `LinkedHashMap`s concurrently can lead to deadlocks. I would therefore only use thread-safe data structures unless we find a good source which confirms that using a plain `WeakHashMap` is safe and could only result in duplicate value creations in the worst case.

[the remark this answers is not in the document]
eyalkoren: How can `LinkedHashMap` (or `WeakHashMap`) can cause deadlock if it doesn't synchronize on anything?? I assume you need to sync on it (or some other monitor) when using it yourself to cause deadlock, but this is not what I am suggesting. I am suggesting using a similar approach to what you did - be aware that you might get duplication in creation or getting null at any operation (which is a risk of any WeakHashMap anyway).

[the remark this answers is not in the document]
felixbarny: I remembered it wrong, it’s not a deadlock but an infinite loop that can happen: https://stackoverflow.com/questions/35534906/java-hashmap-getobject-infinite-loop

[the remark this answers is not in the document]
felixbarny: Seems like the same can indeed happen with a WeakConcurrentMap: https://github.com/spring-projects/spring-loaded/issues/194

PR #177, opened 2018-08-24
Add reference property file with all options to docs

felixbarny:


felixbarny:
property file reference preview

eyalkoren:
I definitely not feel strongly enough about it :)

PR #190, opened 2018-09-03
Client-side load balancing

### LINKED ISSUE #13: Implement client-side load balancing
Let users configure a list of APM servers which are used in a round-robin fashion when reporting transactions.

felixbarny:
closes elastic/apm-agent-java#13


felixbarny:
This is already mergable IMO. If we decide on a different load-balancing algorithm, we can still change that. But for now, I think it's safe to merge, especially because this change is backwards compatible.

PR #203, opened 2018-09-15
Enable complex wildcard patterns

felixbarny:
- match case insensitive by default
- matches are now always allocation free

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 # next version
 
 ## Breaking changes
+ * Wildcard patterns are case insensitive by default. Prepend `(?-i)` to make the matching case sensitive.
 
 ## Features
+ * Wildcard patterns are now not limited to only one wildcard in the middle and can be arbitrarily complex now.
+   Example: `*foo*bar*baz`.
 
 ## Bug Fixes
 
--- a/apm-agent-core/src/main/java/co/elastic/apm/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/configuration/CoreConfiguration.java
@@ -148,20 +148,19 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
         .dynamic(true)
         .tags("security")
         .buildWithDefault(Arrays.asList(
-            WildcardMatcher.valueOf("(?i)password"),
-            WildcardMatcher.valueOf("(?i)passwd"),
-            WildcardMatcher.valueOf("(?i)pwd"),
-            WildcardMatcher.valueOf("(?i)secret"),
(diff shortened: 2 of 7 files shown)


felixbarny:
Sounds very interesting. How do you think this can be optimized further? Some tweaks in the `indexOfIgnoreCase` method?

eyalkoren:
At first I was just looking at the `indexOfIgnoreCase` and thought it could be quite easily optimised by using the [Rabin-Karp algorithm](https://en.wikipedia.org/wiki/Rabin%E2%80%93Karp_algorithm). Basically this algorithm is computing running-hashes so a single iteration should suffice for multiple patterns. It still has a worst-case `O(N*M)` per matching, like the current algorithm, but the current approach **must run it for each pattern**, meaning that if you have `P` defined patterns with an average length `M`, then for an examined string of length `N` you would get `O(N*M*P)`. 
I then thought of using a different algorithm that relies on `Trie`s that I think can yield a worst-case `O(N)` for any `P` and `M`. I will give it some more thought and maybe just write something to try out.

eyalkoren:
So far I couldn't think of a simple way to implement the current requirement of multiple wildcards with a bunch of Tries, but I am currently looking at the [Aho–Corasick algorithm](https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm) that we may be able to apply to get a worst-case `O(N+P*M)` complexity - still linear to `N`.

eyalkoren:
For those you wouldn't have to add the multiple-wildcard capabilities and for those it should be much simpler to implement using a `Trie` that will make it much more efficient because it will compare all at one iteration. So I think it's worth a discussion anyway.

felixbarny:
Can we first merge this and make a task for investigating further optimizations?

On my local machine, the benchmarks have an average runtime of ~1700ns/op. When removing all default matchers, I get ~1400ns/op which is around 18% faster. So there is definitely some room for improvement. Although I'd even consider an overhead of 1700ns/op pretty low.

felixbarny:
> Once you open it for users to define patterns

That's currently already possible. See [`ignore_urls`](https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-ignore-urls), [`ignore_user_agents`](https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-ignore-user-agents), [`url_groups`](https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-url-groups) and [`sanitize_field_names`](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#config-sanitize-field-names). The other agents mostly implement that via regex so using the wildcard matcher, it should be quite a bit faster even.

What do you suggest doing with this PR? Don't you think we can merge it now and try other approaches after that?

PR #204, opened 2018-09-16
Move tags element to be nested under Context rather than Db

eyalkoren:


apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Db.java
```diff
@@ -82,7 +82,7 @@ public Db withStatement(@Nullable String statement) {
         this.statement = statement;
         return this;
     }
-
+
```
felixbarny: remove space

apm-agent-core/src/main/java/co/elastic/apm/report/serialize/DslJsonSerializer.java
```diff
@@ -564,20 +564,37 @@ private boolean isLibraryFrame(String className) {
...
-        writeField("user", db.getUser());
-        writeFieldName("tags");
-        serializeTags(context.getTags());
-        jw.writeByte(OBJECT_END);
+
+        boolean dbContextWritten = serializeDbContext(context.getDb());
+
+        Map<String, String> tags = context.getTags();
+        if (!tags.isEmpty()) {
+            if(dbContextWritten)
```
felixbarny: Currently, methods like `serializeDbContext` add the comma themselves when they are written and the last object, like tags, is always written, even if it's empty. That makes sure the comma is not the last thing in an object.

apm-agent-core/src/test/java/co/elastic/apm/impl/payload/TransactionPayloadJsonSchemaTest.java
```diff
@@ -113,6 +114,55 @@ void testJsonSchemaDslJsonAllValues() throws IOException {
...
+    @Test
+    void testJsonStructure() throws IOException {
+        TransactionPayload payload = createPayloadWithAllValues();
+
+        when(coreConfiguration.isDistributedTracingEnabled()).thenReturn(false);
+        DslJsonSerializer serializer = new DslJsonSerializer(coreConfiguration.isDistributedTracingEnabled(), mock(StacktraceConfiguration.class));
+
+        validateDbSpanSchema(payload, serializer, true);
+
+        for(Span span: payload.getTransactions().get(0).getSpans()) {
```
felixbarny: add space before brace, also on next line

apm-agent-core/src/test/java/co/elastic/apm/impl/payload/TransactionPayloadJsonSchemaTest.java
```diff
@@ -113,6 +114,55 @@ void testJsonSchemaDslJsonAllValues() throws IOException {
...
+                assertThat(db.get("user").textValue()).isEqualTo("readonly_user");
+                JsonNode tags = context.get("tags");
+                if (shouldContainTags) {
+                    assertThat(tags).isNotNull();
+                    assertThat(tags).hasSize(2);
+                    assertThat(tags.get("monitored_by").textValue()).isEqualTo("ACME");
+                    assertThat(tags.get("framework").textValue()).isEqualTo("some-framework");
+                }
+                else
+                {
```
felixbarny: put curly brace in previous line

apm-agent-core/src/test/java/co/elastic/apm/impl/payload/TransactionPayloadJsonSchemaTest.java
```diff
@@ -113,6 +114,55 @@ void testJsonSchemaDslJsonAllValues() throws IOException {
...
+            }
+        }
+    }
+
+    private void validateDbSpanSchema(TransactionPayload payload, DslJsonSerializer serializer, boolean shouldContainTags) throws IOException {
+        final String content = serializer.toJsonString(payload);
+        JsonNode node = objectMapper.readTree(content);
+
+        boolean contextOfDbSpanFound = false;
+        for (JsonNode child: node.get("transactions").get(0).get("spans")) {
```
felixbarny: In intake v2, the spans are not nested inside transactions, but in `payload.spans`

eyalkoren: So what do you suggest?

apm-agent-core/src/main/java/co/elastic/apm/report/serialize/DslJsonSerializer.java
eyalkoren: meaning - you want to always add the `tags` section, even if it mostly empty?

apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Db.java
eyalkoren: r u kidding? it's the change I just did in order to test the GitHub issue...

felixbarny: I didn't see that when looking at the full change set.

apm-agent-core/src/main/java/co/elastic/apm/report/serialize/DslJsonSerializer.java
felixbarny: That would make the code more concise but on the other hand, increases the size of the JSON unnecessarily. I'm torn between the two options... What do you think?

apm-agent-core/src/test/java/co/elastic/apm/impl/payload/TransactionPayloadJsonSchemaTest.java
felixbarny: Look in both locations

apm-agent-core/src/main/java/co/elastic/apm/report/serialize/DslJsonSerializer.java
eyalkoren: I would say not adding by default a section that is normally empty. Serialization code is like that sometimes...

felixbarny: Ok, I'm good with that. Let's leave it as-is.

PR #207, opened 2018-09-17
Unnest exception types matching *Nested*Exception by default

felixbarny:
depends on elastic/apm-agent-java#206


felixbarny:
Instead of

screen shot 2018-09-17 at 18 24 43

(type `org.springframework.web.util.NestedServletException`)


do 
screen shot 2018-09-17 at 18 24 50

(type `java.lang.RuntimeException`)

PR #216, opened 2018-09-19
Proper handling of invalid traceparent header

felixbarny:


apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/TraceContext.java
```diff
@@ -57,24 +57,35 @@
...
+                logger.warn("Only version 00 of the traceparent header is supported, but was '{}'", traceParentHeader);
+                asRootSpan(sampler);
+                return;
+            }
+            parseTraceId(traceParentHeader);
+            if (traceId.isEmpty()) {
+                asRootSpan(sampler);
+                return;
+            }
+            parseParentId(traceParentHeader);
```
axw: should you be checking `parentId.isEmpty()` here also?

apm-agent-core/src/main/java/co/elastic/apm/util/HexUtils.java
```diff
@@ -64,22 +64,34 @@ public static void writeByteAsHex(byte b, StringBuilder sb) {
...
+        if (hi == -1 || lo == -1) {
+            throw new IllegalArgumentException("Not a hex encoded string: " + hexEncodedString + " at offset " + offset);
         }
+        return (byte) ((hi << 4) + lo);
     }
 
-    public static byte getNextByte(String hexEncodedString, int offset) {
-        return (byte) ((Character.digit(hexEncodedString.charAt(offset), 16) << 4)
-            + Character.digit(hexEncodedString.charAt(offset + 1), 16));
+    private static int hexCharToBinary(char ch) {
```
axw: why do this instead of `Character.digit(ch, 16)`? is this faster?
couldn't you just do (in getNextByte):

```java
final int hi = Character.digit(hexEncodedString.charAt(offset), 16);
final int lo = Character.digit(hexEncodedString.charAt(offset + 1), 16);
(rest is the same)
```
?

apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Transaction.java
```diff
@@ -83,7 +83,7 @@ public Transaction(ElasticApmTracer tracer) {
 
     public Transaction start(@Nullable String traceParentHeader, long startTimestampNanos, Sampler sampler) {
         if (traceParentHeader != null) {
-            traceContext.asChildOf(traceParentHeader);
+            traceContext.asChildOf(traceParentHeader, sampler);
```
axw: Your call of course, but it might be a bit neater to leave the concern of what to do in case of an invalid traceparent header to the caller. e.g. you could make `asChildOf` return a bool indicating success, and then:

```java
if (traceParentHeader == null || !traceContext.asChildOf(traceParentHeader)) {
    traceContext.asRootSpan(sampler);
}
```

apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/TraceContext.java
felixbarny: definitely!

apm-agent-core/src/main/java/co/elastic/apm/util/HexUtils.java
felixbarny: TBH, I didn't measure it but I suppose this is faster. Semantically, it's equivalent

apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Transaction.java
felixbarny: nice idea!

Issue #174, opened 2018-08-22
Add context.http.url for HTTP client spans


felixbarny:
See https://github.com/elastic/kibana/issues/20145

eyalkoren:
[Pull Request](https://github.com/elastic/apm-agent-java/pull/218#issue-216964227)

PR #218, opened 2018-09-20
Add http url to span context

eyalkoren:


eyalkoren:
[https://github.com/elastic/apm-agent-java/issues/174](https://github.com/elastic/apm-agent-java/issues/174)

apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/httpclient/ApacheHttpClientInstrumentation.java
```diff
@@ -56,7 +56,7 @@ private static void onBeforeExecute(@Advice.Argument(0) HttpRoute route,
             return;
         }
         final AbstractSpan<?> parent = tracer.getActive();
-        span = HttpClientHelper.startHttpClientSpan(parent, request.getMethod(), route.getTargetHost().getHostName(), SPAN_TYPE_APACHE_HTTP_CLIENT);
+        span = HttpClientHelper.startHttpClientSpan(parent, request.getMethod(), String.valueOf(request.getURI()), route.getTargetHost().getHostName(), SPAN_TYPE_APACHE_HTTP_CLIENT);
```
felixbarny: why do you use `String.valueOf` here? Can the URI return null? In that case, it might be better to change the signature of `startHttpClientSpan` to `@Nullable URI` or `@Nullable String` and only set if not null.

eyalkoren: According to `HttpRequestWrapper`'s constructor it can be set to null. You're right, better not to set a null string.

PR #224, opened 2018-09-24
Add support for JAX-RS-based transaction names

### LINKED ISSUE #67: Add support for JAX-RS-based transaction names

felixbarny:
closes elastic/apm-agent-java#67


apm-agent-core/src/main/java/co/elastic/apm/bci/bytebuddy/CustomElementMatchers.java
```diff
@@ -0,0 +1,78 @@
...
+        for (String applicationPackage : includedPackages) {
+            matcher = matcher.or(nameStartsWith(applicationPackage));
+        }
+        return matcher;
+    }
+
+    public static ElementMatcher.Junction<ClassLoader> classLoaderCanLoadClass(final String className) {
+        return new ElementMatcher.Junction.AbstractBase<ClassLoader>() {
+
+            private final boolean loadableByBootstrapClassLoader = canLoadClass(ClassLoader.getSystemClassLoader().getParent(), className);
```
eyalkoren: I believe `ClassLoader.getSystemClassLoader()` typically returns an instance (used to be `Luncher$AppClassLoader` before Java 9 modules and in my local JDK 10 returns an instance of `ClassLoaders$AppClassLoaders` in the open JDK), for which `getParent()` typically returns a non-null instance (used to be `Launcher$ExtClassLoader`; in my current JDK 10 I get `ClassLoaders$PlatformClassLoader`) which is not the bootstrap CL. If you invoke `getParent` on the latter you would get `null`, which indicates the bootstrap CL. So I believe the `canLoadClass` would yield `true` when it's false and NPE when it should be true.

apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/jaxrs/JaxRsTransactionNameInstrumentation.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public void init(ElasticApmTracer tracer) {
+        applicationPackages = tracer.getConfig(StacktraceConfiguration.class).getApplicationPackages();
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        // setting application_packages makes this matcher more performant but is not required
+        return isInAnyPackage(applicationPackages, ElementMatchers.<TypeDescription>any())
```
eyalkoren: Don't you think it is likely that there are frameworks that use JAX-RS? I am not sure what we want to do if there are, but this means we will miss them, right?

apm-agent-core/src/main/java/co/elastic/apm/bci/bytebuddy/CustomElementMatchers.java
```diff
@@ -0,0 +1,59 @@
...
+                }
+                return result;
+            }
+        };
+    }
+
+    private static boolean canLoadClass(ClassLoader target, String className) {
+        boolean result;
+        try {
+            target.loadClass(className);
```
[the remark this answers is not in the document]
felixbarny: What side effects are you thinking about? What could be an alternative to this approach?

I think a slightly better way would be trying to load the class via `java.lang.Class#forName(java.lang.String, boolean, java.lang.ClassLoader)` and setting the `initialize` flag to `false`.

apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/jaxrs/JaxRsTransactionNameInstrumentation.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return isInAnyPackage(applicationPackages, any())
+            .and(isAnnotatedWith(named("javax.ws.rs.Path")));
+    }
+
+    @Override
+    public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
+        return not(isBootstrapClassLoader())
+            .and(classLoaderCanLoadClass("javax.ws.rs.Path"));
```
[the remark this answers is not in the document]
felixbarny: This filters out class loaders which are not even able to load the `@Path` annotation. By filtering these out early, we don't have to look at their classes at all which speeds up the type matching.

apm-agent-core/src/main/java/co/elastic/apm/bci/bytebuddy/CustomElementMatchers.java
felixbarny: Two ideas:
* We could just require a `boolean` constructor argument whether or not the bootstrap CL should match
* call `getClass().getClassLoader()` on the class to check whether the returned CL is `null`

apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/jaxrs/JaxRsTransactionNameInstrumentation.java
felixbarny: Good point. Not sure if that is a likely case. The use case would be when importing a 3rd party library which also contains JAX-RS resources, right?

apm-agent-core/src/main/java/co/elastic/apm/bci/bytebuddy/CustomElementMatchers.java
felixbarny: oh, wait, `Class.forName` supports a `null` class loader

apm-agent-core/src/main/java/co/elastic/apm/bci/bytebuddy/CustomElementMatchers.java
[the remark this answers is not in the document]
felixbarny: We have direct control over which classes are affected. In this case, it is just the `javax.ws.rs.Path` class, which is also just an annotation.

apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/jaxrs/JaxRsTransactionNameInstrumentation.java
eyalkoren: Yes, that what I meant

apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/jaxrs/JaxRsTransactionNameInstrumentation.java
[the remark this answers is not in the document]
eyalkoren: Did you test that this actually speeds up things? The implementation actually tries to load a class at each check. Regardless, I think the bigger concern is that the loading attempt may alter application startup processes. We should avoid that as agent.

apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/jaxrs/JaxRsTransactionNameInstrumentation.java
eyalkoren: I am not sure if this check provides any value...

apm-agent-core/src/main/java/co/elastic/apm/bci/bytebuddy/CustomElementMatchers.java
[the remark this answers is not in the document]
eyalkoren: OK, so at least you should add an **ALL CAPITALS** warning that this check should be used with caution and only for classes that have no dependencies (like a small interface or annotation)

apm-agent-core/src/main/java/co/elastic/apm/bci/bytebuddy/CustomElementMatchers.java
```diff
@@ -0,0 +1,78 @@
...
+        };
+    }
+
+    private static boolean canLoadClass(@Nullable ClassLoader target, String className) {
+        boolean result;
+        try {
+            Class.forName(className, false, target);
+            result = true;
+        } catch (Exception ignore) {
+            result = false;
```
eyalkoren: Redundant setting to `false`

apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/jaxrs/JaxRsTransactionNameInstrumentation.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    public void init(ElasticApmTracer tracer) {
+        applicationPackages = tracer.getConfig(StacktraceConfiguration.class).getApplicationPackages();
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        // setting application_packages makes this matcher more performant but is not required
+        return isInAnyPackage(applicationPackages, ElementMatchers.<TypeDescription>any())
+            .and(isAnnotatedWith(named("javax.ws.rs.Path")));
```
eyalkoren: Are you sure the class must be annotated with `javax.ws.rs.Path`? Isn't it allowed to only annotate methods?

apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/jaxrs/JaxRsTransactionNameInstrumentation.java
felixbarny: Quite the contrary. This `isInAnyPackage` matcher is a really cheap one because the class file does not have to be parsed at all. In general, the matchers which match based on the class names are the cheapest ones. Without this check, each and every class has to be parsed to see if it is annotated with `@Path`. Prepending the `isInAnyPackage` matcher considerably limits the amount classes which have to be checked with a more expensive matcher :). Also, typically, `application_packages` is only set to one or a handful of packages.

In the case you are importing a 3rd party library containing JAX-RS resources, it probably makes sense to add that to the application_packages namespace anyway. Because then stack traces also highlight those classes. But I can see that this may be confusing for users which wonder why some endpoints are not picked up. It's a tradeoff but I think with proper documentation (right at the supported technologies page), this would be the tradeoff I prefer.

apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/jaxrs/JaxRsTransactionNameInstrumentation.java
felixbarny: Some answers on Stack Overflow suggested`@Path` is required on the class level.

But this quote from the JAX-RS 2.0 spec lets me doubt this:

> A resource class is a Java class that uses JAX-RS annotations to implement a corresponding Web resource.
Resource classes are POJOs that have at least one method annotated with `@Path` or a request method designator.

I'll investigate

felixbarny:
> I think the path is ALWAYS better than code

Well, I guess it depends on who is viewing the data and also how the controllers and methods are named. The verb and method don't always convey the semantics of an operation

Example: `POST /users/$id` can mean a lot of things and not all applications are actually conforming to the semantics of the verbs. Does it mean update user? Or maybe even delete user? Something else? Especially business people may not know the subtle differences between `POST` and `PUT`, for example. So I think sometimes it's easier for them to display the method name, as it might carry more information and tells what is actually happening. Of course, that requires the developers to use meaningful method names.

Maybe we can discuss this in today's agent meeting and see what other agents are doing. Also for Spring MVC, we already use `ClassName#methodName` as the transaction name, so it makes sense to be consistent in the JAX-RS instrumentation.

> I understand we are reluctant to rely on it due to path params, but the support for JAX-RS can rely on the @Path and @PathParams in order to avoid that, can't it?

Sure, we would not have problems with transaction name explosions when relying on the value of `@Path`.

apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/jaxrs/JaxRsTransactionNameInstrumentation.java
eyalkoren: If you have a test app maybe just testing would be fastest

apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/jaxrs/JaxRsTransactionNameInstrumentation.java
eyalkoren: > Without this check, each and every class has to be parsed to see if it is annotated with `@Path`. Prepending the `isInAnyPackage` matcher considerably limits the amount classes which have to be checked with a more expensive matcher :)

Right, my mistake, I confused with the CL matcher that use caches...

felixbarny:
Another important quote from the spec (http://download.oracle.com/otn-pub/jcp/jaxrs-2_0-fr-eval-spec/jsr339-jaxrs-2.0-final-spec.pdf section 3.6 Annotation Inheritance):

> Note that inheritance of class or interface annotations is not supported.

So it seems that the `@Path` annotation has to be on the concrete resource class and a `@Path` on a superclass or an implemented interface is not sufficient. However, method and method parameter-level annotations can be inherited from superclasses/interfaces:

> JAX-RS annotations may be used on the methods and method parameters of a super-class or an implemented interface.

felixbarny:
The performance of matching annotations of a specific type does not seem to be great :/
Enabling the JAX-RS instrumentation increases the type matching overhead by a factor of ~3.

The good news is that if users specify their `application_packages`, the instrumentation will only look for JAX-RS resources in those packages.

|enable_type_matching_name_pre_filtering|excluded_from_instrumentation|Total time spent matching|
|-|-|-|
|true|jax-rs|2,555ms|
|true||6,893ms|
|false|jax-rs|3,497ms|
| false||12,351ms|

felixbarny:
We concluded we should log a warning if the user did not configure the `application_packages`. In addition, we will also extend the default list of excluded packages.

felixbarny:
Jenkins retest this please

PR #228, opened 2018-09-27
Elasticsearch REST client instrumentation: adding support for sync ops and IT

eyalkoren:
This is still WIP, but it's a good point to review.
Several notes:

- The discussion about what to do with tags is still in progress, so I just decided to copy `Span` tags to `ErrorCapture` tags at the moment, just so you can see what info we have. Once decided, I can either put the relevant info on `Span` tags in case of error, or add the tags directly to the `ErrorCapture` when creating it (without relying on copying from parent span)

- I add the query options and parameters as tags as well. I assume that won't be the default behaviour. If you think this may valuable info for some users, we can add a configuration property that enables adding those and is off by default, or we can just overlook those.

- The `onMethodEnter` advice currently introduces variable due to iteration (starting [here](https://github.com/elastic/apm-agent-java/pull/228/files#diff-a50aafc4c23195a6df81ed2942542584R72)), if we decide to keep options and parameter, I will replace them with inlined code or external method, e.g:
```
        // Add request parameters to query info
        span.getContext().getTags().putAll(request.getParameters());

        // Add request options to query info
        request.getOptions().getHeaders().forEach(header -> span.addTag(header.getName(), header.getValue()));
```

- Feel free to suggest different names for span type, tag names etc.

### DIFF
--- a/apm-agent-plugins/apm-es-restclient-plugin/pom.xml
+++ b/apm-agent-plugins/apm-es-restclient-plugin/pom.xml
@@ -13,7 +13,7 @@
     <name>${project.groupId}:${project.artifactId}</name>
 
     <properties>
-        <elasticsearch.version>6.4.1</elasticsearch.version>
+        <version.elasticsearch>6.4.1</version.elasticsearch>
     </properties>
 
     <dependencies>
@@ -21,12 +21,14 @@
         <dependency>
             <groupId>org.elasticsearch.client</groupId>
             <artifactId>elasticsearch-rest-client</artifactId>
-            <version>${elasticsearch.version}</version>
+            <version>${version.elasticsearch}</version>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.elasticsearch.client</groupId>
             <artifactId>elasticsearch-rest-high-level-client</artifactId>
-            <version>${elasticsearch.version}</version>
+            <version>${version.elasticsearch}</version>
+            <scope>provided</scope>
         </dependency>
 
         <!-- Elasticsearch testcontainer -->
@@ -34,6 +36,7 @@
             <groupId>fr.pilato.elasticsearch.testcontainers</groupId>
             <artifactId>testcontainers-elasticsearch</artifactId>
             <version>0.1</version>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 
--- a/apm-agent-plugins/apm-es-restclient-plugin/pom.xml
+++ b/apm-agent-plugins/apm-es-restclient-plugin/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>apm-agent-plugins</artifactId>
         <groupId>co.elastic.apm</groupId>
-        <version>0.8.0-SNAPSHOT</version>
+        <version>0.9.0-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/apm-agent-plugins/apm-es-restclient-plugin/pom.xml
+++ b/apm-agent-plugins/apm-es-restclient-plugin/pom.xml
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>apm-agent-plugins</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>0.8.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>apm-es-restclient-plugin</artifactId>
+    <name>${project.groupId}:${project.artifactId}</name>
+
+    <properties>
+        <elasticsearch.version>6.4.1</elasticsearch.version>
+    </properties>
+
+    <dependencies>
+        <!-- Elasticsearch rest client -->
+        <dependency>
+            <groupId>org.elasticsearch.client</groupId>
+            <artifactId>elasticsearch-rest-client</artifactId>
+            <version>${elasticsearch.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.elasticsearch.client</groupId>
+            <artifactId>elasticsearch-rest-high-level-client</artifactId>
+            <version>${elasticsearch.version}</version>
+        </dependency>
+
+        <!-- Elasticsearch testcontainer -->
+        <dependency>
+            <groupId>fr.pilato.elasticsearch.testcontainers</groupId>
+            <artifactId>testcontainers-elasticsearch</artifactId>
+            <version>0.1</version>
+        </dependency>
+    </dependencies>
+
+</project>
--- a/apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
+++ b/apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
@@ -21,6 +21,7 @@ package co.elastic.apm.es.restclient;
 
 import co.elastic.apm.bci.VisibleForAdvice;
 import co.elastic.apm.report.serialize.DslJsonSerializer;
+import org.elasticsearch.common.io.stream.InputStreamStreamInput;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -36,7 +37,7 @@ public class ESRestClientInstrumentationHelper {
 
     @Nullable
     @VisibleForAdvice
-    public static String readRequestBody(InputStream bodyIS, String endpoint) throws IOException {
+    public static String readRequestBody(InputStream bodyIS, long contentLength, String endpoint) throws IOException {
         String body = null;
         try {
             byte[] data = bodyReadBuffer.get();
@@ -46,16 +47,47 @@ public class ESRestClientInstrumentationHelper {
                 data = new byte[DslJsonSerializer.MAX_LONG_STRING_VALUE_LENGTH];
                 bodyReadBuffer.set(data);
             }
-            int length = bodyIS.read(data, 0, data.length);
 
-            // Using platform's default charset for decoding
-            body = new String(data, 0, length);
-        } catch (IOException e) {
-            LOGGER.info("Failed to read request body for " + endpoint);
-        } finally {
+            try {
+                body = new InputStreamStreamInputWrapper(bodyIS, contentLength).readString();
+            } catch (Exception e) {
+                LOGGER.info("Failed to read request body for " + endpoint);
+            }
+
+        }
+        finally {
             bodyIS.close();
         }
 
         return body;
+    }
+
+    private static class InputStreamStreamInputWrapper extends InputStreamStreamInput {
+        private byte[] contentLengthBytes = new byte[5];
+        private int contentLengthValidReadIndex = 0;
+        private int contentLengthByteReadIndex;
+
+        InputStreamStreamInputWrapper(InputStream is, long contentLength) {
+            super(is);
+            // Encoding the content length into bytes. This size must be returned from the InputStream before the content is returned.
+            for (int i=0; i<contentLengthBytes.length; i++) {
+                contentLengthValidReadIndex++;
+                contentLengthBytes[i] = (byte)(contentLength & 0x7F);
+                contentLength >>>= 7;
+                if (contentLength == 0) {
+                    break;
+                }
+                contentLengthBytes[i] |= 0x80;
+            }
+            contentLengthByteReadIndex = 0;
+        }
+
+        @Override
+        public byte readByte() throws IOException {
+            if(contentLengthByteReadIndex < contentLengthValidReadIndex) {
+                return contentLengthBytes[contentLengthByteReadIndex++];
+            }
+            return super.readByte();
+        }
     }
 }
--- a/apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
+++ b/apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
@@ -41,8 +41,10 @@ public class ESRestClientInstrumentationHelper {
         try {
             byte[] data = bodyReadBuffer.get();
             if (data == null) {
-                bodyReadBuffer.set(new byte[DslJsonSerializer.MAX_LONG_STRING_VALUE_LENGTH]);
-                data = bodyReadBuffer.get();
+                // The DslJsonSerializer.MAX_LONG_STRING_VALUE_LENGTH is actually used to count chars and not bytes, but that's not
+                // that important, the most important is that we limit the payload size we read and decode
+                data = new byte[DslJsonSerializer.MAX_LONG_STRING_VALUE_LENGTH];
+                bodyReadBuffer.set(data);
             }
             int length = bodyIS.read(data, 0, data.length);
 
--- a/apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
+++ b/apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
@@ -1,0 +1,59 @@
+/*-
+ * #%L
+ * Elastic APM Java agent
+ * %%
+ * Copyright (C) 2018 Elastic and contributors
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package co.elastic.apm.es.restclient;
+
+import co.elastic.apm.bci.VisibleForAdvice;
+import co.elastic.apm.report.serialize.DslJsonSerializer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.io.InputStream;
+
+@VisibleForAdvice
+public class ESRestClientInstrumentationHelper {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ESRestClientInstrumentationHelper.class);
+
+    private static ThreadLocal<byte[]> bodyReadBuffer = new ThreadLocal<>();
+
+    @Nullable
+    @VisibleForAdvice
+    public static String readRequestBody(InputStream bodyIS, String endpoint) throws IOException {
+        String body = null;
+        try {
+            byte[] data = bodyReadBuffer.get();
+            if (data == null) {
+                bodyReadBuffer.set(new byte[DslJsonSerializer.MAX_LONG_STRING_VALUE_LENGTH]);
+                data = bodyReadBuffer.get();
+            }
+            int length = bodyIS.read(data, 0, data.length);
+
+            // Using platform's default charset for decoding
+            body = new String(data, 0, length);
+        } catch (IOException e) {
+            LOGGER.info("Failed to read request body for " + endpoint);
+        } finally {
+            bodyIS.close();
+        }
+
+        return body;
+    }
+}
--- a/apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
+++ b/apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
@@ -31,7 +31,7 @@ import java.nio.charset.StandardCharsets;
 
 @VisibleForAdvice
 public class ESRestClientInstrumentationHelper {
-    private static final Logger LOGGER = LoggerFactory.getLogger(ESRestClientInstrumentationHelper.class);
+    private static final Logger logger = LoggerFactory.getLogger(ESRestClientInstrumentationHelper.class);
 
     private static ThreadLocal<byte[]> bodyReadBuffer = new ThreadLocal<>();
     private static final byte CURLY_BRACKET_UTF8 = '{';
@@ -55,7 +55,7 @@ public class ESRestClientInstrumentationHelper {
                 body = new String(data, 0, length, StandardCharsets.UTF_8);
             }
         } catch (IOException e) {
-            LOGGER.info("Failed to read request body for " + endpoint);
+            logger.info("Failed to read request body for " + endpoint);
         } finally {
             bodyIS.close();
         }
--- a/apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
+++ b/apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
@@ -21,23 +21,24 @@ package co.elastic.apm.es.restclient;
 
 import co.elastic.apm.bci.VisibleForAdvice;
 import co.elastic.apm.report.serialize.DslJsonSerializer;
-import org.elasticsearch.common.io.stream.InputStreamStreamInput;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 
 @VisibleForAdvice
 public class ESRestClientInstrumentationHelper {
     private static final Logger LOGGER = LoggerFactory.getLogger(ESRestClientInstrumentationHelper.class);
 
     private static ThreadLocal<byte[]> bodyReadBuffer = new ThreadLocal<>();
+    private static final byte CURLY_BRACKET_UTF8 = '{';
 
     @Nullable
     @VisibleForAdvice
-    public static String readRequestBody(InputStream bodyIS, long contentLength, String endpoint) throws IOException {
+    public static String readRequestBody(InputStream bodyIS, String endpoint) throws IOException {
         String body = null;
         try {
             byte[] data = bodyReadBuffer.get();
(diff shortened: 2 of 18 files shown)


apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -0,0 +1,121 @@
...
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+
+/**
+ * Instrumentation for Elasticsearch RestClient, currently supporting only synchronized queries.
+ * All sync operations go through org.elasticsearch.client.RestClient#performRequest(org.elasticsearch.client.Request)
+ */
+public class ElasticsearchRestClientInstrumentation extends ElasticApmInstrumentation {
+    private static final String ES_REST_CLIENT_INSTRUMENTATION_GROUP = "elasticsearch-restclient";
+    private static final String SPAN_TYPE = "es-restclient";
```
felixbarny: We should align the type with other agents.

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -0,0 +1,121 @@
...
+    @Advice.OnMethodEnter
+    private static void onBeforeExecute(@Advice.Argument(0) Request request,
+                                        @Advice.Local("span") Span span,
+                                        @Advice.Local("esre") ResponseException esre) {
+        if (tracer == null || tracer.getActive() == null) {
+            return;
+        }
+        span = tracer.getActive().createSpan()
+            .withType(SPAN_TYPE)
+            .appendToName(request.getEndpoint()).appendToName(" ").appendToName(request.getMethod())
```
felixbarny: we should align the span.name with other agents

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -0,0 +1,121 @@
...
+            .activate();
+
+
+        // Add request parameters to query info
+        for (Map.Entry<String, String> param: request.getParameters().entrySet()) {
+            span.addTag(param.getKey(), param.getValue());
```
felixbarny: That could be problematic, as the param keys could have a high cardinality, leading to a mapping explosion as all tags are indexed. See https://www.elastic.co/blog/found-crash-elasticsearch, https://www.elastic.co/guide/en/elasticsearch/reference/6.4/mapping.html

apm-agent-plugins/apm-es-restclient-plugin/src/test/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentationIT.java
```diff
@@ -0,0 +1,257 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static co.elastic.apm.es.restclient.ElasticsearchRestClientInstrumentation.ELASTICSEARCH_NODE_KEY;
+import static co.elastic.apm.es.restclient.ElasticsearchRestClientInstrumentation.ERROR_REASON_KEY;
+import static co.elastic.apm.es.restclient.ElasticsearchRestClientInstrumentation.QUERY_STATUS_CODE_KEY;
+import static org.elasticsearch.common.xcontent.XContentFactory.jsonBuilder;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
felixbarny: Why is the method order important? Ideally, the tests should also start with a clean state so that they can be executed individually. This also makes fixing a failing test easier, as you don't have to understand what's happening in the previous tests.

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
felixbarny: Tags are mainly meant for user-defined metadata. The information added by agents are usually added to the JSON schema, which makes it easier to align on what kind of information should be collected across agents.

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
felixbarny: see https://docs.google.com/spreadsheets/d/1SmWeX5AeqUcayrArUauS_CxGgsjwRgMYH4ZY8yQsMhQ

watson:
For reference, the node agent [currently collects](https://github.com/elastic/apm-agent-nodejs/blob/36ad3157998a3d569fbfb0bedb3b8277c9d57ec5/lib/instrumentation/modules/elasticsearch.js#L26-L33) only the query, http method and url. And we only log `_search` query bodies (as we don’t want to store inserts)

apm-agent-plugins/apm-es-restclient-plugin/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+        <elasticsearch.version>6.4.1</elasticsearch.version>
+    </properties>
+
+    <dependencies>
+        <!-- Elasticsearch rest client -->
+        <dependency>
+            <groupId>org.elasticsearch.client</groupId>
+            <artifactId>elasticsearch-rest-client</artifactId>
+            <version>${elasticsearch.version}</version>
+        </dependency>
```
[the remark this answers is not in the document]
eyalkoren: Right! Adding to my es-plugin-showcase blog post notes right away :)

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Instrumentation for Elasticsearch RestClient, currently supporting only synchronized queries.
+ * All sync operations go through org.elasticsearch.client.RestClient#performRequest(org.elasticsearch.client.Request)
+ */
+public class ElasticsearchRestClientInstrumentation extends ElasticApmInstrumentation {
+    private static final String ES_REST_CLIENT_INSTRUMENTATION_GROUP = "elasticsearch-restclient";
+    private static final String SPAN_TYPE = "es-restclient";
+
+    private static final String REST_CLIENT_CLASS_NAME = "org.elasticsearch.client.RestClient";
```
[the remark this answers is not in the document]
eyalkoren: OK, let's agree on that

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -0,0 +1,121 @@
...
+    private static final String PERFORM_REQUEST_METHOD_NAME = "performRequest";
+
+    static final String ELASTICSEARCH_NODE_KEY = "Elasticsearch-node";
+    static final String QUERY_STATUS_CODE_KEY = "Query-status-code";
+    static final String ERROR_REASON_KEY = "Error-reason";
+
+    @Advice.OnMethodEnter
+    private static void onBeforeExecute(@Advice.Argument(0) Request request,
+                                        @Advice.Local("span") Span span,
+                                        @Advice.Local("esre") ResponseException esre) {
```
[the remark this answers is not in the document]
eyalkoren: Did this based on `net.bytebuddy.asm.Advice.Local`'s javadoc, which says:
> ... if a local variable is referenced from an exit advice method, it must also be declared by an enter advice method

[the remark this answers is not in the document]
eyalkoren: Maybe it is needed for correct instrumentation of the stack frame directives

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
eyalkoren: Sure, as I said in the top comment, this is only to show what we have available. Only question is whether we want to make it configurable so that a user can opt to get those collected

apm-agent-plugins/apm-es-restclient-plugin/src/test/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentationIT.java
eyalkoren: Makes sense. The reason for ordering was that I wanted to use each unit test as a small test for a single or few client actions, so I wanted to leave the index creation and deletion out of it. What I can do to accommodate both purposes is for each test:
1. create the index
2. reset the reporter
3. test my action
4. delete the index

eyalkoren:
@watson thanks for referencing. I see that you set `db.context.type` to `elasticsearch`, but can you give me an example to what `db.context.statement` looks like?
Also, I am not sure I understood correctly - do you log full `_search` query bodies?

apm-agent-plugins/apm-es-restclient-plugin/pom.xml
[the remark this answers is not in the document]
felixbarny: looking forward to that :)

apm-agent-plugins/apm-es-restclient-plugin/src/test/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentationIT.java
felixbarny: just as an idea: you could also do a `DELETE _all` in an `@After` method

watson:
@eyalkoren correct, we log the full `_search` query. We currently do not try to filter it. We do allow the users to apply custom filters where they theoretically can perform filtering of the queries, but that would be a pretty advanced filter.

If I recall correctly, if the user performs this query via the Node.js Elasticsearch client:

```js
const response = await client.search({
  q: 'pants'
})
```

The the content of `db.context.statement` would be `{"q":"pants"}`.

eyalkoren:
@watson how do you identify searches? Is it by looking at the path (`path.endsWith('_search')` or `path.contains('_search')`) or is it by monitoring the corresponding client API used only for searches?
I am asking because the latter seems to be a good option, but it requires me to instrument another method, so I would really like to avoid that. I am wondering whether looking at the path is good enough.

Regardless, would the `db.context.statement` be left empty for other operations?

watson:
@eyalkoren All the logic is contained in [these few lines](https://github.com/elastic/apm-agent-nodejs/blob/36ad3157998a3d569fbfb0bedb3b8277c9d57ec5/lib/instrumentation/modules/elasticsearch.js#L26-L33). I check that the path ends with `_search`. We need to monitor all client API's no matter if they are used for searches or not. Only difference is that the ones that are used for searches will have the query logged as well. There's no good way to do that in the Node.js ES client at the moment besides just looking at the path. Other clients might be different though.

But I'm not aware of any downsides of looking just at the path?

Yes we leave out `db.context.statement` if it's not a `_search` query. In fact we also leave out the entire `db` object.

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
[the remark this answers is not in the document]
eyalkoren: Sorry, I am missing something here.
> This is only needed if you want to add a @Local to the exit advice

We are inlining our advice methods, which means that there are no such methods in runtime- their compiled bytecode is injected into the instrumented method. The signature of our advice method is kind of meaningless as nobody invokes them.

> not when creating local variables which are only scoped to the exit advice

This is a directive for BB to create a local variable, I assume for the purpose of properly maintaining stack frame directives for the bytecode verifier. So both entry and exit advices are scoped together under the same method they are injected into because they are both using the same operand stack and local variable array. I believe this is the reason for this requirement.

felixbarny:
I’m not sure that’s the case. Have you tried without @Local?

eyalkoren:
@felixbarny please just quickly review only the changes at [the second commit](https://github.com/elastic/apm-agent-java/pull/228/commits/d58ee2dd882a9305ca6f153d707ce90f3262b87a) when got the time.

No rush, though, as I still have the following stuff to finish:

1. Make `_search` body capture configurable (on by default)
2. Decide what to do with the status code and url - tags? add to schema? use both HTTP and DB for the same span?
3. Figure out why the IT works locally but not on Jenkins (maybe the port it uses cannot become available?)

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
```diff
@@ -0,0 +1,59 @@
...
+    private static ThreadLocal<byte[]> bodyReadBuffer = new ThreadLocal<>();
+
+    @Nullable
+    @VisibleForAdvice
+    public static String readRequestBody(InputStream bodyIS, String endpoint) throws IOException {
+        String body = null;
+        try {
+            byte[] data = bodyReadBuffer.get();
+            if (data == null) {
+                bodyReadBuffer.set(new byte[DslJsonSerializer.MAX_LONG_STRING_VALUE_LENGTH]);
```
felixbarny: That limits the number of bytes as opposed to the number of characters. In case of non-ascii characters, there is a difference. Although, for simplicity's sake it probably still makes sense the way you do it. Even if it means there could be less than 10k chars, but it might be worth adding a comment about that.

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
```diff
@@ -0,0 +1,59 @@
...
+                data = bodyReadBuffer.get();
+            }
+            int length = bodyIS.read(data, 0, data.length);
+
+            // Using platform's default charset for decoding
+            body = new String(data, 0, length);
```
felixbarny: One idea how we might be able to avoid allocations altogether is to use a `CharBuffer statementBuffer` field in `co.elastic.apm.impl.transaction.Db`. Together with [`CharsetDecoder#decode`](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/CharsetDecoder.html#decode(java.nio.ByteBuffer,%20java.nio.CharBuffer,%20boolean)) and your thread-local byte array buffer, we would never create garbage as we can recycle the `CharBuffer` via the `rewind` method as well as the thread-local ByteBuffer. `CharBuffer` also implements the `CharSequence` interface which DSL JSON can directly serialize.

felixbarny: If you prefer keeping the first version simple, we can also revisit that idea later, after merging the PR. This is just meant as an idea 🙂

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
eyalkoren: Well, that's the least of problems here- if the binary data was encoded using a charset other than the platform default, then I would mess it up when decoding it to a String. I wrote the proper handling for dealing with that, which was to get the content-type header from the request, parse it, see if it contains a charset and use it for decoding, but only then I realized that it will be messed up during JSON serialization anyway, so I dropped it for now.

Anyway, I guess the main issue here is size limit, not a specific number, so 10k has no special meaning and I believe it achieves its purpose.

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
eyalkoren: That's a good idea. Let's talk about what to do with locating the proper charset and encoding/decoding because I think that's a bigger issue

eyalkoren:
The [new commit](https://github.com/elastic/apm-agent-java/pull/228/commits/6d21bca98ab3c5e4c867b18e4a10010d8533d15e) demonstrates how it would look like to try and decode properly, more or less. I don't like it but that's what I have at them moment..

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
felixbarny: I'd assume that the ES RestClient uses a fixed encoding, probably UTF-8. Making that assumption very clear, for example by naming the method `readRequestBodyUtf8`, I think It's probably good enough to hard-code the encoding rather than to determine it at runtime.

> but only then I realized that it will be messed up during JSON serialization anyway

not quite sure I understand

eyalkoren: > I'd assume that the ES RestClient uses a fixed encoding, probably UTF-8

Yeah, you are probably right, I treated it as if it's a general-purpose HTTP client, but it doesn't really make sense

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -0,0 +1,138 @@
...
+    @Advice.OnMethodEnter
+    private static void onBeforeExecute(@Advice.Argument(0) Request request,
+                                        @Advice.Local("span") Span span) {
+        if (tracer == null || tracer.getActive() == null) {
+            return;
+        }
+        span = tracer.getActive().createSpan()
+            .withType(SPAN_TYPE)
+            .appendToName("Elasticsearch: ").appendToName(request.getMethod()).appendToName(" ").appendToName(request.getEndpoint())
+            .activate();
```
felixbarny: Do we need to activate the span? Does the user even have the chance to get this span in an interceptor or something?

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -0,0 +1,138 @@
...
+
+        if (request.getEndpoint().endsWith(SEARCH_QUERY_PATH_SUFFIX)) {
```
felixbarny: add `&& span.isSampled()` to avoid overhead on non sampled spans

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
eyalkoren: That's a state setting, so I figure it is always better to maintain it properly. Even your current MDC logging integration relies on current span being activated, doesn't it?

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
eyalkoren: Actually, shouldn't this check be:
```
        if (tracer == null || tracer.getActive() == null || !tracer.getActive().isSampled()) {
            return;
        }
``` 
at the beginning of this method?

felixbarny: Good point. It should probably be both. In case the span limit (default 500 per transaction) is exceeded, the parent might be sampled but the span not.

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
eyalkoren: Looked into doing that, but Db is instantiated per Span, and there can be **A LOT** of those, potentially most or all of which are not DB span. This may lead to big allocation of constant memory, which can be worse than reallocation of recyclable memory. If we really want to do that properly, we need to pool those separately. 
Leaving it as it is at the moment.

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
```diff
@@ -0,0 +1,65 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+
+@VisibleForAdvice
+public class ESRestClientInstrumentationHelper {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ESRestClientInstrumentationHelper.class);
```
felixbarny: use lower case to be consistent with the conventions in the project

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private static ThreadLocal<byte[]> bodyReadBuffer = new ThreadLocal<>();
+    private static final byte CURLY_BRACKET_UTF8 = '{';
```
felixbarny: I think it's called curly brace

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return named("org.elasticsearch.client.RestClient");
+    }
+
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return named("performRequest")
```
felixbarny: Can be replaces with
```
named("performRequest")
    .and(takesArguments(named("org.elasticsearch.client.Request")))
```

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -0,0 +1,147 @@
...
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+
+/**
+ * Instrumentation for Elasticsearch RestClient, currently supporting only synchronized queries.
+ * All sync operations go through org.elasticsearch.client.RestClient#performRequest(org.elasticsearch.client.Request)
+ */
+public class ElasticsearchRestClientInstrumentation extends ElasticApmInstrumentation {
```
felixbarny: some minor formatting issues in this class, maybe just re-format again

apm-agent-plugins/apm-es-restclient-plugin/src/test/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentationIT.java
```diff
@@ -0,0 +1,305 @@
...
+    private void validateSpanContent(Span span, String expectedName, int statusCode, String method) {
+        assertThat(span.getType()).isEqualTo(SPAN_TYPE);
+        assertThat(span.getName().toString()).isEqualTo(expectedName);
+        validateHttpContextContent(span.getContext().getHttp(), statusCode, method);
+
+        assertThat(span.getContext().getDb().getType()).isEqualTo(DB_CONTEXT_TYPE);
+
+        if (expectedName.contains(SEARCH_QUERY_PATH_SUFFIX)) {
+            assertThat(span.getContext().getDb().hasContent()).isTrue();
+        }
```
felixbarny:
```
assertThat(span.getContext().getDb().hasContent()).isEqualTo(expectedName.contains(SEARCH_QUERY_PATH_SUFFIX))
```

to verify that there is not db content for non-search requests

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -0,0 +1,147 @@
...
+                    url = response.getHost().toURI().toString();
+                    statusCode = response.getStatusLine().getStatusCode();
+                } else if(t != null) {
+                    if (t instanceof ResponseException) {
+                        ResponseException esre = (ResponseException) t;
+                        url = esre.getResponse().getHost().toURI().toString();
+                        statusCode = esre.getResponse().getStatusLine().getStatusCode();
+
+                        // Add tags so that they will be copied to error capture
+                        span.addTag(QUERY_STATUS_CODE_KEY, Integer.toString(statusCode));
```
felixbarny: what is the difference to `span.context.http.status_code`?

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -0,0 +1,147 @@
...
+                        span.addTag(ELASTICSEARCH_NODE_URL_KEY, url);
+                        span.addTag(ERROR_REASON_KEY, esre.getResponse().getStatusLine().getReasonPhrase());
```
felixbarny: Does that yield valuable information which is not in the exception message? What are examples of the reason phrase vs the exception message?

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -0,0 +1,147 @@
...
+                    statusCode = response.getStatusLine().getStatusCode();
+                } else if(t != null) {
+                    if (t instanceof ResponseException) {
+                        ResponseException esre = (ResponseException) t;
+                        url = esre.getResponse().getHost().toURI().toString();
+                        statusCode = esre.getResponse().getStatusLine().getStatusCode();
+
+                        // Add tags so that they will be copied to error capture
+                        span.addTag(QUERY_STATUS_CODE_KEY, Integer.toString(statusCode));
+                        span.addTag(ELASTICSEARCH_NODE_URL_KEY, url);
```
felixbarny: isn't that already covered with `span.context.http.url`?

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ESRestClientInstrumentationHelper.java
eyalkoren: I checked, seems to be valid both ways

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
eyalkoren: No, it can't. Did you mean I can remove the num of arguments check?

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
eyalkoren: Didn't find any...Try to select all and apply automatically - didn't find..

apm-agent-plugins/apm-es-restclient-plugin/src/main/java/co/elastic/apm/es/restclient/ElasticsearchRestClientInstrumentation.java
eyalkoren: That was a way of transferring to the error from tag, but irrelevant anymore...

PR #231, opened 2018-10-02
Change span type to db.<db-provider>.query

eyalkoren:


apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/jdbc/helper/JdbcHelperImpl.java
```diff
@@ -51,7 +53,7 @@ public Span createJdbcSpan(@Nullable String sql, Connection connection, @Nullabl
         // getting the meta data can result in another jdbc call
         // if that is traced as well -> StackOverflowError
         // to work around that, isAlreadyMonitored checks if the parent span is a db span and ignores them
-        span.withType("db.unknown.sql");
+        span.withType(computeJdbcSpanTypeName("unknown"));
```
felixbarny: store as a constant to not allocate a new string each time

felixbarny:
Jenkins, test this please

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/jdbc/helper/JdbcHelperImpl.java
eyalkoren: Right, thanks!

PR #232, opened 2018-10-02
Use microsecond epoch timestamps

felixbarny:
- Prepares for https://github.com/elastic/apm-server/issues/1340
- Includes guards against clock drift

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/EpochTickClock.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/EpochTickClock.java
@@ -30,7 +30,7 @@ public class EpochTickClock implements Recyclable {
     }
 
     public void init(long epochMillis, long nanoTime) {
-        nanoTimeOffsetToEpoch = nanoTime - TimeUnit.MILLISECONDS.toNanos(epochMillis);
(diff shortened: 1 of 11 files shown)


felixbarny:
This can be merged before the APM Server changes as this PR does not change the JSON structure yet. Once the Server is ready to accept an epoch microsecond timestamp, the agent just needs some small adjustments.

apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/EpochTickClock.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Initializes and calibrates the clock based on wall clock time
+     */
+    public void init() {
+        init(System.currentTimeMillis(), System.nanoTime());
+    }
+
+    public void init(long epochMillis, long nanoTime) {
+        nanoTimeOffsetToEpoch = nanoTime - TimeUnit.MILLISECONDS.toNanos(epochMillis);
```
eyalkoren: That's correct, but got me confused, as it is not really an offset (it yields a negative value). No need to change for correctness, but I would say it is more readable if you reverse the subtraction order to get an offset to epoch, and then add it to the nano time in [getEpochMicros](https://github.com/elastic/apm-agent-java/pull/232/files#diff-dcb41b5d39f04a99247e0d7aad3ab48bR41) instead of subtracting

PR #237, opened 2018-10-06
Make successive calls to Span#end a noop

felixbarny:
this is required by the OT API

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Span.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Span.java
@@ -59,6 +59,7 @@ public class Span extends AbstractSpan<Span> implements Recyclable {
     }
 
     public Span start(Transaction transaction, @Nullable Span parentSpan, long epochMicros, boolean dropped) {
+        this.finished = false;
         this.transaction = transaction;
         this.clock.init(transaction.clock);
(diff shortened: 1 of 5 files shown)


apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Span.java
```diff
@@ -59,6 +59,7 @@ public Span(ElasticApmTracer tracer) {
     }
 
     public Span start(Transaction transaction, @Nullable Span parentSpan, long epochMicros, boolean dropped) {
+        this.finished = false;
```
eyalkoren: I think this better go into a final non-abstract `start` method in AbstractSpan, like you did with `end`, and the rest can be an implementation of an abstract method defined in super (like `doEnd`)

apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Span.java
```diff
@@ -84,6 +85,7 @@ private Span start(TraceContext parent, long epochMicros, boolean dropped) {
     }
 
     public Span startNoop() {
+        finished = false;
```
eyalkoren: Same comment goes here. This way, `finished` context can be limited to be AbstractSpan private

eyalkoren:
Just one small encapsulation suggestion

PR #240, opened 2018-10-08
Use epoch micros for intake v2

felixbarny:
see elastic/apm-server#1413


felixbarny:
After we only support v2, we should improve the JSON schema tests.

felixbarny:
That was initially the plan, which is why I introduced two distinct configuration options but v2 only supports DT. So in practice both flags would be either true or false at the same time.

eyalkoren:
After we move to using V2, wouldn't DT still remain configurable? When the DT flag is `false` we are using V1.

felixbarny:
After we move to using V2, wouldn't DT still remain configurable? No, because the APM Server only supports the DT for mat for intake v2..

PR #242, opened 2018-10-09
Make it possible to start spans after parent has ended

felixbarny:


apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Span.java
```diff
@@ -153,6 +164,16 @@ public void resetState() {
         traceContext.resetState();
     }
 
+    /**
+     * Returns the {@link Transaction} which this span belongs to.
+     * <p>
+     * Can can return {@code null} in case the span has been started via
```
eyalkoren: Typo: one too many "can" at the beginning of sentence

PR #244, opened 2018-10-16
Store active transaction in its own ThreadLocal

felixbarny:
Deprecate Span#getTransaction (see javadoc)
This change reduces the usage of Span#getTransaction


eyalkoren:
May I suggest a single ThreadLocal Stack of `AbstractSpan`s where the bottom (index 0) is the transaction and the top is the current active span? This way you can always know the full state of ALL open spans.

felixbarny:
That is currently not something we have a use case for but if we have one at a later point, we can always add that behavior. The advantage of this PR is that we don't have to traverse the stack in order to get the transaction. So even if we keep the span stack, we might still want to have a dedicated thread local for the transaction.

eyalkoren:
No need to traverse the stack - just get the one in index 0...
The current use case is having a single ThreadLocal to represent the entire state instead of two that represent a partial state

PR #250, opened 2018-10-18
Keeping PreparedStatements in a WeakHashMap until GCed

eyalkoren:


felixbarny:
How big is the resulting jar now? Is it possible to only take the parts we need from guava?

felixbarny:
Could we for example use [excludes](https://maven.apache.org/plugins/maven-shade-plugin/examples/includes-excludes.html) to exclude parts of guava which are not interesting for us, like `com/google/common/{primitives,escape,graph,eventbus,math,cache,html,io,net,xml}`

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/jdbc/PreparedStatementInstrumentation.java
```diff
@@ -62,7 +62,7 @@
     @Advice.OnMethodEnter
     public static Span onBeforeExecute(@Advice.This PreparedStatement statement) throws SQLException {
         if (tracer != null && jdbcHelper != null) {
-            final String sql = ConnectionInstrumentation.getSqlForStatement(statement);
+            final @Nullable String sql = ConnectionInstrumentation.getSqlForStatement(statement);
```
felixbarny: I think the `@Nullable` here is not necessary

eyalkoren: It's not, but I thought it is a good hint/directive for anyone looking at this code

eyalkoren:
> Weak references come with a cost so it might make sense to remove the entries when Statement#close has been called instead of relying on the GC to clear them.

The cost of weak references is proportional **to the amount of such created**, so I don't believe there is a benefit in maintaining a lower number of them in average. In addition, as I wrote in the task- there is a big cost of actively removing each reference from the map, as this is a write operation on a concurrent map. Relatively, the bulk removal of stale entries done by the map itself seems a lot cheaper.
Since this is also the simpler solution (less instrumentation, less logic), I would opt to this way.

eyalkoren:
> Could we for example use excludes to exclude parts of guava which are not interesting for us, like com/google/common/{primitives,escape,graph,eventbus,math,cache,html,io,net,xml}?

I am pretty sure we can do that. Do we do that with other dependencies (depend on parts of a project/module)? It may probably be more vulnerable to version upgrades, but it's not a big deal

eyalkoren:
> How big is the resulting jar now? Is it possible to only take the parts we need from guava?

Agent jar increased from 4.9MB to 8MB- quite a big difference...

eyalkoren:
Currently only shaded the 'checkframework' stuff.
Started to look into minimizing the jar size. Specifically, the `<minimizeJar>` of the `maven-shade-plugin` looks like an interesting option, but this seems unrelated to this PR.

PR #251, opened 2018-10-18
OT: preserve SpanContext after finish

### LINKED ISSUE #179: Span.context has to be available after Span.finish
Due to object pooling, this is currently not the case

felixbarny:
Adds ability to activate a serialized TraceContext
Enables context propagation in async scenarios
Introduces ChildContextCreator abstraction

TODO javadoc

fixes elastic/apm-agent-java#179

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/impl/ElasticApmTracer.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/impl/ElasticApmTracer.java
@@ -34,6 +34,7 @@ import co.elastic.apm.objectpool.RecyclableObjectFactory;
 import co.elastic.apm.objectpool.impl.QueueBasedObjectPool;
 import co.elastic.apm.report.Reporter;
 import co.elastic.apm.report.ReporterConfiguration;
+import co.elastic.apm.util.HexUtils;
 import org.jctools.queues.atomic.AtomicQueueFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -67,10 +68,10 @@ public class ElasticApmTracer {
     // Maintains a stack of all the activated spans
     // This way its easy to retrieve the bottom of the stack (the transaction)
     // Also, the caller does not have to keep a reference to the previously active span, as that is maintained by the stack
-    private final ThreadLocal<Deque<AbstractSpan<?>>> activeStack = new ThreadLocal<Deque<AbstractSpan<?>>>() {
+    private final ThreadLocal<Deque<Object>> activeStack = new ThreadLocal<Deque<Object>>() {
         @Override
-        protected Deque<AbstractSpan<?>> initialValue() {
-            return new ArrayDeque<AbstractSpan<?>>();
+        protected Deque<Object> initialValue() {
+            return new ArrayDeque<Object>();
         }
     };
     private final CoreConfiguration coreConfiguration;
@@ -123,19 +124,19 @@ public class ElasticApmTracer {
     }
 
     public Transaction startTransaction() {
-        return startTransaction(null);
+        return startTransaction(TraceContext.asRoot(), null);
     }
 
-    public Transaction startTransaction(@Nullable String traceContextHeader) {
-        return startTransaction(traceContextHeader, sampler, -1);
+    public <T> Transaction startTransaction(TraceContext.ChildContextCreator<T> childContextCreator, @Nullable T parent) {
+        return startTransaction(childContextCreator, parent, sampler, -1);
     }
 
-    public Transaction startTransaction(@Nullable String traceContextHeader, Sampler sampler, long epochMicros) {
+    public <T> Transaction startTransaction(TraceContext.ChildContextCreator<T> childContextCreator, @Nullable T parent, Sampler sampler, long epochMicros) {
         Transaction transaction;
         if (!coreConfiguration.isActive()) {
             transaction = noopTransaction();
         } else {
-            transaction = transactionPool.createInstance().start(traceContextHeader, epochMicros, sampler);
+            transaction = transactionPool.createInstance().start(childContextCreator, parent, epochMicros, sampler);
         }
         if (logger.isDebugEnabled()) {
             logger.debug("startTransaction {} {", transaction);
@@ -153,7 +154,7 @@ public class ElasticApmTracer {
 
     @Nullable
     public Transaction currentTransaction() {
-        final AbstractSpan<?> bottomOfStack = activeStack.get().peekFirst();
+        final Object bottomOfStack = activeStack.get().peekFirst();
         if (bottomOfStack instanceof Transaction) {
             return (Transaction) bottomOfStack;
         }
@@ -162,7 +163,7 @@ public class ElasticApmTracer {
 
     @Nullable
     public Span currentSpan() {
-        final AbstractSpan<?> abstractSpan = getActive();
+        final AbstractSpan<?> abstractSpan = activeSpan();
         if (abstractSpan instanceof Span) {
             return (Span) abstractSpan;
         }
@@ -178,18 +179,18 @@ public class ElasticApmTracer {
      * @param parentContext the trace context of the parent
      * @return a new started span
      */
-    public Span startSpan(TraceContext parentContext) {
-        return spanPool.createInstance().start(parentContext);
+    public <T> Span startSpan(TraceContext.ChildContextCreator<T> childContextCreator, T parentContext) {
+        return spanPool.createInstance().start(childContextCreator, parentContext);
     }
 
     /**
      * @param parentContext the trace context of the parent
      * @param epochMicros   the start timestamp of the span in microseconds after epoch
      * @return a new started span
-     * @see #startSpan(TraceContext)
+     * @see #startSpan(TraceContext.ChildContextCreator, Object)
      */
-    public Span startSpan(TraceContext parentContext, long epochMicros) {
-        return spanPool.createInstance().start(parentContext, epochMicros);
+    public <T> Span startSpan(TraceContext.ChildContextCreator<T> childContextCreator, T parentContext, long epochMicros) {
+        return spanPool.createInstance().start(childContextCreator, parentContext, epochMicros);
     }
 
     public Span startSpan(AbstractSpan<?> parent, long epochMicros) {
@@ -208,13 +209,6 @@ public class ElasticApmTracer {
             return createNoopSpan();
         } else {
             span = createRealSpan(transaction, parentSpan, epochMicros);
-        }
-        if (logger.isDebugEnabled()) {
-            logger.debug("startSpan {} {", span);
-            if (logger.isTraceEnabled()) {
-                logger.trace("starting span at",
-                    new RuntimeException("this exception is just used to record where the span has been started from"));
-            }
         }
         return span;
     }
@@ -235,7 +229,7 @@ public class ElasticApmTracer {
             dropped = false;
             transaction.getSpanCount().getStarted().incrementAndGet();
         }
-        span.start(transaction, parentSpan, epochMicros, dropped);
+        span.start(transaction, TraceContext.fromParentSpan(), parentSpan != null ? parentSpan : transaction, epochMicros, dropped);
         return span;
     }
 
@@ -244,7 +238,7 @@ public class ElasticApmTracer {
     }
 
     public void captureException(@Nullable Throwable e) {
-        captureException(System.currentTimeMillis(), e, getActive());
+        captureException(System.currentTimeMillis(), e, activeSpan());
     }
 
     public void captureException(long epochTimestampMillis, @Nullable Throwable e, @Nullable AbstractSpan<?> active) {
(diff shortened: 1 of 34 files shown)


apm-agent-plugins/apm-opentracing-plugin/src/main/java/co/elastic/apm/opentracing/impl/ApmSpanInstrumentation.java
```diff
@@ -74,34 +73,48 @@ public FinishInstrumentation() {
             super(named("finishInternal"));
         }
 
+        @Advice.OnMethodEnter
+        private static void finishInternal(@Advice.FieldValue(value = "dispatcher", readOnly = false) @Nullable Object dispatcher,
+                                           @Advice.Argument(0) long finishMicros) {
+            final Object newDispatcher = doFinishInternal(dispatcher, finishMicros);
+            if (newDispatcher != null) {
+                dispatcher = newDispatcher;
```
eyalkoren: I still think this doesn't solve the async problem. Consider this scenario:
Thread A starts a transaction, then submits a task to a queue, then the task starts being executed on thread B (original transaction is still running on thread A). Now thread B wants to do something, for example - set the operation name:
```
Thread B:
----------
    apmTracer.activeSpan().setOperationName("opName");
        ApmSpanInstrumentation$SetOperationName.setOperationName()
            if (dispatcher instanceof AbstractSpan) {     // returns true
Thread A:
----------
    apmTracer.activeSpan().finish();
        ApmSpanInstrumentation.FinishInstrumentation.finishInternal()
            ApmSpanInstrumentation.FinishInstrumentation.doFinishInternal()
                span.end()     // potentially recycled by thread A or even thread C (disruptor thread)
        dispatcher = newDispatcher;    // this line - replace the underlying dispatcher from span to serialized context
Thread B:
----------
            // continue execution
            AbstractSpan<?> span = (AbstractSpan) dispatcher;    // still the reference to the span
            span.setName(operationName);    // on a potentially recycled span
```

or this scenario:
```
Thread A:
----------
    apmTracer.activeSpan().finish();
        ApmSpanInstrumentation.FinishInstrumentation.finishInternal()
            ApmSpanInstrumentation.FinishInstrumentation.doFinishInternal()
                span.end()     // potentially recycled by thread A or even thread C (disruptor thread)
Thread B:
----------
    apmTracer.activeSpan().setOperationName("opName");
        ApmSpanInstrumentation$SetOperationName.setOperationName()
            if (dispatcher instanceof AbstractSpan) {   // returns true as the dispatcher ref was not yet replaced
                AbstractSpan<?> span = (AbstractSpan) dispatcher;
                span.setName(operationName);    // on a potentially recycled span
Thread A:
----------
        // continue execution
        dispatcher = newDispatcher;    // this line - only now replacing the underlying dispatcher, too late
```

eyalkoren: I still think this doesn't solve the async problem. Consider this scenario:
Thread A starts a transaction, then submits a task to a queue, then the task starts being executed on thread B (original transaction is still running on thread A). Now thread B wants to do something, for example - set the operation name:
```
Thread B:
----------
    apmTracer.activeSpan().setOperationName("opName");
        ApmSpanInstrumentation$SetOperationName.setOperationName()
            if (dispatcher instanceof AbstractSpan) {     // returns true
Thread A:
----------
    apmTracer.activeSpan().finish();
        ApmSpanInstrumentation.FinishInstrumentation.finishInternal()
            ApmSpanInstrumentation.FinishInstrumentation.doFinishInternal()
                span.end()     // potentially recycled by thread A or even thread C (disruptor thread)
        dispatcher = newDispatcher;    // this line - replace the underlying dispatcher from span to serialized context
Thread B:
----------
            // continue execution
            AbstractSpan<?> span = (AbstractSpan) dispatcher;    // still the reference to the span
            span.setName(operationName);    // on a potentially recycled span
```

or this scenario:
```
Thread A:
----------
    apmTracer.activeSpan().finish();
        ApmSpanInstrumentation.FinishInstrumentation.finishInternal()
            ApmSpanInstrumentation.FinishInstrumentation.doFinishInternal()
                span.end()     // potentially recycled by thread A or even thread C (disruptor thread)
Thread B:
----------
    apmTracer.activeSpan().setOperationName("opName");
        ApmSpanInstrumentation$SetOperationName.setOperationName()
            if (dispatcher instanceof AbstractSpan) {   // returns true as the dispatcher ref was not yet replaced
                AbstractSpan<?> span = (AbstractSpan) dispatcher;
                span.setName(operationName);    // on a potentially recycled span
Thread A:
----------
        // continue execution
        dispatcher = newDispatcher;    // this line - only now replacing the underlying dispatcher, too late
```

eyalkoren: I still think this doesn't solve the async problem. Consider this scenario:
Thread A starts a transaction, then submits a task to a queue, then the task starts being executed on thread B (original transaction is still running on thread A). Now thread B wants to do something, for example - set the operation name:
```
Thread B:
----------
    apmTracer.activeSpan().setOperationName("opName");
        ApmSpanInstrumentation$SetOperationName.setOperationName()
            if (dispatcher instanceof AbstractSpan) {     // returns true
Thread A:
----------
    apmTracer.activeSpan().finish();
        ApmSpanInstrumentation.FinishInstrumentation.finishInternal()
            ApmSpanInstrumentation.FinishInstrumentation.doFinishInternal()
                span.end()     // potentially recycled by thread A or even thread C (disruptor thread)
        dispatcher = newDispatcher;    // this line - replace the underlying dispatcher from span to serialized context
Thread B:
----------
            // continue execution
            AbstractSpan<?> span = (AbstractSpan) dispatcher;    // still the reference to the span
            span.setName(operationName);    // on a potentially recycled span
```

or this scenario:
```
Thread A:
----------
    apmTracer.activeSpan().finish();
        ApmSpanInstrumentation.FinishInstrumentation.finishInternal()
            ApmSpanInstrumentation.FinishInstrumentation.doFinishInternal()
                span.end()     // potentially recycled by thread A or even thread C (disruptor thread)
Thread B:
----------
    apmTracer.activeSpan().setOperationName("opName");
        ApmSpanInstrumentation$SetOperationName.setOperationName()
            if (dispatcher instanceof AbstractSpan) {   // returns true as the dispatcher ref was not yet replaced
                AbstractSpan<?> span = (AbstractSpan) dispatcher;
                span.setName(operationName);    // on a potentially recycled span
Thread A:
----------
        // continue execution
        dispatcher = newDispatcher;    // this line - only now replacing the underlying dispatcher, too late
```

eyalkoren: I still think this doesn't solve the async problem. Consider this scenario:
Thread A starts a transaction, then submits a task to a queue, then the task starts being executed on thread B (original transaction is still running on thread A). Now thread B wants to do something, for example - set the operation name:
```
Thread B:
----------
    apmTracer.activeSpan().setOperationName("opName");
        ApmSpanInstrumentation$SetOperationName.setOperationName()
            if (dispatcher instanceof AbstractSpan) {     // returns true
Thread A:
----------
    apmTracer.activeSpan().finish();
        ApmSpanInstrumentation.FinishInstrumentation.finishInternal()
            ApmSpanInstrumentation.FinishInstrumentation.doFinishInternal()
                span.end()     // potentially recycled by thread A or even thread C (disruptor thread)
        dispatcher = newDispatcher;    // this line - replace the underlying dispatcher from span to serialized context
Thread B:
----------
            // continue execution
            AbstractSpan<?> span = (AbstractSpan) dispatcher;    // still the reference to the span
            span.setName(operationName);    // on a potentially recycled span
```

or this scenario:
```
Thread A:
----------
    apmTracer.activeSpan().finish();
        ApmSpanInstrumentation.FinishInstrumentation.finishInternal()
            ApmSpanInstrumentation.FinishInstrumentation.doFinishInternal()
                span.end()     // potentially recycled by thread A or even thread C (disruptor thread)
Thread B:
----------
    apmTracer.activeSpan().setOperationName("opName");
        ApmSpanInstrumentation$SetOperationName.setOperationName()
            if (dispatcher instanceof AbstractSpan) {   // returns true as the dispatcher ref was not yet replaced
                AbstractSpan<?> span = (AbstractSpan) dispatcher;
                span.setName(operationName);    // on a potentially recycled span
Thread A:
----------
        // continue execution
        dispatcher = newDispatcher;    // this line - only now replacing the underlying dispatcher, too late
```

apm-agent-core/src/main/java/co/elastic/apm/impl/ElasticApmTracer.java
```diff
@@ -178,64 +179,37 @@ public Span currentSpan() {
...
+        if (transaction != null) {
+            if (isTransactionSpanLimitReached(transaction)) {
+                dropped = true;
+                transaction.getSpanCount().getDropped().incrementAndGet();
+            } else {
+                dropped = false;
+                transaction.getSpanCount().getStarted().incrementAndGet();
+            }
         } else {
             dropped = false;
```
eyalkoren: Shouldn't be dropped?

felixbarny:
By having two fields, one for the serialized TraceContext and one for the AbstractSpan in ApmSpan, we can avoid all those race conditions by only using the TraceContext field for determining the parent-child relationships.

apm-agent-plugins/apm-opentracing-plugin/src/main/java/co/elastic/apm/opentracing/impl/ApmSpanInstrumentation.java
```diff
@@ -74,14 +73,17 @@ public FinishInstrumentation() {
...
-        @VisibleForAdvice
-        @Advice.OnMethodEnter(inline = false)
-        public static void finishInternal(@Advice.FieldValue(value = "span", typing = Assigner.Typing.DYNAMIC) @Nullable AbstractSpan<?> span,
-                                          @Advice.Argument(0) long finishMicros) {
-            if (span == null) {
-                return;
+        @Advice.OnMethodEnter
+        private static void finishInternal(@Advice.FieldValue(value = "dispatcher", readOnly = false, typing = Assigner.Typing.DYNAMIC) @Nullable AbstractSpan<?> span,
+                                           @Advice.Argument(0) long finishMicros) {
+            if (span != null) {
```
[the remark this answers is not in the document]
felixbarny: It is allowed to call `finish()` multiple times: https://github.com/opentracing/opentracing-java/blob/76d7b8df4e26e6444a9a7ea19f700cb380a03434/opentracing-api/src/main/java/io/opentracing/Span.java#L146

apm-agent-core/src/main/java/co/elastic/apm/impl/ElasticApmTracer.java
```diff
@@ -178,64 +178,37 @@ public Span currentSpan() {
...
```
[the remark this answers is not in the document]
felixbarny: This is basically how the go agent handles that as well. When we don't have access to the transaction, we can't count the spans. But the span count is interpreted a min span count to warn in the UI if there are fewer spans than that. If a span is started after its transaction has ended, we have no way of counting that.

The TraceContext is local to its corresponding span and does not share data with other span contexts.

apm-opentracing/src/main/java/co/elastic/apm/opentracing/ApmSpan.java
```diff
@@ -73,7 +83,9 @@ public void finish() {
 
     @Override
     public void finish(long finishMicros) {
-        finishInternal(finishMicros);
+        synchronized (this) {
```
eyalkoren: Also the other `finish` needs to be synchronized, and can be done on method level instead explicitly on `this`

apm-opentracing/src/main/java/co/elastic/apm/opentracing/ApmScopeManager.java
```diff
@@ -29,29 +29,42 @@
...
-        } else {
+        if (span != null) {
             return new ApmScope(false, new ApmSpan(span));
+        } else {
+            final Object traceContext = getCurrentTraceContext();
+            if (traceContext != null) {
+                return new ApmScope(false, new ApmSpan(new TraceContextSpanContext(traceContext)));
+            }
         }
+        return null;
```
eyalkoren: This can be a problem- since you `synchronize` on the ApmSpan object, you must make sure there is a single object representing a span. This method can return a different ApmSpan object (through an ApmScope), which means that the synchronization will not only be wrong, it may also introduces risk for deadlocks.
Why not holing the ApmScope as a ThreadLocal? Isn't that the scope of an ApmScope?

felixbarny:
> I think all ApmSpan methods that may change the underlying span must be synchronized as well. 

The only methods which are allowed to be executed after `Span#finish()` are `Span#finish()` and `Span#context()`, so I don't we have to handle racy calls to `Span#finish()` and `Span#addTag`, for example.

apm-opentracing/src/main/java/co/elastic/apm/opentracing/ApmScopeManager.java
felixbarny: That would make a lot of things easier but it would not allow getting the active span in case the span has been activated via the internal API and we would essentially have two distinct scope managers which don't interact with each other.

eyalkoren:
You are right, but why not making them synchronized as well, and in them write a warning to log if called when shouldn't? You know, just to be on the safe side if misused...

felixbarny:
It makes the code quite complex and ugly:
```java
if (spanContext != null) {
    final Object traceContext = spanContext.getTraceContext();
    if (traceContext != null) {
        synchronized (traceContext) {
            // do actual thing
        }
    }
}
```

We can also not put the synchronization in the advice classes, as the `synchronized` block needs to start before reading the `dispatcher` field. The problem is that we can't easily change the implementation in the bridge module `apm-opentracing` as people may only update the agent but not the dependency to the bridge. That's why I want to put as few logic into the bridge as possible.

PR #252, opened 2018-10-19
Cache maven dependencies in CI worker image

felixbarny:


.ci/packer_cache.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
fatmcgav: I think the best practise is to use `/usr/bin/env`

So:
```
#!/usr/bin/env bash
```

Issue #225, opened 2018-09-25
Status code sometimes is 200 instead of 500


felixbarny:
Context: https://discuss.elastic.co/t/java-apm-agent-http-status-codes-not-correctly-identified/149156

Can also be reproduced when calling `/is-it-coffee-time` in opbeans-java

eyalkoren:
I assume that the `service` method just throws an Exception, and lets the container handle it. If it is handled through a `ServletFilter`, than we should catch it, but maybe it redirects it to be handled by an error page (maybe even through `sendError`). 
Maybe monitoring `sendError` and `doFilter` would suffice, but I think a more bullet-proof way would be to keep a reference to the `Response` object on `service` and `doFilter` entry and instrument all those `HttpServletResponse` methods that cause a response commit (e.g. `flushBuffer` and maybe all the flushing methods in the OS/Writer related to the response).
PS - `sendRedirect` will also modify the response data, but I am not sure if it is normally called by container code or user code.
PS2 - the way to monitor the commit action would also be useful for other stuff, mainly those that enable monitoring and even modifying the response content (e.g. for automatic RUM agent injection to HTML)

eyalkoren:
I would argue that not only we need to capture error status codes, we need to report them as errors actively (in addition to setting the status code in the `response` part of the transaction)

eyalkoren:
Using @felixbarny example (`/is-it-coffee-time` in opbeans-java) as a test case, this is how Tomcat is handling errors thrown out of the user's servlet:
Following is the stack of where we end the transaction (at the end of the `doFilter` execution):
![transaction.end.png](https://zube.io/files/elastic/a343105e72532b7ddcf6eb68a466a38a-transaction-end.png)

At this state, the status code is still 200. Exception is caught and handled below on the stack, and only then status code is set:
![status-code-setting.png](https://zube.io/files/elastic/efd32ee83138a4b655d84078d5455857-status-code-setting.png)

Even if we instrumented this `setStatus` API, it would be too late, as we already ended the transaction.
In order to create a single transaction with our current model of execution, where the instrumentation of a single method is used for transaction start and end, we would have to implement a solution for Tomcat, meaning - a solution per app server probably. For example, the `CoyoteAdapter.service` at the bottom of the stack would be a perfect candidate, because it is a Servlet-API-like implementation of the `Servlet.service` method, only with `coyote` types rather than the standard API.

[...]

PR #259, opened 2018-10-25
Whenever exception!=null && statusCode==200 set to 500

eyalkoren:
fixes https://github.com/elastic/apm-agent-java/issues/225

### DIFF
--- a/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/servlet/ServletTransactionHelper.java
+++ b/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/servlet/ServletTransactionHelper.java
@@ -123,6 +123,10 @@ public class ServletTransactionHelper {
                         Map<String, String[]> parameterMap, String servletPath, @Nullable String pathInfo) {
         try {
             fillRequestParameters(transaction, method, parameterMap);
+            if(exception != null && status == 200) {
+                // Probably shouldn't be 200 but 5XX, but we are going to miss this...
+                status = 500;
+            }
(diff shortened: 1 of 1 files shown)


apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/servlet/ServletApiAdvice.java
```diff
@@ -148,7 +148,13 @@ public static void onExitServletService(@Advice.Argument(0) ServletRequest servl
                     resp.addHeader(headerName, response.getHeader(headerName));
                 }
 
-                servletTransactionHelper.onAfter(transaction, t, response.isCommitted(), response.getStatus(), request.getMethod(),
+                int statusCode = response.getStatus();
+                if(t != null && statusCode == 200) {
```
felixbarny:
```suggestion
                if (t != null && statusCode == 200) {
```

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/servlet/ServletApiAdvice.java
```diff
@@ -148,7 +148,13 @@ public static void onExitServletService(@Advice.Argument(0) ServletRequest servl
                     resp.addHeader(headerName, response.getHeader(headerName));
                 }
 
-                servletTransactionHelper.onAfter(transaction, t, response.isCommitted(), response.getStatus(), request.getMethod(),
+                int statusCode = response.getStatus();
```
felixbarny: As this code is not Servlet API specific, I'd move this to ServletTransactionHelper

PR #262, opened 2018-10-29
Cache docker containers on CI worker image

felixbarny:
This requires docker to be available while building the worker image. Not sure if that is the case. It will take longer to build the image but CI builds will be faster.


felixbarny:
It only creates `target` subfolders in the git repo. But that one is cleaned up after the build, right?

fatmcgav:
> It only creates `target` subfolders in the git repo. But that one is cleaned up after the build, right?

Yep, the `$PWD` Git repo is cleaned up :)

Issue #211, opened 2018-09-19
Support scopes for public API


felixbarny:
Make it possible to retrieve the current transaction/span when it has been created via the public API.

Context: https://discuss.elastic.co/t/apm-index-logs-default-to-epoch/148868/13

Issue #212, opened 2018-09-19
Add annotations to the public API to create transactions and spans.


felixbarny:
Example:

```java
@Transaction
public void a(String bar) {
    ElasticApm.currentSpan().addTag("foo", bar);
    b();
}

@Span
public void b() {
    ElasticApm.currentSpan().addTag("baz", "qux");
}
```

Another option would be to have a `@Traced` annotation which creates a span if there is already an active transaction or span, and a transaction otherwise.

Context: https://discuss.elastic.co/t/apm-index-logs-default-to-epoch/148868/13

alvarolobato:
Would it make sense to allow adding the tags as a parameter to the span or transaction annotation?

felixbarny:
I think the more common use-case is that the tag value is dynamic, for example, a method parameter. So I don't think adding static tags via the annotation adds a lot of value. And I really don't want to introduce an expression language where you can access method parameters, for example.

PR #265, opened 2018-10-29
Introduce scopes in public api

felixbarny:
closes elastic/apm-agent-java#211


docs/public-api.asciidoc
```diff
@@ -247,14 +255,42 @@ NOTE: Spans created via this method can not be retrieved by calling <<api-curren
...
+// This is also true for methods called within the try block.
+try (final Scope scope = transaction.activate()) {
+    transaction.setName("MyController#myAction");
+    transaction.setType(Transaction.TYPE_REQUEST);
+    // do your thing...
+} catch (Exception e) {
+    transaction.captureException(e);
+    throw e;
+} finally {
+    transaction.end();
```
eyalkoren: It will be great to have annotations do all that, instead of instructing users

felixbarny: That is planned for #212

PR #270, opened 2018-10-31
Add FAQ section

felixbarny:


docs/faq.asciidoc
```diff
@@ -0,0 +1,51 @@
...
+
+That works by leveraging the capability of the JVM to instrument the bytecode of classes so that for the supported technologies,
+there are no code changes required.
+The agent automatically injects small pieces of code before and after interesting events to measure their duration and metadata,
+like the DB statement,
+as well as HTTP related information,
+like the URL, parameters, and headers.
+For example, if the agent detects that a class extending `javax.servlet.HttpServlet` is loaded,
+it injects monitoring code before and after the servlet invocation.
+
```
felixbarny: Maybe add
> These events, called Transactions and Spans, are sent to the APM Server which converts them to a format suitable for Elasticsearch and sends them to an Elasticsearch cluster. You can then use the APM UI in Kibana to gain insight into latency issues and error culprits of your application.

To make clear how the components are working together

docs/faq.asciidoc
```diff
@@ -0,0 +1,51 @@
+[[faq]]
+== Frequently Asked Questions
+
+[float]
+[[faq-how-does-it-work]]
+=== How does the agent work?
+The agent auto-instruments known frameworks and libraries and records interesting events,
```
bmorelli25: _Here's my suggestion to clean up the wording of this answer:_

The agent auto-instruments known frameworks and libraries and records interesting events, like HTTP requests and database queries. To do this, it leverages the capability of the JVM to instrument the bytecode of classes. This means that for the supported technologies, there are no code changes required. 

The agent automatically injects small pieces of code before and after interesting events to measure their duration and metadata (like the DB statement) as well as HTTP related information (like the URL, parameters, and headers). 

For example, if the agent detects that a class extending `javax.servlet.HttpServlet` is loaded, it injects monitoring code before and after the servlet invocation.

These events, called Transactions and Spans, are sent to the APM Server which converts them to a format suitable for Elasticsearch, and sends them to an Elasticsearch cluster. You can then use the APM UI in Kibana to gain insight into latency issues and error culprits within your application.

docs/faq.asciidoc
```diff
@@ -0,0 +1,51 @@
...
+The agent automatically injects small pieces of code before and after interesting events to measure their duration and metadata,
+like the DB statement,
+as well as HTTP related information,
+like the URL, parameters, and headers.
+For example, if the agent detects that a class extending `javax.servlet.HttpServlet` is loaded,
+it injects monitoring code before and after the servlet invocation.
+
+[float]
+[[faq-installation]]
+=== How do I install the agent?
```
bmorelli25: Do you think this question is really necessary in the FAQ? The getting started guide has the same information, and I don't think anyone would go to an FAQ before the getting started page when looking to install the product

docs/faq.asciidoc
```diff
@@ -0,0 +1,51 @@
...
+Yes
+
+[float]
+[[faq-recompile]]
+=== Do I need to re-compile my application?
+No
+
+[float]
+[[faq-unsupported-technologies]]
+=== What to do when the agent does not support the technologies I'm using?
```
bmorelli25:
```suggestion
=== What if the agent doesn't support the technologies I'm using?
```

docs/faq.asciidoc
felixbarny: @makwarth what do you think?

felixbarny:
Thanks for having a look. Removed the how to install section - you are right - it does not make sense in this context. Added a TODO to add an answer to the question "How safe is bytecode instrumentation?".

bmorelli25:
No worries! Pending bytecode answer, this LGTM!

docs/faq.asciidoc
makwarth: Agree

felixbarny:
Merging now, @bmorelli25 if you find anything else don't hesitate to comment anyway, I'll then improve in another PR.

PR #274, opened 2018-10-31
Using DescriptionStrategy.Default.POOL_ONLY

eyalkoren:


eyalkoren:
So far, we used the default strategy of ByteBuddy (called `HYBRID` for inspecting classes and methods that require instrumentation. This strategy would use only static analysis of the class based on class files for non-loaded classes and reflection for already loaded classes. 
The problem with reflection is that even when invoked on loaded classes, but not fully resolved, reflection will force linkage or the class, which can lead to load errors.
This PR is changing that to use a different strategy which never invokes reflection and only relies on static file analyses (called `POOL_ONLY`).
It would probably create hight overhead on server startup, but it is much safer to use. 
We tested startup differences in Wildfly versions 10.0.0 through 13.0.0 and didn't see noticeable difference, although we know this is a limited scenario.

PR #277, opened 2018-10-31
Bugfix in public API for Scope

felixbarny:
Also removes duplication in transaction and span related classes which is long overdue


felixbarny:
jenkins test this please

PR #288, opened 2018-11-05
Fix transaction id copy

eyalkoren:


felixbarny:
Tested successfully locally, CI is stuck atm.

PR #291, opened 2018-11-06
Add `X-Atmosphere-Transport=long-polling` to ignored URLs

TeroPihlaja:
Added `X-Atmosphere-Transport=long-polling` to ignored URLs.

Note: I did not build or test these changes since I don't have a working build environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ## Features
 * Adds `@CaptureTransaction` and `@CaptureSpan` annotations which let you declaratively add custom transactions and spans.
   Note that it is required to configure the `application_packages` for this to work.
+* Added `X-Atmosphere-Transport=long-polling` to ignored URLs
 
 ## Bug Fixes
 
--- a/apm-agent-plugins/apm-web-plugin/src/main/java/co/elastic/apm/web/WebConfiguration.java
+++ b/apm-agent-plugins/apm-web-plugin/src/main/java/co/elastic/apm/web/WebConfiguration.java
@@ -66,7 +66,7 @@ public class WebConfiguration extends ConfigurationOptionProvider {
         .dynamic(true)
         .buildWithDefault(Arrays.asList(
             WildcardMatcher.valueOf("/VAADIN/*"),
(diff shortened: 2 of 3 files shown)


felixbarny:
Hi and thanks for your PR :)

I'm not sure if this is better suited for being in your custom configuration. The defaults should not grow to be too big, as the more matchers it contains, the higher the overhead - even if you don't use one of the frameworks where the matchers would apply.

TeroPihlaja:
@felixbarny That's what I was thinking as well.

BTW: Documentation was not quite clear if I can add a URL to be ignored or if I need to also copy the defaults to the configuration.

TeroPihlaja:
I tried adding that in the configuration but it seems that elastic APM agent is still reporting that.
```
2018-11-06 12:05:27.070 [main] DEBUG co.elastic.apm.configuration.StartupInfo - ignore_urls: '/VAA
DIN/*,(?i)/heartbeat/*,/favicon.ico,*.js,*.css,*.jpg,*.jpeg,*.png,*.webp,*.svg,*.woff,*.woff2,*X-A
tmosphere-Transport=long-polling*' (source: /srv/glassfish/<domain>/elasticapm.properties)
```

reported full url in APM: ```http://<hostname>:4880/<context_root>/PUSH?v-uiId=0&v-csrfToken=c577bc7d-88de-43b8-ab84-05949b61c7eb&X-Atmosphere-tracking-id=f2231fcb-2ab0-44c6-9327-d9a3d6fa3619&X-Atmosphere-Framework=2.2.13.vaadin5-javascript&X-Atmosphere-Transport=long-polling&X-Atmosphere-TrackMessageSize=true&Content-Type=application%2Fjson%3B%20charset%3DUTF-8&X-atmo-protocol=true&_=1541498833158```

Can you check if the format is correct or maybe the equal sign is causing some issue?

felixbarny:
Hmm, a tricky one. The agent does not consider the query parameters for the exclude check.

felixbarny:
And doing that would break existing excludes like `*.jpg`

TeroPihlaja:
Ok. Then this is clear. I will ignore PUSH in the configurations and close this.

Thanks for your help.

PR #314, opened 2018-11-15
[APM-CI] Java agent Pipeline version 0

kuisathaverat:
image

image


Jenkinsfile
```diff
@@ -0,0 +1,306 @@
...
+    timestamps()
+    preserveStashes()
+    //see https://issues.jenkins-ci.org/browse/JENKINS-11752, https://issues.jenkins-ci.org/browse/JENKINS-39536, https://issues.jenkins-ci.org/browse/JENKINS-54133 and jenkinsci/ansicolor-plugin#132
+    //ansiColor('xterm')
+    disableResume()
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+  }
+  parameters {
+    string(name: 'branch_specifier', defaultValue: "", description: "the Git branch specifier to build (branchName, tagName, commitId, etc.)")    
+    string(name: 'GO_VERSION', defaultValue: "1.10.3", description: "Go version to use.")
```
felixbarny: can be removed?

Jenkinsfile
```diff
@@ -0,0 +1,306 @@
...
+            always {
+              sendBenchmarks(file: "${BASE_DIR}/${BULK_UPLOAD_FILE}", index: "benchmark-java", bulk: true)
+              archiveArtifacts(allowEmptyArchive: true, 
+                artifacts: "${BASE_DIR}/${RESULT_FILE},${BASE_DIR}/${BULK_UPLOAD_FILE}", 
+                onlyIfSuccessful: false)
+            }
+          }
+        }
+        /**
+         run Go integration test with the commit version on master branch.
```
felixbarny: ~Go~ Java

Jenkinsfile
```diff
@@ -0,0 +1,306 @@
...
+                booleanParam(name: "ruby_Test", value: false),
+                booleanParam(name: "python_Test", value: false),
+                booleanParam(name: "nodejs_Test", value: false)],
+              wait: true,
+              propagate: true)
+          }
+        }
+        
+        /**
+         run Go integration test with the commit version on a PR.
```
felixbarny: ~Go~ Java

Jenkinsfile
```diff
@@ -0,0 +1,306 @@
...
+                }
+                sh """#!/bin/bash
+                ./scripts/jenkins/run-benchmarks.sh
+                """
+              }
+            }
+          } 
+          post {
+            always {
+              sendBenchmarks(file: "${BASE_DIR}/${BULK_UPLOAD_FILE}", index: "benchmark-java", bulk: true)
```
felixbarny: where is sendBenchmarks defined?

Jenkinsfile
```diff
@@ -0,0 +1,306 @@
...
+                booleanParam(name: "kibana_Test", value: false),
+                booleanParam(name: "server_Test", value: false)],
+              wait: true,
+              propagate: true)
+          }
+        }
+      }
+    }
+        
+    stage('Documentation') {
```
felixbarny: cool, finally a docs build :)

Jenkinsfile
kuisathaverat: The step is defined in the pipeline library, I am testing the step for the Java agent, probably I have to make some adjustments.

kuisathaverat:
>can these files go under a non-root directory, preferably .ci?

I trying to unified were the Jenkins scripts are, on APM server, Go Agent, Hey APM, and Integration test exists a folder scripts/jenkins where all those files are. We can open a team discussion about the location of the files but IMO should be the same on all projects.

Jenkinsfile
kuisathaverat: sure, I needed it for the benchmarks step, I have to correct it to do not need this variable.

Jenkinsfile
kuisathaverat: C&P ™️ 😃

Jenkinsfile
kuisathaverat: C&P ™️ 😃

Jenkinsfile
kuisathaverat: yep, if you see something that I missed or it is needed, please mention it. Also I put the steps to run on master branch, a PR, or both by my own criteria tell me if it is correct.

checkout - all
build - all
test - all
benchmarks - branches (master, 6.x, ...) not in PRs
full integration test - master (disabled for the moment)
short integration test - PRs and other branches that are not master (disabled for the moment)
documentation - master

felixbarny: What do you refer to when you say integration tests? Do you mean testing this (https://github.com/elastic/apm-integration-testing/tree/master/docker/java/spring/testapp) app against a real APM Server and ES?

The Java agent also has some integration tests or maybe you could call them smoke tests for different servers and databases. They are executed on the `verify` goal. `test` only executes the unit tests.

What is the difference between full integration test and short integration test?

kuisathaverat: >What do you refer to when you say integration tests? Do you mean testing this (https://github.com/elastic/apm-integration-testing/tree/master/docker/java/spring/testapp) app against a real APM Server and ES?

yep, https://github.com/elastic/apm-integration-testing/blob/master/scripts/ci/java.sh

>The Java agent also has some integration tests or maybe you could call them smoke tests for different servers and databases. 

smoke test is fine

>They are executed on the verify goal. test only executes the unit tests.

Noted

kuisathaverat: >What is the difference between full integration test and short integration test?

The short  integration test will only test the PR against the stable version of APM Server + Elastic Stack, the other will test against a matrix of versions, this is made on other agents ensure that you version will work with different combinations of APM Server + Elastic Stack. TBD which matrix versions we want to use.

Jenkinsfile
```diff
@@ -0,0 +1,338 @@
...
+      }
+      steps {
+        withEnvWrapper() {
+          unstash 'source'
+          dir("${ELASTIC_DOCS}"){
+            git "https://github.com/elastic/docs.git"
+          }
+          dir("${BASE_DIR}"){    
+            sh """#!/bin/bash
+            make docs
```
felixbarny: there is no make docs in the Java repo :)

felixbarny:
The current pipeline setup takes quite a bit more time to execute. Maybe we can re-shuffle the pipeline setup a bit :)

I don't think we need separate steps for Build and Unit Test as testing also does a build. Similarly, the Smoke Tests also do Unit Tests so I think we can condense all three to one stage.

I think the docs build could also run in parallel to the others. 

Currently, the benchmarks are only run for changes in master. We would at least need a way to tag PR benchmark results so that the benchmark dashboard can filter out non-master results. Mixing results from master and PRs would not be a good idea.

But I had also totally different idea on how to approach benchmarking for PRs. I don't think that all PRs need to execute the benchmarks. We could have a phrase like "Jenkins benchmark this please" to benchmark on demand. We would then execute the benchmarks twice - once from the point, the PR branched off and once with the PR changes. Then we could compare the two results and add both results, including the diff as a GitHub comment.

This is just an idea. We can also go the route of benchmarking for every PR and enable comparison via a dashboard.

kuisathaverat:
>The current pipeline setup takes quite a bit more time to execute. Maybe we can re-shuffle the pipeline setup a bit :)

After the improvements, the preview version takes about 10 min, the old one takes about 14 min, parallelizing another block of test maybe we can decrease another minute or so, but I do not think that worth the effort. Never the less if we categorize/group/tag tests with JUnit annotations it is possible to run them easily by groups.

>Currently, the benchmarks are only run for changes in master. We would at least need a way to tag PR benchmark results so that the benchmark dashboard can filter out non-master results. Mixing results from master and PRs would not be a good idea.

We can store use the PR id (benchmark-java-PR134) to store in a different index with the same pattern (benchmark-java-*), for branches we can do the same (benchmark-java-BRANCH_NAME)

>But I had also totally different idea on how to approach benchmarking for PRs. I don't think that all PRs need to execute the benchmarks. We could have a phrase like "Jenkins benchmark this please" to benchmark on demand. We would then execute the benchmarks twice - once from the point, the PR branched off and once with the PR changes. Then we could compare the two results and add both results, including the diff as a GitHub comment.

It is easy to be able to launch the job to the Jenkins UI for any branch/PR with the parameters that you want, right now, it is possible to do that on this PR.

image

screenshot 2018-11-23 at 11 45 34

felixbarny:
> After the improvements, the preview version takes about 10 min

wow, that's great!

> We can store use the PR id (benchmark-java-PR134) to store in a different index with the same pattern (benchmark-java-*), for branches we can do the same (benchmark-java-BRANCH_NAME)

I would rather add the branch name to the metadata of the benchmark results. That makes it easier to filter out PR benchmark results from the dashboards. How can I get access tot he PR id? Is it an environment variable?

> It is easy to be able to launch the job to the Jenkins UI for any branch/PR with the parameters that you want, right now, it is possible to do that on this PR.

perfect!

kuisathaverat:
>I would rather add the branch name to the metadata of the benchmark results. That makes it easier to filter out PR benchmark results from the dashboards. How can I get access tot he PR id? Is it an environment variable?

sort of, there are some environment variables defined when you process a branch an others when you process a PR

CHANGE_TARGET - is only defined on PRs
GIT_SHA - SHA1 of commit
BRANCH_NAME - the name of the branch, it is defined only on branches

But we could convert a code that I have to make for codecov to a step and we can export it as an environment variable (BRANCH_REF) https://github.com/elastic/apm-pipeline-library/blob/master/vars/codecov.groovy#L20-L44

felixbarny:
what's the status of this? can it be merged?

felixbarny:
🎉

PR #316, opened 2018-11-17
Avoid setting request attributes to reduce allocations

felixbarny:
Adding a request attribute creates a new HashMap$Node which allocates 32b
- Only set transaction request attribute when startAsync is called
- Set excluded as ThreadLocal as opposed to request attribute

### DIFF
--- a/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/servlet/ServletApiAdvice.java
+++ b/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/servlet/ServletApiAdvice.java
@@ -55,6 +55,13 @@ public class ServletApiAdvice {
     @Nullable
     @VisibleForAdvice
     public static ElasticApmTracer tracer;
+    @VisibleForAdvice
+    public static ThreadLocal<Boolean> excluded = new ThreadLocal<>() {
+        @Override
+        protected Boolean initialValue() {
+            return Boolean.FALSE;
+        }
+    };
 
     static void init(ElasticApmTracer tracer) {
         ServletApiAdvice.tracer = tracer;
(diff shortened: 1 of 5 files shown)


apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/servlet/ServletApiAdvice.java
```diff
@@ -121,6 +126,7 @@ public static void onExitServletService(@Advice.Argument(0) ServletRequest servl
         if (tracer == null) {
             return;
         }
+        excluded.set(Boolean.FALSE);
```
eyalkoren: I think for the sake of accurate state management, this should be done only when `transaction != null`.

felixbarny: The transaction is null for excluded requests which is the exact case where we do want to set excluded back to the initial value. We could set it to false only if it's true or if transaction is null but I think it's simpler and not necessarily slower to set it always to false, just to be sure we don't miss anything.

felixbarny: > In the case of nested service calls, you set the excluded to true in the outmost invocation, so you should reset it in the exit of the same method.

I think it's just fine to set it to false in all nested filters and servlets. If it has already been reset to false, resetting it again does not change the state.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/servlet/ServletApiAdvice.java
```diff
@@ -55,6 +55,13 @@
     @Nullable
     @VisibleForAdvice
     public static ElasticApmTracer tracer;
+    @VisibleForAdvice
+    public static ThreadLocal<Boolean> excluded = new ThreadLocal<>() {
```
felixbarny:
```suggestion
    public static ThreadLocal<Boolean> excluded = new ThreadLocal<Boolean>() {
```

Issue #197, opened 2018-09-11
CXFServletExt#doPost generate 500 when INSERT to DB2


lk95jofr:
We are currently using elastic-apm-agent-0.6.2.jar which works fine, collects data from our postgres and DB2 calls.

We wanted to try a new agent and installed, elastic-apm-agent-0.7.0-20180910.093339-71.jar, which resulted in a lot off 500's

How and why I cant actually understand from reading the logs.
I just see that before and after the result is totally different

```xml
Response-Code: 500
Encoding: UTF-8
Content-Type: text/xml
Headers: {}
Payload: <env:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/" xmlns:env="http://schemas.xmlsoap.org/soap/envelope/"><env:Header/><env:Body><env:Fault xmlns:soap="http://schemas
.xmlsoap.org/soap/envelope/">
   <faultcode xmlns:ns1="http://schemas.xmlsoap.org/soap/envelope/">ns1:Server</faultcode>
   <faultstring>Unexpected error</faultstring>
   <detail>
      <LoginExceptionData xmlns="http://our-domain.se/our-login-path/data/v1_0" xmlns:ns2="http://our-domain.se/our-common-element-1/data/v4_6" xmlns:ns3="http://our-domain.se/our-common-element-2/data/v4_6">
         <Errormessage>(class: com/ibm/db2/jcc/am/CallableStatement, method: execute signature: ()Z) Illegal type in constant pool</Errormessage>
      </LoginExceptionData>
   </detail>
</env:Fault></env:Body></env:Envelope>
```

In the APM part of elastiic we did get two new requests
MetricsStreamServlet#doGet
CXFServletExt#doPost

It is the CXFServletExt#doPost that returns 500, it seems to be at time of INSERT

/Fredrik

felixbarny:
Thanks for the report!

This looks interesting:
> class: com/ibm/db2/jcc/am/CallableStatement, method: execute signature: ()Z) Illegal type in constant pool

That sounds like a problem in Byte Buddy. I have now updated to the most recent version. Hopefully, that fixes your issue. Here is a current snapshot: [elastic-apm-agent-0.7.0-SNAPSHOT.jar.zip](https://github.com/elastic/apm-agent-java/files/2370509/elastic-apm-agent-0.7.0-SNAPSHOT.jar.zip)


[...]

PR #320, opened 2018-11-19
Omit instrumenting classes version pre Java 5

felixbarny:
This avoids VerifyErrors like reported [here](https://discuss.elastic.co/t/java-lang-verifyerror-with-java-agent/156882) and in #197.

An alternative would be to use the `TypeConstantAdjustment` visitor. But not instrumenting classes compiled for Java 4 or before seems to be safer.

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/bci/bytebuddy/MinimumClassFileVersionValidator.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/bci/bytebuddy/MinimumClassFileVersionValidator.java
@@ -1,0 +1,61 @@
+package co.elastic.apm.bci.bytebuddy;
+
+import net.bytebuddy.ClassFileVersion;
+import net.bytebuddy.asm.AsmVisitorWrapper;
+import net.bytebuddy.description.field.FieldDescription;
+import net.bytebuddy.description.field.FieldList;
+import net.bytebuddy.description.method.MethodList;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.implementation.Implementation;
+import net.bytebuddy.jar.asm.ClassVisitor;
+import net.bytebuddy.pool.TypePool;
+import net.bytebuddy.utility.OpenedClassReader;
+
+public enum MinimumClassFileVersionValidator implements AsmVisitorWrapper {
+
+    INSTANCE;
+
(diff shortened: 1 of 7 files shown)


apm-agent-core/src/main/java/co/elastic/apm/bci/bytebuddy/MinimumClassFileVersionValidator.java
```diff
@@ -0,0 +1,61 @@
...
+    @Override
+    public ClassVisitor wrap(TypeDescription instrumentedType, ClassVisitor classVisitor, Implementation.Context implementationContext,
+                             TypePool typePool, FieldList<FieldDescription.InDefinedShape> fields, MethodList<?> methods, int writerFlags, int readerFlags) {
+        return new ClassVisitor(OpenedClassReader.ASM_API, classVisitor) {
+            @Override
+            public void visit(int version, int access, String name, String signature, String superName, String[] interfaces) {
+                final ClassFileVersion classFileVersion = ClassFileVersion.ofMinorMajor(version);
+                if (!classFileVersion.isAtLeast(MINIMUM_CLASS_FILE_VERSION)) {
+                    throw UnsupportedClassFileVersionException.INSTANCE;
+                }
```
eyalkoren: This means that the order is correct, meaning that you will avoid applying the `ClassVisitor` coming from the `ForAdvice` transformer whenever version is invalid.

apm-agent-core/src/main/java/co/elastic/apm/bci/bytebuddy/MinimumClassFileVersionValidator.java
```diff
@@ -0,0 +1,61 @@
...
+import net.bytebuddy.utility.OpenedClassReader;
+
+public enum MinimumClassFileVersionValidator implements AsmVisitorWrapper {
+
+    INSTANCE;
+
+    private static final ClassFileVersion MINIMUM_CLASS_FILE_VERSION = ClassFileVersion.JAVA_V5;
+
+    @Override
+    public ClassVisitor wrap(TypeDescription instrumentedType, ClassVisitor classVisitor, Implementation.Context implementationContext,
```
eyalkoren: We talked about switching the order of the calls to `transform`, but maybe what you did is the only right way to go if this is expected to wrap the `ForAdvice` `ClassVisitor`

PR #340, opened 2018-11-24
to change apm.server_urls from  localhost:8300 to localhost:8200

epoll486:
to change elastic.apm.server_urls from  http://localhost:8300 to  http://localhost:8200


felixbarny:
Thanks for spotting this documentation error! There is also a reference to the wrong port in `configuration.asciidoc` and `configuration.asciidoc.ftl`. Could you fix those as well? Could you also please target the PR to `master` instead of `1.x`?

felixbarny:
I have added the changes. Thanks for your contribution!

Issue #339, opened 2018-11-22
apm-reporter not using full server_url for building reporting url


beckjim:
**Describe the bug**
Using elastic-apm-agent-1.0.1.jar (on Payara 5)
We're using a load balancer in front of apm-servers (as we do for elasticsearch), so we configure
`server_urls=http://loadbalancer/apm-server/`
Java agent correctly connects at startup:
`[apm-server-healthcheck] INFO co.elastic.apm.report.ApmServerHealthChecker - Elastic APM server is available: {"ok":{"build_date":"2018-11-16T01:22:04Z","build_sha":"e9b22404feb418663d86f059b833bf1487d9379e","version":"6.5.1"}`

An error occurs as soon as the first events have to be reported:
`[apm-reporter] WARN co.elastic.apm.report.IntakeV2ReportingEventHandler - It seems like you are using a version of the APM Server which is not compatible with this agent. Please use APM Server 6.5.0 or newer.|#]`     
`[apm-reporter] WARN co.elastic.apm.report.IntakeV2ReportingEventHandler - http://loadbalancer/intake/v2/events|#]`

Please note the missing '/apm-server' as part of the reporting URL.

**To Reproduce**
1. Use a loadbalancer with a path other than '/' in front of apm-servers
2. Configre `server_urls=http://loadbalancer/<path>`
2. Trigger java apm agent ro report some incident
3. See error

**Expected behavior**
Java agent should respect full server_url for building reporting URL for intake API.
e.g. reporter should send events to `http://loadbalancer/<path>/intake/v2/events` and not to 
`http://loadbalancer/intake/v2/events`

Cause seems to be in IntakeV2ReportingEventHandler.java:
` public static final String INTAKE_V2_URL = "/intake/v2/events";`
This causes in method HttpURLConnection startRequest() to create an absolut path for the reporting URL, not respecting the fully configured server_url:
`URL url = new URL(serverUrlIterator.get(), INTAKE_V2_URL);`

Otherwise the agent is working really great for far! 
Thanks a lot!

eyalkoren:
Fixed with https://github.com/elastic/apm-agent-java/pull/343

eyalkoren:
@beckjim Thanks a lot for trying our APM solution and for your great feedback!
I hope the upcoming release (1.0.2) will fix this issue for you. Should be out soon.

beckjim:
Perfect, thanks! Will test once it is released.

PR #343, opened 2018-11-25
Handling cases where APM server URL contains path

eyalkoren:
Fixing https://github.com/elastic/apm-agent-java/issues/339

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/report/IntakeV2ReportingEventHandler.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/report/IntakeV2ReportingEventHandler.java
@@ -31,6 +31,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.stagemonitor.util.IOUtils;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.net.ssl.HttpsURLConnection;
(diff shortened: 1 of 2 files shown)


apm-agent-core/src/main/java/co/elastic/apm/report/IntakeV2ReportingEventHandler.java
```diff
@@ -243,6 +245,20 @@ private HttpURLConnection startRequest() {
         }
     }
 
+    @Nonnull
+    URL getUrl() throws MalformedURLException {
+        URL serverUrl = serverUrlIterator.get();
+        String path = serverUrl.getPath();
+        if(path.endsWith("/")) {
```
felixbarny:
```suggestion
        if (path.endsWith("/")) {
```

apm-agent-core/src/main/java/co/elastic/apm/report/IntakeV2ReportingEventHandler.java
```diff
@@ -243,6 +245,20 @@ private HttpURLConnection startRequest() {
...
+            path = path.substring(0, path.length()-1);
```
felixbarny:
```suggestion
            path = path.substring(0, path.length()-1);
```
```suggestion
            path = path.substring(0, path.length() - 1);
```

PR #360, opened 2018-11-29
Add basic metrics support

felixbarny:


felixbarny:
Adds the following system metrics:

```
system.cpu.total.norm.pct
system.process.cpu.total.norm.pct
system.memory.total
system.memory.actual.free
system.process.memory.size
```

And some heap metrics:

```
jvm.memory.heap.used
jvm.memory.heap.committed
jvm.memory.heap.max
jvm.memory.nonheap.used
jvm.memory.nonheap.committed
jvm.memory.nonheap.max
```

felixbarny:
@kuisathaverat Do you know why pipeline Jenkins does not like this PR? I probably confused it with my wild force pushes 😬

kuisathaverat:
it looks like [JENKINS-44598](https://issues.jenkins-ci.org/browse/JENKINS-44598), probably is related to the force push, I never see it before.

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/JvmMemoryMetrics.java
```diff
@@ -0,0 +1,70 @@
...
+public class JvmMemoryMetrics implements LifecycleListener {
+
+    @Override
+    public void start(ElasticApmTracer tracer) {
+        bindTo(tracer.getMetricRegistry());
+    }
+
+    private void bindTo(final MetricRegistry registry) {
+        register(registry, "nonheap", ManagementFactory.getPlatformMXBean(MemoryMXBean.class).getNonHeapMemoryUsage());
+        register(registry, "heap", ManagementFactory.getPlatformMXBean(MemoryMXBean.class).getNonHeapMemoryUsage());
```
eyalkoren: Should be getHeapMemoryUsage :)

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/JvmMemoryMetrics.java
```diff
@@ -0,0 +1,70 @@
...
+import co.elastic.apm.agent.impl.ElasticApmTracer;
+import co.elastic.apm.agent.metrics.DoubleSupplier;
+import co.elastic.apm.agent.metrics.MetricRegistry;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.MemoryMXBean;
+import java.lang.management.MemoryUsage;
+import java.util.Collections;
+
+public class JvmMemoryMetrics implements LifecycleListener {
```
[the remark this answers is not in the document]
felixbarny: We already have the concept of `LifecycleListener`s which provide a callback on application start and on shutdown. I did not create an SPI for metrics as I could just use the already existing one. If you think it would make things more clear, I can add a special SPI just for metrics.

felixbarny:
I have manually checked that `system.cpu.total.norm.pct` returns similar values to the ones emitted by our Go agent.

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
```diff
@@ -0,0 +1,121 @@
...
+    public void add(String name, Map<String, String> tags, DoubleSupplier metric) {
+        MetricSet metricSet = metricSets.get(tags);
+        if (metricSet == null) {
+            metricSets.putIfAbsent(tags, new MetricSet(tags));
+            metricSet = metricSets.get(tags);
+        }
+        metricSet.add(name, metric);
+    }
+
+    public void serialize(JsonWriter jw, StringBuilder replaceBuilder) {
```
eyalkoren: Is it used somewhere?

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
```diff
@@ -0,0 +1,121 @@
...
+    public double get(String name, Map<String, String> tags) {
+        final MetricSet metricSet = metricSets.get(tags);
+        if (metricSet != null) {
+            return metricSet.get(name).get();
+        }
+        return Double.NaN;
+    }
+
+    @Override
+    public String toString() {
```
eyalkoren: Moving the serialization code to the `serialize` package is great, so I think having this `JsonWriter` residual here is a shame. Sorry for nagging, I have future refactors in mind...
What do you say adding a static `asString(MetricsRegistry)` to the serializer?

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/JvmGcMetrics.java
```diff
@@ -0,0 +1,92 @@
...
+        bindTo(tracer.getMetricRegistry());
+    }
+
+    void bindTo(final MetricRegistry registry) {
+        for (final GarbageCollectorMXBean garbageCollectorMXBean : garbageCollectorMXBeans) {
+            final Map<String, String> tags = Collections.singletonMap("name", garbageCollectorMXBean.getName());
+            registry.addUnlessNegative("jvm.gc.count", tags, new DoubleSupplier() {
+                @Override
+                public double get() {
+                    return garbageCollectorMXBean.getCollectionCount();
```
eyalkoren: What would be more interesting - collection count or collection count since last measurement? For graphs, an ever increasing one is not informative as a mostly-stagnant one where glitches are easier to see. This is something that can be calculated in the server or Kibana as well, but since it is not a global metric, I don't think it would be.

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/JvmGcMetrics.java
```diff
@@ -0,0 +1,92 @@
...
+                }
+            });
+            registry.addUnlessNegative("jvm.gc.time", tags, new DoubleSupplier() {
+                @Override
+                public double get() {
+                    return garbageCollectorMXBean.getCollectionTime();
```
eyalkoren: same

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
felixbarny: do you want to create a separate module for serialization?

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
felixbarny: nope, removing it

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/JvmGcMetrics.java
felixbarny: It's quite common to report counters and use a derivative aggregation to graph the deltas.

felixbarny:
> It just occurred to me that the entire metrics harvesting is going to be done on the disruptor publishing thread. Do you think this may be problematic and be better done offline, so not to hold down other event publishing?

I'd think that getting and serializing the metrics is not something that would hold down the reporter thread. I think the bottleneck there is the network anyway. 

> Do you think this may be problematic and be better done offline

Do you have an idea how to do this without serializing the metrics to a buffer which becomes garbage afterwards and without establishing another connection to the APM Server?

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
eyalkoren: Not at the moment :)
I just think it makes sense to avoid dependency on serialization lib here.

felixbarny: I'm all about keeping package dependencies uni-directional!

eyalkoren:
I am not talking about serialization, I am just talking about metrics collection from all MBeans. Anyway, as long as we don't have evidence that this is a problem - it is not.

PR #361, opened 2018-11-29
Test public API in app server integration test

felixbarny:
This currently fails for WildFly, probably because of the bootdelegation

### DIFF
--- a/integration-tests/simple-webapp/pom.xml
+++ b/integration-tests/simple-webapp/pom.xml
@@ -29,6 +29,11 @@
             <artifactId>h2</artifactId>
             <version>${version.h2}</version>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>apm-agent-api</artifactId>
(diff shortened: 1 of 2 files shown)


felixbarny:
Two options to solve this:

1. Add more specific packages than `co.elastic.apm.*` to the bootdelegation:

https://github.com/elastic/apm-agent-java/blob/0451955cecb10321e4c3fa6795f9dc838ddc89bb/apm-agent-core/src/main/java/co/elastic/apm/bci/OsgiBootDelegationEnabler.java#L41

Cons: this list would be quite long and would have to change as we introduce new packages. So that would be quite brittle.

2. Change package structure of the non-api classes. For example to `co.elastic.apm.agent` or `co.elastic.apm.internal` and change the bootdelegation property accordingly so that `co.elastic.apm.api` is not matched anymore.

Cons: Almost all classes have to move but as they are internal, this is would not be a breaking change.

eyalkoren:
Closing. Moved the test change to #376 so it will be included with the fix for it

PR #363, opened 2018-11-30
Reduce log level for transformation errors caused by optional classes

felixbarny:
Logging these as warn has lead to confusion and was perceived as a potential bug/error in the agent repeatedly.
When Byte Buddy introspects a class and tries to resolve a referenced class which is not on the classpath,
it throws an exception.
This also changes the isPublic() matcher to be the last one so that no unnecessary
lookups which can potentially fail are made.


felixbarny:
Whenever we get a `Cannot resolve type description` exception from Byte Buddy, we should check that the matchers don't force resolving a class prematurely as it has been the case in the `AsyncInstrumentation`.

PR #364, opened 2018-11-30
Remove module-info.java to not confuse early Tomcat 8.5 versions

felixbarny:


felixbarny:
One of the reasons this happened is that we add the full agent jar to the class path which can then be discovered by various class path scanners, potentially leading to errors or slowdowns. We should re-evaluate the idea to inject only a minimal jar which then loads the agent in it's own class loader.

elastic-apm-agent/pom.xml
```diff
@@ -31,6 +31,14 @@
                             <createSourcesJar>true</createSourcesJar>
                             <shadeSourcesContent>true</shadeSourcesContent>
                             <shadedArtifactAttached>false</shadedArtifactAttached>
+                            <filters>
+                                <filter>
+                                    <artifact>*:*</artifact>
+                                    <excludes>
+                                        <exclude>META-INF/versions/*/module-info.class</exclude>
```
eyalkoren: I think this structure for locating the module-info class is specific to multi-release jars. Maybe `*/module-info.class` or even `*module-info.class` will be more inclusive, or did you aim for only such jars?

felixbarny: done

PR #366, opened 2018-12-03
Docs: Add sampling and RAM to FAQ

bmorelli25:
@eyalkoren - adapted Felix's responses [here](https://discuss.elastic.co/t/what-is-the-recommended-sampling-percentage-ram/158845/3) to apply broadly and added to FAQ. Please double check for technical accuracy.

### DIFF
--- a/docs/faq.asciidoc
+++ b/docs/faq.asciidoc
@@ -45,9 +45,9 @@ It also respects other agents attached to your application at the same time.
 No
 
 [float]
-[[recommended-sampling-rate]]
-=== What is the recommended sampling rate?
-There is no one-size-fits-all answer to an ideal sampling rate.
+[[recommended-sample-rate]]
+=== What is the recommended sample rate?
+There is no one-size-fits-all answer to an ideal sample rate.
 Sampling comes down to your preferences and your application.
 The more you want to sample, the more network bandwidth and disk space you'll need.
 
@@ -55,6 +55,9 @@ It's important to note that the latency of an application won't be affected much
 even if you sample at 100%.
 However, the background reporter thread has some work to do for serializing and gzipping events.
 If your application is not CPU bound, this shouldn't matter much.
+Note that if the APM Server can't handle all the events,
+the agent will drop data to not crash your application.
+It will then also not serialize and gzip the events.
 
 Sample rate can be changed by altering the <<config-transaction-sample-rate>> configuration. 
 


docs/faq.asciidoc
```diff
@@ -44,6 +44,26 @@ It also respects other agents attached to your application at the same time.
...
+=== What is the recommended sampling rate?
+There is no one-size-fits-all answer to an ideal sampling rate.
+Sampling comes down to your preferences and your application.
+The more you want to sample, the more network bandwidth and disk space you'll need.
+
+It's important to note that the latency of an application won't be affected much by the agent (in the order of single-digit microseconds),
+even if you sample at 100%.
+However, the background reporter thread has some work to do for serializing and gzipping events.
+If your application is not CPU bound, this shouldn't matter much.
+
```
eyalkoren: I think adding the last part is good as well:
>Note that if the APM Server can't handle all the events, the agent will drop data to not crash your application. It will then also not serialize and gzip the events.

docs/faq.asciidoc
```diff
@@ -44,6 +44,26 @@ It also respects other agents attached to your application at the same time.
...
+Sample rate can be changed by altering the <<config-transaction-sample-rate>> configuration.
```
eyalkoren: I don't think `<<config-transaction-sample-rate>>` can't work. You are trying to refer to another document, so maybe something like:
```
Sample rate can be changed by altering the <<config-core#config-transaction-sample-rate,`transaction_sample_rate` configuration option>>.
```
to try and link to the exact anchor, and if that doesn't work maybe:
```
Sample rate can be changed by altering the `transaction_sample_rate` <<config-core,configuration option>>.
```
In order to try these out, I assume you need to build locally and see how it works, but I am almost certain that the current way will result in a broken link.

bmorelli25: It actually does work :)

Asciidoc automatically builds the correct link (`../config-core.html#config-transaction-sample-rate`) from the `<<config-transaction-sample-rate>>`

Here's what it looks like:
screen shot 2018-12-04 at 7 05 48 am

docs/faq.asciidoc
bmorelli25: Added!

bmorelli25:
Thanks @eyalkoren! See my comments above. As for sample vs sampling - I did a little research and "sample" seems to be the prevalent term, therefor, I've changed the title and references to "sample". I'm not sure there's a right answer here, so if you feel strongly the other way, just let me know.

docs/faq.asciidoc
eyalkoren: Ohh, it resolves directly from an external anchor, nice!!
Well done then, sorry to doubt, just didn't want to get this nasty broken link notification in dev channel :)

bmorelli25: No worries at all! I appreciate you digging into everything!

Issue #325, opened 2018-11-21
Read docker container id from filesystem


eyalkoren:
See https://github.com/elastic/apm/issues/22

eyalkoren:
Implemented in https://github.com/elastic/apm-agent-java/issues/325

PR #368, opened 2018-12-04
Finding and sending container ID

eyalkoren:
Implementing https://github.com/elastic/apm-agent-java/issues/325

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/SystemInfo.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/SystemInfo.java
@@ -20,12 +20,23 @@
 package co.elastic.apm.agent.impl.payload;
 
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
 import java.net.InetAddress;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
 
 /**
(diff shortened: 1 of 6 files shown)


apm-agent-core/src/main/java/co/elastic/apm/impl/payload/SystemInfo.java
```diff
@@ -58,6 +79,73 @@ static String getNameOfLocalHost() {
...
+                    containerId = parseContainerId(line);
+                    if(containerId != null) {
+                        break;
+                    }
+                }
+            }
+        } catch (Throwable e) {
+            logger.warn("Failed to read/parse container ID from '/proc/self/cgroup'");
+        }
+        logger.debug("container ID is: " + containerId);
```
felixbarny: Use a placeholder instead of string concatenation

```suggestion
        logger.debug("container ID is: {}", containerId);
```

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/SystemInfo.java
```diff
@@ -58,6 +79,73 @@ static String getNameOfLocalHost() {
...
+                    idPart = idPart.substring(0, idPart.length() - ".scope".length());
+                }
+
+                if(idPart.matches("^[0-9a-fA-F]{64}$")) {
+                    containerId = idPart;
+                }
+            }
+        }
+        if(containerId == null) {
+            logger.info("Could not parse container ID from '/proc/self/cgroup' line: " + line);
```
felixbarny:
```suggestion
            logger.info("Could not parse container ID from '/proc/self/cgroup' line: {}", line);
```

Issue #371, opened 2018-12-05
IndexOutOfBoundsException when adding a header to Response on WildFly


eyalkoren:
Related to https://discuss.elastic.co/t/apm-java-agent-indexoutofboundsexception/159419.
User has Wildfly containers that recently got IndexOutOfBoundsException.
Working with Java agent 1.0.1 and APM-Server 6.5.1
```
[0m[31m12:25:47.699 ERROR [io.undertow.request] (default task-15) UT005023: Exception handling request to /api/households/5380009081/sync: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
at java.util.ArrayList.rangeCheck(ArrayList.java:657)
at java.util.ArrayList.get(ArrayList.java:433)
at co.elastic.apm.util.PotentiallyMultiValuedMap.add(PotentiallyMultiValuedMap.java:57)
at co.elastic.apm.impl.context.Response.addHeader(Response.java:64)
at org.jboss.resteasy.plugins.server.servlet.HttpServletDispatcher.service(HttpServletDispatcher.java:52)
at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
at io.undertow.servlet.handlers.ServletHandler.handleRequest(ServletHandler.java:85)
at io.undertow.servlet.handlers.security.ServletSecurityRoleHandler.handleRequest(ServletSecurityRoleHandler.java:62)
at io.undertow.servlet.handlers.ServletDispatchingHandler.handleRequest(ServletDispatchingHandler.java:36)
at org.wildfly.extension.undertow.security.SecurityContextAssociationHandler.handleRequest(SecurityContextAssociationHandler.java:78)
at io.undertow.server.handlers.PredicateHandler.handleRequest(PredicateHandler.java:43)
at io.undertow.servlet.handlers.security.SSLInformationAssociationHandler.handleRequest(SSLInformationAssociationHandler.java:131)
at io.undertow.servlet.handlers.security.ServletAuthenticationCallHandler.handleRequest(ServletAuthenticationCallHandler.java:57)
at io.undertow.server.handlers.PredicateHandler.handleRequest(PredicateHandler.java:43)
at io.undertow.security.handlers.AbstractConfidentialityHandler.handleRequest(AbstractConfidentialityHandler.java:46)
at io.undertow.servlet.handlers.security.ServletConfidentialityConstraintHandler.handleRequest(ServletConfidentialityConstraintHandler.java:64)
[...]

PR #385, opened 2018-12-16
Async Servlet state management changes

eyalkoren:
### General
Investigation started due to #371 
A race condition introduced due to implementation failure of some fine-grained definitions of the `HttpServletRequest.startAsync` API in some application servers requires a shift in how we instrument this. The full explanation is below, but the short version is- `AsyncContext.isAsyncStarted` is not fully reliable when invoked at the end of `service/doFilter` in which `HttpServletRequest.startAsync` was used, due to race condition between both threads handling the request. 
Basic rationale behind the suggested solution: if `HttpServletRequest.startAsync` is used, then the transaction data should not be updated by the `service/doFilter` instrumentation route, but only through the `AsyncListener` we register. This should hold for all `AsyncContext` APIs and regardless of multiple dispatches. Furthermore, we assume that if an `AsyncListener` was registered- it must be invoked when response gets committed, so we rely on it to clear states.

### Detailed explanation
The javadoc for `AsyncContext.isAsyncStarted` says that it will return `false` after `AsyncContext.dispatch` or `AsyncContext.complete` are invoked. Looking further into the `dispatch` and `complete` javadoc shows that this is a bit more complicated, as it says:

> If this method is called before the container-initiated dispatch that called startAsync has returned to the container, the dispatch operation will be delayed until after the container-initiated dispatch has returned to the container.

This means that querying `AsyncContext.isAsyncStarted` should be safe at the end of `service`/`doFilter`, as the container should wait for those to return before actually invoking `dispatch/complete`.  When testing multiple WildFly versions, I saw that while this is indeed the case with `dispatch`, it is not so in the case of `complete`, at least in WildFly versions
8 through 11. Specifically, the `AsyncContext` offers a `start` API for conveniently invoke logic on a separate thread, where `complete` is likely to be called, making it a common use case.

### DIFF
--- a/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/helper/ApmAsyncListener.java
+++ b/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/helper/ApmAsyncListener.java
@@ -80,7 +80,7 @@ public class ApmAsyncListener implements AsyncListener {
     // (see class-level Javadoc)
     private void endTransaction(AsyncEvent event) {
         // To ensure transaction is ended only by a single event
-        if(EVENT_COUNTER_UPDATER.getAndIncrement(this) > 0) {
+        if (EVENT_COUNTER_UPDATER.getAndIncrement(this) > 0) {
             return;
         }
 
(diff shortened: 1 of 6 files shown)


felixbarny:
//cc @adriancole

adriancole:
thanks for the ping

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/helper/ApmAsyncListener.java
```diff
@@ -82,7 +79,14 @@ public void onStartAsync(AsyncEvent event) {
     // because only the onExitServletService method may contain references to the servlet API
     // (see class-level Javadoc)
     private void endTransaction(AsyncEvent event) {
+        // To ensure transaction is ended only by a single event
+        if(EVENT_COUNTER_UPDATER.getAndIncrement(this) > 0) {
```
felixbarny:
```suggestion
        if (EVENT_COUNTER_UPDATER.getAndIncrement(this) > 0) {
```

PR #388, opened 2018-12-17
Wrap Runnable in AsyncContext.start

eyalkoren:


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/InScopeRunnableWrapper.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public InScopeRunnableWrapper withTransaction(Transaction transaction) {
+        this.transaction = transaction;
+        return this;
+    }
+
+    @Override
+    public void run() {
+        Scope scope = null;
```
felixbarny: I think this can be simplified with
```
if (transaction != null) {
    transaction.activate();
    try {
        delegate.run();
    } catch (Exception e) {
        transaction.captureException(e);
        throw e;
    } finally {
        transaction.deactivate();
    }
}
```

eyalkoren: Well, I don't want to give up the try/catch around `activate`/`deactivate` - it's just like your other PR to catch errors during advices - our code should never prevent user code from running. I can give up the usage of scope though.

Your suggestion to capture the Exception and report as error is tempting, but is not safe here. Notice that this code is not taking responsibility on ending and reporting the Transaction, it's just responsible for scoping in that thread. For example, the usage of `AsyncContext.onComplete` would end the transaction in this PR (and `AsyncContext.onError` will report errors). So I don't know what the status of the transaction is at this stage.

felixbarny: Good point on the problems with `captureException`. However, `AbstractSpan#activate` already catches exceptions.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/AsyncInstrumentation.java
```diff
@@ -54,87 +54,130 @@
...
+
+        @VisibleForAdvice
+        public static class AsyncContextStartAdvice {
+
+            @Advice.OnMethodEnter(suppress = Throwable.class)
+            private static void onEnterAsyncContextStart(@Advice.Argument(value = 0, readOnly = false) @Nullable Runnable runnable) {
+                if (tracer != null) {
+                    final Transaction transaction = tracer.currentTransaction();
+                    if (transaction != null && runnable != null && asyncHelper != null) {
+                        runnable = asyncHelper.getForClassLoaderOfClass(AsyncContext.class).wrapRunnable(runnable, transaction);
```
felixbarny: wouldn't it be easier to just call the `wrapRunnable` method on the `tracer` instance? Then you could also remove the `wrapRunnable` from the helper.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/InScopeRunnableWrapper.java
```diff
@@ -0,0 +1,96 @@
...
+@VisibleForAdvice
+public class InScopeRunnableWrapper implements Runnable, Recyclable {
+    private final Logger logger = LoggerFactory.getLogger(InScopeRunnableWrapper.class);
+
+    @Nullable
+    private Runnable delegate;
+    @Nullable
+    private AbstractSpan<?> span;
+    @Nullable
+    private ElasticApmTracer tracer;
```
felixbarny: make final and remove `@Nullable`

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/InScopeRunnableWrapper.java
```diff
@@ -0,0 +1,96 @@
...
+
+        try {
+            //noinspection ConstantConditions
+            delegate.run();
+        } finally {
+            try {
+                if (span != null) {
+                    span.deactivate();
+                }
+                //noinspection ConstantConditions
```
felixbarny: can be removed when making tracer nonnull

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/InScopeRunnableWrapper.java
```diff
@@ -0,0 +1,96 @@
...
+                try {
+                    logger.warn("Failed to activate span");
+                } catch (Throwable t) {
+                    // do nothing, just never fail
+                }
+            }
+        }
+
+        try {
+            //noinspection ConstantConditions
```
felixbarny: it's safer to add a null check than to suppress these inspections

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/InScopeRunnableWrapper.java
```diff
@@ -0,0 +1,96 @@
...
+    @Nullable
+    private AbstractSpan<?> span;
+    @Nullable
+    private ElasticApmTracer tracer;
+
+    InScopeRunnableWrapper(ElasticApmTracer tracer) {
+        this.tracer = tracer;
+    }
+
+    public InScopeRunnableWrapper wrap(Runnable delegate) {
```
felixbarny: Maybe add `AbstractSpan` as a second parameter and remove the `withScope` method? Calling one without the other would lead to an invalid state anyway.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/AsyncInstrumentation.java
eyalkoren: When implementing I imagined it will do some additional stuff, but probably no need for that as it is now

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/InScopeRunnableWrapper.java
eyalkoren: Hmm, didn't realize we don't recycle this reference... 
Will do

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/InScopeRunnableWrapper.java
eyalkoren: I know it's not null. I don't like to add null checks just to satisfy inspections... There are already too many non-required null checks.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/InScopeRunnableWrapper.java
eyalkoren: Right, part of the separation I did for the instance creation and other stuff, which we already decide I will remove :)

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/InScopeRunnableWrapper.java
felixbarny: Is it valid to deactivate a finished span? Ideally, `#end` should be the last call on a span because after that the span is recycled and could be reused any time. Deactivating after that is borderline but probably unproblematic if the state isn't changed after `#end`.

PR #394, opened 2018-12-19
Add ability to get tracing headers from API

### LINKED ISSUE #334: Add manual inter-process context propagation capability to the public API
Maybe something like supported in the [OT inject-extract API](https://opentracing.io/guides/java/inject-extract/) - a key-value map that can be obtained from context for injection in client side and can be parsed into context in the server side

felixbarny:
closes elastic/apm-agent-java#334

### DIFF
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
@@ -20,6 +20,8 @@
 package co.elastic.apm.api;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Map;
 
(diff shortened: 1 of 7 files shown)


apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
```diff
@@ -192,4 +194,9 @@
      */
     boolean isSampled();
 
+    void addTraceHeaders(@Nullable Map<? super String, ? super String> headers);
```
felixbarny: TODO: Javadoc

felixbarny:
Superseded by https://github.com/elastic/apm-agent-java/pull/396

PR #398, opened 2018-12-21
Add trace_methods option

### LINKED ISSUE #347: Add a config to enable tracing of specific methods
Make it possible to instrument a method in a concrete class. This would be equivalent to annotating the method with `@CaptureSpan` but also allows to instrument methods of 3rd party libraries.

felixbarny:
Defines a syntax for matching methods which should be traced

closes #347

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -20,6 +20,8 @@
 package co.elastic.apm.agent.configuration;
 
 import co.elastic.apm.agent.bci.ElasticApmInstrumentation;
+import co.elastic.apm.agent.bci.methodmatching.MethodMatcher;
+import co.elastic.apm.agent.bci.methodmatching.configuration.MethodMatcherValueConverter;
 import co.elastic.apm.agent.configuration.validation.RegexValidator;
 import co.elastic.apm.agent.matcher.WildcardMatcher;
 import co.elastic.apm.agent.matcher.WildcardMatcherValueConverter;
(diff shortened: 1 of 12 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -238,6 +240,33 @@ public static String getAllInstrumentationGroupNames() {
...
+            " - `org.example.MyClass#myMethod(java.lang.String)`\n" +
+            " - `org.example.MyClass#myMe*od(java.lang.String, int)`\n" +
+            " - `private org.example.MyClass#myMe*od(java.lang.String, *)`\n" +
+            " - `* org.example.MyClas*#myMe*od(*.String, int[])`\n" +
+            " - `public org.example.services.*.*Service#*`\n" +
+            "\n" +
+            "NOTE: Only use wildcards if necessary.\n" +
+            "The more methods you match to more overhead will be caused by the agent.\n" +
+            "Also note that there is a maximum amount of spans per transaction (see <<config-transaction-max-spans, `transaction_max_spans`>>).")
+        .buildWithDefault(Collections.singletonList(MethodMatcher.of("org.springframework.web.servlet.DispatcherServlet#render(*, *, *)")));
```
felixbarny: Remove default, that was just for testing purposes. When actually instrumenting `DispatcherServlet#render`, we probably want to set a custom type. Maybe template or render.

apm-agent-core/src/test/java/co/elastic/apm/agent/bci/methodmatching/MethodMatcherInstrumentationTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import org.junit.jupiter.api.Test;
+
+import java.lang.reflect.Method;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class MethodMatcherInstrumentationTest {
```
felixbarny: @eyalkoren This tests modifiers and arguments so I don't think we need to do this again in `TraceMethodInstrumentationTest`

eyalkoren: Not really- it tests the matcher code, not the instrumentation code.

felixbarny: Wouldn't testing it in `TraceMethodInstrumentationTest` also just test the matcher? Anyway, it probably makes sense to test one simple case for the modifier and the arguments in `TraceMethodInstrumentationTest` as well.

eyalkoren: It is not testing the code at `TraceMethodInstrumentation.getMethodMatcher`

felixbarny: actually it's doing exactly that :)

https://github.com/elastic/apm-agent-java/pull/398/files/e4293fca3fdff5371f8c24520e8f3ed4b26c190c#diff-340c555184d8a9790f0bc3f0e2f4020cR71

eyalkoren: Your'e right, missed that 😊

PR #399, opened 2018-12-24
Error handling of Servlet Async API

eyalkoren:
Status code and error reporting
Pool ApmAsynListener objects
Extend integration tests to capturing errors and JSP transactions

### DIFF
--- a/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/helper/AsyncContextAdviceHelperImpl.java
+++ b/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/helper/AsyncContextAdviceHelperImpl.java
@@ -21,26 +21,50 @@ package co.elastic.apm.agent.servlet.helper;
 
 import co.elastic.apm.agent.impl.ElasticApmTracer;
 import co.elastic.apm.agent.impl.transaction.Transaction;
+import co.elastic.apm.agent.objectpool.Allocator;
+import co.elastic.apm.agent.objectpool.ObjectPool;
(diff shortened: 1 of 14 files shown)


eyalkoren:
Jenkins test this please

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/helper/AsyncContextAdviceHelperImpl.java
```diff
@@ -21,26 +21,50 @@
...
     private final ServletTransactionHelper servletTransactionHelper;
     private final ElasticApmTracer tracer;
 
     public AsyncContextAdviceHelperImpl(ElasticApmTracer tracer) {
         this.tracer = tracer;
         servletTransactionHelper = new ServletTransactionHelper(tracer);
+
+        int maxPooledElements = tracer.getConfigurationRegistry().getConfig(ReporterConfiguration.class).getMaxQueueSize() * 2;
+        asyncListenerObjectPool = QueueBasedObjectPool.ofRecyclable(
+            AtomicQueueFactory.<ApmAsyncListener>newQueue(createBoundedMpmc(maxPooledElements)),
```
felixbarny: The number of required `ApmAsyncListener` instances is independent of the size of the disruptor. It's probably dependant on the number of container threads. The default for Tomcat is 200 IIRC. I'm not sure whether there is an extra thread pool for the handling of async requests but I suppose there is not.

Maybe hard-code to 256?

eyalkoren: You are right it is not the right configuration to rely on. We can start with hard code and see if it bites us.

Issue #389, opened 2018-12-18
Add `transaction.sampled` to Errors


eyalkoren:
Implementing https://github.com/elastic/apm/issues/30

PR #405, opened 2019-01-02
Setting transaction.sampled in errors

eyalkoren:


apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
```diff
@@ -165,6 +165,9 @@ private void handleEvent(ReportingEvent event, long sequence, boolean endOfBatch
             }
             connection = startRequest();
             payloadSerializer.serializeMetaDataNdJson(metaData);
+            if (logger.isTraceEnabled()) {
+                logger.trace(payloadSerializer.toString());
```
felixbarny: The payload is already logged when the request ends. What's the benefit of logging the whole payload for every event?

eyalkoren: Getting it in JSON format, see what we actually send the server

felixbarny: But that's already done here: https://github.com/elastic/apm-agent-java/blob/035ba3a2ef30262d6e1983daf36d92128bd76ce2/apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java#L111

eyalkoren: Only if flushed, which never happens if the connection with the server is down: 
https://github.com/elastic/apm-agent-java/blob/035ba3a2ef30262d6e1983daf36d92128bd76ce2/apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java#L282

eyalkoren: Anyway, not sure the right way to go, I'll remove it anyway and we will find a way to do it properly

eyalkoren:
related to #389

PR #409, opened 2019-01-07
added span to DispatcherServlet.render

### LINKED ISSUE #287: Add span for DispatcherServlet#render
We should create a span for the method `org.springframework.web.servlet.DispatcherServlet#render` so Spring users can analyze how much time it takes to render a template.

nugusbayevkk:
closes #287


apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentation.java
```diff
@@ -0,0 +1,93 @@
...
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    private static void beforeExecute(@Advice.Argument(0) ModelAndView modelAndView,
+                                      @Advice.Local("span") Span span) {
+        if (tracer == null || tracer.activeSpan() == null)
+            return;
+        final AbstractSpan<?> parent = tracer.activeSpan();
+        span = parent
+            .createSpan()
+            .withType(SPAN_TYPE_DISPATCHER_SERVLET_RENDER)
+            .appendToName(RENDER_METHOD).appendToName(" ").appendToName(modelAndView.getViewName())
```
felixbarny: As per javadoc, `modelAndView.getViewName()` can return `null`. That case should probably be handled. Either just use `Render` as the name then or maybe use `modelAndView.getView().getClass().getSimpleName()`. However, `getSimpleName()` allocates memory which is best to avoid by coping the contents after the last dot/dollar sign up to the end of the string to the name `StringBuilder`. Is it common that the `View` is implemented as an anonymous inner class? In that case, the simple class name would not be helpful.

So maybe as a first step, just add the view name unless `null`.

felixbarny:
also, please add an entry to `CHANGELOG.md` 🙂

felixbarny:
Jenkins test this please

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentation.java
```diff
@@ -0,0 +1,93 @@
...
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    private static void beforeExecute(@Advice.Argument(0) ModelAndView modelAndView,
+                                      @Advice.Local("span") Span span) {
+        if (tracer == null || tracer.activeSpan() == null)
+            return;
+        final AbstractSpan<?> parent = tracer.activeSpan();
+        span = parent
+            .createSpan()
+            .withType(SPAN_TYPE_DISPATCHER_SERVLET_RENDER)
+            .appendToName(RENDER_METHOD).appendToName(" ").appendToName(modelAndView.getViewName() != null ? modelAndView.getViewName() : "")
```
eyalkoren: If view name is not available, maybe a better name would be `DispatcherServlet#render` to be more Spring MVC indicative

apm-agent-plugins/apm-spring-webmvc-plugin/src/test/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentationTest.java
```diff
@@ -0,0 +1,103 @@
...
+        public String test() {
+            return "message-view";
+        }
+    }
+
+    @Test
+    public void testCallOfDispatcherServletRender() throws Exception {
+        this.mockMvc.perform(get("/test"));
+        assertEquals(1, reporter.getTransactions().size());
+        assertEquals(2, reporter.getSpans().size());
```
eyalkoren: what's the other span?

nugusbayevkk: i will add check on that span isAlreadyMonitored.

eyalkoren: So what is it? If it's not related to this test (e.g. from the setup step or something) - just start the test with `reporter.reset();` to start clean. Otherwise, if it's something we don't want to report we shouldn't.

nugusbayevkk: it is span of DispatcherServlet#render. I don't understand why it's appear twice on test.

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentation.java
nugusbayevkk: i can't reproduce the case when ModelAndView#getViewName returns null, in DispatcherServlet there is a check, if View is null, viewName is set from the request URI:
```
	protected String getDefaultViewName(HttpServletRequest request) throws Exception {
		return (this.viewNameTranslator != null ? this.viewNameTranslator.getViewName(request) : null);
	}
```

apm-agent-plugins/apm-spring-webmvc-plugin/src/test/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentationTest.java
eyalkoren: Can you debug that? Because we wouldn't want to send twice if only one is valid.

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentation.java
eyalkoren: This can also return null..

apm-agent-plugins/apm-spring-webmvc-plugin/src/test/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentationTest.java
nugusbayevkk: I’ll debug it. But there is strange behavior when running application, span sended once to apm server.

eyalkoren: Maybe one of the spans is related to some lifecycle thing that is not related to the request (although I can't see why `render` be invoked in such route)? When you debug, check if one of these spans (the fault span) doesn't have the transaction set as parent.

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentation.java
eyalkoren: Exactly what I meant, thanks!
Only I would go with
```
    private static final String RENDER_METHOD = "Render";
    private static final String DISPATCHER_SERVLET_RENDER_METHOD = "DispatcherServlet#render";

    private static void beforeExecute(@Advice.Argument(0) ModelAndView modelAndView,
                                      @Advice.Local("span") Span span) {
        if (tracer == null || tracer.activeSpan() == null || isAlreadyMonitored(tracer.activeSpan()))
            return;
        final AbstractSpan<?> parent = tracer.activeSpan();
        
        span = parent.createSpan().withType(SPAN_TYPE_DISPATCHER_SERVLET_RENDER);
        if (modelAndView == null || modelAndView.getViewName() == null)
            span.appendToName(DISPATCHER_SERVLET_RENDER_METHOD);
        else
            span.appendToName(RENDER_METHOD).appendToName(" ").appendToName(modelAndView.getViewName());

        span.activate();
    }
```
Just because I am allergic to code duplications 😊

apm-agent-plugins/apm-spring-webmvc-plugin/src/test/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentationTest.java
nugusbayevkk: I found the reason, render method called twice from TestDispatcherServlet, too:  .and(hasSuperType(named("org.springframework.web.servlet.DispatcherServlet"))).

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentation.java
felixbarny: Maybe use `DISPATCHER_SERVLET_RENDER_METHOD` in both cases and add `modelAndView.getViewName()` if available to be consistent.

Also, in the actual code, please use braches for the if/else statements.

apm-agent-plugins/apm-spring-webmvc-plugin/src/test/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentationTest.java
felixbarny: maybe, instead of `hasSuperType(named("org.springframework.web.servlet.DispatcherServlet"))`, just do `named("org.springframework.web.servlet.DispatcherServlet")`

eyalkoren: Great, so we have two options:

1. If we expect that `org.springframework.web.servlet.DispatcherServlet#render` is always invoked even when overridden (meaning - overrides always invoke `super#render`) and if we think it is good enough to trace only it (i.e. the duration measurement is good enough etc.) - then just change the matching code above so that only this exact method is traced. That's the easiest, most performant and cleanest solution.
2. Otherwise, we need to guard from nested spans, see https://github.com/elastic/apm-agent-java/blob/5de143707981b2db950120d1519c4327c4aad2da/apm-agent-plugins/apm-httpclient-core/src/main/java/co/elastic/apm/agent/http/client/HttpClientHelper.java#L54 as example. 
In this case, there may or may not be a preference to the outer or inner invocation, so needs to be tested.

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentation.java
```diff
@@ -0,0 +1,94 @@
...
+                .end();
+        }
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return nameStartsWith("org.springframework")
+            .and(not(isInterface()))
+            .and(declaresMethod(getMethodMatcher()))
+            .and(named("org.springframework.web.servlet.DispatcherServlet"));
```
felixbarny: as this is an exact class name match, I don't think we need the other matchers

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentation.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+
+public class DispatcherServletRenderInstrumentation extends ElasticApmInstrumentation {
+
+    private static final String SPAN_TYPE_DISPATCHER_SERVLET_RENDER = "template.dispatcher-servlet.render";
+    private static final String DISPATCHER_SERVLET_RENDER_METHOD = "DispatcherServlet#render";
+
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    private static void beforeExecute(@Advice.Argument(0) ModelAndView modelAndView,
```
felixbarny: Annotate with `@Nullable` to document the parameter's nullability and to make tools (like IntelliJ's constant conditions inspection) aware of it.

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentation.java
```diff
@@ -0,0 +1,94 @@
...
+                                      @Advice.Local("span") Span span) {
+        if (tracer == null || tracer.activeSpan() == null)
```
felixbarny: add braces to statement

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/spring/webmvc/DispatcherServletRenderInstrumentation.java
```diff
@@ -0,0 +1,94 @@
...
+            return;
+        final AbstractSpan<?> parent = tracer.activeSpan();
+
+        span = parent.createSpan().withType(SPAN_TYPE_DISPATCHER_SERVLET_RENDER).withName(DISPATCHER_SERVLET_RENDER_METHOD);
+
+        if (modelAndView != null && modelAndView.getViewName() != null)
```
felixbarny: add braces here as well

nugusbayevkk:
thanks all for the help!

PR #410, opened 2019-01-08
Fix JsonWriter buffer leak when APM server not available

eyalkoren:


apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
```diff
@@ -164,27 +166,52 @@ private void handleEvent(ReportingEvent event, long sequence, boolean endOfBatch
...
-            connection = startRequest();
-            payloadSerializer.serializeMetaDataNdJson(metaData);
-        }
         try {
-            writeEvent(event);
-        } catch (Exception e) {
-            // end request on error to start backing off
-            flush();
+            if (connection == null) {
+                long millisToGracePeriodEnd = gracePeriodEnd - System.currentTimeMillis();
```
felixbarny: I think it would be easier to wait in `onConnectionError` that way we also don't need the `gracePeriodEnd` instance variable.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
```diff
@@ -83,6 +84,7 @@
     private int errorCount;
     private long gracePeriodEnd;
     private boolean shutDown;
+    private volatile boolean closed;
```
felixbarny: Maybe reuse `shutDown` and make that volatile? Closed seems to be used for the same condition.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
```diff
@@ -164,27 +166,52 @@ private void handleEvent(ReportingEvent event, long sequence, boolean endOfBatch
...
-            writeEvent(event);
-        } catch (Exception e) {
-            // end request on error to start backing off
-            flush();
+            if (connection == null) {
+                long millisToGracePeriodEnd = gracePeriodEnd - System.currentTimeMillis();
+                if (millisToGracePeriodEnd > 0) {
+                    // back off because there are connection issues with the apm server
+                    try {
+                        synchronized (WAIT_LOCK) {
```
felixbarny: Why not use `this` as the monitor? So that other classes can't hold it?

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
eyalkoren: I started with that but then changed it for two reasons:

1. `shutDown` is queried on each event, so making it volatile will have more impact
2. there will be a strange race condition due to https://github.com/elastic/apm-agent-java/blob/64a6c21d3b3cfe50b1fd58e7d8bbe62daccd13b7/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java#L218 

Looking back at it, maybe both reasons are not valid:
1. we actually want to avoid handling events after close, so maybe this query is a good one
2. in this case, this race condition is not a problem as we wait on the disruptor for 5 seconds, but it may be if are not fully aware to that (e.g. if you switch https://github.com/elastic/apm-agent-java/blob/64a6c21d3b3cfe50b1fd58e7d8bbe62daccd13b7/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java#L225 and https://github.com/elastic/apm-agent-java/blob/64a6c21d3b3cfe50b1fd58e7d8bbe62daccd13b7/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java#L221 in order)

what do you think?

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
eyalkoren: Yes, the disruptor holds reference to it, I want that to be fully under my control

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
eyalkoren: Good idea, will change that

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
felixbarny: I think it should wait for the shutdown event to be handled. This ensures that events which are currently in the ring buffer are handled, if possible. If it can't (for example, because the event handler waits because of a grace period), the `shutDown` variable should be set to `true` by the thread handling the shutdown. Maybe we should wait on a future for the shutdown event to be handled, then close the event handler and after that shut down the disruptor.

eyalkoren: If you want the thread initiating the shutdown to change this flag, then consolidating to one flag is good and no need to wait on a future as we already wait on the disruptor. In the current state- this race condition is a good one, just as you described - either thread can change it but we make sure it is changed by the end of the close method. 
And maybe it is also a good thing that we stop handling event after close is called (even if we pay the price of frequent queries for a volatile).

So, bottom line - I you say we should change it to be one flag that can be changed by close, right?

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
```diff
@@ -361,6 +355,20 @@ private void onConnectionError(@Nullable Integer responseCode, long droppedEvent
...
         }
+
+        long backoffTimeSeconds = getBackoffTimeSeconds(errorCount++);
+        logger.info("Backing off for {} seconds (±10%)", backoffTimeSeconds);
+        final long backoffTimeMillis = TimeUnit.SECONDS.toMillis(backoffTimeSeconds);
+        if (backoffTimeMillis > 0) {
+            // back off because there are connection issues with the apm server
+            try {
+                synchronized (WAIT_LOCK) {
+                    WAIT_LOCK.wait(backoffTimeMillis);
```
felixbarny:
```suggestion
                    WAIT_LOCK.wait(backoffTimeMillis + getRandomJitter(backoffTimeMillis));
```

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
```diff
@@ -164,27 +157,35 @@ private void handleEvent(ReportingEvent event, long sequence, boolean endOfBatch
...
-            // end request on error to start backing off
-            flush();
-            onConnectionError(null, currentlyTransmitting, 0);
+            if (connection == null) {
+                connection = startRequest();
+                payloadSerializer.serializeMetaDataNdJson(metaData);
+            }
+            try {
+                writeEvent(event);
+            } catch (Exception e) {
```
felixbarny: I think we can get rid of this try/catch when catching `Exception` instead of `IOException` in the next catch block and by moving `flush()` to that as well. If the connection is not established, `flush()` is a noop anyway.

eyalkoren: Not the most accurate error handling anyway, but I will improve logging to compensate that

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
eyalkoren: Oops... Good catch!

felixbarny:
A side effect of blocking, instead of discarding events during the grace period is that a full ring buffer can never recover if the server persists to be unavailable. That also means we can't publish flush events. I can't remember why we decided to block instead of moving this if outside the `if (connection == null)` which would also have solved the memory leak:

```java
if (gracePeriodEnd > System.currentTimeMillis()) {
    // back off because there are connection issues with the apm server
    dropped++;
    return;
}
```

eyalkoren:
Discarding events during grace periods means two things that I assume should be avoided:
1. keeping the consumer threads busy doing nothing instead of being idle
2. losing events even for short connection failures

What is the problem with a ring buffer that can't recover when the server is unavailable? And what is the problem with not being able to publish the shutdown event if server is unavailable (besides the fact that the `publish` is blocking, but using `tryPublish` overcomes that)?

felixbarny:
> losing events even for short connection failures

Ah, that's a very good point. Didn't remember that reason.

Besides the fact that it's blocking on `publish`, which can be solved with a `tryPublish` I don't think there is any problem with waiting in the reporter during a grace period. I just didn't remember why we thought blocking was better and it was not documented in the PR or the code.

felixbarny:
> And what is the problem with not being able to publish the shutdown event if server is unavailable

One potential problem, which is probably just an edge case, is that if the APM Server is available but the ring buffer is still full due to a traffic spike. In that case, the agent can't flush the remaining data. But again, this is quite an edge case and the impact does also not seem severe. So I like the solution we currently have :)

PR #412, opened 2019-01-08
Exclude constructors in TraceMethodInstrumentation

felixbarny:
because Byte Buddy complains about catching exceptions in constructors:
java.lang.IllegalStateException: Cannot catch exception during constructor call


apm-agent-core/src/test/java/co/elastic/apm/agent/bci/methodmatching/TraceMethodInstrumentationTest.java
```diff
@@ -62,35 +63,52 @@ void tearDown() {
...
         assertThat(reporter.getTransactions()).isEmpty();
     }
 
-    // not traced because visibility modifier does not match
-    public void traceMeNot() {
+    // Byte Buddy can't catch exceptions (onThrowable = Throwable.class) during a constructor call:
+    @Test
+    void testNotMatched_Constructor() {
+        new TestExcludeConstructor();
+        assertThat(reporter.getTransactions()).isEmpty();
```
eyalkoren: Just to have feedback that the test is valid, add here a method invocation on that instance and check that it is instrumented

PR #413, opened 2019-01-08
Adjust license headers for 2019

felixbarny:


eyalkoren:
🎉 Happy new year
@elastic/apm-agent-devs @elastic/apm-server @elastic/apm-ui - maybe you need to do the same?

PR #416, opened 2019-01-11
Explain reasoning of the public API

felixbarny:


eyalkoren:
Great stuff! I would add, just for emphasizing this point, something like:
> Having this separation means that if the agent needs to be removed promptly (e.g. because of a problem)- this can be done immediately by restarting the app without the `javaagent` flag. No need to delay until code is edited, compiled and tested.

Issue #145, opened 2018-08-01
In process context propagation for Executors


felixbarny:
To propagate the trace context to different threads, we should by default instrument `ExecutorService`s to propagate the currently active span. The user should not have to manually wrap the `Executor` in order to enable the context propagation for it.

Prior art:
* https://github.com/talsma-ict/context-propagation
* https://github.com/census-instrumentation/opencensus-java/blob/master/contrib/agent/src/main/java/io/opencensus/contrib/agent/instrumentation/ExecutorInstrumentation.java
* https://github.com/census-instrumentation/opencensus-java/blob/master/contrib/agent/src/main/java/io/opencensus/contrib/agent/instrumentation/ThreadInstrumentation.java
* https://github.com/opentracing-contrib/java-concurrent
* https://github.com/openzipkin/brave/blob/3774c2c0681304c4b9e1bf29f701a484b5f65e6f/brave/src/main/java/brave/propagation/CurrentTraceContext.java#L200
* https://github.com/glowroot/glowroot/blob/master/agent/plugins/executor-plugin/src/main/java/org/glowroot/agent/plugin/executor/ExecutorAspect.java

One design decision is whether we want to wrap `Runnable`s  and store the context as an instance variable or if we want to instrument implementations of `Runnable` and add an instance variable via BCI. Wrapping has the disadvantage that we are creating garbage and type casts to a specific sub-type of `Runnable` would fail. The disadvantage of adding the instance variable via BCI is that this approach would not work with runtime attachment of the agent. Yet another option would be to store the trace context in a map which associates Runnables with the corresponding trace context.

Another question is whether we want to generically instrument all runnables or only those submitted to an `Executor`. We could store the trace context when the runnable is created and restore it in the run method. The assumption here is that the `Runnable` is instantiated in the Thread which holds the current trace context. But this assumption is not always true, for example when the `Runnable` is stored in an instance variable.

Relevant interfaces are: `java.util.concurrent.Executor`, `java.util.concurrent.ExecutorService`, `java.util.concurrent.ScheduledExecutorService`, `java.lang.Runnable` and `java.util.concurrent.Callable`

This should fix https://discuss.elastic.co/t/java-apm-agent-sends-no-timeline-spans/141150, where Hystrix is used for fault tolerance of JDBC queries.

PR #417, opened 2019-01-11
Introduce TraceContextHolder, unifying context propagation

felixbarny:
Introducing `TraceContextHolder` in the internal API as a common
superclass of both `TraceContext` and `AbstractSpan`. Given an instance
of this class, you can create child spans, capture exceptions and manage
activations/scopes.

This abstraction reliefs clients from having to differ between the case
when the current activation is a `TraceContext` vs an `AbstractSpan`.

A `TraceContext` would be active when the current thread does not own
the lifecycle of the parent span. Otherwise an `AbstractSpan` would be
active.

There is rarely a case where a client would need to get the currently
active span to set the name or tags. This propagation is normally done
with `@Advice.Local` variables from enter to exit advices. Should the
need for that arise however, and the client can ensure that the current
activation is actually an `AbstractSpan`, all they need to do is to cast
`tracer.getActive()` from `TraceContextHolder` to `AbstractSpan`.

The public API does not differ between `Span` and `TraceContextHolder`.
`ElasticApm.currentSpan()` always returns a `Span`. In case the current
activation is a `TraceContextHolder`, methods like `Span#setTag` are
silent noops. Calling those methods on a `Span` whose lifecycle is not
owned by the caller is illegal anyway. However, calling
`ElasticApm.currentSpan().createSpan()` is always allowed.

Taking away `ElasticApm.currentSpan()` is not an option, as a common use
case is to rename the spans created by the auto instrumentation or to
add custom tags.

Also, `ElasticApm.currentSpan().createSpan()` makes for a much nicer and
more object-oriented API than starting a span on a tracer, passing in
the parent as a start option.

In preparation of elastic/apm-agent-java#145


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContextHolder.java
```diff
@@ -0,0 +1,134 @@
...
+ * when the current activation is a {@link TraceContext} vs an {@link AbstractSpan}.
+ * </p>
+ * <p>
+ * A {@link TraceContext} would be active when the current thread does not own the lifecycle of the parent span.
+ * Otherwise an {@link AbstractSpan} would be active.
+ * </p>
+ *
+ * @param <T> the type, used to enable fluent method chaining
+ */
+public abstract class TraceContextHolder<T extends TraceContextHolder> {
```
eyalkoren: You can make `Recyclable` instead of declaring in both subclasses

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/plugin/api/AbstractSpanInstrumentation.java
```diff
@@ -170,8 +183,10 @@ public ActivateInstrumentation() {
 
         @VisibleForAdvice
         @Advice.OnMethodEnter(suppress = Throwable.class)
-        public static void addTag(@Advice.FieldValue(value = "span", typing = Assigner.Typing.DYNAMIC) AbstractSpan<?> span) {
-            span.activate();
+        public static void addTag(@Advice.FieldValue(value = "span", typing = Assigner.Typing.DYNAMIC) TraceContextHolder<?> holder) {
```
eyalkoren: change method name to `activate`

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/plugin/api/AbstractSpanInstrumentation.java
```diff
@@ -170,8 +183,10 @@ public ActivateInstrumentation() {
...
+            if (holder instanceof AbstractSpan) {
```
eyalkoren: Why not activate for any holder type?

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/plugin/api/AbstractSpanInstrumentation.java
```diff
@@ -182,9 +197,9 @@ public IsSampledInstrumentation() {
 
         @VisibleForAdvice
         @Advice.OnMethodExit(suppress = Throwable.class)
-        public static void addTag(@Advice.FieldValue(value = "span", typing = Assigner.Typing.DYNAMIC) AbstractSpan<?> span,
+        public static void addTag(@Advice.FieldValue(value = "span", typing = Assigner.Typing.DYNAMIC) TraceContextHolder<?> holder,
```
eyalkoren: change method name to `isSampled`

felixbarny:
Thanks for the excellent review! 👌

PR #419, opened 2019-01-13
added support for OkHttp(sync calls)

### LINKED ISSUE #329: Support OkHttp

nugusbayevkk:
closes #329


apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientInstrumentation.java
```diff
@@ -0,0 +1,100 @@
...
+    private static final String SPAN_TYPE_OK_HTTP_CLIENT = HTTP_CLIENT_SPAN_TYPE_PREFIX + "okhttp";
+
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    private static void onBeforeExecute( @Advice.FieldValue(value = "originalRequest", typing = Assigner.Typing.DYNAMIC, readOnly = false) @Nullable Object originalRequest,
+                                         @Advice.Local("span") Span span) throws IOException {
+
+        if (tracer == null || tracer.activeSpan() == null) {
+            return;
+        }
+        final AbstractSpan<?> parent = tracer.activeSpan();
```
eyalkoren: This API is about to change so it will require a small resolving

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientInstrumentation.java
```diff
@@ -0,0 +1,100 @@
...
+
+        if (originalRequest == null) {
+            return;
+        }
+
+        if (originalRequest instanceof com.squareup.okhttp.Request) {
```
eyalkoren: We can't have dependency on both `com.squareup.okhttp.Request` and `okhttp3.Request` at the same class/method, as they most likely won't be both available in the classpath in real time, which means this class will not be resolved properly. We need to have one Instrumentation class for each version

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientInstrumentation.java
```diff
@@ -0,0 +1,100 @@
...
+            com.squareup.okhttp.Request request = (com.squareup.okhttp.Request) originalRequest;
+            span = HttpClientHelper.startHttpClientSpan(parent, request.method(), request.uri(), request.url().getHost(), SPAN_TYPE_OK_HTTP_CLIENT);
+            originalRequest = ((com.squareup.okhttp.Request) originalRequest).newBuilder().addHeader(TraceContext.TRACE_PARENT_HEADER, span.getTraceContext().getOutgoingTraceParentHeader().toString()).build();
```
eyalkoren: I imagine this produces garbage and side effects, what about the other solution of adding the header during request build? It will complicate things but will be safer and more performant

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientInstrumentation.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
+    public static void onAfterExecute(@Advice.Local("span") @Nullable Span span,
+                                      @Advice.Thrown @Nullable Throwable t) {
+        if (span != null) {
+            span.captureException(t)
+                .deactivate()
+                .end();
```
eyalkoren: Try to get the status code and set in `span.context.http.statusCode`

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientInstrumentation.java
nugusbayevkk: i tried add header before executing of RequestBuilder.build, but that method called twice, i wrote this in https://github.com/elastic/apm-agent-java/issues/329 .

eyalkoren: Looks like when you call `org.testcontainers.shaded.okhttp3.Request.Builder#header` method, it invokes the `org.testcontainers.shaded.okhttp3.Headers.Builder#set` method, which seems to replace former value, so shouldn't be a problem if called twice.
I didn't look at version 2, but try to look for the same thing.
Some other options: 
1. look into headers and see if already added
2. try to remove and add
3. flag after first header addition somehow, through request or ThreadLocal for example

nugusbayevkk:
@eyalkoren please check changes.

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpRequestBuilderInstrumentation.java
```diff
@@ -0,0 +1,77 @@
...
+            return;
+        }
+        if (thiz == null || headers == null) {
+            return;
+        }
+        final AbstractSpan<?> currentSpan = tracer.currentSpan();
+        if (thiz instanceof Request.Builder) {
+            if (headers instanceof Headers.Builder) {
+                if (((Headers.Builder) headers).get(TraceContext.TRACE_PARENT_HEADER) == null) {
+                    ((Headers.Builder) headers).set(TraceContext.TRACE_PARENT_HEADER, currentSpan.getTraceContext().getOutgoingTraceParentHeader().toString());
```
eyalkoren: This probably happens before the actual span is started (in `Call.execute`), so the `parentId` part of the outgoing header would not be the one we want to send. We need some APIs we are now missing. For example, a way to create a child `TraceContext` of the active span and use it for the header creation, then make it available for the to-be-created span. We are currently missing both of these internal APIs- create a standalone context and start a span with a readymade context. Let me see think about that and see how we do that.
In any case, since we are not entirely sure there will be a span at all, we will need to make sure this does not leak.

Regardless, I just looked at `okhttp3.Request` and it seems to have a package-private constructor, so it seems to be safe to assume it is always built through the Builder. Is it the same for okhttp 2 as well? Is there another standard way to create `Request`s?

eyalkoren:
@nugusbayevkk regarding the add the header addition - as using the build-instrumentation approach would greatly complicate things, let's go back to your original suggestion of creating a new builder from the request, adding the header and create a new request and live with these allocations. If we see it is a problem in the future we will revisit. Sorry for the hassle and thanks for taking the time to investigate. 
I believe that once you revert to that and apply the other suggestions (mostly restructure)- we can merge.

nugusbayevkk:
@eyalkoren I changed to previous version and addressed your comments

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttp3ClientInstrumentation.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return named("execute").and(returns(named("okhttp3.Response")));
+    }
+
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Arrays.asList("http-client", "okhttp3");
```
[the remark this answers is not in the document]
nugusbayevkk: ok, i will update it soon.

nugusbayevkk:
image
"Span detail" of okhttp client execute call:
image
and "Span detail" of okhttp3 client execute call:
image

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 
 ## Bug Fixes
 
+# 1.3.1
```
felixbarny: remove

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
...
+
+## Features
```
felixbarny: remove

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
...
+ * Added support for sync calls of OkHttp client
```
felixbarny: move to `next > Features`:

```
# next

## Features

 * Added support for sync calls of OkHttp client
```

apm-agent-plugins/apm-okhttp-plugin/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+        <groupId>co.elastic.apm</groupId>
+        <version>1.3.1-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>apm-okhttp-plugin</artifactId>
+    <name>${project.groupId}:${project.artifactId}</name>
+
+    <dependencies>
+        <dependency>
```
felixbarny: this dependency can be removed as it's inherited from `apm-agent-plugins/pom.xml`

apm-agent-plugins/apm-okhttp-plugin/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp</artifactId>
+            <version>${version.okhttp}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
```
felixbarny: I think this dependency is unused, can it be removed?

apm-agent-plugins/apm-okhttp-plugin/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+            <artifactId>logging-interceptor</artifactId>
+            <version>${version.okhttp}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.squareup.okio</groupId>
```
felixbarny: Isn't okio a transitive dependency of okhttp? Why does it have to be declared explicitly here?

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttp3ClientInstrumentation.java
```diff
@@ -0,0 +1,108 @@
...
+            if (originalRequest == null) {
+                return;
+            }
+
+            final TraceContextHolder<?> parent = tracer.getActive();
+
+            if (originalRequest instanceof okhttp3.Request) {
+                okhttp3.Request request = (okhttp3.Request) originalRequest;
+                span = HttpClientHelper.startHttpClientSpan(parent, request.method(), request.url().uri(), request.url().host(), SPAN_TYPE_OK_HTTP_CLIENT);
+                originalRequest = ((okhttp3.Request) originalRequest).newBuilder().addHeader(TraceContext.TRACE_PARENT_HEADER, span.getTraceContext().getOutgoingTraceParentHeader().toString()).build();
```
felixbarny: Add a null check for span. `HttpClientHelper.startHttpClientSpan` may return `null`.

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientInstrumentation.java
```diff
@@ -0,0 +1,109 @@
...
+            final TraceContextHolder<?> parent = tracer.getActive();
+
+            if (originalRequest == null) {
+                return;
+            }
+
+            if (originalRequest instanceof com.squareup.okhttp.Request) {
+                com.squareup.okhttp.Request request = (com.squareup.okhttp.Request) originalRequest;
+                span = HttpClientHelper.startHttpClientSpan(parent, request.method(), request.uri(), request.url().getHost(), SPAN_TYPE_OK_HTTP_CLIENT);
+                originalRequest = ((com.squareup.okhttp.Request) originalRequest).newBuilder().addHeader(TraceContext.TRACE_PARENT_HEADER, span.getTraceContext().getOutgoingTraceParentHeader().toString()).build();
```
felixbarny: same, please add a null check

apm-agent-plugins/apm-okhttp-plugin/src/test/java/co/elastic/apm/agent/okhttp/OkHttpClientInstrumentationTest.java
```diff
@@ -0,0 +1,46 @@
...
+    public void setUp() {
+        client = new OkHttpClient();
+    }
+
+    @Override
+    protected void performGet(String path) throws Exception {
+        Request request = new Request.Builder()
+            .url(path)
+            .build();
+        client.newCall(request).execute();
```
felixbarny:
```suggestion
        client.newCall(request).execute().close();
```

apm-agent-plugins/apm-okhttp-plugin/src/test/java/co/elastic/apm/agent/okhttp/OkHttp3ClientInstrumentationTest.java
```diff
@@ -0,0 +1,45 @@
...
+    public void setUp() {
+        client = new OkHttpClient();
+    }
+
+    @Override
+    protected void performGet(String path) throws Exception {
+        Request request = new Request.Builder()
+            .url(path)
+            .build();
+        client.newCall(request).execute();
```
felixbarny:
```suggestion
        client.newCall(request).execute().close();
```

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientInstrumentation.java
```diff
@@ -0,0 +1,109 @@
...
+    public Class<?> getAdviceClass() {
+        return OkHttpClientExecuteAdvice.class;
+    }
+
+    @VisibleForAdvice
+    public static class OkHttpClientExecuteAdvice {
+
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        private static void onBeforeExecute( @Advice.FieldValue(value = "originalRequest", typing = Assigner.Typing.DYNAMIC, readOnly = false) @Nullable Object originalRequest,
+                                             @Advice.Local("span") Span span) throws IOException {
```
felixbarny: remove unnecessary throws declaration
```suggestion
                                             @Advice.Local("span") Span span) {
```

nugusbayevkk: exception is thrown by com.squareup.okhttp.Request.uri() .

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:compile (default-compile) on project apm-okhttp-plugin: Compilation failure
[ERROR] /home/nugusbayevkk/Documents/PROJECTS/elastic/apm-agent-java/apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientInstrumentation.java:[70,98] unreported exception java.io.IOException; must be caught or declared to be thrown

apm-agent-plugins/apm-okhttp-plugin/src/test/java/co/elastic/apm/agent/okhttp/OkHttpClientInstrumentationTest.java
nugusbayevkk: method close does not exist.

apm-agent-plugins/apm-okhttp-plugin/pom.xml
nugusbayevkk: it's not necessary anymore.

apm-agent-plugins/apm-okhttp-plugin/src/test/java/co/elastic/apm/agent/okhttp/OkHttp3ClientInstrumentationTest.java
nugusbayevkk: added call method ResponseBody.close().

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientInstrumentation.java
```diff
@@ -0,0 +1,111 @@
...
+            }
+            final TraceContextHolder<?> parent = tracer.getActive();
+
+            if (originalRequest == null) {
+                return;
+            }
+
+            if (originalRequest instanceof com.squareup.okhttp.Request) {
+                com.squareup.okhttp.Request request = (com.squareup.okhttp.Request) originalRequest;
+                span = HttpClientHelper.startHttpClientSpan(parent, request.method(), request.uri(), request.url().getHost(), SPAN_TYPE_OK_HTTP_CLIENT);
```
felixbarny: By calling `request.httpUrl().host()` you save an allocation of `java.net.URL`
```suggestion
                span = HttpClientHelper.startHttpClientSpan(parent, request.method(), request.uri(), request.httpUrl().host(), SPAN_TYPE_OK_HTTP_CLIENT);
```

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttp3ClientInstrumentation.java
```diff
@@ -0,0 +1,110 @@
...
+
+            if (originalRequest == null) {
+                return;
+            }
+
+            final TraceContextHolder<?> parent = tracer.getActive();
+
+            if (originalRequest instanceof okhttp3.Request) {
+                okhttp3.Request request = (okhttp3.Request) originalRequest;
+                span = HttpClientHelper.startHttpClientSpan(parent, request.method(), request.url().uri(), request.url().host(), SPAN_TYPE_OK_HTTP_CLIENT);
```
felixbarny: A disadvantage of calling `request.url().uri()` is that it allocates a new builder, reencodes the uri and allocates an URI. It could also throw a Runtime exception if there is a problem while encoding. An alternative could be `request.url().toString()` but that one does also include the query parameters. I don't think we'd want that. So I guess for now this is good enough but we should keep it in mind that there is a potential for improvement.

nugusbayevkk: right now HttpClientHelper.startHttpClientSpan 3rd argument requires URI object, so should we pass argument as new URI(request.url().toString()) ? Also here we can catch java.net.URISyntaxException.

felixbarny:
Let's wait for the CI then we can merge :)
Thanks for going through all the suggestions 👍

PR #422, opened 2019-01-14
Jsf support

eyalkoren:
Extending Servlet instrumentation to include all implementations of `javax.servlet.Servlet`
JSF transaction name is the path, which typically indicates the Facelet view
In addition, Lifecycle `execute` and `render` are captured
Extended the integration test suite to facilitate addition of new test apps (and added a basic JSF app)
Capturing looks like this:
screen shot 2019-01-14 at 14 28 18


apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletApiAdvice.java
```diff
@@ -134,11 +134,18 @@ public static void onExitServletService(@Advice.Argument(0) ServletRequest servl
...
         }
-        if (thiz instanceof HttpServlet && servletRequest instanceof HttpServletRequest) {
+        if (servletRequest instanceof HttpServletRequest) {
             Transaction currentTransaction = tracer.currentTransaction();
             if (currentTransaction != null) {
                 final HttpServletRequest httpServletRequest = (HttpServletRequest) servletRequest;
-                ServletTransactionHelper.setTransactionNameByServletClass(httpServletRequest.getMethod(), thiz.getClass(), currentTransaction.getName());
+                if (thiz instanceof HttpServlet) {
+                    ServletTransactionHelper.setTransactionNameByServletClass(httpServletRequest.getMethod(), thiz.getClass(), currentTransaction.getName());
+                } else if ("javax.faces.webapp.FacesServlet".equals(thiz.getClass().getName())) {
```
felixbarny: Could this be done in the lifecycle instrumentation or by instrumenting `javax.faces.webapp.FacesServlet` directly? I would prefer this class not to have knowledge of specific frameworks. Ideally, that knowledge should only be in the plugins.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletInstrumentation.java
```diff
@@ -63,14 +63,16 @@ public void init(ElasticApmTracer tracer) {
...
-            .and(hasSuperType(named("javax.servlet.http.HttpServlet")));
+            .and(hasSuperType(named("javax.servlet.Servlet")));
     }
 
     @Override
     public ElementMatcher<? super MethodDescription> getMethodMatcher() {
         return named("service")
-            .and(takesArgument(0, named("javax.servlet.http.HttpServletRequest")))
-            .and(takesArgument(1, named("javax.servlet.http.HttpServletResponse")));
+            .and(
```
felixbarny: Doesn't that mean that HTTP servlets are instrumented twice now? 
- `javax.servlet.http.HttpServlet#service(javax.servlet.ServletRequest, javax.servlet.ServletResponse)`
- `javax.servlet.http.HttpServlet#service(javax.servlet.http.HttpServletRequest, javax.servlet.http.HttpServletResponse)`

felixbarny:
Instead of checking in the binaries, couldn't we have two versions of the webapp? One targeted for application servers which ship their own implementation of JSF and one for servlet containers like Tomcat, Jetty and Undertow.

We could have a `simple-webapp-servlet-container` module (or similar) which has a dependency on `simple-webapp` but additionally has dependencies to the JSF implementations.

I feat that after some time more and more binaries would get checked in, slowing down git operations.

apm-agent-plugins/apm-jsf-plugin/src/main/java/co/elastic/apm/agent/jsf/JsfLifecycleInstrumentation.java
```diff
@@ -0,0 +1,160 @@
...
+                } finally {
+                    span.deactivate().end();
+                }
+
+            }
+        }
+    }
+
+    public static class JsfLifecycleRenderInstrumentation extends JsfLifecycleInstrumentation {
+        private static final String SPAN_TYPE = "jsf-render";
```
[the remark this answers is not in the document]
eyalkoren: Right, added it as a placeholder waiting for the span type thing and forgot about it. Good catch!
What does `template` stands for though?

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletApiAdvice.java
eyalkoren: I would prefer it won't be here too. I didn't do it in a separate advice as it either means no general instrumentation for `javax.servlet.Servlet` or chained instrumentation of code from two advices, where order matters.
However, doing it from the Lifecycle is a good idea! I will do that.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletInstrumentation.java
eyalkoren: Good catch!

apm-agent-plugins/apm-jsf-plugin/src/main/java/co/elastic/apm/agent/jsf/JsfLifecycleInstrumentation.java
```diff
@@ -0,0 +1,160 @@
...
+    public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
+        return not(isBootstrapClassLoader())
+            .and(classLoaderCanLoadClass("javax.faces.lifecycle.Lifecycle"));
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return hasSuperType(named("javax.faces.lifecycle.Lifecycle"));
+    }
+
```
felixbarny: Maybe add

```
    @Override
        return nameContains("Lifecycle");
    }
```

to speed up type matching?

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/HttpServletInstrumentation.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Instruments HTTP servlets to create transactions.
+ * <p>
+ * If the transaction has already been recorded with the help of {@link FilterChainInstrumentation},
+ * it does not record the transaction again.
+ * But if there is no filter registered for a servlet,
+ * this makes sure to record a transaction in that case.
+ * </p>
+ */
+public class HttpServletInstrumentation extends ElasticApmInstrumentation {
```
felixbarny: Why do we need `HttpServletInstrumentation` and `ServletInstrumentation`?

eyalkoren: Wasn't this what you requested?

felixbarny: Are you referring to https://github.com/elastic/apm-agent-java/pull/422#discussion_r248605696? Now, `HttpServlet`s are still instrumented twice, aren't they? I think we only need the `ServletInstrumentation`.

eyalkoren: Right, always calls from it, removing .....

felixbarny:
Since merging this PR, the integration tests are failing on pipeline Jenkins.

//cc @kuisathaverat

PR #428, opened 2019-01-17
Allow package-only matchers

felixbarny:


felixbarny:
A last set of suggested changes: https://github.com/nugusbayevkk/apm-agent-java/pull/1

If you like them, just merge that PR. I guess then we can also merge this PR :)

PR #432, opened 2019-01-19
Don't trace proxies and type initializers for trace_methods

felixbarny:
Also clarify documentation that package matching is recursive


apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -260,7 +261,7 @@ public static String getAllInstrumentationGroupNames() {
             " - `org.example.MyClass#myMe*od(java.lang.String, int)`\n" +
             " - `private org.example.MyClass#myMe*od(java.lang.String, *)`\n" +
             " - `* org.example.MyClas*#myMe*od(*.String, int[])`\n" +
-            " - `public org.example.services.*.*Service#*`\n" +
+            " - `public org.example.services.*Service#*`\n" +
             "\n" +
             "NOTE: Only use wildcards if necessary.\n" +
             "The more methods you match to more overhead will be caused by the agent.\n" +
```
eyalkoren:
```suggestion
            "The more methods you match the more overhead will be caused by the agent.\n" +
```

PR #435, opened 2019-01-21
Disable stack traces for trace_methods spans

felixbarny:


felixbarny:
> So we never collect? It may be a good default, but don't you want that configurable?

Yes, eventually it probably makes sense to let the user configure it. But currently, it's still unclear how. Until we decide on that, not capturing stack traces at all for `trace_method` spans seems like a better default.

felixbarny:
Some ideas are:
* having a configuration option which enables/disables stack trace collection for all `trace_methods` spans
* extend the syntax of `trace_methods` to configure stack trace collection
* disable stack trace collection if the corresponding `trace_methods` matcher contains a wildcard

felixbarny:
We decided not to do this but instead mention the potential overhead in the `trace_methods` docs and link to the [`span_frames_min_duration`](https://www.elastic.co/guide/en/apm/agent/java/current/config-stacktrace.html#config-span-frames-min-duration).

PR #437, opened 2019-01-22
Disable jvm.gc metrics

felixbarny:
The APM 6.6.0 UI has problems handling metrics with tags.
So for the moment tagged metrics will be disabled.
Once you are on Kibana 6.6.1, you can safely enable the metrics again.


eyalkoren:
Configuration looks good, but just saw the build failed due to NPE...

felixbarny:
Jenkins retest this please

PR #447, opened 2019-01-25
Set type for unsampled transactions

felixbarny:


felixbarny:
Because we report non-sampled transactions, we need to set their type. Otherwise, they are not properly shown in the UI.

Issue #357, opened 2018-11-28
Search for a concurrent map with weak keys


felixbarny:
Requirements
- Weak keys
- Thread safe (Not `WeakHashMap`)
- Highly performant in multi-threaded scenarios (not Guava or `Collections.synchronizedMap`)
- No allocations on `get()` (not [`WeakConcurrentMap`](https://github.com/raphw/weak-lock-free/blob/master/src/main/java/com/blogspot/mydailyjava/weaklockfree/WeakConcurrentMap.java#L51))

felixbarny:
Relates to #317

tylerbenson:
Good luck.  You basically excluded all the options I am aware of.

felixbarny:
Hehe, thx :smile:
What I want to test is the map backing caffeine cache.

ben-manes:
Caffeine is like `WeakConcurrentMap` by allocating a wrapper for the key comparison. I think this isn't a problem because of escape analysis, since the object is on the same thread. But I don't recall testing that and escape analysis may not work across method cals. However the allocation should be cheap due to short lived. Your best bet is to write a JMH benchmark. The big difference is that Caffeine does not require create any dedicated threads and performs maintenance cycles after a threshold of operations occur.

Guava gets around that by forking the hash table so that it can use a pluggable `Equivalence`. However, that is a big commitment which they did in a misguided enthusiasm towards soft references. In parallel I was exploring concurrent LRU and, when reality set in on the Guava team, we tried to port my work into theirs. So Guava is optimized for reference caching first, whereas Caffeine is size first. We could fork the hash table just for reference caching, but I really don't want to commit towards maintaining even more complexity.

felixbarny:
Thanks a lot for your input.

> I think this isn't a problem because of escape analysis, since the object is on the same thread.

In my benchmarks, the escape analysis didn't seem to kick in, unfortunately.

> However the allocation should be cheap due to short lived.

True, but if an agent increases the allocation rate of short-lived objects, it may push the GC over its limits when the monitored application itself allocates so much that the GC is just able to keep up. In such situations, short-lived objects might end up in the old gen, increasing the frequency and duration of major GCs. That's one reason why I'm striving for a garbage free agent.

[...]

PR #448, opened 2019-01-25
Use WeakConcurrentMap instead of Guava

felixbarny:
closes elastic/apm-agent-java#357


felixbarny:
I'm merging now so I can move forward with #430. Feel free to propose changes though.

PR #454, opened 2019-01-27
add span.subtype and span.action

eyalkoren:
Implemented according to https://github.com/elastic/apm/issues/8#issuecomment-454433475


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -84,11 +105,59 @@ public Span withName(@Nullable String name) {
         return this;
     }
 
+    /**
+     * Keywords of specific relevance in the span's domain (eg: 'db', 'template', 'ext', etc)
+     */
+    public Span withType(@Nullable String type) {
+        if (!isSampled()) {
```
felixbarny: I think checking for sampled is not really necessary, I'd remove it. It can lead to problems (see #447). If we want to track metrics for spans we also might expect the type to be set.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
```diff
@@ -104,6 +111,17 @@ public Transaction withName(@Nullable String name) {
         return this;
     }
 
+    /**
+     * Keyword of specific relevance in the service's domain (eg:  'request', 'backgroundjob')
+     */
+    public Transaction withType(@Nullable String type) {
+        if (!isSampled()) {
```
felixbarny: remove (see #447)

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
```diff
@@ -524,10 +524,44 @@ private void serializeSpan(final Span span) {
...
 
+    /**
+     * TODO: remove in 2.0
+     * To be removed for agents working only with APM server 7.0 or higher, where schema contains span.type, span.subtype and span.action
+     * @param span serialized span
+     */
+    private void serializeSpanType(Span span) {
+        writeFieldName("type");
+        String type = span.getType();
+        if (type != null) {
```
felixbarny: If only the type is set, you can serialize it without checking for dots. You can then also remove the Utils.

```
if (type != null && subtype == null && action == null) {
    writeStringValue(type);
}
```

eyalkoren: That's what I thought of doing, but didn't seem right. Our model now includes these three fields for span and our entire code should assume that from now going forward. For example, we may decide to collect metrics for each DB type separately, or whatever. So an old API usage is supported, but I fit it into the model as soon as possible.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
eyalkoren: will do

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/plugin/api/AbstractSpanInstrumentation.java
```diff
@@ -80,8 +83,27 @@ public SetTypeInstrumentation() {
...
+
+    public static class SetTypesInstrumentation extends AbstractSpanInstrumentation {
+        public SetTypesInstrumentation() {
+            super(named("doSetTypes"));
+        }
+
+        @VisibleForAdvice
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        public static void setType(@Advice.Argument(0) Object span,
+                                   @Advice.Argument(1) String type,
```
felixbarny: I'd mark type, subtype and action as `@Nullable`.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
eyalkoren: Started with that as well 😄 (I mean- three setters), but it also didn't feel right, as it is only true now when I do this change. Otherwise, type, subtype and action are separate parts of the model, where you can set either elsewhere and in different times.
I got a good example for that when implemented the setting-through-tags for the OT bridge- each is done separately.
So I rather stick with that, but we can talk about it more. 
Anyway, I will add it as a static method to Span

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
eyalkoren: removing for `withSubtype` and `withAction` as well

felixbarny:
Also, document the new tags in `opentracing.asciidoc` and add a changelog.

docs/public-api.asciidoc
```diff
@@ -355,31 +355,66 @@ preferably before other JS libraries:
 
 See the {apm-rum-ref}[JavaScript RUM agent documentation] for more information.
 
+[float]
+[[api-transaction-start-span-with-type]]
+==== `Span startSpan(String type, String subtype, String action)`
```
felixbarny: Add subtype and action docs to `@CaptureSpan`

apm-agent-plugins/apm-httpclient-core/src/main/java/co/elastic/apm/agent/http/client/HttpClientHelper.java
```diff
@@ -27,23 +27,25 @@
...
-    public static Span startHttpClientSpan(TraceContextHolder<?> parent, String method, @Nullable String uri, String hostName, String spanType) {
+    public static Span startHttpClientSpan(TraceContextHolder<?> parent, String method, @Nullable String uri, String hostName, String client) {
         Span span = null;
         if (!isAlreadyMonitored(parent)) {
             span = parent
                 .createSpan()
-                .withType(spanType)
+                .withType(EXTERNAL_TYPE)
+                .withSubtype(HTTP_SUBTYPE)
+                .withAction(client)
```
felixbarny: The client is not really an action. Maybe just leave it out.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -31,6 +31,27 @@
 
     private static final Logger logger = LoggerFactory.getLogger(Span.class);
 
+    /**
+     * General type describing this span (eg: 'db', 'ext', 'template', etc)
+     * (Required)
+     */
+    @Nullable
+    private volatile String type;
```
felixbarny: Why do the types have to be `volatile`?

docs/public-api.asciidoc
eyalkoren: ooooff, right 🤦‍♂️

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
eyalkoren: Basically everything in all reported objects should be volatile because set by one or more threads and read (for serialization) by a different thread.
However, since we are not doing that now, and since those are only populated on creation- I will remove

felixbarny: The disruptor ensures visibility on the reporter thread. But for some properties it's important that when a span is created on one thread and ended on another one, visibility is ensured. The `finished` flag is an example for that.

felixbarny:
I'm fine with that. We can still change our minds at a later point if we want.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
```diff
@@ -51,6 +51,13 @@
      */
     private boolean noop;
 
+    /**
+     * Keyword of specific relevance in the service's domain (eg:  'request', 'backgroundjob')
+     * (Required)
+     */
+    @Nullable
+    private volatile String type;
```
felixbarny: another unnecessary volatile

eyalkoren: No, this one may in fact be necessary

Issue #445, opened 2019-01-24
Rejected metric events due to NaN


felixbarny:
```
co.elastic.apm.agent.report.IntakeV2ReportingEventHandler - {"accepted":1,"errors":[{"message":"Problem validating JSON document against schema: I[#] S[#] doesn't validate with \"metricset#\"\n  I[#] S[#/allOf/0] allOf failed\n    I[#/samples/system.cpu.total.norm.pct/value] S[#/allOf/0/properties/samples/patternProperties/%5E%5B%5E%2A%22%5D%2A$/properties/value/type] expected number, but got string","document":"{\"metricset\":{\"timestamp\":1548335970013000,\"samples\":{\"system.process.cpu.total.norm.pct\":{\"value\":0.08920415620004567},\"jvm.memory.heap.used\":{\"value\":471631432.0},\"jvm.memory.non_heap.used\":{\"value\":97109528.0},\"jvm.memory.heap.max\":{\"value\":3817865216.0},\"system.process.memory.size\":{\"value\":10466623488.0},\"jvm.memory.non_heap.committed\":{\"value\":101154816.0},\"jvm.memory.heap.committed\":{\"value\":1030750208.0},\"jvm.memory.non_heap.max\":{\"value\":-1.0},\"system.memory.actual.free\":{\"value\":107765760.0},\"system.memory.total\":{\"value\":17179869184.0},\"system.cpu.total.norm.pct\":{\"value\":\"NaN\"}}}}"}]}
```

The problem is that `system.cpu.total.norm.pct` is `{\"value\":\"NaN\"}`

The value is read from this MBean: `https://docs.oracle.com/javase/7/docs/jre/api/management/extension/com/sun/management/OperatingSystemMXBean.html#getSystemCpuLoad()`

Questions:
- What's the correct way to encode `NaN` in JSON
- Can the APM Server deal with `NaN` values?
- What's the implications in the APM UI?
- Should we omit `NaN` values or serialize as `0`?

This seems to happen when the monitored system is idle. When applying load, the value is non-`NaN` again. Tested with Java 8 on Mac OS.

jalvz:
Here `NaN` is encoded as a string (`"NaN"`) and therefore rejected as we only allow numbers.
JSON spec doesn't contemplate `NaN` afaik, so the best think would be to filter those out (nulls are ok, and the UI should know how to treat them).
[...]

PR #461, opened 2019-01-29
Serializing null for non finit metric values

eyalkoren:


felixbarny:
Seems like the APM Server does not like null values:

```
java.io.IOException: Server returned HTTP response code: 400 for URL: https://apm.europe-west3.gcp.cloud.es.io:443/intake/v2/events
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1894)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1492)
	at sun.net.www.protocol.https.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:263)
	at co.elastic.apm.agent.report.IntakeV2ReportingEventHandler.flush(IntakeV2ReportingEventHandler.java:283)
	at co.elastic.apm.agent.report.IntakeV2ReportingEventHandler.handleEvent(IntakeV2ReportingEventHandler.java:152)
	at co.elastic.apm.agent.report.IntakeV2ReportingEventHandler.onEvent(IntakeV2ReportingEventHandler.java:140)
	at co.elastic.apm.agent.report.IntakeV2ReportingEventHandler.onEvent(IntakeV2ReportingEventHandler.java:58)
	at co.elastic.apm.agent.shaded.lmax.disruptor.BatchEventProcessor.processEvents(BatchEventProcessor.java:168)
	at co.elastic.apm.agent.shaded.lmax.disruptor.BatchEventProcessor.run(BatchEventProcessor.java:125)
2019-02-22 19:03:39.195 [apm-reporter] WARN co.elastic.apm.agent.report.IntakeV2ReportingEventHandler - {"accepted":14,"errors":[{"message":"Problem validating JSON document against schema: I[#] S[#] doesn't validate with \"metricset#\"\n  I[#] S[#/allOf/0] allOf failed\n    I[#/samples/system.cpu.total.norm.pct/value] S[#/allOf/0/properties/samples/patternProperties/%5E%5B%5E%2A%22%5D%2A$/properties/value/type] expected number, but got null","document":"{\"metricset\":{\"timestamp\":1550858576638000,\"samples\":{\"system.process.cpu.total.norm.pct\":{\"value\":0.125},\"jvm.memory.heap.used\":{\"value\":1071700752.0},\"jvm.memory.non_heap.used\":{\"value\":93600808.0},\"jvm.memory.heap.max\":{\"value\":3817865216.0},\"jvm.gc.alloc\":{\"value\":4364208480.0},\"jvm.memory.non_heap.committed\":{\"value\":97607680.0},\"system.cpu.total.norm.pct\":{\"value\":null},\"system.process.memory.size\":{\"value\":10463731712.0},\"jvm.memory.heap.committed\":{\"value\":1654652928.0},\"jvm.memory.non_heap.max\":{\"value\":-1.0},\"system.memory.actual.free\":{\"value\":103964672.0},\"system.memory.total\":{\"value\":17179869184.0}}}}"}]}
```

eyalkoren:
@elastic/apm-server we did this (sending `null` in metrics when the infrastructure query returns `NaN`) based on #445. Is there something we should do to fix the error we get above or would you change to allow `null`s? (the latter should be validated with UI as well I guess).

simitt:
When sending a sample, a [value is required](https://github.com/elastic/apm-server/blob/master/docs/spec/metricsets/sample.json#L9), so sending _null_ values for samples is not allowed. This is consistent handling throughout the Intake API: required attributes aren't allowed to be null, optional attributes are. 
I assume the `value` has been set as _required_, as it wouldn't make sense to send up a metric sample without a value (@graphaelli please correct me if I'm wrong). From my perspective the Intake API makes sense as is, and it probably also makes sense from an agent perspective to not send data that won't be stored anyways. However, if this is a problem for any agent, I don't see a problem loosening this and making the value optional and therefore nullable.

eyalkoren:
@simitt to me it makes perfect sense to avoid sending it when we have no value if this is not stored anyway, as long as the server doesn't expect to get specific metric keys. 

@elastic/apm-ui just to make sure- if we omit specific metric samples occasionally (when we don't have a valid value for them)- would aggregations made for graph creation be affected by the lack of documents?

sqren:
@eyalkoren This shouldn't affect the UI too much. Obviously there'll be fewer data points, so instead of having a datapoint every 30s there'll be a gap of 60s. The UI should handle this fine, and it is only if the user zooms in a lot that they'll notice it.

eyalkoren:
Issue above fixed by #501 - avoid sending samples with `NaN` instead of sending `null`

PR #464, opened 2019-01-31
[ci] add clean arg to maven wrapper call

jonahbull:


felixbarny:
If the `packer_cache.sh` script exits w/a non-zero return code it'll retry up to 5 times before giving up. This is why we have to make sure to always clean.

PR #472, opened 2019-02-01
Fix multi-threaded use of HttpUrlConnection

### LINKED ISSUE #467: HttpUrlConnection instrumentation in multi-threaded use cases
When starting a span in one therad via `HttpUrlConnection#connect()` and ending it in another thread via `HttpUrlConnection#getInputStream()`, the span's activation and deactivation happen in different threads, corrupting the activation state.

felixbarny:
fixes elastic/apm-agent-java#467

### DIFF
--- a/apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
+++ b/apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
@@ -20,10 +20,11 @@
 package co.elastic.apm.agent.urlconnection;
 
 import co.elastic.apm.agent.bci.ElasticApmInstrumentation;
+import co.elastic.apm.agent.bci.VisibleForAdvice;
 import co.elastic.apm.agent.http.client.HttpClientHelper;
 import co.elastic.apm.agent.impl.transaction.Span;
 import co.elastic.apm.agent.impl.transaction.TraceContext;
-import co.elastic.apm.agent.impl.transaction.TraceContextHolder;
+import com.blogspot.mydailyjava.weaklockfree.WeakConcurrentMap;
 import net.bytebuddy.asm.Advice;
 import net.bytebuddy.description.NamedElement;
 import net.bytebuddy.description.method.MethodDescription;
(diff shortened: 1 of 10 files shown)


apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
```diff
@@ -63,43 +67,52 @@
...
                                 @Nullable @Advice.Thrown Throwable t,
-                                @Advice.FieldValue("responseCode") int responseCode) {
-            if (tracer == null) {
+                                @Advice.FieldValue("responseCode") int responseCode,
+                                @Nullable @Advice.Local("span") Span span,
+                                @Advice.Origin String signature) {
+            if (span == null) {
                 return;
+            } else {
             }
```
eyalkoren: ?

apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
```diff
@@ -121,15 +134,9 @@ public static void exit(@Advice.This HttpURLConnection thiz,
...
-            if (tracer == null) {
-                return;
-            }
-            final TraceContextHolder<?> active = tracer.getActive();
-            if (active instanceof Span) {
-                final Span span = (Span) active;
-                if (span.isOriginatedBy(thiz)) {
-                    span.captureException(t).deactivate().end();
-                }
+            Span span = inFlightSpans.get(thiz);
```
eyalkoren: Don't we want to remove from the map?

eyalkoren:
Does it fix the Tomcat 7 IT?

apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
felixbarny: good catch!

apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
felixbarny: very good question 😄

felixbarny:
> Does it fix the Tomcat 7 IT?
No, I don't think so. The Tomcat test failure was related to the Executor instrumentation.

PR #492, opened 2019-02-17
Adding HelperClassManager for any class loader

eyalkoren:
Contains integer boxing, but I believe it is simple, robust, performant and safe enough for us to use it.

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
@@ -257,7 +257,7 @@ public abstract class HelperClassManager<T> {
         }
     }
 
-    static Class injectClass(@Nullable ClassLoader targetClassLoader, ProtectionDomain pd, String className, boolean isBootstrapClass) throws IOException, ClassNotFoundException {
+    static Class injectClass(@Nullable ClassLoader targetClassLoader, @Nullable ProtectionDomain pd, String className, boolean isBootstrapClass) throws IOException, ClassNotFoundException {
         if (targetClassLoader == null) {
             if (isBootstrapClass) {
                 return Class.forName(className, false, null);
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
@@ -20,6 +20,7 @@
 package co.elastic.apm.agent.bci;
 
 import co.elastic.apm.agent.impl.ElasticApmTracer;
+import com.blogspot.mydailyjava.weaklockfree.WeakConcurrentMap;
 import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.dynamic.ClassFileLocator;
 import net.bytebuddy.dynamic.loading.ByteArrayClassLoader;
@@ -36,10 +37,9 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.WeakHashMap;
 
 /**
  * This class helps to overcome the fact that the agent classes can't access the classes they want to instrument.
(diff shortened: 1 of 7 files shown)


eyalkoren:
Jenkins test this please

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
```diff
@@ -128,52 +167,169 @@ public T getForClassLoaderOfClass(Class<?> classOfTargetClassLoader) {
...
+            failed = true;
+        }
+    }
+
+    public static class ForAnyClassLoader<T> extends HelperClassManager<T> {
+
+        // doesn't need to be concurrent - invoked only from a synchronized context
+        static final Map<Integer, WeakReference<List<Object>>> clId2helperImplListMap = new HashMap<>();
+
+        final ConcurrentHashMap<Integer, WeakReference<T>> clId2helperMap;
```
felixbarny: Why don't you use WeakConcurrentMaps here?

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
```diff
@@ -128,52 +167,169 @@ public T getForClassLoaderOfClass(Class<?> classOfTargetClassLoader) {
...
+        final ConcurrentHashMap<Integer, Boolean> failedClassLoaderSet;
```
felixbarny: Have you considered `Collections.newSetFromMap` or `WeakConcurrentSet` here?

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
```diff
@@ -128,52 +167,169 @@ public T getForClassLoaderOfClass(Class<?> classOfTargetClassLoader) {
...
             }
         }
+    }
 
-        private static List<String> asList(String implementation, String[] additionalHelpers) {
-            final ArrayList<String> list = new ArrayList<>(additionalHelpers.length + 1);
-            list.add(implementation);
-            list.addAll(Arrays.asList(additionalHelpers));
-            return list;
+    static Class injectClass(@Nullable ClassLoader targetClassLoader, ProtectionDomain pd, String className, boolean isBootstrapClass) throws IOException, ClassNotFoundException {
```
felixbarny: I'm not sure I got this right. Is my understanding of this correct?
This injects the `HelperHolder` directly into the target CL, right? You do this so that there is a strong reference from the target CL to the helper classes, which are loaded from a child of the target CL. This prevents the helper classes to be collected, because they are otherwise only weakly referenced by from this class, which resides in the system/bootstrap CL.

What I don't quite understand is, when you already inject the `HelperHolder` directly into the target CL, why don't you just inject the helper classes directly into the target CL. Then you also don't need the `HelperHolder` anymore.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
eyalkoren: Yes, I did, but it seems to be more costly in the overall

felixbarny: I'd be surprised if `WeakConcurrentMap`s are slower. What is it you are concerned about? The weak references for the key? The lookup should be just as fast and even allocation free. It would also simplify `loadAndReferenceHelper` as it takes care of cleaning the stale entries.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
eyalkoren: No. I assume `WeakConcurrentSet` has no added value if I reuse the boxed CL hash code.
As for `Collections.newSetFromMap` - good idea, seems to be no added overhead and with the set API

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
eyalkoren: Because injecting (i.e. defining not as intended) a class exposes you to JVM quirks, like how and when exactly the class and its dependencies are resolved, how loaded classes are searched for when referenced, what effect is there that the JVM only has a "defining loader" to the class and not an "initiating loader" and other things I did not think about. 
All those risks are considerably reduced if you are injecting a minimal class with no dependencies and immediately losing its reference so that it will never be used. This way all the code we run is properly loaded using the JVM mainstream mechanisms.

felixbarny: Is it to avoid that the helper classes in the system/bootstrap CL are referenced by making sure to only reference those defined by the `ChildFirstClassLoader`?

eyalkoren: BTW, what you suggested doesn't solve the problem of the reference- how would I prevent my helper instances from being GCed?

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
eyalkoren: I thought of using `WeakReference`s as keys, but these maps shouldn't grow a lot as they have entry-per-class-loader. Changed to `WeakConcurrentMap`

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
eyalkoren: No need for it anymore now that I use weak keys for the class loader in the other map- I can just map the CL to a `WeakReference` wrapping null for caching

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/HelperClassManager.java
eyalkoren: This is so that injected classes are never really used or even referenced. Not referencing the incorrect instance is achieved by using a ref to a bootstrap-loaded interface instead of a concrete instance.

PR #495, opened 2019-02-19
Stabilizing ScopeManagementTest

eyalkoren:


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -91,6 +91,7 @@
     private final List<ActivationListener> activationListeners;
     private final MetricRegistry metricRegistry;
     private Sampler sampler;
+    boolean assertionsEnabled = false;
 
     ElasticApmTracer(ConfigurationRegistry configurationRegistry, Reporter reporter, Iterable<LifecycleListener> lifecycleListeners, List<ActivationListener> activationListeners) {
```
felixbarny: My suggestion would be to add a constructor parameter `boolean strictMode` which binds to a final field. In strict mode, the tracer throws `IllegalStateExceptions`. Unit tests run in strict mode by default. This could either be determined by checking if assertions are enabled or just by making sure to set `strictMode(true)` for the tracer builder in unit tests. The `ScopeManagementTest` would then just build a tracer in non-strict mode to test how the tracer behaves when it's used illegally.

eyalkoren: I believe it was my suggestion :)
Anyway, I opted for this option, where we don't add test-related API to builder. I looked it up and it seems there is not a very big performance gain in making the field final, so I opted for the non-final field where I also have the flexibility of modifying it for specific tests, like I did.

felixbarny: Yes, and you convinced me about it ;) Anyway, it's not something I feel strongly about. I also don't have performance concerns as it is only invoked in case of wrong api usage anyway.

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ScopeManagementTest.java
```diff
@@ -54,56 +51,77 @@ void setUp() {
...
 
+    /**
+     * Disables assertions in {@link ElasticApmTracer}, runs the test and restores original setting
+     *
+     * @param test
+     */
+    void runTestWithAssertionsDisabled(Runnable test) {
+        boolean assertionsEnabled = tracer.assertionsEnabled;
+        tracer.assertionsEnabled = false;
+        test.run();
```
felixbarny: wrap in try/finally

PR #497, opened 2019-02-20
Add support for Number and boolean tags

### LINKED ISSUE #494: allow different types in addTag
**Is your feature request related to a problem? Please describe.**
We are currently instrumenting a java application running Quartz-Jobs which in turn get 1 or more jobs out of a job queue for bulk processing. I would like to store the information how many jobs are being processed as a tag on the transaction. For example I could do something like:
transaction.addTag("job_count",5);

Then, I could search in Kibana APM Gui for context.tags.job_count>=1.

**Describe the solution you'd like**
I would like to have a transaction.addTag (and Span.addTag) with numeric types as value.

**Describe alternatives you've considered**
Currently we store the number as text which only helps to search something like:
not context.tags.job_count:0

felixbarny:
closes elastic/apm-agent-java#494

### DIFF
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/SpanImpl.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/SpanImpl.java
@@ -57,7 +57,7 @@ class SpanImpl extends AbstractSpanImpl {
     @Nonnull
     @Override
     public Span addLabel(String key, String value) {
-        doAddTag(key, value);
+        doAddStringLabel(key, value);
         return this;
     }
 
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/SpanImpl.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/SpanImpl.java
@@ -54,4 +54,18 @@ class SpanImpl extends AbstractSpanImpl {
         return this;
     }
 
+    @Nonnull
+    @Override
+    public Span addTag(String key, Number value) {
+        doAddNumberTag(key, value);
+        return this;
+    }
+
+    @Nonnull
+    @Override
+    public Span addTag(String key, boolean value) {
+        doAddBooleanTag(key, value);
+        return this;
+    }
+
 }
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/SpanImpl.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/SpanImpl.java
@@ -56,15 +56,22 @@ class SpanImpl extends AbstractSpanImpl {
 
     @Nonnull
     @Override
-    public Span addTag(String key, Number value) {
-        doAddNumberTag(key, value);
+    public Span addLabel(String key, String value) {
+        doAddTag(key, value);
         return this;
     }
 
     @Nonnull
     @Override
-    public Span addTag(String key, boolean value) {
-        doAddBooleanTag(key, value);
+    public Span addLabel(String key, Number value) {
+        doAddNumberLabel(key, value);
+        return this;
+    }
+
+    @Nonnull
+    @Override
+    public Span addLabel(String key, boolean value) {
+        doAddBooleanLabel(key, value);
         return this;
     }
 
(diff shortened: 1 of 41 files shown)


felixbarny:
hmm, good point about calling them labels. but that could also cause some confusion.

eyalkoren:
Yes, but I imagine it won't be long before we all align on `labels` in the API, so maybe you can deprecate the one `addTag` method and introduce the new three `addLabel` now, instead of deprecating them all in the near future

apm-agent-api/src/main/java/co/elastic/apm/api/SpanImpl.java
```diff
@@ -56,15 +56,22 @@ public Span addTag(String key, String value) {
 
     @Nonnull
     @Override
-    public Span addTag(String key, Number value) {
-        doAddNumberTag(key, value);
+    public Span addLabel(String key, String value) {
+        doAddTag(key, value);
```
eyalkoren: Why not change `doAddTag` to `doAddLabel`? It means we will have duplicated instrumentation (for both), but better lose any unnecessary `tag` reminders.

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/plugin/api/AbstractSpanInstrumentation.java
```diff
@@ -173,45 +173,45 @@ public static void getTraceId(@Advice.FieldValue(value = "span", typing = Assign
...
         }
 
         @VisibleForAdvice
         @Advice.OnMethodEnter(suppress = Throwable.class)
-        public static void addTag(@Advice.FieldValue(value = "span", typing = Assigner.Typing.DYNAMIC) TraceContextHolder<?> context,
-                                  @Advice.Argument(0) String key, @Nullable @Advice.Argument(1) String value) {
+        public static void addLabel(@Advice.FieldValue(value = "span", typing = Assigner.Typing.DYNAMIC) TraceContextHolder<?> context,
+                                    @Advice.Argument(0) String key, @Nullable @Advice.Argument(1) String value) {
             if (value != null && context instanceof AbstractSpan) {
                 ((AbstractSpan) context).addTag(key, value);
```
eyalkoren: I suggest lose the tag notion internally as well, leave only OpenTracing stuff with it (because we have no choice...). This is our terminology going forward.

docs/public-api.asciidoc
```diff
@@ -262,17 +262,21 @@ transaction.setType(Transaction.TYPE_REQUEST);
 
 [float]
 [[api-transaction-add-tag]]
-==== `Transaction addTag(String key, value)`
-A flat mapping of user-defined tags with string values.
-Note: the tags are indexed in Elasticsearch so that they are searchable and aggregatable.
+==== `Transaction addLabel(String key, value)`
+A flat mapping of user-defined labels with string values.
```
eyalkoren: Not only `string values` anymore 🙂

docs/public-api.asciidoc
```diff
@@ -527,17 +531,21 @@ span.setName("SELECT FROM customer");
 
 [float]
 [[api-span-add-tag]]
-==== `Span addTag(String key, value)`
-A flat mapping of user-defined tags with string values.
-Note: the tags are indexed in Elasticsearch so that they are searchable and aggregatable.
+==== `Span addLabel(String key, value)`
+A flat mapping of user-defined labels with string values.
```
eyalkoren: same

apm-agent-api/src/main/java/co/elastic/apm/api/SpanImpl.java
felixbarny: I'd like to avoid having different instrumentation for different API versions if possible. The problem is that we can't easily unit test for previous API versions. We could potentially have a separate test module for each API version we want to test but I don't think it's worth the complexity.

eyalkoren: What's the problem of testing? The old API is still there, only deprecated. You can write tests using it.

felixbarny: you're right. done.

PR #500, opened 2019-02-25
Fix logging integration

### LINKED ISSUE #499: No MDC Data in logs with enable_log_correlation=true after a RestTemplate call
**Describe the bug**
When the property enable_log_correlation is configured with true value in the apm agent (for Java), traceId and transactionId are inserted in MDC and these values are written in log registries, but when you call a RestTemplate the next logs doesn't print the MDC values (traceId and transactionId).

**To Reproduce**
1. Configure APM Agent with the following params (used in Eclipse IDE)

-javaagent:${workspace_loc:elastic_apm_micros_spring2}/configuration/apm/elastic-apm-agent-1.4.0.jar
-Delastic.apm.service_name=bankservice
-Delastic.apm.service_version=1.0.0
-Delastic.apm.environment=dev
-Delastic.apm.sanitize_field_names=pass*,pwd*
-Delastic.apm.capture_headers=true
-Delastic.apm.enable_log_correlation=true
-Delastic.apm.application_packages=org.jrovira.blog.apm
-Delastic.apm.server_urls=http://localhost:8200

2. The source code with Spring Framework 2.1.2.RELEASE and JDK 1.8

@PostMapping("/account")
public ResponseEntity<String> processAccount(@RequestParam String name, @RequestParam String nif) {
	LOGGER.info("test1");
        restTemplate.getForEntity(riskUrl, Boolean.class, nif);
        LOGGER.info("test2");
}

3. The logback configuration

&lt;appender name=&quot;STDOUT&quot; class=&quot;ch.qos.logback.core.ConsoleAppender&quot;&gt;
		&lt;encoder class=&quot;net.logstash.logback.encoder.LogstashEncoder&quot;&gt;
			&lt;includeCallerData&gt;false&lt;/includeCallerData&gt;
			&lt;includeContext&gt;false&lt;/includeContext&gt;
			&lt;shortenedLoggerNameLength&gt;36&lt;/shortenedLoggerNameLength&gt;
			&lt;customFields&gt;
				{
					&quot;micro_name&quot;: &quot;${microName}&quot;
				}
			&lt;/customFields&gt;
		&lt;/encoder&gt;
	&lt;/appender&gt;

4. 

The console output

{...,"message":"test1",...,"transaction.id":"612a533941b93965","trace.id":"484302c7d391d4df487d39e238ded4b1",...}
{...,"message":"test2",...}

in the second log doesn't appear the transaction.id and trace.id fields.

**Expected behavior**
The second log should be 

{...,"message":"test2",...,"transaction.id":"612a533941b93965","trace.id":"484302c7d391d4df487d39e238ded4b1",...}

The complete source code of my app is in https://bitbucket.org/jrovirablog/elastic_apm_micros_spring2/commits/f4fe94b424abe3ee416816893e4d8e5a1c6e583e

Thanks for all.

eyalkoren:
Fixes elastic/apm-agent-java#499 and also enables proper logging integration for async spans/transactions as well.
In the sake of simplification, I added a `span.id` in all cases (even from transactions), so ID will be duplicated in logs created when the active context is a transaction. This means that log lines will be consistent for all types, so aggregations done for `span.id` will include transaction-related events as well. We need to decide whether this is a good thing (supports aggregations for all spans AND transactions easily) or a bad thing (doesn't support aggregations for spans AND NOT transactions).

### DIFF
--- a/apm-agent-plugins/apm-slf4j-plugin/src/main/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListener.java
+++ b/apm-agent-plugins/apm-slf4j-plugin/src/main/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListener.java
@@ -39,7 +39,6 @@ public class Slf4JMdcActivationListener implements ActivationListener {
     private static final String TRACE_ID = "trace.id";
     private static final String SPAN_ID = "span.id";
     private static final String TRANSACTION_ID = "transaction.id";
-    private static final String NA_SPAN_ID = "NA";
 
     private final WeakKeySoftValueLoadingCache<ClassLoader, MethodHandle> mdcPutMethodHandleCache =
         new WeakKeySoftValueLoadingCache<>(new WeakKeySoftValueLoadingCache.ValueSupplier<ClassLoader, MethodHandle>() {
--- a/apm-agent-plugins/apm-slf4j-plugin/src/main/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListener.java
+++ b/apm-agent-plugins/apm-slf4j-plugin/src/main/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListener.java
@@ -22,7 +22,6 @@ package co.elastic.apm.agent.slf4j;
 import co.elastic.apm.agent.cache.WeakKeySoftValueLoadingCache;
 import co.elastic.apm.agent.impl.ActivationListener;
 import co.elastic.apm.agent.impl.ElasticApmTracer;
-import co.elastic.apm.agent.impl.transaction.Span;
 import co.elastic.apm.agent.impl.transaction.TraceContext;
 import co.elastic.apm.agent.impl.transaction.TraceContextHolder;
 import co.elastic.apm.agent.logging.LoggingConfiguration;
@@ -31,6 +30,8 @@ import javax.annotation.Nullable;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
+import java.util.ArrayDeque;
+import java.util.Deque;
 
 public class Slf4JMdcActivationListener implements ActivationListener {
 
@@ -40,6 +41,15 @@ public class Slf4JMdcActivationListener implements ActivationListener {
     private static final String TRACE_ID = "trace.id";
     private static final String SPAN_ID = "span.id";
     private static final String TRANSACTION_ID = "transaction.id";
+    private static final String NA_SPAN_ID = "NA";
+
+    private final ThreadLocal<Deque<String>> spanIdStack = new ThreadLocal<Deque<String>>() {
+        @Override
+        protected Deque<String> initialValue() {
+            return new ArrayDeque<String>();
+        }
+    };
+
     private final WeakKeySoftValueLoadingCache<ClassLoader, MethodHandle> mdcPutMethodHandleCache =
         new WeakKeySoftValueLoadingCache<>(new WeakKeySoftValueLoadingCache.ValueSupplier<ClassLoader, MethodHandle>() {
             @Nullable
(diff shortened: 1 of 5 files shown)


apm-agent-plugins/apm-slf4j-plugin/src/main/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListener.java
```diff
@@ -68,37 +78,79 @@ public MethodHandle get(ClassLoader classLoader) {
...
+                }
             }
         }
     }
 
     @Override
-    public void onDeactivate() throws Throwable {
-        if (config != null && config.isLogCorrelationEnabled()) {
-            MethodHandle remove = mdcRemoveMethodHandleCache.get(Thread.currentThread().getContextClassLoader());
+    public void onDeactivate(TraceContextHolder<?> context) throws Throwable {
```
felixbarny: We can't add the `TraceContextHolder` here as it might already be ended and recycled.

apm-agent-plugins/apm-slf4j-plugin/src/main/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListener.java
```diff
@@ -68,37 +78,79 @@ public MethodHandle get(ClassLoader classLoader) {
...
+        if (config != null && config.isLogCorrelationEnabled() && context.isSampled()) {
+            String formerSpanId = spanIdStack.get().poll();
```
felixbarny: Instead of maintaining another stack here, couldn't we just get `tracer.getActive()` and call `isSampled()` and `traceContext.getId().toString()` on that instance?

apm-agent-plugins/apm-slf4j-plugin/src/main/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListener.java
eyalkoren: This method is called from `TraceContextHolder#deactivate`, when the `this` instance is still the active one on the thread stack, so I it is very valid to for checking this state. You are right however that I shouldn't observe it, so I shouldn't call `isSampled` for example.
So, I will add to documentation that I can only use the reference but not observe and certainly not update its state

eyalkoren: Another option is to transfer on the TraceContextHolder instance hash code...

apm-agent-plugins/apm-slf4j-plugin/src/main/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListener.java
eyalkoren: This contradicts what you mentioned in the above comment :)
It may be already ended and recycled, so not allowed to observe at this point.

felixbarny: What I meant was to get the parent of the context which is being deactivated. You reversed the order in the PR so that it deactivates after calling the listeners. I think we should make it more obvious from the method name if the deactivation happens before or after. For example we should call it either `beforeDeactivate` or `afterDeactivate`.

apm-agent-plugins/apm-slf4j-plugin/src/main/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListener.java
felixbarny: > You are right however that I shouldn't observe it, so I shouldn't call `isSampled` for example.

By providing the context as an argument, we are more likely to make mistakes as that. We should only add the argument if the need for it actually arises. Even then it might more sense to just have the identity hash as a parameter.

apm-agent-plugins/apm-slf4j-plugin/src/main/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListener.java
eyalkoren: Ahh, I see. 
To me, the right way to undo/reverse `A -> B -> C -> D` is `!D -> !C -> !B -> !A`. So, for this matter, if `A` is `TraceContextHolder#activate` and `B` is `ActivationListener#onActivate`, It seemed to make sense to reverse the order. 
What if I change `ActivationListener#onActivate` to `beforeActivate` and `ActivationListener#onDeactivate` to `afterDeactivate`? Then I can remove the additional stack as you suggest and have consistent code in both methods: 
```
    public void beforeActivate(TraceContextHolder<?> context) throws Throwable {
        ...
        if (tracer != null) {
            if (tracer.getActive() == null) {
                put.invokeExact(TRACE_ID, traceContext.getTraceId().toString());
                put.invokeExact(TRANSACTION_ID, traceContext.getTransactionId().toString());
            }
            put.invokeExact(SPAN_ID, traceContext.getId().toString());
        }
    }

    public void afterDeactivate(TraceContextHolder<?> context) throws Throwable {
        ...
        if (tracer != null) {
            TraceContextHolder active = tracer.getActive();
            if (active == null) {
                remove.invokeExact(TRACE_ID);
                remove.invokeExact(TRANSACTION_ID);
            } else if (active.isSampled()) {
                put.invokeExact(SPAN_ID, active.getTraceContext().getId().toString());
            }
        }
    }
```

eyalkoren:
We decided to:
1. log info even for unsampled events
2. log the span ID for transactions

PR #501, opened 2019-02-26
Fix metrics NaN issue

eyalkoren:
Not the prettiest, but given the way our metrics serialization is working, it is a simple solution without extra allocation. 
It may send empty metrics sets if ALL have `NaN` values but I decided it doesn't make sense to compromise performance or code simplicity for this unexpected extreme (should't cause an error, just send empty data event)

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/MetricRegistrySerializer.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/MetricRegistrySerializer.java
@@ -69,29 +69,39 @@ public class MetricRegistrySerializer {
         final int size = samples.size();
         if (size > 0) {
             final Iterator<Map.Entry<String, DoubleSupplier>> iterator = samples.entrySet().iterator();
-            Map.Entry<String, DoubleSupplier> kv = iterator.next();
-            serializeSample(kv.getKey(), kv.getValue().get(), jw);
-            for (int i = 1; i < size; i++) {
-                jw.writeByte(JsonWriter.COMMA);
+            String key = null;
+            double value = Double.NaN;
+            Map.Entry<String, DoubleSupplier> kv;
+            while (iterator.hasNext() && !isValid(value)) {
                 kv = iterator.next();
-                serializeSample(kv.getKey(), kv.getValue().get(), jw);
+                key = kv.getKey();
+                value = kv.getValue().get();
+            }
+            if (isValid(value)) {
+                serializeSample(key, value, jw);
+                while (iterator.hasNext()) {
+                    kv = iterator.next();
+                    value = kv.getValue().get();
+                    if (isValid(value)) {
+                        jw.writeByte(JsonWriter.COMMA);
+                        serializeSample(kv.getKey(), value, jw);
+                    }
+                }
             }
         }
         jw.writeByte(JsonWriter.OBJECT_END);
+    }
+
+    private static boolean isValid(double value) {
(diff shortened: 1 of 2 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/MetricRegistrySerializer.java
```diff
@@ -69,29 +69,39 @@ private static void serializeSamples(Map<String, DoubleSupplier> samples, JsonWr
...
+            Map.Entry<String, DoubleSupplier> kv;
+            while (iterator.hasNext() && !isValid(value)) {
                 kv = iterator.next();
-                serializeSample(kv.getKey(), kv.getValue().get(), jw);
+                key = kv.getKey();
+                value = kv.getValue().get();
+            }
+            if (isValid(value)) {
+                serializeSample(key, value, jw);
+                while (iterator.hasNext()) {
```
felixbarny: I found the code a bit hard to follow. Especially because there is not a clear separation of where the first value is serialized and where the rest is serialized. I think a simple change would help: move the `while (iterator.hasNext())` and the corresponding scope out of the `if (isValid(value))` scope and move the serialization of the first value to the first loop. You can then also use separate local variables which are scoped inside the loop. T

```java
            // serialize first valid value
            while (iterator.hasNext() && !isValid(value)) {
                Map.Entry<String, DoubleSupplier> kv = iterator.next();
                double value = kv.getValue().get();
                if (isValid(value)) {
                    serializeSample(kv.getKey(), value, jw);
                }
            }

            // serialize rest
            while (iterator.hasNext()) {
                Map.Entry<String, DoubleSupplier> kv = iterator.next();
                double value = kv.getValue().get();
                if (isValid(value)) {
                    jw.writeByte(JsonWriter.COMMA);
                    serializeSample(kv.getKey(), value, jw);
                }
            }
```

This has equal semantics, but you can reason about the logic for serializing the first and subsequent values independently. Reducing the scope of the local variables also supports that. Another benefit is that the serialization of the first vs the subsequent values look really similar so it's easier to spot what the differences are. Another minor benefit is that this reduces the nesting level by one.

eyalkoren:
@kuisathaverat is there a way to trigger a build in the CI through GH comment?

kuisathaverat:
jenkins run the tests

mierzej91:
Is this fix implemented? I used APM agent in version 1.18.0 and I still got this error

`{ "message": "failed to validate metricset: error validating JSON: I[#] S[#] doesn't validate with \"metricset#\"\n  I[#] S[#/allOf/5] allOf failed\n    I[#/samples/example.met.processing/value] S[#/allOf/5/properties/samples/patternProperties/%5E%5B%5E%2A%22%5D%2A$/properties/value/type] expected number, but got string",

  "document": "{\"metricset\":{\"timestamp\":1601377271199000,\"tags\":{\"command\":\"ExamleCommand\",\"status\":\"*\"},\"samples\":{\"example.met.processing\":{\"value\":\"NaN\"}}}}"}`

eyalkoren:
@mierzej91 this was merged long ago.

Can https://github.com/elastic/apm-agent-java/pull/1419 be related to what you are seeing? If so, please try to [download](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/PR-1419/2/artifacts) and test the related bugfix and see if the problem is resolved.

cc @SylvainJuge

Issue #384, opened 2018-12-14
Automatically track spans in Servlet RequestDispatcher forwards/includes


jeremyfreedman-natgeo:
This is based on a discussion I had with Felix here:
https://discuss.elastic.co/t/apm-java-agent-startup-error-with-adobe-experience-manager

It would be great if the java APM agent automatically created spans anytime the Servlet RequestDispatcher does a forward or include.

At least for the product I am using this for (Adobe Experience Manager), providing those spans exposes great timing information about components that are rendered as part of a page.

eyalkoren:
related to https://github.com/elastic/apm-agent-java/issues/429

eyalkoren:
Implementation started at #557 with instrumenting the RequestDispatcher. Instead, we thought of using our existing doFilter and service instrumentation, ie do everything through `ServletApiAdvice`:
On method enter and exit, add a `@Nullabale @Advice.Local("span") Span span`.
On method enter, we will rely on `javax.servlet.ServletRequest#getDispatcherType` to decide if this is a transaction, or a forward/include/error span. If the DispatcherType is `FORWARD`, `INCLUDE` or `ERROR`, AND `tracer.currentTransaction()!=null`, we will start and activate a span, unless the current span has the same servletPath and pathInfo.
On method exit, if the span is not null, deactivate it and end it.

lreuven:
@eyalkoren any updated on this one?

eyalkoren:
@lreuven No, needs to be implemented as described above

PR #503, opened 2019-02-26
Task/429 servlet transaction

### LINKED ISSUE #429: Use JSP name for JSP-transactions instead of servlet name
For example- `GET TobBar.jsp` and `POST TopBar.jsp` instead of `TopBar_jsp#doGet` and `TopBar_jsp#doPost`

nugusbayevkk:
closes #429 and #384
I added instrumentation for getRequestDispatcher method of ServletRequest as a helper to store forward or include path.


nugusbayevkk:
screenshot from kibana[apm-server]:

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardRequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,103 @@
...
+import javax.servlet.ServletRequest;
+import javax.servlet.http.HttpServletRequest;
+import java.util.Arrays;
+import java.util.Collection;
+
+import static net.bytebuddy.matcher.ElementMatchers.*;
+
+public class ForwardRequestDispatcherInstrumentation extends ElasticApmInstrumentation {
+
+    private static final String SPAN_TYPE = "template";
```
felixbarny: I'm not sure if `template` is appropriate here. We don't actually know if there is any template involved, right?

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardRequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,103 @@
...
+                                          @Advice.Argument(0) @Nullable ServletRequest request) {
+            if (tracer == null || tracer.getActive() == null) {
+                return;
+            }
+            final TraceContextHolder<?> parent = tracer.getActive();
+
+            if (request != null && request instanceof HttpServletRequest) {
+                HttpServletRequest httpServletRequest = (HttpServletRequest) request;
+                String forwardServletPath = (String) httpServletRequest.getAttribute(RequestDispatcher.FORWARD_SERVLET_PATH);
+                span = parent.createSpan()
```
felixbarny: If the `RequestDispatcher` delegates calls to another `RequestDispatcher`, it would create two spans. The simplest way to avoid this is to check if the parent span has the same type as the one you are about to create.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/IncludeRequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return not(isInterface())
+            .and(nameContainsIgnoreCase("dispatcher"));
+    }
+
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return named("include").and(takesArguments(2))
```
felixbarny: Instead of duplicating the instrumentation, I'd rather match "include" and "forward" and inject the actual method name by adding a `@Advice.Origin("#m") String method` parameter to the `@OnMethodEnter` method.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardRequestDispatcherInstrumentation.java
felixbarny: However, it might be a valid case to forward to a servlet which does another forward.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardRequestDispatcherInstrumentation.java
nugusbayevkk: yes, you are right. i had a question: what span_type we can use?
I found "template" and "db: span types in project.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/RequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,73 @@
...
+    public Class<?> getAdviceClass() {
+        return RequestDispatcherInstrumentation.RequestDispatcherAdvice.class;
+    }
+
+    public static class RequestDispatcherAdvice extends IncludeRequestDispatcherInstrumentation {
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        private static void beforeExecute(@Advice.Argument(0) @Nullable String path,
+                                          @Advice.This @Nullable Object thiz) {
+            if (thiz != null && thiz instanceof HttpServletRequest) {
+                ((HttpServletRequest) thiz).setAttribute(RequestDispatcher.FORWARD_SERVLET_PATH, (path != null) ? path : "");
```
felixbarny: Isn't that already done by the container?

felixbarny:
superseded by https://github.com/elastic/apm-agent-java/pull/557

PR #508, opened 2019-03-01
[APM-CI] Store artifacts in Jenkins

kuisathaverat:
Store the APM Java agent jar file on each Jenkins build


felixbarny:
Thanks :)

eyalkoren:
@kuisathaverat @felixbarny do you think we should upload `apm-agent-attach-*.jar` as well?

felixbarny:
If you think it's useful, we can definitely do that :)

kuisathaverat:
I take note, I'll make it in a gap between things

PR #511, opened 2019-03-05
Consistently rename tags to labels in DslJsonSerializer

felixbarny:
Avoid allocation of tags iterator if there are no tags


apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
```diff
@@ -658,7 +658,7 @@ private void serializeSpanContext(SpanContext context) {
                 jw.writeByte(COMMA);
             }
             writeFieldName("tags");
```
eyalkoren: Should we mark such places for a reminder to change on 2.0?

felixbarny: There is currently no concrete plan when the intake API will be changed.

PR #515, opened 2019-03-07
Elasticsearch client async instr

### LINKED ISSUE #263: Elasticsearch support for asynchronous operations
We are [currently supporting only synchronous calls](https://github.com/elastic/apm-agent-java/pull/228).
A good place to capture probably all async operations would be `org.elasticsearch.client.RestClient#performRequestAsync(Request, ResponseListener)`, where we have the request available and we can hook on the `ResponseListener` to capture both responses and errors.
However, the ES client is relying on Apache HttpAsyncClient, so in order to make the async correlation we will probably first want to have a basic support for that.

eyalkoren:
Closes #263


apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-5_6/src/main/java/co/elastic/apm/agent/es/restclient/v5_6/ElasticsearchClientAsyncInstrumentation.java
```diff
@@ -0,0 +1,101 @@
...
+                    wrapped = true;
+                }
+            }
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
+        private static void onAfterExecute(@Advice.Argument(5) ResponseListener responseListener,
+                                           @Advice.Local("span") @Nullable Span span,
+                                           @Advice.Local("wrapped") boolean wrapped,
+                                           @Advice.Local("helper") @Nullable ElasticsearchRestClientInstrumentationHelper<HttpEntity, Response, ResponseListener> helper, @Advice.Thrown @Nullable Throwable t) {
```
felixbarny: Wrap the Throwable argument to new line

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-5_6/src/main/java/co/elastic/apm/agent/es/restclient/v5_6/ElasticsearchClientAsyncInstrumentation.java
```diff
@@ -0,0 +1,101 @@
...
+            if (span != null) {
+                // Deactivate in this thread. Span will be ended and reported by the listener
+                span.deactivate();
+
+                if (!wrapped) {
+                    // Listener is not wrapped- we need to end the span so to avoid leak and report error if occurred during method invocation
```
felixbarny: I might be wrong but if there was an exception in the method invocation, wrapped wouldn't wrapped still be `true`? So this is basically only in case `helper.wrapResponseListener()` threw an exception. If that's the case, couldn't we just do an instance-of check and remove the wrapped flag altogether?
```suggestion
                    // Listener is not wrapped- we need to end the span so to avoid leak and report error if occurred in the enter advice
```

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/es/restclient/ElasticsearchRestClientInstrumentationHelperImpl.java
```diff
@@ -0,0 +1,145 @@
...
+        span.activate();
+
+        if (span.isSampled()) {
+            span.getContext().getHttp().withMethod(method);
+            if (endpoint.endsWith(SEARCH_QUERY_PATH_SUFFIX)) {
+                if (httpEntity != null && httpEntity.isRepeatable()) {
+                    try {
+                        IOUtils.readUtf8Stream(httpEntity.getContent(), span.getContext().getDb().withStatementBuffer());
+                    } catch (IOException e) {
+                        // We can't log from here
```
felixbarny: we actually can :)

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/es/restclient/ElasticsearchRestClientInstrumentationHelperImpl.java
```diff
@@ -0,0 +1,145 @@
...
+    @Override
+    @Nullable
+    public Span createClientSpan(String method, String endpoint, @Nullable HttpEntity httpEntity) {
+        final TraceContextHolder<?> activeSpan = tracer.getActive();
+        if (activeSpan == null || !activeSpan.isSampled()) {
+            return null;
+        }
+
+        // Don't record nested spans. In 5.x clients the instrumented sync method is calling the instrumented async method
+        if (activeSpan instanceof Span && ELASTICSEARCH.equals(((Span) activeSpan).getSubtype())) {
```
felixbarny: Maybe not for now but I think we should add a `external` flag to spans. The tracer should return `null` when trying to create a child span of an external span.

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/es/restclient/ElasticsearchRestClientInstrumentationHelperImpl.java
```diff
@@ -0,0 +1,145 @@
...
+                url = response.getHost().toURI();
+                statusCode = response.getStatusLine().getStatusCode();
+            } else if (t != null) {
+                if (t instanceof ResponseException) {
+                    ResponseException esre = (ResponseException) t;
+                    url = esre.getResponse().getHost().toURI();
+                    statusCode = esre.getResponse().getStatusLine().getStatusCode();
+
+                        /*
+                        // Add tags so that they will be copied to error capture
```
felixbarny: I would prefer to remove this and rather create a task to enhance the intake API.

felixbarny:
As the span for this is started on one thread and ended on another - are all necessary properties of span volatile? Or should we maybe rather do a memory fence when activating on another thread so that we don't have to declare the main properties volatile at all?

What I mean by that is to write to a single volatile flag which is only used to ensure visibility. That works as writing to a volatile variable introduces total ordering of reads/writes.

To cite Java concurrency in practice:
> The visibility effects of volatile variables extend beyond the value of the volatile variable itself. When thread A writes to a volatile variable and subse- quently thread B reads that same variable, the values of all variables that were visible to A prior to writing to the volatile variable become visible to B after reading the volatile variable. So from a memory visibility perspective, writing a volatile variable is like exiting a synchronized block and reading a volatile variable is like entering a synchronized block. 

Although relying on that could make the code more fragile and harder to understand.

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-5_6/src/main/java/co/elastic/apm/agent/es/restclient/v5_6/ElasticsearchClientAsyncInstrumentation.java
eyalkoren: You are not wrong, this is a state for us to know whether we wrapped the listener or not. If we did not wrap it, we need to end the span properly no matter what and while doing so it would be informative to report an Exception if such occurred **during the method invocation**.

> If that's the case, couldn't we just do an instance-of check and remove the wrapped flag altogether

We couldn't because the `ResponseListenerWrapper` class the advice will see is the one visible to the application, meaning the one loaded (or most likely - failed to be loaded) from BootstrapCL, whereas the one we actually used is loaded from a child CL of the app CL.

We could use `listener.getClass().getName().equals(...)`, but what's the benefit over using a boolean?

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/es/restclient/ElasticsearchRestClientInstrumentationHelperImpl.java
eyalkoren: We sure can, noticed that and forgot to log 🤦‍♂️

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/es/restclient/ElasticsearchRestClientInstrumentationHelperImpl.java
eyalkoren: I agree, but in this case this is to avoid nested ES client spans, so I would leave it anyway as it deals with a specific problem.

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/es/restclient/ElasticsearchRestClientInstrumentationHelperImpl.java
eyalkoren: I will, maybe just leave a comment to remember that `esre.getResponse().getStatusLine().getReasonPhrase()` provides a useful info.

eyalkoren:
We now have `AbstractSpan#isLifecycleManagingThreadSwitch` that effectively does that for ExecutorServices and for Servlet's `AsyncContext#start` API. We can extend that to be more general - any action causing span transfer between threads will write to it before doing so and `TraceContextHolder#activate` will start by reading it

PR #516, opened 2019-03-10
Loading plugins from System CL and parent hierarchy

eyalkoren:
Related Discuss topic: https://discuss.elastic.co/t/custom-plugin-for-java-agent/171249.
Letting agent load plugins located in the classpath, in the platform/ext libs or bootclasspath.
Here is an attach jar for testing: [apm-agent-attach-1.4.1-SNAPSHOT.jar.zip](https://github.com/elastic/apm-agent-java/files/2949676/apm-agent-attach-1.4.1-SNAPSHOT.jar.zip)

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
@@ -97,7 +97,7 @@ public class ElasticApmAgent {
     @Nonnull
     private static Iterable<ElasticApmInstrumentation> loadInstrumentations(ElasticApmTracer tracer) {
         final ArrayList<ElasticApmInstrumentation> instrumentations = new ArrayList<ElasticApmInstrumentation>();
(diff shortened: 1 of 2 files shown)


felixbarny:
But that won't help if the plugins are loaded from the web application class loader, right?

We we would allow that, there would be some obstacles:
* What if there are breaking changes between the attached agent and the plugin? We can currently break the internal API even on bugfix releases.
* In case there are multiple applications deployed to the same server and one application wants to register a plugin, the other application may get `ClassNotFoundException`s

eyalkoren:
> In case there are multiple applications deployed to the same server and one application wants to register a plugin, the other application may get ClassNotFoundExceptions

I didn't suggest app-context plugins, only System CL or ext/platform..

> What if there are breaking changes between the attached agent and the plugin? We can currently break the internal API even on bugfix releases.

The idea of this PR is not to enhance our public API or anything of that sort. It is just to ease plugin loading for edge cases like the one in the forum topic linked above. Anyone that writes code that relies on our internal APIs should be aware of such risk. 
The alternative is to tell such users to append the plugin jar to the boot classpath (eg by editing the agent jar's `MANIFEST.MF`), so it is a bit more work for them but the problem will remain the same.

PR #525, opened 2019-03-14
Channelling span creation through a common path

eyalkoren:
Maybe related to https://discuss.elastic.co/t/kibana-apm-not-showing-spans-which-are-visible-in-discover-too-many-spans/171690, where we ignore the span limit setting


felixbarny:
Could that be the reason why the span limit does not work?

eyalkoren:
That's what I am hoping, will ask him to try it out

PR #526, opened 2019-03-14
Support async OkHttp calls

### LINKED ISSUE #425: Support async Okhttp client requests

felixbarny:
closes elastic/apm-agent-java#425


apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientAsyncInstrumentation.java
```diff
@@ -0,0 +1,166 @@
...
+
+            CallbackWrapper(Span span, Callback delegate) {
+                this.span = span;
+                this.delegate = delegate;
+            }
+
+            @Override
+            public void onFailure(Request req, IOException e) {
+                try {
+                    span.captureException(e).end();
```
eyalkoren: Is `onResponse` called after this as well? If not- can you find the response code in `e` maybe?

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientAsyncInstrumentation.java
```diff
@@ -0,0 +1,166 @@
...
+                span.deactivate();
+            }
+        }
+    }
+
+    public static class CallbackWrapperCreator implements WrapperCreator<Callback> {
+
+        @Override
+        public Callback wrap(final Callback delegate, Span span) {
+            return new CallbackWrapper(span, delegate);
```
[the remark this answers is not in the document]
felixbarny: We allocate anyway by calling `originalRequest.newBuilder().addHeader`. We can think about creating a follow-up PR to remove both allocations.

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientAsyncInstrumentation.java
felixbarny: I think in those cases there is no response, for example, because of a network error. If there is an error status code, `onResponse` would be called.

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientAsyncInstrumentation.java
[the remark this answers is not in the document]
eyalkoren: We went over the `newBuilder` and got to the conclusion it is better than all other alternatives we reviewed. I don't think they are related.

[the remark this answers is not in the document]
felixbarny: I'm just a bit skeptical that it's worth the hassle of recycling the `Callback` wrappers if we can't make tracing OkHttp calls garbage-free anyway.

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttp3ClientAsyncInstrumentation.java
```diff
@@ -0,0 +1,165 @@
...
+
+    public static class CallbackWrapperCreator implements WrapperCreator<Callback> {
+
+        @Override
+        public Callback wrap(final Callback delegate, Span span) {
+            return new CallbackWrapper(span, delegate);
+        }
+
+        private static class CallbackWrapper implements Callback {
+            private final Span span;
```
eyalkoren: I wrote this comment in the initial review and it got lost somehow, noticed just now. 

Like we did with the ES client, we need to enforce proper visibility. Make this `span` field volatile and switch the order of assignments in the constructor (so that it is written to after `delegate` is assigned on the initiating thread). In both `onFailure` and `onResponse` the `span` is read first, so this should suffice. Otherwise, you can make both `span` and `delegate` volatile.

felixbarny: We don't need to in this case, because the fields are `final` 🙂

eyalkoren: Right 🤦‍♂️ , in my head we were recycling them 😄 . 
I don't have support for async discussions yet, hence the visibility problem 🤓

eyalkoren:
Jenkins run the tests

PR #529, opened 2019-03-18
Add more comprehensive inject/extract example to API docs

felixbarny:


felixbarny:
Related: https://discuss.elastic.co/t/java-agent-how-to-manually-instrument-an-rpc-framework-which-is-not-already-supported/172730

docs/public-api.asciidoc
```diff
@@ -497,7 +527,22 @@ Example:
 
 [source,java]
 ----
-span.injectTraceHeaders(request::addHeader);
+// Hook into a callback provided by the RPC framework that is called on outgoing requests
```
eyalkoren: There is also same documentation in the `Span API` section

PR #530, opened 2019-03-18
Override span.id instead of removing and adding

felixbarny:
Each remove or add operation results in duplicating the MDC map
Therefore, this change minimizes allocations while retaining the same semantics


eyalkoren:
I didn't do this because I didn't find an explicit guarantee in the API that old values are always replaced if already exist for a key, but we can probably assume that it follows the common Map convention...

PR #536, opened 2019-03-19
Don't instrument Hazelcast Executors

felixbarny:
Context: https://discuss.elastic.co/t/cant-run-apm-java-agent-in-alfresco-cluster-environment/172962


felixbarny:
But that would prevent context propagation if the `Runnable` implements `Serializable` and submitting that to a conventional `Executor`.

eyalkoren:
If someone bothered to mark `Runnable`s as `Serializable`, we probably shouldn't wrap them...

felixbarny:
But it doesn't depend on the `Runnable` but rather on the `Executor` if it can deal with those. If we exclude `Serializable` `Runnable`s, I'm sure we'll sooner or later get an issue like `Context propagation does sometimes not work`. Debugging that won't be easy.

PR #539, opened 2019-03-19
Ensure proper visibility of events' data before sending

eyalkoren:
Hopefully this will fix the sporadic integration testing failures 🤞 

May not be bullet-proof though, we need to think whether a deeper read/write is required (i.e. inner objects)...


felixbarny:
The disruptor already guarantees visibility in the `com.lmax.disruptor.EventHandler` thread as it acts as a memory barrier.

eyalkoren:
How's that done? Can you refer me to the relevant code?

felixbarny:
See this blog: https://dzone.com/articles/dissecting-disruptor

eyalkoren:
Ahh, [this one](https://github.com/LMAX-Exchange/disruptor/blob/5764f3b01faab603322c61cede957b7e79f58c8b/src/main/java/com/lmax/disruptor/Sequence.java#L30) is written to after events are ready and read from before serialization...

PR #540, opened 2019-03-20
Document when options and APIs were added

felixbarny:
Makes use of the added[version] asciidoc extension and ConfigurationOption tags


felixbarny:
Screen Shot 2019-03-20 at 10 57 20

felixbarny:
Also shows other configuration option tags like `performance` or `security` next to the key.

PR #541, opened 2019-03-20
Allow setting manual start and end timestamps

### LINKED ISSUE #236: Java API to allow setting external start and end timestamps on spans and transactions
In my case, I am using asynchronous notifications with callbacks that can be processed at a later time than the original event occurred. In my case, I am able to retrieve the event timestamp from the notification parameter passed into the callback. I would like to be able to specify my own timestamp at start of transaction or a span, as well at an end of transaction or a span.

My callback code example:
```java
public class TraceMessageProcessorNotificationListener
		implements MessageProcessorNotificationListener<MessageProcessorNotification> {

	@Override
	public void onNotification(MessageProcessorNotification notification) {

		switch (notification.getAction()) {
		case MessageProcessorNotification.MESSAGE_PROCESSOR_PRE_INVOKE:
			spanUtils.startSpan(notification); // <-- Set start span time explicitly here
			break;

		case MessageProcessorNotification.MESSAGE_PROCESSOR_POST_INVOKE:
			spanUtils.endSpan(notification); // <-- Set end span time explicitly here
			break;
		}
	}
}

### PR DESCRIPTION
**Pull request by felixbarny**
closes elastic/apm-agent-java#236

### DIFF
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
@@ -190,7 +190,7 @@ public interface Span {
      * <p>
      * The type, subtype and action strings are used to group similar spans together, with increasing resolution.
      * For instance, all DB spans are given the type `db`; all spans of MySQL queries are given the subtype `mysql` and all spans
-     * describing queries are give the action `query`.
+     * describing queries are given the action `query`.
      * </p>
      * <p>
      * In the above example `db` is considered the general type. Though there are no naming restrictions for the general types,
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
@@ -146,6 +146,15 @@ public interface Span {
     Span addLabel(String key, boolean value);
 
     /**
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: apm-agent-api/src/main/java/co/elastic/apm/api/Span.java [hunk 1]
```diff
@@ -180,7 +189,7 @@
      * </p>
      * <p>
      * The type, subtype and action strings are used to group similar spans together, with increasing resolution.
-     * For instance, all DB spans are given the type `db`; all spans of MySQL queries are given the subtype `mysql` and all spans 
+     * For instance, all DB spans are given the type `db`; all spans of MySQL queries are given the subtype `mysql` and all spans
      * describing queries are give the action `query`.
```
eyalkoren:
```suggestion
     * describing queries are given the action `query`.
```

felixbarny:
Test results:

> 1. start time is smaller than parent's start time

The start time of the child will be adjusted by the UI so that it starts at the same time as the parent

> 2. end time is smaller than start time.

The UI filters out those transactions

I don't think we need that validation - we can't correct the wrong data anyway. I'd like to proceed with this and possibly add validation later if there is confusion among users.

PR #544, opened 2019-03-22
Add support for providing config on attach

felixbarny:


fredgalvao:
Just tested this with the fresh new 6.7.0 stack through 1.4.1-SNAPSHOT, works perfectly. :tada:

eyalkoren:
There is an official 1.5.0 as of yesterday, just FYI

PR #548, opened 2019-03-28
Adding configuration for class loading matching and logging class source

eyalkoren:
Related to https://discuss.elastic.co/t/kibana-apm-not-showing-capturespan-and-capturetransaction/173026 - even if not fixes this one, class loading has side effects and needs to be configurable

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
@@ -164,6 +164,7 @@ public class ElasticApmAgent {
         final Logger logger = LoggerFactory.getLogger(ElasticApmAgent.class);
         logger.debug("Applying advice {}", advice.getClass().getName());
         advice.init(tracer);
+        final boolean classLoadingMatchingPreFilter = tracer.getConfig(CoreConfiguration.class).isClassLoadingMatchingPreFilter();
         final boolean typeMatchingWithNamePreFilter = tracer.getConfig(CoreConfiguration.class).isTypeMatchingWithNamePreFilter();
         final ElementMatcher.Junction<ClassLoader> classLoaderMatcher = advice.getClassLoaderMatcher();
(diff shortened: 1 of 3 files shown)


felixbarny:
Why do you think disabling the class loader matcher may help in the linked case? Is the class loader pre matcher problematic when there are annotations on the class path twice? What negative side effects are there when loading a class which would otherwise not be loaded?

eyalkoren:
>Why do you think disabling the class loader matcher may help in the linked case?

The user gets `Caused by: java.lang.LinkageError: loader (instance of  org/apache/catalina/loader/ParallelWebappClassLoader): attempted  duplicate class definition for name: "javax/ws/rs/Path"` in his logs. I didn't bother to look for the `WebappClassLoaderBase` code to see why it doesn't find the defined class and tries to define it again during the execution of `findClassInternal`. It could even be something weird that prevents it from finding a defined class even if it uses the same jar. However, it happens with the agent and doesn't happen without it, and that's the only reason I could see how that might be, other than a BB bug that we never encountered before.

> What negative side effects are there when loading a class which would otherwise not be loaded?

For example- it might have dependencies that are either missing or that you are going to cause loading on a wrong timing (where other dependencies are not available or where a static init does something). It is always risky to change the order/location of class loading in an existing environment. In many cases it is exposes problems that were not occurring just due to order of execution. I would even argue that you may want to set this config option to `false` by default, but we can wait and see if we get more of those. We currently load very limited number of classes, but should be aware of that.

felixbarny:
> Caused by: java.lang.LinkageError: loader (instance of org/apache/catalina/loader/ParallelWebappClassLoader): attempted duplicate class definition for name: "javax/ws/rs/Path"

That really sounds like the class loader matcher may be causing this. But I think it's probably worth verifying that first. So let's wait for the user's feedback?

eyalkoren:
> But I think it's probably worth verifying that first. So let's wait for the user's feedback?

Sure, we can wait, but I think we need this (hidden) config and logging regardless (as I explained in the former comment), which is why I did this PR without waiting for the response.

PR #553, opened 2019-04-02
Use only ascii chars in log statements

felixbarny:


felixbarny:
See https://discuss.elastic.co/t/invalid-json-somewhere-in-pipeline/174820

PR #554, opened 2019-04-02
Avoid that the reporter blocks server shutdown

felixbarny:
This only happens after the APM Server is not available for a while. In that case, the reporter queue fills up and can never really drain because the reporter blocks its thread to throttle APM Server connection retries. On shutdown, the reporter tries to flush by registering a flush event. The registration blocks until a new slot in the ring buffer becomes available. But due to the throttling only once every 36 seconds, an event is picked up from the ring buffer. So a shutdown takes around 512 * 36 seconds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Features
 
 ## Bug Fixes
+* Avoid that the agent blocks server shutdown in case the APM Server is not available
 
 # 1.5.0
 
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java
@@ -160,7 +160,10 @@ public class ApmServerReporter implements Reporter {
      */
     @Override
     public Future<Void> flush() {
-        disruptor.publishEvent(FLUSH_EVENT_TRANSLATOR);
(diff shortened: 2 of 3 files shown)


felixbarny:
The bug likely was introduced here: https://github.com/elastic/apm-agent-java/pull/410

eyalkoren:
On shutdown, the reporter tries to publish a shutdown event and then calls the `Disruptor.shutdown` with does a busy spin to see if there are still events to publish and times out after 5 seconds:
https://github.com/elastic/apm-agent-java/blob/24424d6f38531865c99ef013caa49f7108e3785a/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java#L221

Then it wakes up the Handler thread, which may try to flush current buffered serialized data, but not flush all events in buffer:
https://github.com/elastic/apm-agent-java/blob/24424d6f38531865c99ef013caa49f7108e3785a/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java#L225

This also changes the state of the Handler so that it shouldn't continue trying to send events, but in any case, this thread is a daemon thread, so it shouldn't prevent shutdown.

The only blocking thing here seems to be the `publish`, which I assume would be fixed by changing to `tryPublish` as you did. So I see how the shutdown is delayed by the `publish` call (which doesn't guarantee success after 36 seconds if other publishers are trying to send events as well) plus the 5 seconds wait, but not longer.

What am I missing here?

felixbarny:
> The only blocking thing here seems to be the publish, which I assume would be fixed by changing to tryPublish as you did.

That's correct. I think with this change we're all good.

PR #555, opened 2019-04-02
Add support for WebLogic

### LINKED ISSUE #109: support for weblogic servers
I was wondering if you have ever tried this agent with oracle weblogic and if not if there are any plans to do so in the future.
Thank you in advance.

felixbarny:
closes elastic/apm-agent-java#109

### DIFF
--- a/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
+++ b/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
@@ -148,19 +148,11 @@ public class ServletTransactionHelper {
     public void onAfter(Transaction transaction, @Nullable Throwable exception, boolean committed, int status, String method,
                         @Nullable Map<String, String[]> parameterMap, String servletPath, @Nullable String pathInfo, @Nullable String contentTypeHeader) {
         try {
-            fillRequestParameters(transaction, method, parameterMap, contentTypeHeader);
-            if(exception != null && status == 200) {
-                // Probably shouldn't be 200 but 5XX, but we are going to miss this...
-                status = 500;
-            }
-            fillResponse(transaction.getContext().getResponse(), committed, status);
-            transaction.withResult(ResultUtil.getResultByHttpStatus(status));
-            transaction.withType("request");
-            if (transaction.getName().length() == 0) {
-                applyDefaultTransactionName(method, servletPath, pathInfo, transaction.getName());
-            }
-            if (exception != null) {
-                transaction.captureException(exception);
+            // thrown the first time a JSP is invoked in order to register it
+            if (exception != null && "weblogic.servlet.jsp.AddToMapException".equals(exception.getClass().getName())) {
+                transaction.ignoreTransaction();
+            } else {
+                doOnAfter(transaction, exception, committed, status, method, parameterMap, servletPath, pathInfo, contentTypeHeader);
             }
         } catch (RuntimeException e) {
             // in case we screwed up, don't bring down the monitored application with us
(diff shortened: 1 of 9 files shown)


felixbarny:
TODO
- [ ] create a hub.docker.com user for CI, accept the conditions for `store/oracle/weblogic:12.2.1.3-dev`, login with this user on CI in order to have access to the image. @kuisathaverat could you help here?
- [ ] only enable on specific environments, like CI, where we have access to the image

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
```diff
@@ -148,19 +148,11 @@ public static void setUsernameIfUnset(@Nullable String userName, TransactionCont
...
-            fillResponse(transaction.getContext().getResponse(), committed, status);
-            transaction.withResult(ResultUtil.getResultByHttpStatus(status));
-            transaction.withType("request");
-            if (transaction.getName().length() == 0) {
-                applyDefaultTransactionName(method, servletPath, pathInfo, transaction.getName());
-            }
-            if (exception != null) {
-                transaction.captureException(exception);
+            // thrown the first time a JSP is invoked in order to register it
+            if (exception != null && "weblogic.servlet.jsp.AddToMapException".equals(exception.getClass().getName())) {
```
eyalkoren: We agreed not to add platform/library/framework-specific code into general agent code

felixbarny: The case we discussed was to not add JSF specific stuff in the servlet plugin. But this is still Servlet, though also WebLogic specific. We already have application server specifics, for example in

https://github.com/elastic/apm-agent-java/blob/39ee1f05e213c090d7c27b18b30b39d3bc09998d/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ServiceNameUtil.java#L58-L71

and

https://github.com/elastic/apm-agent-java/blob/39ee1f05e213c090d7c27b18b30b39d3bc09998d/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/helper/ApmAsyncListener.java#L119-L122

So I think servlet container/application server specifics are ok to have directly in the servlet plugin.

felixbarny:
jenkins run the tests please

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
eyalkoren: I think they are all similar.
If this makes everything much simpler then I don't have a problem with it, but if it's easy to separate then it is probably better.
Maybe when it gets too loaded with such stuff we can add an app-server-specific API for which a factory chooses an implementation based on the app-server on startup.
I don't think you should waste efforts on this, only move if very simple.

felixbarny:
jenkins run the tests please

felixbarny:
@kuisathaverat is there an environment variable like `CI=true` set on Jenkins so that I can easily disable this test by default if not run on CI?

You ticked the box for docker login. Does that mean that the CI can pull the `store/oracle/weblogic:12.2.1.3-dev` image and that the test should be able to succeed now? It does not seem like it's working: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/PR-555/3/pipeline/80/#step-128-log-747

> WebLogicIT.<init>:39 » ContainerFetch Can't get Docker image: RemoteDockerImag...

Also important, will the packer image builder be able to pull the `store/oracle/weblogic:12.2.1.3-dev` image? Do we need to make sure to `docker login` here too? (/cc @fatmcgav)

kuisathaverat:
>is there an environment variable like CI=true set on Jenkins so that I can easily disable this test by default if not run on CI?

The most commonly used are the following

```
BUILD_ID
BUILD_NUMBER
```

check that it is not blank is enough

>Also important, will the packer image builder be able to pull the store/oracle/weblogic:12.2.1.3-dev image? Do we need to make sure to docker login here too?

I think so

kuisathaverat:
I've made the PR to make the login in docker.io https://github.com/elastic/apm-agent-java/pull/560

kuisathaverat:
jenkins run the test please

PR #557, opened 2019-04-02
Create spans for forward/include

### LINKED ISSUE #429: Use JSP name for JSP-transactions instead of servlet name

kananindzya:
closes elastic/apm-agent-java#384


apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/RequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,73 @@
...
+    public Class<?> getAdviceClass() {
+        return RequestDispatcherInstrumentation.RequestDispatcherAdvice.class;
+    }
+
+    public static class RequestDispatcherAdvice extends RequestDispatcherInstrumentation {
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        private static void beforeExecute(@Advice.Argument(0) @Nullable String path,
+                                          @Advice.This @Nullable Object thiz) {
+            if (thiz instanceof HttpServletRequest) {
+                ((HttpServletRequest) thiz).setAttribute(RequestDispatcher.FORWARD_SERVLET_PATH, (path != null) ? path : "");
```
felixbarny: Isn't that property set by the servlet container by default? Do we really have to set it explicitly via instrumentation?

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardIncludeRequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public ElementMatcher<? super NamedElement> getTypeMatcherPreFilter() {
+        return nameContainsIgnoreCase("dispatcher");
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return nameContainsIgnoreCase("dispatcher");
```
felixbarny: `hasSuperClass(named("javax.servlet.RequestDispatcher"));`

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardIncludeRequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,111 @@
...
+                span.captureException(t)
+                    .deactivate()
+                    .end();
+            }
+        }
+    }
+
+    @Override
+    public ElementMatcher<? super NamedElement> getTypeMatcherPreFilter() {
+        return nameContainsIgnoreCase("dispatcher");
```
felixbarny: `nameContains("Dispatcher");`

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardIncludeRequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public static class ForwardIncludeRequestDispatcherAdvice extends ForwardIncludeRequestDispatcherInstrumentation {
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        private static void beforeExecute(@Advice.Local("span") @Nullable Span span,
+                                          @Advice.Origin("#m") String method,
+                                          @Advice.Argument(0) @Nullable ServletRequest request) {
+            if (tracer == null || tracer.getActive() == null) {
+                return;
+            }
+            final TraceContextHolder<?> parent = tracer.getActive();
```
felixbarny: The instrumented RequestDispatcher might call another, delegate RequestDispatcher which would lead to the creation of two identical spans. See for example https://github.com/spring-projects/spring-security/blob/9c1eac79e2abb50f7b01e77c2418566f2a30532f/web/src/main/java/org/springframework/security/web/firewall/RequestWrapper.java#L151-L155.

It's a tough one to detect, because there can be legitimate cases where you have a `FORWARD` span as a direct parent of another `FORWARD` span. 🤔

apm-agent-plugins/apm-servlet-plugin/src/test/java/co/elastic/apm/agent/servlet/ApmFilterTest.java
```diff
@@ -235,7 +235,7 @@ void captureTransactionNameBasedOnServlet() throws IOException, ServletException
         filterChain = new MockFilterChain(new TestServlet());
         filterChain.doFilter(new MockHttpServletRequest("GET", "/foo"), new MockHttpServletResponse());
         assertThat(reporter.getTransactions()).hasSize(1);
-        assertThat(reporter.getFirstTransaction().getName().toString()).isEqualTo("ApmFilterTest$TestServlet#doGet");
+        assertThat(reporter.getFirstTransaction().getName().toString()).isEqualTo("GET ApmFilterTest$TestServlet");
```
felixbarny: I'm not sure if I like this format better. For other transaction names, we use either `$method $path`, like `GET /test` or `$class#$method`, like `MyServlet#doGet`. This is some sort of mix, like `$method $class`.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/RequestDispatcherInstrumentation.java
kananindzya: no, it isn't set by default. I tested on tomcat 8.5.5

felixbarny: That would be very surprising. The javadoc for `javax.servlet.RequestDispatcher#FORWARD_SERVLET_PATH` states:
> The name of the request attribute under which the original
servlet path is made available to the target of a 
{@link #forward(ServletRequest,ServletResponse) forward}

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardIncludeRequestDispatcherInstrumentation.java
felixbarny: You could probably check if the parent span has the same name as this span would have.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardIncludeRequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,111 @@
...
+
+            if (request != null && request instanceof HttpServletRequest) {
+                HttpServletRequest httpServletRequest = (HttpServletRequest) request;
+                String servletPath = "";
+                if (FORWARD.equalsIgnoreCase(method)) {
+                    servletPath = (String) httpServletRequest.getAttribute(RequestDispatcher.FORWARD_SERVLET_PATH);
+                    span = parent.createSpan().withType(SPAN_TYPE_REQUEST_DISPATCHER_FORWARD).withName(FORWARD);
+                } else {
+                    servletPath = (String) httpServletRequest.getAttribute(RequestDispatcher.INCLUDE_SERVLET_PATH);
+                    span = parent.createSpan().withType(SPAN_TYPE_REQUEST_DISPATCHER_INCLUDE).withName(INCLUDE);
```
felixbarny: Instead of a dot-separated span type, use the `withSubtype()` and `withAction()` methods.

kananindzya: as I understand it, it will look like this:
```
withType("servlet")
.withSubType("request-dispatcher")
.withAction("forward")
```
?

felixbarny: looks good

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardIncludeRequestDispatcherInstrumentation.java
kananindzya: @felixbarny how access to parentSpan via  
```
final TraceContextHolder<?> parent = tracer.getActive();?
```

felixbarny: You can do an `instanceof AbstractSpan` check and cast

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardIncludeRequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,129 @@
...
+                } else {
+                    servletPath = (String) httpServletRequest.getAttribute(RequestDispatcher.INCLUDE_SERVLET_PATH);
+                    pathInfo = (String) httpServletRequest.getAttribute(RequestDispatcher.INCLUDE_PATH_INFO);
+                    span = parent.createSpan().withType(SPAN_TYPE).withSubtype(SPAN_SUBTYPE).withAction(INCLUDE_SPAN_ACTION).withName(INCLUDE);
+                }
+                if (servletPath != null && !servletPath.isEmpty()) {
+                    span.appendToName(" ").appendToName(servletPath);
+                }
+                if (pathInfo != null && !pathInfo.isEmpty()) {
+                    span.appendToName("#").appendToName(pathInfo);
```
kananindzya: In all integration tests this value is already null or empty.

kananindzya:
@felixbarny  please check pull request, i deleted instrumentation on RequestDispatcher. 
And added integration tests on forward, include.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardIncludeRequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,129 @@
...
+            }
+            final TraceContextHolder<?> parent = tracer.getActive();
+
+            if (request != null && request instanceof HttpServletRequest) {
+                HttpServletRequest httpServletRequest = (HttpServletRequest) request;
+                String servletPath = "";
+                String pathInfo = "";
+
+                if (FORWARD.equalsIgnoreCase(method)) {
+                    servletPath = (String) httpServletRequest.getAttribute(RequestDispatcher.FORWARD_SERVLET_PATH);
```
felixbarny: Didn't you say these properties are not set yet? Or are they only set on the real application servers but were not on the unit tests?

integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/tests/ServletApiTestApp.java
```diff
@@ -93,6 +97,63 @@ private void testHttpUrlConnection(AbstractServletContainerIntegrationTest test)
         assertThat(span.get("context").get("http").get("status_code").intValue()).isEqualTo(200);
     }
 
+    private void testTransactionReportingWithForward(AbstractServletContainerIntegrationTest test) throws Exception {
+        String pathToTest = "/simple-webapp" + "/forward";
+        boolean isExistForwardSpan = false;
+        boolean isExistDbH2QuerySpan = false;
+        List<String> spanTypes = Arrays.asList("db.h2.query", "servlet.request-dispatcher.forward");
+        List<String> expectedValues = Arrays.asList("FORWARD", "FORWARD /echo");
```
felixbarny: ah, so it works on some application server but on others, it doesn't?

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardIncludeRequestDispatcherInstrumentation.java
kananindzya: I tested  with Apache Tomcat Version 8.5.5, and those properties were not there.
They were not on the unit tests, too.
I don't know, if I should  instrument RequestDispatcher .

integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/tests/ServletApiTestApp.java
kananindzya: no, it doesn't work on all application servers.
For this reason I added 2 expected values, but "FORWARD" - is actual.

felixbarny:
Some ideas about alternatives which enable the capturing of the path:
* Just instrument `javax.servlet.ServletRequest#getRequestDispatcher` and return a  `RequestDispatcherWrapper`. Similar to `RequestStreamRecordingInstrumentation`.
  * Pro: easy to implement
  * Con: Allocates memory, does not capture `ERROR` requests
* Enhance `ServletApiAdvice` to create spans for `DispatcherType`s `FORWARD`, `INCLUDE` and potentially `ERROR`
  * Pro: we can simply read the path from the `request.getServletPath()` and `request.getPathInfo()`
  * Con: Makes the complex `ServletApiAdvice` even more complex

Another option is to just omit the path (the current state of the PR).

@kananindzya @eyalkoren WDYT?

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardIncludeRequestDispatcherInstrumentation.java
eyalkoren: I think there was a confusion here- the attribute `RequestDispatcher.FORWARD_SERVLET_PATH` provides the servlet path of the original request, but here we are interested in the new request paths, so just `HttpServletRequest#getServletPath` and `HttpServletRequest#getPathInfo`.

integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/tests/ServletApiTestApp.java
eyalkoren: Your ForwardingServlet forwards to `/servlet` but you expect `/echo`. Don't allow two options unless you must. See if there is a way to look for a definitive name, otherwise we are not testing ourselves properly.

apm-agent-plugins/apm-servlet-plugin/src/test/java/co/elastic/apm/agent/servlet/ServletInstrumentationTest.java
```diff
@@ -113,12 +116,18 @@ void testNoopInstrumentation() throws Exception {
 
     @Test
     void testForward() throws Exception {
+        List<String> expectedValues = Arrays.asList("FORWARD", "FORWARD /test");
```
eyalkoren: The current code would probably yield the name `FORWARD /forward`, but after fixing this should work with a single expected value, not two options.

eyalkoren:
@felixbarny I just saw your comment now.

What I thought when reviewing was to create the span as it is created now in this PR with the basic name only, but update its name in the `service` that is ran for the forwarded/included servlet, either by setting the span as a Request attribute and assume the servlet container will copy it to the new request it generates (needs to be tested) or by looking for the active span and if it is the FORWARD/INCLUDE one update its name (would not work if spans are created in between).

However, 
> Enhance ServletApiAdvice to create spans for DispatcherTypes FORWARD, INCLUDE and potentially ERROR

may actually be a better option - you mean that we rely on the `RequestDispather#FORWARD_XXX/INCLUDE_XXX/ERROR_XXX` attributes to decide if and which type of span this is? 

> Con: Makes the complex ServletApiAdvice even more complex

I don't think so. Seems pretty straightforward.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardIncludeRequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,129 @@
...
+import javax.servlet.ServletRequest;
+import javax.servlet.http.HttpServletRequest;
+import java.util.Arrays;
+import java.util.Collection;
+
+import static net.bytebuddy.matcher.ElementMatchers.*;
+
+public class ForwardIncludeRequestDispatcherInstrumentation extends ElasticApmInstrumentation {
+
+    private static final String SPAN_TYPE = "servlet";
```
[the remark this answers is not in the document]
felixbarny: I think `app` might be a good type. For breakdown graphs, the transaction's self-time will also be attributed as `app` and it's also the default type for the `@CaptureSpan` annotation.

felixbarny:
>  you mean that we rely on the RequestDispather#FORWARD_XXX/INCLUDE_XXX/ERROR_XXX attributes to decide if and which type of span this is?

No, I meant relying on `javax.servlet.ServletRequest#getDispatcherType`.

But that can easily become messy - especially when there's a filter chain which may or may not invoke the target servlet and taking into account that there may be multiple forwards/includes.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ForwardIncludeRequestDispatcherInstrumentation.java
```diff
@@ -0,0 +1,129 @@
...
+                    span.appendToName(" ").appendToName(servletPath);
+                }
+                if (pathInfo != null && !pathInfo.isEmpty()) {
+                    span.appendToName(" ").appendToName(pathInfo);
+                }
+
+                if (parent instanceof AbstractSpan) {
+                    AbstractSpan parentSpan = (AbstractSpan) parent;
+                    if (parentSpan.getName().toString().equals(span.getName().toString())) {
+                        return;
```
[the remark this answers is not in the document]
felixbarny: Good question. I don't really know. Theoretically, that could happen, although not very likely.

eyalkoren:
@kananindzya So let's do this:
Instead of instrumenting `RequestDispatcher`, we will use our existing `doFilter` and `service` instrumentation, ie do everything through `ServletApiAdvice`. 

On method enter and exit, add a `@Nullabale @Advice.Local("span") Span span`. 
On method enter, we will rely on `javax.servlet.ServletRequest#getDispatcherType` to decide if this is a transaction, or a forward/include/error span. If the `DispatcherType` is `FORWARD`, `INCLUDE` or `ERROR`, AND `tracer.currentTransaction()==null`, we will start and activate a span, unless the current span has the same servletPath and pathInfo. 
On method exit, if the span is not null, deactivate it and end it.

WDYT?

integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/tests/ServletApiTestApp.java
kananindzya: my mistake, I deleted 2nd element.

eyalkoren:
@kananindzya we would like to implement that based on the logic described in the last comment.
Is it OK if I close this PR?

PR #561, opened 2019-04-05
[APM-CI] force pull weblogic image from CI

kuisathaverat:
testcontainers it is failing to pull the image so I guess it needs the image pulled before to use it, so until we could have that Docker image pre-pulled on VMs, we can force to pull it.
 
related to https://github.com/elastic/apm-agent-java/pull/555

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -158,6 +158,7 @@ pipeline {
             unstash 'build'
             dir("${BASE_DIR}"){
               dockerLogin(secret: "${DOCKERHUB_SECRET}", registry: "docker.io")
+              sh(label: 'pull weblogic Docker image', script: 'docker pull store/oracle/weblogic:12.2.1.3-dev')
               sh './scripts/jenkins/smoketests-02.sh'
             }
           }


felixbarny:
Maybe there is also a way to `docker login` with Testcontainers. That way, we wouldn't have to hard-code the containers we need to login for.

felixbarny:
Hm, on my local machine Testcontainers was able to fetch the image which logged in. 

Maybe the login didn't quite work yet. See this part of the logs:

```
[2019-04-05T11:43:52.896Z] [main] INFO org.testcontainers.dockerclient.DockerClientProviderStrategy - Found Docker environment with Environment variables, system properties and defaults. Resolved: 
[2019-04-05T11:43:52.896Z]     dockerHost=unix:///var/run/docker.sock
[2019-04-05T11:43:52.896Z]     apiVersion='{UNKNOWN_VERSION}'
[2019-04-05T11:43:52.896Z]     registryUrl='https://index.docker.io/v1/'
[2019-04-05T11:43:52.896Z]     registryUsername='jenkins'
[2019-04-05T11:43:52.896Z]     registryPassword='null'
[2019-04-05T11:43:52.896Z]     registryEmail='null'
[2019-04-05T11:43:52.896Z]     dockerConfig='DefaultDockerClientConfig[dockerHost=unix:///var/run/docker.sock,registryUsername=jenkins,registryPassword=<null>,registryEmail=<null>,registryUrl=https://index.docker.io/v1/,dockerConfigPath=/var/lib/jenkins/.docker,sslConfig=<null>,apiVersion={UNKNOWN_VERSION},dockerConfig=<null>]'
```

It says `registryUsername='jenkins'`. This is not the username which is stored in the `secret/apm-team/ci/elastic-observability-dockerhub` vault secret.

felixbarny:
The `docker login` also seems to cause some build failures due to timeouts: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/PR-562/3/pipeline/89#step-136-log-3

kuisathaverat:
the docker login works, if it doesn't work you could not make `docker pull` with the docker CLI, and you can 

Screenshot 2019-04-08 at 12 35 36

The thing is that the docker configuration file is in the WORKSPACE of the build and this is the same that the HOME environment variable points, it is not the real user HOME, I mean it is not the same that it is in the /etc/passwd file, I guess testcontainers is ignoring the value of the HOME environment variable and get the configuration in another way, I saw that there is a way to force testcontainers to grab a docker config file`DOCKER_CONFIG_FILE=/some/location/config.json`

https://www.testcontainers.org/modules/docker_compose/

kuisathaverat:
> The `docker login` also seems to cause some build failures due to timeouts: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/PR-562/3/pipeline/89#step-136-log-3

it is not related to our changes, this would happen always we have infra issues, because of that, we should include the image in the local cache as we've made with the other images. We've to confirm first that testcontainers don't pull the image if it is pulled locally.

kuisathaverat:
also seems more people has problems wit authenticated registries see the comments here https://github.com/testcontainers/testcontainers-java/issues/968 and https://github.com/testcontainers/testcontainers-java/pull/1007

PR #562, opened 2019-04-08
Allow constructor arguments in service implementations

felixbarny:


felixbarny:
It allows fields to be `@Nonnull` and `final` so we have to do less null checks. In contrast to the `init` method, it can be verified that the constructor is called before the other methods.

eyalkoren:
Maybe it can even help me with my unit tests, let's see after you merge 🙂

PR #563, opened 2019-04-08
Check if class loader can load class via resource lookup

felixbarny:
this does not actually load the class thus preventing side-effects

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/CustomElementMatchers.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/CustomElementMatchers.java
@@ -23,6 +23,7 @@ import co.elastic.apm.agent.matcher.WildcardMatcher;
 import com.blogspot.mydailyjava.weaklockfree.WeakConcurrentMap;
 import net.bytebuddy.description.NamedElement;
 import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.dynamic.ClassFileLocator;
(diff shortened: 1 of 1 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/CustomElementMatchers.java
```diff
@@ -87,18 +87,19 @@ public boolean matches(@Nullable ClassLoader target) {
     private static boolean canLoadClass(@Nullable ClassLoader target, String className) {
         boolean result;
         try {
-            Class<?> clazz = Class.forName(className, false, target);
+            if (target == null) {
+                target = ClassLoader.getSystemClassLoader();
```
eyalkoren:
```suggestion
                URL resource = String.class.getResource("/" + className.replace('.', '/') + ".class");
```

What you did will return `true` for the bootstrap class loader and any class loaded from the system classpath, which means this filter will be inefficient for lots of classes. 
A false positive is not so bad here (just a matter of startup time), so we can live with that. However, what I suggested will behave better in some JVMs (at least openjdk 10+)

eyalkoren:
```suggestion
                target = ClassLoader.getSystemClassLoader().getParent();
```

This will reduce a lot of false positives, for environments where dependencies we are filtering for were appended to the system classpath. 
You should check that this doesn't yield `null`, but normally it shouldn't.

PR #565, opened 2019-04-08
[APM-CI][Java] Add WebLogic container to packer cache

felixbarny:


felixbarny:
@kuisathaverat do you think that should work? Do we need to `docker login` on the host which builds the CI worker images?

kuisathaverat:
yep, if we use our Docker registry as cache, I am waiting for a namespace for Observability team once we have it, we would put the image in our repository and from there to our packer. I will move this issue to my kanban

kuisathaverat:
@felixbarny it is ready now, we have the image in the private repo. I'd make a PR to test the cache.

felixbarny:
Oh, cool! That was quick :D

PR #567, opened 2019-04-08
[improvement,feature] Support groovy code with the annotation API:

fredgalvao:
As per https://discuss.elastic.co/t/kibana-apm-not-showing-capturespan-and-capturetransaction/173026 and a few other tests I did myself in the last few weeks, apm-agent-java couldn't support code that went through groovyc with the annotation API (`CaptureSpan`, `CaptureTransaction`). I initially only saw that the annotations weren't making it to the final `.class` file, but couldn't investigate properly at the time.

It was only recently that I noticed that there was no `RetentionPolicy` for these two annotations, which was making the default `RetentionPolicy#CLASS` the chosen one, which can't really be seen during runtime by bytebuddy for things it doesn't fully understand. See [this](https://stackoverflow.com/a/30032884/10700198) for more details on this, even though I can't really explain why bytebuddy would see it for java classes during runtime...

Changing the `RetentionPolicy` to `RUNTIME` enables everything I wanted and could test:
- tested on both:
  - grails[2.2.4], groovy[2.0.8]
  - grails[3.3.1], groovy[2.4.11]
- tested on both:
  - `@CompileStatic`
  - Normal(dynamic) compilation
- tested on both:
  - Spring components/artifacts
  - Raw code

Tests were already failing on my machine on master, so I really don't know if I could ever break anything with this change, but I'm willing to bet it will be inoffensive.

### DIFF
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/CaptureSpan.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/CaptureSpan.java
@@ -19,8 +19,10 @@
  */
 package co.elastic.apm.api;
 
+import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
 /**
  * Annotating a method with {@code @}{@link CaptureSpan} creates a {@link Span} as the child of the currently active span or transaction
@@ -34,6 +36,7 @@ import java.lang.annotation.RetentionPolicy;
  * </p>
  */
 @Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
 public @interface CaptureSpan {
 
     /**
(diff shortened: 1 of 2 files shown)


elasticcla:
Hi @fredgalvao, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/apm-agent-java/pull/567.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

fredgalvao:
Sorry for the diff email thing. I've now fixed it.

felixbarny:
Oh, boy, I can’t believe this oversight :facepalm:

Thanks a lot for investigating. I’m wondering how this has ever worked. While at it, could you also add the @Target(METHOD) annotation?

fredgalvao:
My pleasure to help ;)

> I’m wondering how this has ever worked.

#metoo

> While at it, could you also add the @Target(METHOD) annotation?

Done.

Issue #440, opened 2019-01-23
Discard fast trace_methods spans


danielfariati:
**Is your feature request related to a problem? Please describe.**
It would be nice to have the option to discard spans that were too fast when enabling _trace_methods_.
That way, we would avoid sending too much data to the server for calls that aren't that relevant, when using _trace_methods_ to find performance bottlenecks.

**Describe the solution you'd like**
A simple configuration property would solve this. Something like _trace_methods_min_duration_.

danielfariati:
The only problem with that, would be fast methods being called a lot of times.
As a single execution would be fast, every span would be discarded.
So, the performance bottleneck in this case would be a little obfuscated.
I don't think it is REALLY a problem, as the method doing the calls would still show up as a span (if it is also included in the _trace_methods_ regex).

eyalkoren:
@danielfariati thanks for the input. This indeed sounds like a nice enhancement. Unfortunately it is not trivial to achieve with the way the agent works. Some things to consider:
1. the trace_methods span can create a child span that makes a quick-enough HTTP call, or even just a DB query we don't want to discard - this breaks the chain as there will be a missing parent
2. even if we wanted to "fix" the parent reference in the child span when the discarded span is finished, this could be too late, as the way the agent currently work is just serialize and send spans when they are finished, so the child span will not be available
3. how would we present that? Just ignore? this mean non-deterministic results

It would be nice if you would specify what is the use case you are thinking this will be useful for and what is the main concern you would like to address by having that.
Thanks.

eyalkoren:
One option to deal with problems 1 and 2:
Never discard spans on a branch that lead to "important spans" (eg DB, HTTP). The way to know whether a method is on such a branch is by marking spans on the recorded span stack. Imagine the following span stack state: `A->B->C->D->E` where `E` is an "important span". When `E` exits it marks its parent as "cannot-be-discarded", denoted here with `*`. Our stack would be then: `A->B->C->D*`. When `D` exits we know it cannot be discarded because it is marked as such and we transfer this mark to the parent method on the stack, so after `D` exits (and being reported), the stack would look like this: `A->B->C*`.  If now some new executions occur, the stack would look like this: `A->B->C*->F->G->H`. `F`, `G` and `H` are still discardable and so it will remain until the stack state will once again be `A->B->C*` or another "important span" occurs.

danielfariati:
[...]

PR #577, opened 2019-04-11
Discard fast methods coming from trace_methods

eyalkoren:
Closes elastic/apm-agent-java#440 

Made a few decisions:
1. Never discard transactions
2. Never discard branches leading to async branches so not to break span parent referencing
3. Off by default (config option set to 0)
4. When on (ie threshold > 0) - branches leading to spans ending with Exceptions are not discarded

The last one is explicitly related to a threshold setting, however, it makes sense that if someone opts to lose non-interesting, error traces are still valuable


apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -295,6 +292,25 @@
...
+            "duration. When set to a value greater than 0, spans representing methods traced based on `trace_methods` will be discarded " +
+            "by default.\n" +
+            "Such methods will be traced and reported if one of the following applies:\n" +
+            " - This method's duration crossed the configured threshold.\n" +
+            " - This method ended with Exception.\n" +
+            " - A method executed as part of the execution of this method crossed the threshold or ended with Exception.\n" +
+            " - A \"forcibly-traced method\" (e.g. DB queries, HTTP exits, custom) was executed during the execution of this method.\n" +
+            "Set to 0 to disable.\n" +
+            "\n" +
+            "NOTE: Transaction are never discarded, regardless of their duration. This configuration affects only spans.\n" +
```
felixbarny:
```suggestion
            "NOTE: Transactions are never discarded, regardless of their duration. This configuration affects only spans.\n" +
```

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -295,6 +292,25 @@
...
+            "In order not to break span references, all spans leading to an async operations are never discarded, regardless \n" +
+            "of their duration.\n")
+        .buildWithDefault(TimeDuration.of("0ms"));
```
felixbarny: Should we enhance TimeDuration to add support for `us` (µs)?

apm-agent-plugins/apm-java-concurrent-plugin/src/test/java/co/elastic/apm/agent/concurrent/AsyncTraceMethodInstrumentationTest.java
```diff
@@ -0,0 +1,149 @@
...
+        new TestAsyncTraceMethodsClass().invokeAsync();
+        assertThat(reporter.getTransactions()).hasSize(1);
+        assertThat(reporter.getSpans()).hasSize(4);
+    }
+
+
+    private static class TestAsyncTraceMethodsClass {
+
+        /**
+         * Calling this method results in this method call tree:
```
felixbarny: seems like the formatting got messed up, maybe your local history still contains the old version

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
eyalkoren: Not sure if sub-millisecond is enhancing enough. I imagine someone using `trace_methods` with wildcards and hoping to get performance costs down shouldn't configure microseconds thresholds.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -261,7 +258,7 @@
     private final ConfigurationOption<List<MethodMatcher>> traceMethods = ConfigurationOption
         .builder(new ListValueConverter<>(MethodMatcherValueConverter.INSTANCE), List.class)
         .key("trace_methods")
-        .tags("added[1.3.0,Enhancements in 1.4.0]")
+        .tags("added[1.3.0,Enhancements in 1.4.0 and 1.6.0]")
```
felixbarny: `1.7.0` now
But I'm not sure if that can be considered an enhancement to the `trace_methods` directly.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
```diff
@@ -293,6 +296,14 @@ void setRecorded(boolean recorded) {
         }
     }
 
+    public void setDiscard(boolean discard) {
+        this.discard = discard;
```
felixbarny: We have a similar flag in `Transaction` called `noop`. Maybe we should rather pull that up to `AbstractSpan`?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -475,6 +475,10 @@ public void deactivate(TraceContextHolder<?> holder) {
         }
         final Deque<TraceContextHolder<?>> stack = activeStack.get();
         assertIsActive(holder, stack.poll());
+        if (!stack.isEmpty() && !holder.isDiscard()) {
+            //noinspection ConstantConditions
+            stack.peek().setDiscard(false);
```
felixbarny: don't we have to set discard to false on all elements in the stack?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
```diff
@@ -95,6 +95,8 @@ public boolean asChildOf(TraceContext child, Object ignore) {
     private final Id transactionId = Id.new64BitId();
     private final StringBuilder outgoingHeader = new StringBuilder(TRACE_PARENT_LENGTH);
     private byte flags;
+    private boolean discard;
```
felixbarny: also copy on `asChildOf`?

eyalkoren:
> I think this approach may not be generic enough as there might be async spans without a prior wrapping.

Isn't this the same with reference counting, which relies on wrapping as well?

> Also, async spans which don't outlive their parent can safely be discarded when they are fast.

May not be consistent - can sometimes outlive parent and sometimes not.

In general, this PR is an enhancement for `trace_methods` users that use it with wildcards. It is still not the recommended way of using this feature.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
eyalkoren: > But I'm not sure if that can be considered an enhancement to the `trace_methods` directly.

Why is that? It is only related to `trace_methods` and has no effect when not used. It extends the configuration capabilities of this feature.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
eyalkoren: Seems like currently they effectively serve a similar purpose. I will move it and if we see a new need to differentiate `noop` from `discard` we can reiterate later.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
eyalkoren: Not really. All spans are created as non-discardable, except from the `trace_methods` spans. This means there is no race between `deactivate` and `end` and it should be safe to do it this way.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
eyalkoren: No. Only `trace_methods` spans should be created as discardable. Their children shouldn't, unless the are such as well.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
eyalkoren: Actually, this breaks the logic used for this flag- `noop` transactions are not-sampled, and this is an inheritable trait, so all decedents will be non-sampled. However, marking a span as `discard` should not be inherited and should not affect parent (as opposed to the `discard` flag).

eyalkoren:
Jenkins run the tests

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
felixbarny: gotcha

PR #580, opened 2019-04-12
[APM-CI][Java][WIP] Release stage

kuisathaverat:
* Release process from the CI
* Add some editorconfig rules for Jenkinsfiles


Jenkinsfile
```diff
@@ -267,6 +268,41 @@ pipeline {
...
+        GNUPGHOME="${env.TMP_WORKSPACE}/keyring"
+      }
+      input {
+        message 'Should we release a new version?'
+        ok 'Yes, we should.'
+      }
+      when {
+        beforeAgent true
+        anyOf {
+          tag "v\\d+\\.\\d+\\.\\d+*"
```
v1v: This won't work as it does use the `GLOB` comparator by default, you need to use something like
```suggestion
          tag pattern: 'v\\d+\\.\\d+\\.\\d+.*', comparator: 'REGEXP'
```

## Further details:

The current when implementation for the release stage in the CI uses `GLOB` rather than `REGEX` refers to:
- https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/b257a9e40fac92c9880d438a9be787f9a8d23ed8/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/when/impl/TagConditional.java#L83

and [docs](https://jenkins.io/doc/book/pipeline/syntax/#when) see tag section in the [when docs](https://jenkins.io/doc/book/pipeline/syntax/#when) page.

```
The optional parameter comparator may be added after an attribute to specify how any patterns are evaluated for a match: EQUALS for a simple string comparison, GLOB (the default) for an ANT style path glob (same as for example changeset), or REGEXP for regular expression matching. For example: when { tag pattern: "release-\\d+", comparator: "REGEXP"}
```

v1v:
In case we do this approach, should we remove the jjb for the release within the infra repo which refers to [elastic+apm-agent-java+release](https://internal-ci.elastic.co/job/elastic+apm-agent-java+release/)

kuisathaverat:
superseded by https://github.com/elastic/apm-agent-java/issues/991

PR #583, opened 2019-04-15
Changing Byte Buddy Location Strategy to WEAK

eyalkoren:
According to https://stackoverflow.com/questions/55616650/byte-buddy-locationstrategy-types - this should be fine, since we are using a per-class-loader cache for the resolved `TypeDescription`s, and therefore preferable in our case as it avoids class loader leaks.


eyalkoren:
Jenkins test this

PR #587, opened 2019-04-16
Recycle via reference counting

felixbarny:
- Recycling is not done explicitly but implicitly when the reference count goes to 0
- This also simplifies async context propagation
  - There's no need to copy the context to the other thread
  - The Span is just activated on the other thread
      If it's ended before the child starts it is still not recycled because the reference count is not at zero yet.
- This is in also in preparation for breakdown metrics
  - You need to have a reference to the parent in order to calculate the self-time of spans
  - In order to be able to calculate the self-time in async scenarios, it's required to activate the span as opposed to just it's TraceContext


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -219,4 +232,24 @@ public Span withStacktrace(Throwable stacktrace) {
...
+    }
+
+    @Override
+    public void decrementReferences() {
+        if (transaction != null) {
+            transaction.decrementReferences();
+        }
+        final int referenceCount = references.decrementAndGet();
+        if (referenceCount == 0) {
+            tracer.recycle(this);
```
eyalkoren: If end happens here as well, you won't need to manage transaction references I think as it will keep the transaction alive

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -58,14 +58,25 @@
...
+    private Transaction transaction;
 
     public Span(ElasticApmTracer tracer) {
         super(tracer);
     }
 
     public <T> Span start(TraceContext.ChildContextCreator<T> childContextCreator, T parentContext, long epochMicros, boolean dropped) {
-        onStart();
         childContextCreator.asChildOf(traceContext, parentContext);
+        if (parentContext instanceof Transaction) {
```
eyalkoren: Since this is where you make the parent and transaction references, I think this is the best place to update reference counters. This way you won't miss.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/async/SpanInScopeCallableWrapper.java
```diff
@@ -46,6 +46,7 @@ public SpanInScopeCallableWrapper(ElasticApmTracer tracer) {
     public SpanInScopeCallableWrapper<V> wrap(Callable<V> delegate, AbstractSpan<?> span) {
         this.delegate = delegate;
         this.span = span;
+        span.incrementReferences();
```
eyalkoren: move to `call` for wrappers created but not invoked

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
```diff
@@ -213,4 +213,17 @@ public String getType() {
     public String toString() {
         return String.format("'%s' %s", name, traceContext);
     }
+
+    @Override
+    public void incrementReferences() {
+        super.incrementReferences();
```
eyalkoren: is this override redundant?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
```diff
@@ -382,6 +382,11 @@ public Span createSpan() {
         return tracer.startSpan(fromParent(), this);
     }
 
+    @Override
+    public Span createSpan(long epochMicros) {
+        return tracer.startSpan(fromParent(), this, epochMicros);
```
eyalkoren: Another one that escapes the parent reference counting?

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java
```diff
@@ -115,7 +115,7 @@ public Thread newThread(Runnable r) {
     @Override
     public void report(Transaction transaction) {
         if (!tryAddEventToRingBuffer(transaction, TRANSACTION_EVENT_TRANSLATOR)) {
-            transaction.recycle();
+            transaction.decrementReferences();
```
eyalkoren: Also won't be needed if spans are created with reference count of 0 and ended implicitly when references go down to 0

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/ExecutorInstrumentation.java
```diff
@@ -61,7 +61,8 @@
     public ElementMatcher<? super TypeDescription> getTypeMatcher() {
         return hasSuperType(named("java.util.concurrent.Executor"))
             // hazelcast tries to serialize the Runnables/Callables to execute them on remote JVMs
-            .and(not(nameStartsWith("com.hazelcast")));
+            .and(not(nameStartsWith("com.hazelcast")))
+            .and(not(nameStartsWith("org.apache.felix.resolver")));
```
eyalkoren: wrong PR 😄

apm-agent-plugins/apm-java-concurrent-plugin/src/test/java/co/elastic/apm/agent/concurrent/FailingExecutorInstrumentationTest.java
```diff
@@ -20,8 +20,8 @@
 package co.elastic.apm.agent.concurrent;
 
 import co.elastic.apm.agent.AbstractInstrumentationTest;
-import co.elastic.apm.agent.impl.async.ContextInScopeCallableWrapper;
-import co.elastic.apm.agent.impl.async.ContextInScopeRunnableWrapper;
```
eyalkoren: Did you delete these two, or are they still needed?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
felixbarny: I don't quite understand

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -20,39 +20,47 @@
...
 
 public abstract class AbstractSpan<T extends AbstractSpan> extends TraceContextHolder<T> {
     private static final Logger logger = LoggerFactory.getLogger(AbstractSpan.class);
     protected static final double MS_IN_MICROS = TimeUnit.MILLISECONDS.toMicros(1);
     protected final TraceContext traceContext;
 
-    // used to mark this span as expected to switch lifecycle-managing-thread, eg span created by one thread and ended by another
-    private volatile boolean isLifecycleManagingThreadSwitch;
-
     /**
```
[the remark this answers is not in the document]
felixbarny: Yes, the `wrap()` -> `incrementReferences()` -> `incrementAndGet()` call chain should act as a memory barrier

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -113,8 +121,8 @@ public void resetState() {
         name.setLength(0);
         timestamp = 0;
         duration = 0;
-        isLifecycleManagingThreadSwitch = false;
         traceContext.resetState();
+        references.set(0);
```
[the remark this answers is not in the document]
felixbarny: The check in `createTransaction` is important but I think we can remove this one. `resetState()` is only called when the reference count decrements to 0.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -136,8 +147,15 @@ public Span createSpan(long epochMicros) {
...
 
-    protected void onStart() {
+    /**
+     * Called after the span has been started and its parent references are set
+     */
+    protected void onAfterStart() {
         this.finished = false;
+        // this final reference is decremented when the span is reported
+        // or even after its reported and the last child span is ended
+        references.set(0);
```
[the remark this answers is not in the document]
felixbarny: I think this is a leftover. We can remove because the `ElasticApmTracer.createTransaction()` method verifies that the reference count is 0.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -126,8 +134,11 @@ public Span createSpan() {
         return createSpan(traceContext.getClock().getEpochMicros());
     }
 
+    @Override
     public Span createSpan(long epochMicros) {
-        return tracer.startSpan(this, epochMicros);
+        final Span span = tracer.startSpan(this, epochMicros);
+        onChildStart(span, epochMicros);
```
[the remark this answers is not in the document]
felixbarny: Why's that? If the parent is a `TraceContext`, `Span.transaction` and `Span.parent` is `null` but children of that span just reference-count normally.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -178,6 +190,9 @@ public void doEnd(long epochMicros) {
         if (type == null) {
             type = "custom";
         }
+        if (parent != null) {
+            parent.onChildEnd(this, epochMicros);
+        }
         this.tracer.endSpan(this);
```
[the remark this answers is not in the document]
felixbarny: But a span can end before the reference count is 0.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/async/SpanInScopeRunnableWrapper.java
```diff
@@ -45,6 +45,7 @@ public SpanInScopeRunnableWrapper(ElasticApmTracer tracer) {
     public SpanInScopeRunnableWrapper wrap(Runnable delegate, AbstractSpan<?> span) {
         this.delegate = delegate;
         this.span = span;
+        span.incrementReferences();
```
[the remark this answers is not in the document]
felixbarny: In those cases the GC will take care of this transaction. We can't move it to run as run might be called after the span is ended and sent to the APM Server. If we don't increment the references when we wrap, indicating that we expect this reference to be used later, it might already be recycled when run starts.

Visualization:

```
█████
└─────────────█████
    ^         ^
    submit()  run()
    wrap()
```

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -227,7 +227,16 @@ public void onChange(ConfigurationOption<?> configurationOption, Double oldValue
...
 
     public Transaction noopTransaction() {
-        return transactionPool.createInstance().startNoop();
+        return createTransaction().startNoop();
+    }
+
+    private Transaction createTransaction() {
+        Transaction transaction = transactionPool.createInstance();
+        while (transaction.getReferenceCount() != 0) {
+            logger.warn("Tried to start a transaction with a non-zero reference count {} {}", transaction.getReferenceCount(), transaction);
```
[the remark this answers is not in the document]
felixbarny: If the reference count is higher than 0 the transaction would not have been recycled. That can happen if we miss to deactivate, for example. But in this case, we have the safety net of the GC which will collect the instance if it's not reachable anymore.

But it could also happen that we deactivate again when the reference count is already 0 and we have already recycled the transaction. The pattern would be

```java
t.activate();
t.deactivate();
t.deactivate();
```

That actually happened before https://github.com/elastic/apm-agent-java/pull/605. In this case the reference count is also not 0. It's better not to try reusing this instance again as it might still be in use. So we would just let GC take care of this transaction which collects it once not referenced anymore.

Bottom line - I don't think there is an increased risk of memory leaks.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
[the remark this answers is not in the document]
eyalkoren: Not very important to remove, just wanted to know which scenarios this is for.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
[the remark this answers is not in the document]
eyalkoren: Sorry, I meant the other things that happen in end except from the timestamping, ie the parent reference update and the queueing for serialization and sending.
In other words, I suggested a seamless buffering of referenced spans without actually managing a buffer. This opens all kind of options for us for retrospective decisions. The problem with that is missed reference accounting (this is real memory leak) and very long waiting tasks like messaging or timers etc. This can be solved with timeouts, but surely complicates things.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/async/SpanInScopeRunnableWrapper.java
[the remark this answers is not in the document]
felixbarny: Not sure if an optimization for that (edge) case is worth the effort. I think we can add that later if we need. Maybe a better way would be to instrument the corresponding `Future`'s cancel method.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
eyalkoren: Deferred to be considered as a future enhancement

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/async/SpanInScopeCallableWrapper.java
eyalkoren: We can't because of spans being called after parent is finished.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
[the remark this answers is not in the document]
eyalkoren: Activation on the worker thread is reading the volatile from the `references` AtomicInteger, so should act as a full memory barrier.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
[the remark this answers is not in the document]
felixbarny: We can always add that later if needed. Let's defer this for now.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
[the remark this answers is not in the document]
felixbarny: They could break the reference chain so that breakdowns may not work if there's an OT span in between.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
felixbarny: yes, but it helps when debugging

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
felixbarny: In this case, there is no parent span to count references for

apm-agent-plugins/apm-java-concurrent-plugin/src/test/java/co/elastic/apm/agent/concurrent/FailingExecutorInstrumentationTest.java
felixbarny: I think we can remove those

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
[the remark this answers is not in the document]
felixbarny: > Actually, this seems the right place to write an error to the log

I initially had an `assert references.get() == 0`. But there are some tests which test the reset itself. So it is somewhat legal to reset without recycling a span. As you said recycling can only ever be triggered when the reference count goes to 0 so I think it does not really make sense to add a warning on recycling when the count is not zero as well.

apm-agent-plugins/apm-java-concurrent-plugin/src/test/java/co/elastic/apm/agent/concurrent/FailingExecutorInstrumentationTest.java
felixbarny: I was wrong. We still need that for `TraceContext.withActive`

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -229,19 +225,25 @@ public void setStartTimestamp(long epochMicros) {
...
-
-    void onChildEnd(Span span, long epochMicros) {
-        decrementReferences();
-    }
-
     public void incrementReferences() {
         references.incrementAndGet();
-        logger.trace("increment references to {} ({})", this, references);
+        if (logger.isDebugEnabled()) {
+            logger.trace("increment references to {} ({})", this, references);
```
eyalkoren: `logger.debug`?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -229,19 +225,25 @@ public void setStartTimestamp(long epochMicros) {
...
+                logger.trace("incrementing references at",
+                    new RuntimeException("This is an expected exception. Is just used to record where the reference count has been incremented."));
+            }
+        }
     }
 
-    public abstract void decrementReferences();
+    public void decrementReferences() {
+        if (logger.isDebugEnabled()) {
+            logger.trace("decrement references to {} ({})", this, references);
```
eyalkoren: `logger.debug`?

Issue #556, opened 2019-04-02
Add support for timing breakdowns


felixbarny:
Implementation of agent breakdown metrics (https://github.com/elastic/apm/issues/78)

Checklist/TL;DR version of the design:

- [x] [Calculate self-time of transactions and spans](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.ondan294nbpt)
  - [x] Prerequisite: Spans have a reference to their parent
  - [x] Add a child duration reentrant timer to transactions and spans
- [x] Add [feature flag](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.12wulrccoi9k)
- Metrics registry enhancements
  - [x] Add support for [timers](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.qk36t41ikda3)
  - [x] [Limit the max amount of timers](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.tzyh7loxf2is)
  - [x] Add support for counters
  - [x] Optional: [guard against concurrent reads/writes](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.aler9ixddp0e)
- [x] [Track span self-times on the transaction](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.g5pdtefj3ag)
  - [x] Prerequisite: Spans have a reference to their transaction
  - On span end:
    - [x] track the span self-time on the corresponding transaction
  - On transaction end
    - [x] track all self-times on the metric registry
    - [x] track the transaction's self time with the `span.type` `app`
    - [x] track the transaction's duration on the metrics registry
- [Reporting metrics](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.gkxkryn9w02s)
  - [x] Add support for top-level labels for `transaction.name`, `transaction.type`, `span.type` and `span.subtype`
[...]

PR #588, opened 2019-04-16
Timing breakdown

felixbarny:
closes elastic/apm-agent-java#556
depends on #587, https://github.com/elastic/apm-server/issues/2086

TODO
- [x] guard against concurrent reads (report metrics to APM Server) and writes (tracking of metrics)
- [x] Javadoc
- [x] ~Report `self_time` on transactions/spans?~
- [x] Feature flag
- [x] Limit timers to 1000
- [x] Thread-safe publication for `Labels#immutableCopy`

Open questions
- [x] resetting timers ~vs monotonically increment~ `Timer` counters
- [x] decide on a final format for top-level metric tags (https://github.com/elastic/apm-server/issues/2086)


apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/Labels.java
```diff
@@ -0,0 +1,259 @@
...
+        return new Labels();
+    }
+
+    public static Labels of(String key, CharSequence value) {
+        final Labels labels = new Labels();
+        labels.add(key, value);
+        return labels;
+    }
+
+    public static Labels of(Map<String, ? extends CharSequence> labelMap) {
```
felixbarny: unused

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/Labels.java
```diff
@@ -0,0 +1,259 @@
...
+import co.elastic.apm.agent.objectpool.Recyclable;
+
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+public class Labels implements Recyclable {
```
felixbarny: Labels allow for `CharSequence`s as a value, thus avoiding allocations for `transaction.name.toString()` when tracking breakdown metrics for a transaction.

Iterations over the labels also don't allocate an Iterator, in contrast to `Map.entrySet().iterator()`.

felixbarny:
@bmorelli25 the last commit contains documentation for the metrics. I'd love to get some feedback from you 🙂

docs/metrics.asciidoc
```diff
@@ -205,3 +205,65 @@ An approximation of the total amount of memory,
...
+[float]
+[[metrics-application]]
+=== Application Metrics
+
+*`transaction.duration`*::
++
+--
+type: simple timer
+
+This timer tracks the transaction durations and allows to create graphs displaying a weighted average.
```
bmorelli25:
```suggestion
This timer tracks the duration of transactions and allows for the creation of graphs displaying a weighted average.
```

docs/metrics.asciidoc
```diff
@@ -205,3 +205,65 @@ An approximation of the total amount of memory,
...
+
+
+*`transaction.breakdown.count`*::
++
+--
+type: long
+
+format: count (delta)
+
+Tracks for how many transactions breakdown metrics (`span.self_time`) have been created.
```
bmorelli25: This sentence is also confusing for me. What about:
```suggestion
The number of transactions for which breakdown metrics (`span.self_time`) have been created.
```

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
```diff
@@ -221,9 +239,63 @@ public void incrementReferences() {
...
+            timer = new Timer();
+            Timer racyTimer = spanTimings.putIfAbsent(spanType.immutableCopy(), timer);
+            if (racyTimer != null) {
+                timer = racyTimer;
+            }
+        }
+        timer.update(duration);
+        if (finished) {
+            // in case end()->trackMetrics() has been called concurrently
+            // don't leak timers
```
eyalkoren: Still not thread safe:
```
Thread A:
      AbstractSpan#end() -> Transaction#doEnd -> trackMetrics() // all timers are reset, finished == false

Thread B:
      AbstractSpan#end() -> Span#doEnd // creates an put a Timer, doesn't clean because finished == false

Thread A:
      AbstractSpan#end() -> Transaction#doEnd (continue) 
      AbstractSpan#end() -> sets finished = true 
```

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
```diff
@@ -40,93 +45,196 @@
...
 
-    public Map<Map<String, String>, MetricSet> getMetricSets() {
-        return metricSets;
+    public void report(MetricsReporter metricsReporter) {
+        try {
+            phaser.readerLock();
+            ConcurrentMap<Labels.Immutable, MetricSet> temp = inactiveMetricSets;
+            inactiveMetricSets = activeMetricSets;
+            activeMetricSets = temp;
+            phaser.flipPhase();
```
eyalkoren: This is blocking the Event handler- we should try and avoid that. I think we need to use a dedicated thread to prepare serialized metrics events.

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
```diff
@@ -40,93 +45,196 @@
...
+            phaser.readerUnlock();
+        }
+    }
+
+    public void updateTimer(String timerName, Labels labels, long durationUs) {
+        updateTimer(timerName, labels, durationUs, 1);
+    }
+
+    public void updateTimer(String timerName, Labels labels, long durationUs, long count) {
+        long criticalValueAtEnter = phaser.writerCriticalSectionEnter();
```
eyalkoren: I think in the current scenario this is unnecessary, as it is guarded from the outside. Maybe that's the way to go - pass the guarding responsibility to the MethodRegistry user 🤔

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricSet.java
```diff
@@ -43,26 +44,73 @@
  * </pre>
  */
 public class MetricSet {
-    private final Map<String, String> labels;
-    private final ConcurrentMap<String, DoubleSupplier> samples = new ConcurrentHashMap<>();
+    private final Labels.Immutable labels;
+    private final ConcurrentMap<String, DoubleSupplier> gauges;
+    // low load factor as hash collisions are quite costly when tracking breakdown metrics
```
eyalkoren: Can this be improved by using a cache of immutable label objects for all combinations of the span types and subtypes we use, where equation is quickly done through reference comparison?

apm-agent-core/src/main/java/co/elastic/apm/agent/util/KeyListConcurrentHashMap.java
```diff
@@ -0,0 +1,101 @@
...
+            keyList.remove(key);
+        }
+        return remove;
+    }
+
+    /**
+     * Returns a mutable {@link List}, roughly equal to the {@link #keySet()}.
+     * <p>
+     * Note that in concurrent scenarios, the key list may be a subset of the values of the respective {@link #keySet()}.
+     * Entries added via the {@code compute*} family of methods are not reflected in the list.
```
eyalkoren: What are the `compute` methods?
I think this list in general may not reflect the underlying `keySet()` as it is not thread-safe with regard to the underlying map, right?

apm-agent-core/src/main/java/co/elastic/apm/agent/util/KeyListConcurrentHashMap.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A subclass of {@link ConcurrentHashMap} which maintains a {@link List} of all the keys in the map.
+ * It can be used to iterate over the map's keys without allocating an {@link java.util.Iterator}
+ */
+public class KeyListConcurrentHashMap<K, V> extends ConcurrentHashMap<K, V> {
+    private final List<K> keyList = Collections.synchronizedList(new ArrayList<K>());
```
eyalkoren: Make sure the map can only grow in size- throw unsupported operation exceptions from `remove` and `clear`

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/MetricRegistrySerializer.java
```diff
@@ -91,23 +93,121 @@ private static void serializeSamples(Map<String, DoubleSupplier> samples, JsonWr
...
-    private static void serializeSample(String key, double value, JsonWriter jw) {
-        jw.writeString(key);
-        jw.writeByte(JsonWriter.SEMI);
-        jw.writeByte(JsonWriter.OBJECT_START);
-        DslJsonSerializer.writeFieldName("value", jw);
+    private static void serializeTimer(String key, Timer timer, JsonWriter jw) {
+        serializeValue(key, ".count", timer.getCount(), jw);
+        jw.writeByte(JsonWriter.COMMA);
+        serializeValue(key, ".sum", timer.getTotalTimeMs(), jw);
+        timer.resetState();
```
eyalkoren: All these resets seem awkward here, much better fit in `metricSet.onAfterReport()`. Is this done here in order to prevent additional iteration?

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/MetricRegistrySerializer.java
```diff
@@ -91,23 +93,121 @@ private static void serializeSamples(Map<String, DoubleSupplier> samples, JsonWr
...
+                if (value.hasContent()) {
+                    jw.writeByte(JsonWriter.COMMA);
+                    serializeTimer(kv.getKey(), value, jw);
+                }
+            }
+        }
+        return hasSamples;
+    }
+
+    private static void serializeCounters(Map<String, AtomicLong> timers, boolean hasSamples, JsonWriter jw) {
```
eyalkoren:
```suggestion
    private static void serializeCounters(Map<String, AtomicLong> counters, boolean hasSamples, JsonWriter jw) {
```

apm-agent-core/src/main/java/co/elastic/apm/agent/util/KeyListConcurrentHashMap.java
felixbarny: For example `java.util.concurrent.ConcurrentHashMap#computeIfAbsent`

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
```diff
@@ -221,9 +239,63 @@ public void incrementReferences() {
...
+        labels.transactionName(name).transactionType(type);
+        final MetricRegistry metricRegistry = tracer.getMetricRegistry();
+        long criticalValueAtEnter = metricRegistry.writerCriticalSectionEnter();
+        try {
+            metricRegistry.updateTimer("transaction.duration", labels, getDuration());
+            if (collectBreakdownMetrics) {
+                metricRegistry.incrementCounter("transaction.breakdown.count", labels);
+                List<Labels> keyList = spanTimings.keyList();
+                for (int i = 0; i < keyList.size(); i++) {
+                    Labels spanType = keyList.get(i);
```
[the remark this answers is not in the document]
felixbarny: It can't - at least when setting `finished= true` before tracking the metrics.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
felixbarny: will set `finished = true` earlier

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/Labels.java
```diff
@@ -0,0 +1,380 @@
...
+        private final String spanSubType;
+
+        public Immutable(Labels labels) {
+            super(new ArrayList<>(labels.getKeys()), copy(labels.getValues()));
+            final CharSequence transactionName = labels.getTransactionName();
+            this.transactionName = transactionName != null ? transactionName.toString() : null;
+            this.transactionType = labels.getTransactionType();
+            this.spanType = labels.getSpanType();
+            this.spanSubType = labels.getSpanSubType();
+            this.hash = labels.hashCode();
```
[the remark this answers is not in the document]
felixbarny: could you explain?

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/Labels.java
```diff
@@ -0,0 +1,380 @@
...
+        private final String transactionName;
+        @Nullable
+        private final String transactionType;
+        @Nullable
+        private final String spanType;
+        @Nullable
+        private final String spanSubType;
+
+        public Immutable(Labels labels) {
+            super(new ArrayList<>(labels.getKeys()), copy(labels.getValues()));
```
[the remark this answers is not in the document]
felixbarny: That's why I'm creating a new array list based on `labels.getKeys()`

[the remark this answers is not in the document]
eyalkoren: You are! Completely ignored that, maybe because of constructor colour 🙂

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/Labels.java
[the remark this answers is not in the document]
eyalkoren: You are constructing an Immutable probably from a Mutable object, which may change during construction, for example (if used concurrently). Maybe in the future we will add some manipulation to the `labels` fields before assigning them in the Immutable, so this will also make it invalid.
Even if you can't see a scenario where this may currently actually break, it seems more correct to calculate an Object's hash code based on its own fields, and not another Object's fields.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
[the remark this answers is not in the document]
eyalkoren: Yes, moving `finished=true` assignment to before `doEnd` should fix that 👍

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricSet.java
felixbarny: Labels also contain the transaction name. If the transaction name is chosen poorly, it may have unlimited distinct values (for example if it contains an URL path parameter)

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/Labels.java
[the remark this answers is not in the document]
felixbarny: Copying is not thread-safe in either case - if there are labels added concurrently to creating an immutable copy there can be inconsistencies. Bottom line is that `Labels.Mutable` is not thread-safe while `Labels.Immutable` is.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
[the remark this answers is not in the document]
felixbarny: hm, there's still a possibility of a race condition then:

| Thread | action| finished| 
|--------|-|-------|
| B| calls `transaction.incrementTimer`| false |
| A | set finished = true | true |
| B | has already passed the finished check and updates the timer map | true|
| A | at the time the timer is read, only the `totalTime` field has been updated by thread B. `count` is still `0`. | true |

->  `incrementTimer` has to be executed within a critical section and `finished` can only be set to `true` once there is no one in the critical section.

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
felixbarny: we agreed to do that in a separate PR

apm-agent-core/src/main/java/co/elastic/apm/agent/util/KeyListConcurrentHashMap.java
felixbarny: Synchronizing on add and remove operations should also make it thread-safe, right? That would make the map more applicable for other use-cases where modifying the list does not happen under contention.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/MetricRegistrySerializer.java
felixbarny: Yes, that the idea was to prevent another iteration. But it's probably better to trade off in favor of readability here.

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/Labels.java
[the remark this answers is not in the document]
eyalkoren: > Copying is not thread-safe in either case - if there are labels added concurrently to creating an immutable copy there can be inconsistencies

This is not about maintaining consistency between the original and copy- you can't guarantee consistency between Mutable and Immutable, regardless of concurrency (the mutable may change well after copying). 

I demonstrated a case where the `hashCode` of the Object would not reflect it's fields- if after you set `this.spanType = labels.getSpanType()`, then other thread changes `labels#spanType`, then you set `this.hash = labels.hashCode()` - this is what you'd get.

Bottom line- more correct to store an Immutable's final `hashCode` based on its own final fields.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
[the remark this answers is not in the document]
eyalkoren: There are two issues here- this conversation is about the thread safety of the usage of `spanTimings`, which should be achieved by moving the `finished` flag setting (so that `keyList` won't change while iterating over it).

The other is thread safety of the Timers, as I referred separately in https://github.com/elastic/apm-agent-java/pull/588/files#r294697740, https://github.com/elastic/apm-agent-java/pull/588/files#r294701992 and https://github.com/elastic/apm-agent-java/pull/588/files#r294707481. You can achieve Timer atomicity with critical sections, but I think it is better to make it atomic inherently, so that it can't be written to or read from concurrently (then in race conditions you read full state - either set or not-yet-set, which is OK). In addition, you would still need to make sure it is reset properly either way.

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricSet.java
eyalkoren: Can the structure suggested in https://github.com/elastic/apm-agent-java/pull/588/files#r294691938 solve this?

apm-agent-core/src/main/java/co/elastic/apm/agent/util/KeyListConcurrentHashMap.java
felixbarny: see https://github.com/elastic/apm-agent-java/pull/588#discussion_r296233760

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
felixbarny: should be fixed now

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
```diff
@@ -221,9 +239,63 @@ public void incrementReferences() {
...
+                    final Timer timer = spanTimings.get(spanType);
+                    if (timer.getCount() > 0) {
+                        labels.spanType(spanType.getSpanType()).spanSubType(spanType.getSpanSubType());
+                        metricRegistry.updateTimer("span.self_time", labels, timer.getTotalTimeUs(), timer.getCount());
```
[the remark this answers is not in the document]
felixbarny: let's reconsider this for a refinement PR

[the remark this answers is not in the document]
felixbarny: should now be fixed

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
```diff
@@ -221,9 +239,63 @@ public void incrementReferences() {
...
+                        timer.resetState();
```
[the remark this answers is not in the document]
felixbarny: should now be fixed

Issue #512, opened 2019-03-05
JAX-RS class hierarchy scanning is incomplete and creates transaction name explosion when using EJB


mdindoffer:
**Summary**
The JAX-RS class hierarchy scanning introduced in PR #493   does not work correctly when using EJB.
EJB uses a `$$$view<number>` naming scheme when creating proxies, which do not get filtered out and leads to explosion of transaction names:
 
![screenshot from 2019-03-04 15-18-38_crop](https://user-images.githubusercontent.com/12681399/53796890-5990f700-3f35-11e9-85c9-60a1c90980cc.png)
![screenshot from 2019-03-04 15-18-10_crop_blurred](https://user-images.githubusercontent.com/12681399/53800135-52b9b280-3f3c-11e9-88db-636beb1902cc.png)

**Additional Info**
I've tried to fix it by simply adding a condition to the `JaxRsTransactionNameInstrumentation` matcher:
```java
.and(not(ElementMatchers.<TypeDescription>nameContains("$view")))
```
That gets rid of the `$$$view` explosion, but all the requests still end up in the HttpServlet30Dispatcher transaction:
![screenshot_2019-03-05 httpservlet30dispatcher_dopost_blurred](https://user-images.githubusercontent.com/12681399/53800839-f22b7500-3f3d-11e9-9c4a-506bdf0b4fa7.png)
This shows that the current solution from PR elastic/apm-agent-java#493 is not sufficient. The problem probably comes up when dealing with generic hierarchies.

**A Reproducible scenario**
We use a generic structure that might look complicated and intimidating at first glance, but that's just because of the number of generic types used. We have an abstract generic endpoint, a concrete endpoint specialization, a generic endpoint/client interface and a concrete endpoint/client interface.
To illustrate (Wildfly 9.0.2 btw):
1. Concrete JaxRS Endpoint:
```java
import javax.ejb.Singleton;

@Singleton
public class ConcreteEndpoint
        extends AbstractRequestEndpoint<ClassT, ClassU, ClassE, ClassV>
        implements ConcreteClient
```
2. Concrete endpoint/client JaxRS web service interface definition:
```java
@Path( "/" )
public interface ConcreteClient extends BaseRequestClient<ClassT, ClassU>
```
3. Abstract JaxRS endpoint:
```java
public abstract class AbstractRequestEndpoint<T extends Class1<?>, U extends Class2<?>, E extends Class3, V extends Class4<E, U>>
        implements BaseRequestClient<T, U>
[...]

PR #600, opened 2019-04-18
Exclude $view proxies from JAX-RS instrumentation

mdindoffer:
Excluding `$view` proxy classes prevents explosion of transaction names in certain scenarios, see #512.

### DIFF
--- a/apm-agent-plugins/apm-jaxrs-plugin/src/test/java/co/elastic/apm/agent/jaxrs/JaxRsTransactionNameInstrumentationTest.java
+++ b/apm-agent-plugins/apm-jaxrs-plugin/src/test/java/co/elastic/apm/agent/jaxrs/JaxRsTransactionNameInstrumentationTest.java
@@ -104,12 +104,31 @@ public class JaxRsTransactionNameInstrumentationTest extends JerseyTest {
         assertThat(actualTransactions.get(2).getName().toString()).isEqualTo("ResourceWithPathOnAbstract#testMethod");
     }
 
+    @Test
+    public void testProxyClassInstrumentationExclusion() {
+        when(config.getConfig(JaxRsConfiguration.class).isEnableJaxrsAnnotationInheritance()).thenReturn(true);
+        ElasticApmAgent.initInstrumentation(tracer, ByteBuddyAgent.install());
+
+        doRequest("testViewProxy");
+        doRequest("testProxyProxy");
+
+        List<Transaction> actualTransactions = reporter.getTransactions();
+        assertThat(actualTransactions).hasSize(2);
+        assertThat(actualTransactions.get(0).getName().toString()).isEqualTo("unnamed");
+        assertThat(actualTransactions.get(1).getName().toString()).isEqualTo("unnamed");
+    }
+
 
     /**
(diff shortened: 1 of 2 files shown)


apm-agent-plugins/apm-jaxrs-plugin/src/test/java/co/elastic/apm/agent/jaxrs/JaxRsTransactionNameInstrumentationTest.java
```diff
@@ -106,6 +106,20 @@ public void testJaxRsTransactionNameWithJaxrsAnnotationInheritance() {
...
+    @Test
+    public void testProxyClassInstrumentationExclusion() {
+        when(config.getConfig(JaxRsConfiguration.class).isEnableJaxrsAnnotationInheritance()).thenReturn(true);
+        ElasticApmAgent.initInstrumentation(tracer, ByteBuddyAgent.install());
+
+        doRequest("testViewProxy");
+        doRequest("testProxyProxy");
+
+        List<Transaction> actualTransactions = reporter.getTransactions();
+        assertThat(actualTransactions).hasSize(2);
```
eyalkoren: Why do we capture them at all? Through a base servlet or servlet filter?

felixbarny: I guess these are just EJB proxies for the actual resource class. But maybe we should rather exclude those proxies globally here:

https://github.com/elastic/apm-agent-java/blob/4ab2b6f358eb040df407d830831d9d03a07eef67/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java#L229-L246

by adding `WildcardMatcher.caseSensitiveMatcher("*$Proxy")` and `WildcardMatcher.caseSensitiveMatcher("*$view")`

eyalkoren: I meant- why are they captured in the test?

felixbarny: In a real-world scenario, the transaction would be created by the servlet instrumentation. The JAX-RS instrumentation does not create transactions, it only sets the name. In this test, the transaction is manually created by the private `doRequest` method.

WDYT about moving it to the default `classes_excluded_from_instrumentation`?

eyalkoren: Ahh, missed the manual transaction creation.

I am not sure about moving to `classes_excluded_from_instrumentation`. Here it causes a transaction name explosion. Maybe in some scenarios it will capture things that otherwise won't be captured? Like some implementation that creates `$Proxy` implementations to interface stubs without a resource class in between? If you don't think this is a valid concern, go ahead and move it there.

Issue #602, opened 2019-04-23
Classloader deadlock with FileSystems.getDefault


mhoglan:
**Describe the bug**
Deadlock occurs periodically due to class initialization where FileSystems.getDefault() is called from inside a lock on the runtime but that is blocked due to another thread which is also requesting the lock on the runtime while it holds the FileSystems lock.  Looks like it occurs during to the APM Server Healthcheck.

There is an outstanding JDK bug around this:
https://bugs.openjdk.java.net/browse/JDK-8194653

It is fixed in later versions of the JDK and is backported to JDK8 in 8u231 which is ~Oct 2019;  

Looking at related projects that reference the bug, they also fixed it by preloading by calling `FileSystems.getDefault()` before the other threads.

Tomcat
https://bz.apache.org/bugzilla/show_bug.cgi?id=63251

Similar issue was reported here for the elastic apm-java-agent:
https://discuss.elastic.co/t/apm-java-agent-creating-deadlock-in-web-logic/173744

Tried the 1.5.0 and 1.6.0 releases of the agent and same behavior.

**To Reproduce**
Periodically has been occurring in build pipelines.  

**Expected behavior**
To not deadlock :)

**Debug logs**
```
/ # jstack  7
Picked up JAVA_TOOL_OPTIONS: -XX:+UseContainerSupport -XX:MaxRAMPercentage=75.0 -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/heapdump -Djava.security.egd=file:/dev/./urandom
2019-04-13 19:31:23
Full thread dump OpenJDK 64-Bit Server VM (25.201-b08 mixed mode):

"Attach Listener" #10 daemon prio=9 os_prio=0 tid=0x00005581950d9800 nid=0x3c runnable [0x0000000000000000]
   java.lang.Thread.State: RUNNABLE

"apm-server-healthcheck" #9 daemon prio=5 os_prio=0 tid=0x000055819503e800 nid=0x11 in Object.wait() [0x00007f971186d000]
   java.lang.Thread.State: RUNNABLE
	at java.nio.file.FileSystems.getDefault(FileSystems.java:176)
	at java.nio.file.Paths.get(Paths.java:138)
	at sun.misc.Launcher$ExtClassLoader.findLibrary(Launcher.java:235)
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1830)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
[...]

PR #603, opened 2019-04-23
Fix classloader deadlock with FileSystems.getDefault() by preloading

mhoglan:
Fixes: #602


eyalkoren:
Jenkins run the tests

Issue #584, opened 2019-04-15
Improper life cycle management in async servlet API


felixbarny:
The `/async-start-servlet` test case logs warnings. The transaction is activated 3 times but deactivated 4 times.

```
[2019-04-15T11:43:21.683Z] [jetty] [qtp1983025922-26] DEBUG co.elastic.apm.agent.impl.ElasticApmTracer - Deactivating 00
-00000000000000000000000000000000-9c3a3bf78adb2a20-00 on thread 26
[2019-04-15T11:43:21.683Z] [jetty] [qtp1983025922-26] WARN co.elastic.apm.agent.impl.ElasticApmTracer - Deactivating a span ('' 00-00000000000000000000000000000000-9c3a3bf78adb2a20-00) which is not the currently active span (null). This can happen when not properly deactivating a previous span.
[2019-04-15T11:43:21.683Z] [jetty] [qtp1983025922-26] DEBUG co.elastic.apm.agent.impl.ElasticApmTracer - Deactivating 00-00000000000000000000000000000000-9c3a3bf78adb2a20-00 on thread 26
[2019-04-15T11:43:21.683Z] [jetty] [qtp1983025922-26] WARN co.elastic.apm.agent.impl.ElasticApmTracer - Deactivating a span ('' 00-00000000000000000000000000000000-9c3a3bf78adb2a20-00) which is not the currently active span (null). This can happen when not properly deactivating a previous span.
```

We should enable assertions in the servlet containers so this fails integration tests

eyalkoren:
Actually the lifecycle is mostly managed properly - activated and deactivated properly on the initiating thread and ended exactly once in all scenarios. 
On the worker thread, things are a bit more complicated- some servlet containers wrap the already-wrapped-Runnable (the one we wrap during `AsyncContext#start`) and submit to an `ExecutorService`. If we instrument this `ExecutorService`, then the servlet-container-Runnable-wrapper escapes our check and gets wrapped again. In such scenarios, the Transaction gets activated and deactivated twice **properly**. The problem is with https://github.com/elastic/apm-agent-java/blob/5dce1606f5a328d113216659716ff5cce1c63485/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java#L482 where the first activation clears the stack and the second gets it empty.

We should probably get rid of this stack clearance. Instead, we can force a certain order of deactivate/end only for Transactions, and make the stack check-and-clear on `Transaction#end`.

[...]

PR #605, opened 2019-04-23
Fixing some async mechanisms lifecycle issues

eyalkoren:
Fixes elastic/apm-agent-java#584

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
@@ -495,12 +495,6 @@ public class ElasticApmTracer {
         }
         final Deque<TraceContextHolder<?>> stack = activeStack.get();
         assertIsActive(holder, stack.poll());
-        if (holder == stack.peekLast()) {
-            // if this is the bottom of the stack
-            // clear to avoid potential leaks in case some spans didn't deactivate properly
-            // makes all leaked spans eligible for GC
-            stack.clear();
-        }
     }
 
     private void assertIsActive(Object span, @Nullable Object currentlyActive) {
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
@@ -95,6 +95,14 @@ public class ElasticApmTracer {
             return new ArrayDeque<TraceContextHolder<?>>();
         }
     };
+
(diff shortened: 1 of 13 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -193,6 +201,18 @@ public void onChange(ConfigurationOption<?> configurationOption, Double oldValue
         return startTransaction(childContextCreator, parent, sampler, -1, initiatingClassLoader);
     }
 
+    public void avoidWrappingOnThread() {
```
felixbarny: I think we should make it so that it also works when things are wrapped multiple times. Avoiding the wrapping as you did is ok but should "just" be an optimization. It should not be required to call these methods.

eyalkoren: I agree, which is why I added another guard where if wrapped twice, same span won't be activated and deactivated twice using this flag: https://github.com/elastic/apm-agent-java/pull/605/files#diff-d1e413fd7a24f92f13f6094cbf2dd787R37

In addition, I am relying on your elimination of https://github.com/elastic/apm-agent-java/blob/master/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java#L478, which, if moved, things would not be optimal but work properly. This is the only thing that currently makes multiple activation/deactivation of a span not fully safe.

felixbarny: Could you remove that in your pr as well? Reason is that we probably want to release this PR with a bug fix release soon whereas the other PR will be released later in a feature release.

eyalkoren: Sure, good point.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/async/SpanInScopeRunnableWrapper.java
```diff
@@ -53,34 +50,16 @@ public SpanInScopeRunnableWrapper wrap(Runnable delegate, AbstractSpan<?> span)
     // In this case, this class acts as the boundary of user and agent code so we have to do the tedious exception handling here
     @Override
     public void run() {
+        System.out.println("Wrapper run");
```
felixbarny: leftover?

eyalkoren: 🤦‍♂️ of course! 
Debugging async scenarios is a lot of fun 😅

PR #606, opened 2019-04-23
Add support for custom context

### LINKED ISSUE #604: Add custom context
Add these API methods:

* `context.addCustomContext(String key, String value)`
* `context.addCustomContext(String key, Number value)`
* `context.addCustomContext(String key, boolean value)`

See https://github.com/elastic/apm/issues/7#issuecomment-484057924

felixbarny:
closes elastic/apm-agent-java#604


apm-agent-api/src/main/java/co/elastic/apm/api/Transaction.java
```diff
@@ -79,6 +79,48 @@
...
+     * Non-indexed means the data is not searchable or aggregatable in Elasticsearch,
+     * and you cannot build dashboards on top of the data.
+     * However, non-indexed information is useful for other reasons,
+     * like providing contextual information to help you quickly debug performance issues or errors.
+     *
+     * @param key   The custom context key.
+     * @param value The custom context value.
+     */
+    @Nonnull
+    Transaction addCustomContext(String key, String value);
```
felixbarny: An alternative would be to add a `getContext` method and add the `addCustom()` methods to the `TransactionContext`. Also kinda depends on how we want to make the other context objects accessible. Also, as of 7.x there is no `context` namespace in ES documents anymore. Context only exists in the intake API now. Custom is stored under `transaction.custom`. So maybe we should just name the method `addCustom`?

eyalkoren: I think it is good as it is. The term `context` is appropriate here regardless of the schema we use.
With `getContext` do you mean we add an API context struct for data storage? Not sure this is necessary.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
```diff
@@ -715,6 +715,11 @@ private void serializeContext(final TransactionContext context, TraceContext tra
         }
         serializeRequest(context.getRequest());
         serializeResponse(context.getResponse());
+        if (context.hasCustom()) {
+            writeFieldName("custom");
+            serializeLabels(context.getCustomIterator(), replaceBuilder, jw);
```
eyalkoren: Maybe change the method name to `serializeDictionary` or something of that sort instead?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/TransactionContext.java
```diff
@@ -55,17 +59,34 @@ public void copyFrom(TransactionContext other) {
...
+    public Object getCustom(String key) {
+        return custom.get(key);
     }
 
     public Response getResponse() {
         return response;
     }
 
+    public void addCustom(String key, String value) {
+        custom.put(key, value);
```
eyalkoren: No size limitation at all on those?

apm-agent-api/src/main/java/co/elastic/apm/api/TransactionImpl.java
```diff
@@ -77,6 +77,24 @@ public Transaction addLabel(String key, boolean value) {
         return this;
     }
 
+    @Nonnull
+    @Override
+    public Transaction addCustomContext(String key, String value) {
+        return this;
```
eyalkoren: What about your nice comments referring to the corresponding instrumentation? I like those 😄

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/TransactionContext.java
felixbarny: the size limitation is enforced centrally on serialization

eyalkoren: Ahh. Don't we want to allow a bigger size for these? Probably a common use for non-indexed text is big texts that users still want to keep, like HTTP bodies.

felixbarny: Oh, good point. We'll want to use the 10k char limit on those values!

felixbarny:
> I assume there was a discussion about that and this is aligned with other agents, I just don't recall- what is the benefit of custom Numbers and Boolean? Would it make a big difference if indexed (as labels)?

Do you mean why we can't simply convert numbers and booleans as strings and store them as keywords?

It's because you can't graph strings if you want the actual number to be on the y-axis. In other words, you can't use [metric aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics.html) on strings.

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/plugin/api/TransactionInstrumentation.java
```diff
@@ -101,4 +102,25 @@ private static void ensureParentId(@Advice.FieldValue(value = "span", typing = A
...
     }
+
+    public static class AddCustomContextInstrumentation extends TransactionInstrumentation {
+        public AddCustomContextInstrumentation() {
+            super(named("addCustomContext"));
+        }
+
+        @VisibleForAdvice
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        public static void addLabel(@Advice.FieldValue(value = "span", typing = Assigner.Typing.DYNAMIC) TraceContextHolder<?> context,
```
eyalkoren:
```suggestion
        public static void addCustomContext(@Advice.FieldValue(value = "span", typing = Assigner.Typing.DYNAMIC) TraceContextHolder<?> context,
```

felixbarny: oops, a classical copy/pasta 🍝

PR #608, opened 2019-04-24
Add note that number and boolean labels require APM Server 6.7+

felixbarny:


apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
```diff
@@ -61,39 +61,32 @@
...
+     *     <li>Elasticsearch field: {@code labels} (previously {@code context.tags} in stack version {@code < 7.0})</li>
+     * </ul>
      * <p>
-     * Note: the labels are indexed in Elasticsearch so that they are searchable and aggregatable.
-     * By all means,
-     * you should avoid that user specified data,
-     * like URL parameters,
-     * is used as a label key as it can lead to mapping explosions.
+     * Label values can be a string, boolean, or number.
+     * Because labels are stored in the same place in Elasticsearch, each value must have the same data type.
```
eyalkoren:
```suggestion
     * Because labels for a given key are stored in the same place in Elasticsearch, all label values of a given key must have the same data type.
```

felixbarny: I mostly copy/pasted that from here: https://github.com/elastic/apm-server/pull/2129
Should we propose the change there as well?

eyalkoren: I don't know.. What do you think? 
I suggested this change because it wasn't clear enough for me. I actually don't know what 
> stored in the same place in Elasticsearch

really means...

felixbarny: I think your changes made it a lot clearer! Would you mind proposing the same changes on the apm-server PR?

felixbarny:
related to https://github.com/elastic/apm/issues/42#issuecomment-486073492

docs/public-api.asciidoc
```diff
@@ -287,16 +287,24 @@ transaction.setType(Transaction.TYPE_REQUEST);
 [float]
 [[api-transaction-add-tag]]
 ==== `Transaction addLabel(String key, value)` added[1.5.0]
```
felixbarny:
```suggestion
==== `Transaction addLabel(String key, value)` added[1.5.0,Number and boolean labels require APM Server 6.7]
```

PR #619, opened 2019-04-30
Add a step to build to create a dependency list

eyalkoren:


eyalkoren:
@tomcallahan please see [the created dependency list](https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/job/PR-619/1/artifact/src/github.com/elastic/apm-agent-java/target/site/aggregate-third-party-report.html). I will go over it to verify we didn't miss anything and that everything is in accordance to requirements, for now I need your approval about its format and the location we store it after every build.

felixbarny:
Is there a way to exclude test dependencies or to only create the report for the `elastic-apm-agent` subproject, which is the main binary and the only binary with vendored code.

eyalkoren:
There is a way, but I think it needs to include them all (with different definitions as to which license is allowed where). One of the things I will need to look into.

Jenkinsfile
```diff
@@ -73,6 +73,36 @@ pipeline {
               onlyIfSuccessful: true
           }
         }
+        /**
```
kuisathaverat: if we want to speed up the build we would move it to the parallel block of stages

kuisathaverat: it takes 21m 22s Wow!!!

eyalkoren: Oops, it ran the tests 🤦‍♂ 
Fixing..

eyalkoren: I can't find a way to skip phases and run the install plugin directly through `mvn install:insall` based on the unstashed state from the build phase (artifacts created through `mvn package`). I am not sure there is an option to do this in a multi-module maven project at all. 
So I will just do the additional `install` phase in the build stage and create the list and publish from there.
@kuisathaverat you see any good reason why I shouldn't do so? Shouldn't add a lot to the existing stage.

eyalkoren: Testing, let's see how it looks like

eyalkoren: Added less than one minute to the Build stage. I think this is a good solution.

eyalkoren:
@kuisathaverat what do you say? Any reason not to do it this way?

PR #621, opened 2019-05-02
Adding NOTICE and LICENSE

eyalkoren:
- Adding NOTICE per artefact containing redistributed dependencies with copyright notices and relevant licenses
- Adding NOTICE and LICENSE to all artefacts
- Adding `<licenses>` section to pom.xml files (seems to be expected by the maven-license-plugin, so probably a convention)
- Fixing source jar packaging
- Uncommenting disabled SystemMetricsTest


apm-agent-attach/resources/NOTICE
```diff
@@ -1,3 +1,11 @@
+Elastic APM Java Agent Attach Tool
+Copyright 2018-2019 Elasticsearch B.V.
+
+This product includes software licensed under the Apache License 2.0 from the following sources:
+ - ByteBuddy - Copyright 2014 - 2019 Rafael Winterhalter
+ - https://github.com/kohlschutter/junixsocket - Copyright (c) 2009-2018 Dr. Christian Kohlschütter
```
felixbarny: this is a `provided`-scoped dependency so I think we can remove this line as we don't bundle/vendor it

eyalkoren:
@tomcallahan we have 4 artefacts, for each I created a relevant NOTICE reflecting only the redistributed libraries:
1. Our API jar for manual instrumentation (has no dependencies) - [NOTICE.txt](https://github.com/elastic/apm-agent-java/files/3138136/NOTICE.txt)
2. OpenTracing (standard APM API) bridge - [NOTICE.txt](https://github.com/elastic/apm-agent-java/files/3138147/NOTICE.txt)
3. A utility to dynamically attach the agent - [NOTICE.txt](https://github.com/elastic/apm-agent-java/files/3138154/NOTICE.txt)
4. Our actual agent jar - [NOTICE.txt](https://github.com/elastic/apm-agent-java/files/3138159/NOTICE.txt)

Please review the structure of those.

pom.xml
```diff
@@ -423,7 +410,7 @@
                 <configuration>
                     <verbose>false</verbose>
                     <licenseName>apache2_license</licenseName>
-                    <licenseResolver>${license.dir.uri}</licenseResolver>
+                    <licenseResolver>file://${apm-agent-parent.base.dir}/licenses</licenseResolver>
```
felixbarny: ah, didn't know about that notation nice!

apm-agent-attach/resources/NOTICE
```diff
@@ -178,15 +186,15 @@
...
+      boilerplate notice, with the fields enclosed by brackets "{}"
       replaced with your own identifying information. (Don't include
       the brackets!)  The text should be enclosed in the appropriate
       comment syntax for the file format. We also recommend that a
       file or class name and description of purpose be included on the
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2018 - 2019 Elasticsearch BV
+   Copyright {yyyy} {name of copyright owner}
```
felixbarny: 🤔

NOTICE
```diff
@@ -0,0 +1,284 @@
+Elastic APM Java Agent
+Copyright 2018-2019 Elasticsearch B.V.
+
+###############################################################################
+
+This product includes software developed by The Apache Software Foundation
```
felixbarny: I don't think we include software developed by The Apache Software Foundation. Or do we?

apm-agent-attach/resources/NOTICE
eyalkoren: Copied from their repo 🤷‍♂

eyalkoren: I think this is GitHub misinterpreting the diff - some weird merge between LICENSE and NOTICE

NOTICE
eyalkoren: I think you are right. Not sure why I left this. Probably too many edits 🤓

eyalkoren:
Last change added ~20 seconds to Smoke Tests 01 stage- I can live with that 😄

apm-agent-api/pom.xml
```diff
@@ -15,7 +22,34 @@
...
+            </resource>
+        </resources>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-source-plugin</artifactId>
+                <version>3.0.1</version>
+                <executions>
+                    <execution>
+                        <id>generate-source-jar</id>
```
felixbarny: does that fix the shaded source jar?

eyalkoren: No, this one was created properly. The addition of this plugin in the parent project maybe took care of that, not sure. However, once I removed it from there, I had to add it here and to the opentracing module

tomcallahan:
These look close.  Under Apache2, we have to present any NOTICE content that is in the NOTICE files of those (vendored) dependencies.  Most of these do not appear to have NOTICE files, but I noticed that stagemonitor does.  It may be appropriate to exclude items from that NOTICE that we haven't actually copied, but we are obligated to include this for instance:

```stagemonitor
Copyright 2014-2017 iSYS Software GmbH

This product includes software developed at
iSYS Software GmbH (http://www.isys-software.de/).
```

Can you clarify that these dependencies are vendored?

eyalkoren:
Merging as is - source jar produced properly when I run it locally, but is created on Jenkins only with the shaded sources (without our sources 😕 ). Will open a different issue for that

PR #622, opened 2019-05-02
WIP: Add support for org.asynchttpclient

thomasklinger1234:
**WORK IN PROGRESS**

This PR adds tracing support for [org.asynchtclient](https://mvnrepository.com/artifact/org.asynchttpclient/async-http-client/2.8.1). 

## TODOs

* Test cases for redirects are not pasing. Need to investigate how the client configuration is supposed to be or work around by adding some tricks into the APM instrumentation to keep the tracing headers

### DIFF
--- a/apm-agent-plugins/apm-asynchttpclient-plugin/pom.xml
+++ b/apm-agent-plugins/apm-asynchttpclient-plugin/pom.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>apm-agent-plugins</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.6.1-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>apm-asynchttpclient-plugin</artifactId>
+    <name>${project.groupId}:${project.artifactId}</name>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>8</source>
+                    <target>8</target>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+    <properties>
+        <version.asynchttpclient>2.8.1</version.asynchttpclient>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>apm-httpclient-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.asynchttpclient</groupId>
+            <artifactId>async-http-client</artifactId>
+            <version>${version.asynchttpclient}</version>
+        </dependency>
+    </dependencies>
+
+</project>
--- a/apm-agent-plugins/apm-asynchttpclient-plugin/pom.xml
+++ b/apm-agent-plugins/apm-asynchttpclient-plugin/pom.xml
@@ -39,6 +39,13 @@
             <artifactId>async-http-client</artifactId>
             <version>${version.asynchttpclient}</version>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>apm-httpclient-core</artifactId>
+            <version>${project.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>
--- a/apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
+++ b/apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
@@ -1,0 +1,105 @@
+/*-
+ * #%L
+ * Elastic APM Java agent
+ * %%
+ * Copyright (C) 2018 - 2019 Elastic and contributors
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package co.elastic.apm.agent.asynchttpclient;
+
+import co.elastic.apm.agent.bci.ElasticApmInstrumentation;
+import co.elastic.apm.agent.bci.VisibleForAdvice;
+import co.elastic.apm.agent.http.client.HttpClientHelper;
+import co.elastic.apm.agent.impl.transaction.Span;
+import co.elastic.apm.agent.impl.transaction.TraceContext;
+import co.elastic.apm.agent.impl.transaction.TraceContextHolder;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.implementation.bytecode.assign.Assigner;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.asynchttpclient.Request;
+import org.asynchttpclient.Response;
+
+import javax.annotation.Nullable;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.concurrent.ExecutionException;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.returns;
+
+public class AsyncHttpClientInstrumentation extends ElasticApmInstrumentation {
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return named("org.asynchttpclient.AsyncHttpClient");
+    }
+
+    @Override
+    public ElementMatcher<? super MethodDescription>  getMethodMatcher() {
+        return named("execute")
+            .and(returns(named("org.asynchttpclient.ListenableFuture")));
+    }
+
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Arrays.asList("http-client", "org.asynchttpclient");
+    }
+
+    @VisibleForAdvice
+    public static class AsyncHttpClientExecuteAdvice {
+
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        private static void onBeforeExecute(
+            @Advice.Argument(typing = Assigner.Typing.DYNAMIC, readOnly = true, value = 1) Request request,
+            @Advice.Local("span") Span span) {
+
+            if (tracer == null || tracer.getActive() == null) {
+                return;
+            }
+
+            final TraceContextHolder<?> parent = tracer.getActive();
+
+            span = HttpClientHelper.startHttpClientSpan(parent, request.getMethod(), request.getUrl(), request.getUri().getHost());
+
+            if(span != null) {
+                span.activate();
+                request.getHeaders().add(TraceContext.TRACE_PARENT_HEADER, span.getTraceContext().getOutgoingTraceParentHeader().toString());
+            }
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
+        public static void onAfterExecute(@Advice.Return @Nullable org.asynchttpclient.ListenableFuture<Response> response,
+                                          @Advice.Local("span") @Nullable Span span,
+                                          @Advice.Thrown @Nullable Throwable t) {
+            if (span != null) {
+                try {
+                    if (response != null) {
+                        int statusCode = response.get().getStatusCode();
+                        span.getContext().getHttp().withStatusCode(statusCode);
+                    }
+                    span.captureException(t);
+                } catch (InterruptedException e) {
+                    span.captureException(e);
+                } catch (ExecutionException e) {
+                    span.captureException(e);
+                } finally {
+                    span.deactivate().end();
+                }
+            }
+        }
+    }
+}
--- a/apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
+++ b/apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
@@ -7,9 +7,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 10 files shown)


apm-agent-plugins/apm-asynchttpclient-plugin/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+    <parent>
+        <artifactId>apm-agent-plugins</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.6.1-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>apm-asynchttpclient-plugin</artifactId>
+    <name>${project.groupId}:${project.artifactId}</name>
+    <build>
```
felixbarny: I think this can be removed. It's important that the agent is compiled with target 7 (inherited from parent).

apm-agent-plugins/apm-asynchttpclient-plugin/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>apm-httpclient-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.asynchttpclient</groupId>
+            <artifactId>async-http-client</artifactId>
+            <version>${version.asynchttpclient}</version>
+        </dependency>
```
felixbarny: set `<scope>provided</scope>`

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+
+import static net.bytebuddy.matcher.ElementMatchers.*;
+
+public class AsyncHttpClientInstrumentation extends ElasticApmInstrumentation {
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return hasSuperType(named("org.asynchttpclient.AsyncHttpClient"));
```
felixbarny: Maybe it makes sense to target the implementation `org.asynchttpclient.DefaultAsyncHttpClient` here. That will be much faster on matching (match the exact name of the class vs traversing the inheritance hierarchy of all loaded classes). You are also relying on that fact that all methods are funneled through `executeRequest(Request, AsyncHandler)` which seems implementation specific.

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+            if(span != null) {
+                span.activate();
+                request.getHeaders().add(TraceContext.TRACE_PARENT_HEADER, span.getTraceContext().getOutgoingTraceParentHeader().toString());
+            }
+            if(!request.getHeaders().contains(TraceContext.TRACE_PARENT_HEADER) && parent != null) {
+                request.getHeaders().add(TraceContext.TRACE_PARENT_HEADER, parent.getTraceContext().getOutgoingTraceParentHeader().toString());
+            }
+
+            // TODO: Optimize the new call away
+            asyncHandler = new AsyncHandlerWrapperCreator().wrap(asyncHandler, span, request);
```
felixbarny: move up to `if (span != null)`

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            final TraceContextHolder<?> parent = tracer.getActive();
+            span = HttpClientHelper.startHttpClientSpan(parent, request.getMethod(), request.getUri().toFullUrl(), request.getUri().getHost());
+
+            if(span != null) {
+                span.activate();
+                request.getHeaders().add(TraceContext.TRACE_PARENT_HEADER, span.getTraceContext().getOutgoingTraceParentHeader().toString());
+            }
+            if(!request.getHeaders().contains(TraceContext.TRACE_PARENT_HEADER) && parent != null) {
```
felixbarny: `parent` can't be null (as checked on line 73)

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        public static void onBeforeExecute(@Advice.Argument(value = 0, readOnly = true) Request request,
+                                           @Advice.Argument(value = 1, readOnly = false) AsyncHandler asyncHandler,
+                                           @Advice.Local("span") Span span) {
+            if(tracer == null || tracer.getActive() == null) {
+                return;
+            }
+
+            final TraceContextHolder<?> parent = tracer.getActive();
+            span = HttpClientHelper.startHttpClientSpan(parent, request.getMethod(), request.getUri().toFullUrl(), request.getUri().getHost());
```
felixbarny: I'd rather use `toUrl()` than `toFullUrl()` as it always uses a cached string.

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+        public static void onAfterExecute(@Advice.Local("span") Span span,
+                                          @Advice.Thrown Throwable t) {
+            if(span != null) {
+                span.deactivate();
+            }
+        }
+    }
+
+    public static class AsyncHandlerWrapperCreator {
+        public AsyncHandler<Object> wrap(final AsyncHandler asyncHandler, Span span, Request originalRequest) {
```
felixbarny:
```suggestion
        public <T> AsyncHandler<T> wrap(final AsyncHandler<T> asyncHandler, Span span, Request originalRequest) {
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+            return new AsyncHandlerWrapper(asyncHandler, span, originalRequest);
```
felixbarny:
```suggestion
            return new AsyncHandlerWrapper<T>(asyncHandler, span, originalRequest);
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        private static class AsyncHandlerWrapper implements AsyncHandler {
+            private Span span;
+            private AsyncHandler delegate;
```
felixbarny:
```suggestion
            private AsyncHandler<T> delegate;
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+            private Request originalRequest;
```
felixbarny: Can this be removed? Seems unused.

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+    }
+
+    public static class AsyncHandlerWrapperCreator {
+        public AsyncHandler<Object> wrap(final AsyncHandler asyncHandler, Span span, Request originalRequest) {
+            return new AsyncHandlerWrapper(asyncHandler, span, originalRequest);
+        }
+
+        private static class AsyncHandlerWrapper implements AsyncHandler {
+            private Span span;
```
felixbarny: The properties should either be final or volatile as they may be read from another thread

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            @Override
+            public void onThrowable(Throwable t) {
+                span.captureException(t).end();
+                delegate.onThrowable(t);
+            }
+
+            @Override
+            public Object onCompleted() throws Exception {
```
felixbarny:
```suggestion
            public <T> onCompleted() throws Exception {
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+    }
+
+    public static class AsyncHandlerWrapperCreator {
+        public AsyncHandler<Object> wrap(final AsyncHandler asyncHandler, Span span, Request originalRequest) {
+            return new AsyncHandlerWrapper(asyncHandler, span, originalRequest);
+        }
+
+        private static class AsyncHandlerWrapper implements AsyncHandler {
```
felixbarny:
```suggestion
        private static class AsyncHandlerWrapper<T> implements AsyncHandler<T> {
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientRedirectInstrumentation.java
```diff
@@ -0,0 +1,71 @@
...
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.asynchttpclient.Request;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public class AsyncHttpClientRedirectInstrumentation extends ElasticApmInstrumentation {
```
felixbarny: Seems like this is not needed, the redirect tests work without it.

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+
+            final TraceContextHolder<?> parent = tracer.getActive();
+            span = HttpClientHelper.startHttpClientSpan(parent, request.getMethod(), request.getUri().toFullUrl(), request.getUri().getHost());
+
+            if(span != null) {
+                span.activate();
+                request.getHeaders().add(TraceContext.TRACE_PARENT_HEADER, span.getTraceContext().getOutgoingTraceParentHeader().toString());
+            }
+            if(!request.getHeaders().contains(TraceContext.TRACE_PARENT_HEADER) && parent != null) {
+                request.getHeaders().add(TraceContext.TRACE_PARENT_HEADER, parent.getTraceContext().getOutgoingTraceParentHeader().toString());
```
felixbarny: Seems like this is not needed, the redirect tests work without it.

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            @Override
+            public State onHeadersReceived(HttpHeaders headers) throws Exception {
+                logger.info("ON_HEADERS_RECEIVED: " + headers.toString());
+                return State.CONTINUE;
+            }
+        };
+
+        asyncHttpClient()
```
felixbarny:
```suggestion
        client
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+
+import static org.asynchttpclient.Dsl.asyncHttpClient;
+
+public class AsyncHttpClientInstrumentationTest extends AbstractHttpClientInstrumentationTest {
+
+    private AsyncHttpClient client;
+
+    @BeforeEach
```
felixbarny:
```suggestion
    @Before
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
```diff
@@ -0,0 +1,108 @@
...
+    void setUp() {
```
felixbarny:
```suggestion
    public void setUp() {
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
```diff
@@ -0,0 +1,108 @@
...
+        AsyncHttpClientConfig config = Dsl.config()
+            .setFollowRedirect(true)
+            .build();
+        client = asyncHttpClient(config);
+    }
+
+    @AfterEach
```
felixbarny:
```suggestion
    @After
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
```diff
@@ -0,0 +1,108 @@
...
+    void tearDown() {
```
felixbarny:
```suggestion
    public void tearDown() throws Exception {
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
```diff
@@ -0,0 +1,108 @@
...
+        try {
```
felixbarny: simply `client.close();` should be good enough

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
```diff
@@ -0,0 +1,108 @@
...
+                return State.CONTINUE;
+            }
+        };
+
+        asyncHttpClient()
+            .executeRequest(request).get();
+    }
+
+    @Test
+    void testGetAdviceClassReturnsAsyncInstrumentationClass() {
```
felixbarny: Not sure this test adds a lot. If it would return a different advice class the other tests wouldn't work as well.

docs/supported-technologies.asciidoc
```diff
@@ -197,6 +197,11 @@ The spans are named after the schema `<method> <host>`, for example `GET elastic
  inherently support context propagation as they are using `HttpUrlConnection` underneath.
 |1.4.0
 
+|Asynchronous Http Client
+|
```
felixbarny: This is the row for the library version
```suggestion
| 2.8
```

docs/supported-technologies.asciidoc
```diff
@@ -197,6 +197,11 @@ The spans are named after the schema `<method> <host>`, for example `GET elastic
...
+|
+| 2.8.1
```
felixbarny: And this one is for the agent version where the support for the library has been released
```suggestion
| 1.7.0
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+            @Override
+            public State onStatusReceived(HttpResponseStatus responseStatus) throws Exception {
+                span.getContext().getHttp().withStatusCode(responseStatus.getStatusCode());
+                return delegate.onStatusReceived(responseStatus);
+            }
+
+            @Override
+            public State onHeadersReceived(HttpHeaders headers) throws Exception {
+                if(!headers.contains(TraceContext.TRACE_PARENT_HEADER)) {
+                    headers.add(TraceContext.TRACE_PARENT_HEADER, "");
```
felixbarny: I think this is a callback when response headers are received and thus can be removed.

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
```diff
@@ -0,0 +1,108 @@
...
+ * limitations under the License.
+ * #L%
+ */
+package co.elastic.apm.agent.asynchttpclient;
+
+import co.elastic.apm.agent.httpclient.AbstractHttpClientInstrumentationTest;
+import io.netty.handler.codec.http.HttpHeaders;
+import org.asynchttpclient.*;
+import org.junit.Assert;
+import org.junit.jupiter.api.AfterEach;
```
felixbarny: The superclass is a JUnit 4 test, so JUnit Jupiter (5) annotations don't work here.

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
felixbarny: This will cause classloader/class visibility issues. The agent code lives within the system/bootstrap class loader, while the HTTP client might live in a Webapp class loader. For example when deploying a war file to tomcat which has the asynchttpclient in the libs folder. In that situation, the agent does not have access to the asynchttpclient classes, as they are loaded in a child classloader. Use `HelperClassManager.ForAnyClassLoader.of` to work around that problem. Take a look at how it's done in `ApacheHttpAsyncClientInstrumentation`.

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
felixbarny: The default methods are not overridden which would break the delegation to them. Also, new default methods might be added in future versions of this library. I think it's more feasible to not wrap the `AsyncHandler` but rather to instrument the relevant methods of types which implement `AsyncHanlder`. That also has the advantage that we don't have to use the `HelperClassManager`.

The downside is that we have to traverse the inheritance hierarchy of all classes to determine whether a particular class implements `AsyncHandler`. Therefore, it makes sense to add a classloader matcher like this to skip classloaders which can't load asynchttpclient at all:

```java
@Override
public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
    return not(isBootstrapClassLoader())
        .and(classLoaderCanLoadClass("org.asynchttpclient.AsyncHandler"));
}
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
felixbarny: Probably obsolete with https://github.com/elastic/apm-agent-java/pull/622#pullrequestreview-233394323

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
felixbarny: Hm, but the question then is how to access the span from within the listener 🤔

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+            // TODO: Optimize the new call away
+            asyncHandler = new AsyncHandlerWrapperCreator().wrap(asyncHandler, span, request);
+        }
+
+        @Nullable
+        @VisibleForAdvice
+        @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
+        public static void onAfterExecute(@Advice.Local("span") Span span,
+                                          @Advice.Thrown Throwable t) {
+            if(span != null) {
```
felixbarny: end the span and record the exception if `t != null`

felixbarny:
Superseded by https://github.com/elastic/apm-agent-java/pull/625. Let's continue the discussion there 🙂

PR #625, opened 2019-05-03
Add support for AsyncHttpClient

felixbarny:
based on https://github.com/elastic/apm-agent-java/pull/622


docs/supported-technologies.asciidoc
```diff
@@ -197,6 +197,11 @@ The spans are named after the schema `<method> <host>`, for example `GET elastic
  inherently support context propagation as they are using `HttpUrlConnection` underneath.
 |1.4.0
 
+|AsyncHttpClient
+| 2.x
```
felixbarny: I manually tested 2.0-2.8

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,180 @@
...
+
+        private final ElementMatcher<? super MethodDescription> methodMatcher;
+
+        protected AbstractAsyncHandlerInstrumentation(ElementMatcher<? super MethodDescription> methodMatcher) {
+            this.methodMatcher = methodMatcher;
+        }
+
+        @Override
+        public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+            return hasSuperType(named("org.asynchttpclient.AsyncHandler"));
```
felixbarny: Probably quite a costly matcher. Maybe we should only look within `org.asynchttpclient` and `application_packages` if defined.

eyalkoren: Costly even with the provided `getClassLoaderMatcher()` implementation?

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java
```diff
@@ -0,0 +1,180 @@
...
+
+        public AsyncHandlerOnStatusReceivedInstrumentation() {
+            super(named("onStatusReceived").and(takesArgument(0, named("org.asynchttpclient.HttpResponseStatus"))));
+        }
+
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        private static void onMethodEnter(@Advice.This AsyncHandler<?> asyncHandler, @Advice.Argument(0) HttpResponseStatus status) {
+            final Span span = handlerSpanMap.get(asyncHandler);
+            if (span != null) {
+                span.getContext().getHttp().withStatusCode(status.getStatusCode());
```
eyalkoren: Nice!

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
```diff
@@ -0,0 +1,61 @@
...
+    public void tearDown() throws IOException {
+        client.close();
+    }
+
+    @Override
+    protected void performGet(String path) throws Exception {
+        client.executeRequest(new RequestBuilder()
+            .setUrl(path)
+            .build()).get();
+    }
```
eyalkoren: What about the APIs demonstrated at https://static.javadoc.io/org.asynchttpclient/async-http-client/2.0.0/org/asynchttpclient/AsyncHttpClient.html?
Are they all go in implementation through `DefaultAsyncHttpClient#executeRequest`? Same for the synchronous one?
If so, can you add them to the tests?

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java
felixbarny: That certainly makes it less severe. Only if there are lots of classes in (child) classloaders which load `asynchttpclient` matching will be slow.

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
felixbarny: Yes, all funnel through that method. But you are right - better to add some more tests.

felixbarny:
@thomasklinger1234 have you already had the chance to try that out? I hope it's fine I went ahead and created another PR based on yours.

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java
eyalkoren: You mean like if it is on the main app server lib for example?
Well, if you are worried about that, your suggestion seems efficient and good... :)

felixbarny: I'm a bit worried that there are frameworks which use asynchttpclient under the hood and which define their own `AsyncHandlers`. We wouldn't match those then. I'd say let's continue with how it is now and optimize if we see this becomes a problem.

PR #629, opened 2019-05-08
Calculate free memory based on /proc/meminfo on Linux

### LINKED ISSUE #570: Free memory metrics don't include buff/cache
**Describe the bug**
The Java APM agent fails to capture accurate memory and CPU metrics in a containerised env.  The cpu and memory statistics for the underlying host are reported NOT the container. This manifests itself in a k8 env but is probably also an issue in a simple docker deployment.

**To Reproduce**

1. Deploy an app running in a k8 env - the opbeans example is sufficient. k8 configuration can be provided if needed.
2. Check metrics tab of agent

image

As illustrated above, memory is always almost 100%. The docs suggest the values of the underlying host are used - in this case 30GB.

```
{
          "system" : {
            "memory" : {
              "actual" : {
                "free" : 8.82143232E8
              },
              "total" : 3.1629697024E10
            }
          }
        }

```

```
dalemcdiarmid@gke-eden-test-default-pool-18a6be9a-8qkm ~ $ free
              total        used        free      shared  buff/cache   available
Mem:       30888376     5974440     1329996        3956    23583940    24656560
```

**Expected behaviour**

CPU and memory should represent those of the pod with respect to their limits. Similar to how metricbeat collects.

**Debug logs**
NA

The approach of each agent seems to be different here - the node agent uses cgroups for example.

felixbarny:
Use OperatingSystemMXBean#getFreePhysicalMemorySize on other platforms

closes elastic/apm-agent-java#570

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/SystemMetrics.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/SystemMetrics.java
@@ -134,10 +134,10 @@ public class SystemMetrics implements LifecycleListener {
         if (meminfo.canRead()) {
             metricRegistry.addUnlessNan("system.memory.actual.free", Collections.<String, String>emptyMap(), new DoubleSupplier() {
                 final List<WildcardMatcher> relevantLines = Arrays.asList(
-                    caseSensitiveMatcher("MemAvailable:*"),
-                    caseSensitiveMatcher("MemFree:*"),
-                    caseSensitiveMatcher("Buffers:*"),
-                    caseSensitiveMatcher("Cached:*"));
+                    caseSensitiveMatcher("MemAvailable:*kB"),
+                    caseSensitiveMatcher("MemFree:*kB"),
+                    caseSensitiveMatcher("Buffers:*kB"),
+                    caseSensitiveMatcher("Cached:*kB"));
 
                 @Override
                 public double get() {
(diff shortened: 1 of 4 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/SystemMetrics.java
```diff
@@ -116,12 +131,42 @@ public double get() {
...
+                    caseSensitiveMatcher("Cached:*"));
+
+                @Override
+                public double get() {
+                    Map<String, Long> memInfo = new HashMap<>();
+                    try (BufferedReader fileReader = new BufferedReader(new FileReader(meminfo))) {
+                        for (String memInfoLine = fileReader.readLine(); memInfoLine != null && !memInfoLine.isEmpty(); memInfoLine = fileReader.readLine()) {
+                            if (WildcardMatcher.isAnyMatch(relevantLines, memInfoLine)) {
+                                final String[] memInfoSplit = StringUtils.split(memInfoLine, ' ');
+                                memInfo.put(memInfoSplit[0], Long.parseLong(memInfoSplit[1]));
```
axw: `* 1024`?
maybe also check it ends with "kB", and ignore otherwise?

felixbarny: 🤦‍♂️

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/SystemMetrics.java
```diff
@@ -78,15 +87,21 @@
 
     @Nullable
     private final Method virtualProcessMemory;
+    private final File meminfo;
```
eyalkoren:
```suggestion
    private final File meminfoFile;
```
Just because we have a `memInfo` variable later...

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/SystemMetrics.java
```diff
@@ -116,12 +131,44 @@ public double get() {
...
+            metricRegistry.addUnlessNan("system.memory.actual.free", Collections.<String, String>emptyMap(), new DoubleSupplier() {
+                final List<WildcardMatcher> relevantLines = Arrays.asList(
+                    caseSensitiveMatcher("MemAvailable:*kB"),
+                    caseSensitiveMatcher("MemFree:*kB"),
+                    caseSensitiveMatcher("Buffers:*kB"),
+                    caseSensitiveMatcher("Cached:*kB"));
+
+                @Override
+                public double get() {
+                    Map<String, Long> memInfo = new HashMap<>();
```
eyalkoren: All this happens on the reporter thread. We need to reconsider this, not as part of this PR...

PR #632, opened 2019-05-10
Sanitize URLs of HTTP client spans

felixbarny:


eyalkoren:
> Since this is mostly for paranoia, I don't think the allocations are an issue, and can be optimised later if necessary.

As long as this sanitation is on by default, I don't think that this is relevant. If this is always on, the relevant question is whether there will be cases where performance will be greatly impacted by this.

> We should probably strive for consistency. The Python agent is retaining the username, but masking the password with ********. The Go agent elides user-info entirely. I'd suggest going with the Python approach - WDYT?

I am not sure. This means that those URLs will not be aggregate-able. It will not only affect users ability to visualize now, but I assume it may be less optimal for our own future usages, such as service maps. I'd say if we want to show the user name, better add a `http.user` field.

axw:
> As long as this sanitation is on by default, I don't think that this is relevant. If this is always on, the relevant question is whether there will be cases where performance will be greatly impacted by this.

Sure, I was suggesting that it's unlikely to be common as it's bad practice to include in the URL (per RFC2396, it's "NOT RECOMMENDED"). I don't think we have a good way to measure this.

> I'd say if we want to show the user name, better add a http.user field.

That SGTM. You previously suggested we mask user-info with `***:***`, which then won't aggregate with a URL with no user-info. Maybe it would be better to have another separate field, like `http.auth.type: basic`?

eyalkoren:
I honestly don't know how frequent the use of this is. If this is a rare corner case, no need to sweat over it at all...

felixbarny:
I'm not aware of any Java HTTP client which actually parses the user info part of the URL and transforms it into an `Authorization` header. So I agree that this is only for paranoia for a very slim edge case. That's why the impl checks if it contains an `@` at all before doing any parsing. Therefore, having user info on the URL does not necessarily mean that we can set `http.auth.type: basic` as the user info is ignored by most HTTP clients which require the `Authorization` header to be set explicitly.

PR #633, opened 2019-05-16
Better span names for SQL spans

### LINKED ISSUE #128: Parse SQL queries to set a better span name
See also https://github.com/elastic/apm-agent-go/issues/103#issuecomment-405158846

felixbarny:
closes elastic/apm-agent-java#128

### DIFF
--- a/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/Scanner.java
+++ b/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/Scanner.java
@@ -376,7 +376,7 @@ public class Scanner {
         UPDATE;
 
         private static final Token[] EMPTY = {};
-        private static final Token[][] keywordsByLength = {
+        private static final Token[][] KEYWORDS_BY_LENGTH = {
             {},
             {},
             {AS, OR},
@@ -389,9 +389,8 @@ public class Scanner {
         };
 
         public static Token[] getKeywordsByLength(int length) {
-            final Token[] tokens = keywordsByLength[length];
-            if (tokens != null) {
-                return tokens;
+            if (length < KEYWORDS_BY_LENGTH.length) {
+                return KEYWORDS_BY_LENGTH[length];
             }
             return EMPTY;
         }
--- a/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/Scanner.java
+++ b/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/Scanner.java
@@ -1,0 +1,404 @@
+/*-
+ * #%L
+ * Elastic APM Java agent
+ * %%
+ * Copyright (C) 2018 - 2019 Elastic and contributors
+ * %%
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * #L%
+ */
+package co.elastic.apm.agent.jdbc.signature;
+
+import javax.annotation.Nullable;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Scanner {
+
+    @Nullable
+    private String input;
+    private int start; // text start pos in bytes
+    private int end; // text end pos in bytes
+    private int pos; // read pos in bytes
+
+    public void setQuery(String sql) {
+        this.input = sql;
+        start = 0;
+        end = 0;
+        pos = 0;
+    }
+
+    public Token scanWhile(Token token) {
+        for (Token t = scan(); t != Token.EOF; t = scan()) {
+            if (t != token) {
+                return t;
+            }
+        }
+        return Token.EOF;
+    }
+
+    public boolean scanUntil(Token token) {
+        for (Token t = scan(); t != Token.EOF; t = scan()) {
+            if (t == token) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public boolean scanToken(Token token) {
+        for (Token t = scan(); t != Token.EOF; t = scan()) {
+            if (t == token) {
+                return true;
+            } else if (t != Token.COMMENT) {
+                return false;
+            }
+        }
+        return false;
+    }
+
+    public Token scan() {
+        if (!hasNext()) {
+            return Token.EOF;
+        }
+        char c = next();
+        while (Character.isSpaceChar(c) && hasNext()) {
+            c = next();
+        }
+        start = pos - 1;
+        if (c == '_' || Character.isLetter(c)) {
+            return scanKeywordOrIdentifier(c != '_');
+        } else if (Character.isDigit(c)) {
+            return scanNumericLiteral();
+        }
+        switch (c) {
+            case '\'':
+                // Standard string literal
+                return scanStringLiteral();
+            case '"':
+                // Standard double-quoted identifier.
+                //
+                // NOTE(axw) MySQL will treat " as a
+                // string literal delimiter by default,
+                // but we assume standard SQL and treat
+                // it as a identifier delimiter.
+                return scanQuotedIdentifier('"');
+            case '[':
+                // T-SQL bracket-quoted identifier
+                return scanQuotedIdentifier(']');
+            case '`':
+                // MySQL-style backtick-quoted identifier
+                return scanQuotedIdentifier('`');
+            case '(':
+                return Token.LPAREN;
+            case ')':
+                return Token.RPAREN;
+            case '-':
+                if (isNextChar('-')) {
+                    // -- comment
+                    next();
+                    return scanSimpleComment();
+                }
+                return Token.OTHER;
+            case '/':
+                if (isNextChar('*')) {
+                    // /* comment */
+                    next();
+                    return scanBracketedComment();
+                }
+                return Token.OTHER;
+            case '.':
+                return Token.PERIOD;
+            case '$':
+                if (!hasNext()) {
+                    return Token.OTHER;
+                }
+                char next = peek();
+                if (Character.isDigit(next)) {
+                    while (hasNext()) {
+                        if (!Character.isDigit(peek())) {
+                            break;
+                        } else {
+                            next();
+                        }
+                    }
+                    return Token.OTHER;
+                } else if (next == '$' || next == '_' || Character.isLetter(next)) {
+                    // PostgreSQL supports dollar-quoted string literal syntax, like $foo$...$foo$.
+                    // The tag (foo in this case) is optional, and if present follows identifier rules.
+                    while (hasNext()) {
+                        c = next();
+                        if (c == '$') {
+                            // This marks the end of the initial $foo$.
+                            final String text = text();
+                            int i = input.indexOf(text, pos);
+                            if (i >= 0) {
+                                end = i + text.length();
+                                pos = i + text.length();
+                                return Token.STRING;
+                            }
+                            return Token.OTHER;
+                        } else if (Character.isLetter(c) || Character.isDigit(c) || c == '_') {
+                            // Identifier char, consume
+                        } else if (Character.isSpaceChar(c)) {
+                            end--;
+                            return Token.OTHER;
+                        }
+                    }
+                    // Unknown token starting with $ until EOF, just ignore it.
+                    return Token.OTHER;
+                }
+            default:
+                return Token.OTHER;
+        }
+    }
+
+    private Token scanKeywordOrIdentifier(boolean maybeKeyword) {
+        while (hasNext()) {
+            char c = peek();
+            if (Character.isDigit(c) || c == '_' || c == '$') {
+                maybeKeyword = false;
+            } else if (!Character.isLetter(c)) {
+                break;
+            }
+            next();
+        }
+        if (!maybeKeyword) {
+            return Token.IDENT;
+        }
+        List<Token> keywordsByLength = Token.getKeywordsByLength(textLength());
+        for (int i = 0; i < keywordsByLength.size(); i++) {
+            Token token = keywordsByLength.get(i);
+            if (isTextEqualIgnoreCase(token.name())) {
+                return token;
+            }
+        }
+        return Token.IDENT;
+    }
+
+    private Token scanNumericLiteral() {
+        boolean hasPeriod = false;
+        boolean hasExponent = false;
+        while (hasNext()) {
+            char c = peek();
+            if (Character.isDigit(c)) {
+                next();
+                continue;
+            }
+            switch (c) {
+                case '.':
+                    if (hasPeriod) {
+                        return Token.NUMBER;
+                    }
+                    next();
+                    hasPeriod = true;
+                    break;
+                case 'e':
+                case 'E':
+                    if (hasExponent) {
+                        return Token.NUMBER;
+                    }
+                    next();
+                    hasExponent = true;
+                    if (isNextChar('+') || isNextChar('-')) {
+                        next();
+                    }
+                    break;
+                default:
+                    return Token.NUMBER;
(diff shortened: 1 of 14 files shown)


apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
```diff
@@ -0,0 +1,132 @@
...
+import static co.elastic.apm.agent.jdbc.signature.Scanner.Token.FROM;
+import static co.elastic.apm.agent.jdbc.signature.Scanner.Token.IDENT;
+import static co.elastic.apm.agent.jdbc.signature.Scanner.Token.LPAREN;
+import static co.elastic.apm.agent.jdbc.signature.Scanner.Token.RPAREN;
+
+public class SignatureParser {
+
+    private final Scanner scanner = new Scanner();
+
+    public void querySignature(String query, StringBuilder signature) {
```
eyalkoren: Shouldn't we cache the resulted name for a given query string? It is likely that the same string object will be reused, but even if not, I assume hash code calculation is faster than this scan (can be benchmarked).

felixbarny: Good idea. But we also need to limit the size of the cache. When using algorithms like LRU for the cache and if there's a lot of dynamically generated SQL there will be a lot of thrashing and contention on that cache. I suspect that the contention is more expensive than just scanning the SQL. So what do you think about just caching the first ~200 SQL statements?

/cc @axw

felixbarny: Some challenges: We should probably use a map with weak keys so that we don't leak strings which would otherwise not be referenced as it takes away space in the cache and the same query may not be used again afterwards. However, if we do that for an application where a large portion of SQLs are dynamically created, we end up allocating a lot of strings for the cached signature. If not cached, the signature is just appended to a StringBuilder but we need to convert that to a String when we want to cache it.

A few ideas:
- When we hit a limit of total cache size, just disable the cache, assuming the application creates a lot of dynamic SQLs.
- Don't use a weak map and just cache the first x SQLs. Hit rate may not be great though.

axw: @felixbarny do you mean just stop caching new statements once you've hit 200? Seems fine, but then you've still got to deal with concurrent map updates.

I'm not really convinced that scanning each time is going to be a problem in practice. Applications that make repeat queries should be using prepared statements. If they're not, then I would take that as either an indication of one of the following:

 - performance of that particular bit of code is either cold / unimportant
 - it's hot and the devs don't know it; it will be (slightly?) exacerbated, but also highlighted by the agent as something to fix
 - it's a dynamically generated query, so will end up overflowing the cache anyway, and cache miss for every >200th query

So... I'd just benchmark it and see whether it's worth it.

I just went and benchmarked the base overhead for Go:

```
BenchmarkStmtQueryContext/baseline-8                      200000              9332 ns/op            1008 B/op         16 allocs/op
BenchmarkStmtQueryContext/baseline_prepared-8             300000              4790 ns/op             912 B/op         10 allocs/op
BenchmarkStmtQueryContext/instrumented-8                  200000              9213 ns/op             822 B/op         16 allocs/op
BenchmarkStmtQueryContext/instrumented_prepared-8         300000              4885 ns/op             721 B/op         10 allocs/op
BenchmarkQuerySignature-8                                1000000              1552 ns/op          77.94 MB/s          32 B/op          1 allocs/op
```

The first four are repeatedly executing "SELECT * FROM foo" against an in-memory SQLite database. The "prepared" variants use a prepared statement, the others do not. In either case, the query is short and simple, and the overhead is ~100ns/query.

The last benchmark is just repeatedly scanning a longer query: "SELECT *,(SELECT COUNT(*) FROM table2 WHERE table2.field1 = table1.id) AS count FROM table1 WHERE table1.field1 = 'value'", and we can see the overhead there is more substantial.

You would have to benchmark against using a concurrent hash map to know if it's worth it, and at what query length threshold.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/Scanner.java
```diff
@@ -0,0 +1,404 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class Scanner {
+
+    @Nullable
+    private String input;
+    private int start; // text start pos in bytes
```
axw: are these really in bytes? seems like the methods you use (e.g. StringBuilder.append) deal with character indices?

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/Scanner.java
```diff
@@ -0,0 +1,404 @@
...
+
+    public Token scan() {
+        if (!hasNext()) {
+            return Token.EOF;
+        }
+        char c = next();
+        while (Character.isSpaceChar(c) && hasNext()) {
+            c = next();
+        }
+        start = pos - 1;
```
axw: I just found a bug in the Go agent while reading your code :)
If you are dealing in bytes, then I think this should be `- number_of_bytes_in(c)` rather than `- 1`

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/Scanner.java
felixbarny: that's a copy/paste error 😅 
it's actually the char position

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
felixbarny: > do you mean just stop caching new statements once you've hit 200
That's option 2 which only really makes sense when not using weak keys.

Option 1 is to really deactivate caching altogether. The reasoning is that when using a weak reference to the statement we may do a lot of allocations, as described in my previous comment.

eyalkoren: > Applications that make repeat queries should be using prepared statements. 

The current code will still run every time. Do you suggest caching for the PreparedStatement objects instead? This is an option and we actually have such cache already. However, it will cover less cases.

I assume most applications use repeated queries, so I do believe this is something we should do in one for or another.

> When we hit a limit of total cache size, just disable the cache, assuming the application creates a lot of dynamic SQLs.

This can work, just make sure you either count yourself, or you count removals, as the weak-key maps clean GC keys, so map size is not a good indication.

axw: > That's option 2 which only really makes sense when not using weak keys.
>
> Option 1 is to really deactivate caching altogether. The reasoning is that when using a weak reference to the statement we may do a lot of allocations, as described in my previous comment.

@felixbarny sorry, I was replying to your previous message - hadn't refreshed and seen your follow up. Makes sense.

> The current code will still run every time. Do you suggest caching for the PreparedStatement objects instead? This is an option and we actually have such cache already. However, it will cover less cases.

@eyalkoren ah, I had meant to check if prepared statements were being cached, but forgot. This is essentially what I do in Go, and it works well. I have to override the prepared statement object anyway, so it's natural to attach the span name to the overridden prepared statement object. I don't know what the performance implication for the Java agent would be, so yet another thing to compare.

apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/signature/ScannerTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Test
+    void testScanner() {
+        for (JsonNode testCase : testCases) {
+            System.out.println(testCase);
+            scanner.setQuery(testCase.get("input").textValue());
+            for (JsonNode token : testCase.get("tokens")) {
+                assertThat(scanner.scan()).isEqualTo(Scanner.Token.valueOf(token.get("kind").textValue()));
+                assertThat(scanner.text()).isEqualTo(token.get("text").textValue());
+            }
```
axw: you should also check that the scanner returns Token.EOF after the loop

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/Scanner.java
```diff
@@ -0,0 +1,404 @@
...
+        }
+        return false;
+    }
+
+    public Token scan() {
+        if (!hasNext()) {
+            return Token.EOF;
+        }
+        char c = next();
+        while (Character.isSpaceChar(c) && hasNext()) {
```
axw: I think there's a problem here: what happens if the whole string is whitespace? You'll exit the loop, but keep handling the whitespace character below. I think you should `if (!hasNext()) {return Token.EOF}` inside the loop.

apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/signature/ScannerTest.java
felixbarny: ha, just did that 😄

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/Scanner.java
felixbarny: good catch! but I should only do that if the next char is a whitespace again

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/Scanner.java
```diff
@@ -0,0 +1,404 @@
...
+        INTO,
+        OR,
+        REPLACE,
+        SELECT,
+        SET,
+        TABLE,
+        TRUNCATE, // Cassandra/CQL-specific
+        UPDATE;
+
+        private static final Map<Integer, List<Token>> keywordsByLength = new HashMap<>();
```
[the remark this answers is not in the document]
felixbarny: Good idea, that also eliminates the allocation for boxing the int. However, that seems to be handled by escape analysis. Still seems worth it.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
felixbarny: > This can work, just make sure you either count yourself,

What do you mean with "count yourself"? I'm only one 😜 

> or you count removals, as the weak-key maps clean GC keys, so map size is not a good indication.

Do you mean that after x keys have been GC'd, we should stop caching?

Only caching the span name for prepared statements sounds good. But prepared statements queries may still be constructed dynamically, causing unnecessary allocations for the span name. For example when there's a search which multiple inputs which correspond to different properties (first name, last name, age range, ...) the query may be constructed based on which fields are filled out but still uses prepared statements to leverage parameter placeholders.

apm-agent-plugins/apm-jdbc-plugin/src/test/resources/scanner_tests.json
```diff
@@ -0,0 +1,213 @@
...
+        "kind": "PERIOD",
+        "text": "."
+      },
+      {
+        "kind": "NUMBER",
+        "text": "3"
+      }
+    ]
+  }
+]
```
axw: FYI I'm in the process of adding a new test case to cover the bug I found in the Go agent:

```json
{
       "name": "unicode",
       "input": "选择 FROM foo",
       "tokens": [
               {
                       "kind": "IDENT",
                       "text": "选择"
               },
               {
                       "kind": "FROM",
                       "text": "FROM"
               },
               {
                       "kind": "IDENT",
                       "text": "foo"
               }
       ]
}
```

felixbarny: phew, works with the Java impl

felixbarny:
Implemented a cache which disables once it hits 512 entries, assuming the application creates a lot of dynamic SQL which would make a cache ineffective. More rationale in the Javadocs. I'd appreciate feedback so we can move on with this 🙂

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
```diff
@@ -32,10 +36,60 @@
...
+    private static final int QUERY_LENGTH_CACHE_UPPER_THRESHOLD = 10_000;
     private final Scanner scanner = new Scanner();
+    /**
+     * Not using weak keys as the DISABLE_CACHE_THRESHOLD may not be reached when clearing stale entries.
+     * That means that we are leaking Strings but as the size of the map is limited that should not be an issue.
+     * Another reason is that ORMs like Hibernate generate equal SQL strings for the same query but don't reuse the same string instance.
+     * When relying on weak keys, we would not leverage any caching benefits if the query string is collected.
+     */
+    @Nullable
+    private ConcurrentMap<String, String> signatureCache = new ConcurrentHashMap<>(DISABLE_CACHE_THRESHOLD);
```
axw: each thread has its own cache? or should this be static?

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
```diff
@@ -32,10 +36,60 @@
...
+    /**
+     * We don't want to keep alive references to huge query strings
+     */
+    private static final int QUERY_LENGTH_CACHE_UPPER_THRESHOLD = 10_000;
     private final Scanner scanner = new Scanner();
+    /**
+     * Not using weak keys as the DISABLE_CACHE_THRESHOLD may not be reached when clearing stale entries.
+     * That means that we are leaking Strings but as the size of the map is limited that should not be an issue.
+     * Another reason is that ORMs like Hibernate generate equal SQL strings for the same query but don't reuse the same string instance.
+     * When relying on weak keys, we would not leverage any caching benefits if the query string is collected.
```
eyalkoren: That what I meant by saying you should count yourself (so - not to count you as a PreparedStatement 😛 ), eg count removals. But I assume if we have one cache for the entire JVM this is a non-issue

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
```diff
@@ -32,10 +36,60 @@
...
+     */
+    @Nullable
+    private ConcurrentMap<String, String> signatureCache = new ConcurrentHashMap<>(DISABLE_CACHE_THRESHOLD);
+
+    public void querySignature(String query, StringBuilder signature, boolean preparedStatement) {
```
eyalkoren: I imagine that even if the cache gets full over time, it still contains the **hot** queries, so it still makes sense to do lookups, while avoiding caching of new signatures. In other words, can I think it's better to use the conditions only for deciding whether to add to the cache, but not remove the cache altogether.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
felixbarny: good catch. should be static

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
felixbarny: That means we'll have to keep track via an AtomicBoolean whether or not the cache is active, right? But we can probably use `AtomicBoolean#getOpaque` to check if the cache is active or not. It has weaker guarantees than volatile but it guarantees to be eventually consistent.

felixbarny: Wait, I don't need an extra flag at all when not using a weak map. And we have already concluded weak maps are not a good fit because ORMs don't use the same string instances but generate new but equal strings.

PR #640, opened 2019-05-21
Remote config

### LINKED ISSUE #636: Remote config
See elastic/apm#76. We will update the agent to periodically query config from the APM Server.

felixbarny:
closes elastic/apm-agent-java#636


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
```diff
@@ -100,8 +105,13 @@ public ElasticApmTracer build() {
             final List<ConfigurationSource> configSources = getConfigSources(agentArguments);
             configurationRegistry = getDefaultConfigurationRegistry(configSources);
         }
+        final ApmServerClient apmServerClient = new ApmServerClient(configurationRegistry.getConfig(ReporterConfiguration.class));
+        final DslJsonSerializer payloadSerializer = new DslJsonSerializer(configurationRegistry.getConfig(StacktraceConfiguration.class));
+        final MetaData metaData = MetaData.create(configurationRegistry, null, null);
+        configurationRegistry.addConfigurationSource(new ApmServerConfigurationSource(payloadSerializer, metaData, apmServerClient));
+        configurationRegistry.reloadDynamicConfigurationOptions();
```
felixbarny: This block the startup until there's a response from the APM Server.

Pros:
- Does not start the agent without applying the remote configuration
  Avoids that the agent starts with 100% sampling and only bring it down to the remotely configured level after 30s
- We could also load non-dynamic/reloadable options remotely

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ApmServerConfigurationSource.java
```diff
@@ -0,0 +1,108 @@
...
+                connection.setRequestProperty("If-None-Match", etag);
+            }
+            payloadSerializer.setOutputStream(connection.getOutputStream());
+            payloadSerializer.serializeMetadata(metaData);
+            payloadSerializer.flush();
+
+            final int status = connection.getResponseCode();
+            if (status == 404) {
+                logger.info("This version of the APM Server does not allow fetching configuration. " +
+                    "Update to APM Server 7.? to take advantage of centralized configuration.");
```
felixbarny:
```suggestion
                    "Update to APM Server 7.3 to take advantage of centralized configuration.");
```

jalvz:
depends on https://github.com/elastic/apm-server/pull/2095

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ApmServerConfigurationSource.java
```diff
@@ -0,0 +1,108 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(ApmServerConfigurationSource.class);
+
+    private final DslJson<Object> dslJson = new DslJson<>();
+    private final byte[] buffer = new byte[4096];
+    private final PayloadSerializer payloadSerializer;
+    private final MetaData metaData;
+    private final ApmServerClient apmServerClient;
+    @Nullable
+    private String etag;
+    private Map<String, String> config = Collections.emptyMap();
```
felixbarny:
```suggestion
    private volatile Map<String, String> config = Collections.emptyMap();
```

felixbarny: Not needed, as the configuration is only ready from the same thread. The values are then updated on the `ConfigurationOption` class which only parses the String if it differs from the previous value.

apm-agent-core/src/test/java/co/elastic/apm/agent/metrics/builtin/SystemMetricsTest.java
```diff
@@ -37,6 +37,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
 
+@Disabled
```
eyalkoren: Revert 😄

apm-agent-core/src/test/java/co/elastic/apm/agent/configuration/ApmServerConfigurationSourceTest.java
```diff
@@ -0,0 +1,88 @@
...
+    public void setUp() throws Exception {
+        config = SpyConfiguration.createSpyConfig();
+        apmServerClient = new ApmServerClient(config.getConfig(ReporterConfiguration.class), List.of(new URL("http", "localhost", mockApmServer.port(), "/")));
+        mockApmServer.stubFor(post(urlEqualTo("/config/v1/agents")).willReturn(ResponseDefinitionBuilder.okForJson(Map.of("foo", "bar")).withHeader("ETag", "foo")));
+        mockApmServer.stubFor(post(urlEqualTo("/config/v1/agents")).withHeader("If-None-Match", equalTo("foo")).willReturn(status(304)));
+        configurationSource = new ApmServerConfigurationSource(new DslJsonSerializer(mock(StacktraceConfiguration.class)), MetaData.create(config, null, null), apmServerClient);
+    }
+
+    @Test
+    public void testLoadRemoteConfig() throws Exception {
```
eyalkoren: Depending on the changes in the communication, add tests. For example- getting a 404 with no `Cache-Control` header should cause change in the polling interval.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerHealthChecker.java
```diff
@@ -24,86 +24,65 @@
...
-        try {
-            URL url = new URL(reporterConfiguration.getServerUrls().get(0).toString() + "/");
-            if (logger.isDebugEnabled()) {
-                logger.debug("Starting healthcheck to {}", url);
-            }
-            connection = (HttpURLConnection) url.openConnection();
-            if (!reporterConfiguration.isVerifyServerCert()) {
-                if (connection instanceof HttpsURLConnection) {
-                    trustAll((HttpsURLConnection) connection);
+        for (int i = 0; i < reporterConfiguration.getServerUrls().size() && !success; i++) {
```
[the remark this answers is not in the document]
felixbarny: That would be very tricky as even if we start the agent, we might want to read some configuration option in a constructor and store it in a final field. That doesn't work when we don't have the final configuration yet.
Or we only ever allow configuring dynamic configuration options centrally.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
felixbarny: That should be aligned with the other agents. I've added a comment, listing the options here: https://github.com/elastic/apm/issues/76#issuecomment-509709435

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerHealthChecker.java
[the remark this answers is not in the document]
eyalkoren: What I meant is this:
- Do everything same as you did so far - initialize all local configurations on startup
- When done- set `active` to `false` (it's dynamic)
- Do healthcheck
   - If healthcheck fails - keep inactive
   - if healthcheck succeeds - try to get remote config
      - if succeeds - apply configurations and switch to active (in the future active will be one of the configs coming from this source)
      - if fails - switch to active

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
eyalkoren: Added a comment too, tried to push this to a conclusion.

apm-agent-core/src/test/java/co/elastic/apm/agent/report/ApmServerClientTest.java
```diff
@@ -0,0 +1,148 @@
...
+        assertThatThrownBy(() -> apmServerClient.executeForAllUrls("/not-found", connection -> {
+            connection.getInputStream();
+            return null;
+        }))
+            .isInstanceOf(FileNotFoundException.class)
+            .matches(t -> t.getSuppressed().length == 1, "should have a suppressed exception");
+        apmServer1.verify(getRequestedFor(urlEqualTo("/not-found")));
+        apmServer2.verify(getRequestedFor(urlEqualTo("/not-found")));
+        // two failures -> urls wrap
+        assertThat(apmServerClient.getCurrentUrl().getPort()).isEqualTo(apmServer1.port());
```
eyalkoren: If you agree `executeForAllUrls` shouldn't affect state, the related test should be testing failure URLs and verifying state didn't change

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
```diff
@@ -0,0 +1,258 @@
...
+        if (previousException == null) {
+            throw new IllegalStateException("Expected previousException not to be null");
+        }
+        throw previousException;
+    }
+
+    public void executeForAllUrls(String path, ConnectionHandler<Void> connectionHandler) throws Exception {
+        int expectedErrorCount = errorCount.get();
+        Exception previousException = null;
+        for (URL serverUrl : getPrioritizedUrlList()) {
```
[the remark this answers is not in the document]
felixbarny: I don't really like the name either.. Do you have a suggestion?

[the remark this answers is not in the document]
eyalkoren: It's not about the name, I meant- you can just use `serverUrls` here

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
```diff
@@ -0,0 +1,258 @@
...
+                }
+            }
+        }
+        if (previousException == null) {
+            throw new IllegalStateException("Expected previousException not to be null");
+        }
+        throw previousException;
+    }
+
+    public void executeForAllUrls(String path, ConnectionHandler<Void> connectionHandler) throws Exception {
```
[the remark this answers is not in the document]
felixbarny: With the new changes, the logic is quite different now.

PR #642, opened 2019-05-23
Fix source jar

### LINKED ISSUE #630: Fix source jar created on Jenkins build
After multiple tests, current situation is:
The elastic-apm-agent pom.xml has two related plugins involved with the source jar build-
1. the maven-source-plugin is invoked at the `prepare-package` phase
2. the maven-shade-plugin creates a source jar with the shaded sources

I am not entirely sure the first is required here, it is required for the attach source jar.

Invoking the exact same command as invoked by Jenkins - `./mvnw clean install -DskipTests=true -Dmaven.javadoc.skip=true` creates the jar as expected. However, on Jenkins, the jar is created only with the shaded sources and without our sources (at least - this is what gets stored for the build artifacts).

eyalkoren:
Fixes elastic/apm-agent-java#630


eyalkoren:
Closing this as we don't seem to get to that. The issue is still ready.

PR #646, opened 2019-05-27
Cache AttachmentProvider

felixbarny:


apm-agent-attach/src/main/java/co/elastic/apm/attach/CachedAttachmentProvider.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package co.elastic.apm.attach;
+
+import net.bytebuddy.agent.ByteBuddyAgent;
+
+/**
+ * Successive attachments would lead to the VirtualMachine class to be loaded from multiple class loaders
+ * (see net.bytebuddy.agent.ByteBuddyAgent.AttachmentProvider.ForStandardToolsJarVm#attempt()).
+ * That leads to a UnsatisfiedLinkError on Java 7 and 9 because the native library libattach can only be loaded by one class loader.
+ * By caching the {@link net.bytebuddy.agent.ByteBuddyAgent.AttachmentProvider.Accessor}, the same
```
eyalkoren: Your sentence ends unexpectedly 🙂

apm-agent-attach/src/main/java/co/elastic/apm/attach/CachedAttachmentProvider.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public class CachedAttachmentProvider implements ByteBuddyAgent.AttachmentProvider {
+    private final Accessor accessor;
+
+    public CachedAttachmentProvider(ByteBuddyAgent.AttachmentProvider provider) {
+        this.accessor = provider.attempt();
```
eyalkoren: This means only one AttachementProvider can be attempted. Is that the case always?

felixbarny: There is a `AttachmentProvider.Compound` which tries multiple ones and returns the first successful attempt.

apm-agent-attach/src/main/java/co/elastic/apm/attach/CachedAttachmentProvider.java
felixbarny: 😱

PR #649, opened 2019-05-29
Fix service loading for WebLogic 12.2.1.2.0

felixbarny:
The issue has to do with how `ServiceLoader`s behave on that particular version of WebLogic. I'm not sure how WebLogic can even mess with such low-level stuff...

When `ServiceLoader`s are asked to load a resource for the bootstrap class loader (which is represented with `null`), they actually use `ClassLoader.getSystemClassLoader().loadClass(String)` and rely on the parent delegation to load the classes from the bootstrap CL.

Our agent is actually present on both the system and the bootstrap CL (which is not ideal tbh) and on this specific WebLogic version, the boot delegation for `ClassLoader.getSystemClassLoader().loadClass(String)` does not seem to work for some reason. The workaround is to load the classes via `java.lang.Class#forName(java.lang.String, boolean, java.lang.ClassLoader)` which allows the `ClassLoader` to be `null` which then reliably loads the classes from the bootstrap classloader.

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
@@ -11,9 +11,9 @@
  * the Apache License, Version 2.0 (the "License"); you may
  * not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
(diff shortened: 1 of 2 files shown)


felixbarny:
TIL: it's possible to override the system class loader via the system property `java.system.class.loader`, which is what this WebLogic version does. See https://docs.oracle.com/javase/7/docs/api/java/lang/ClassLoader.html#getSystemClassLoader()

Issue #466, opened 2019-01-31
Support basic JMS scenarios


eyalkoren:
This is about the basic scenarios, so, for example, not supporting a pattern of writing a message to a queue and waiting for a response message on a different queue.
Furthermore, this issue only concerns Java agent support for the JMS API, we need to see how things get presented in the UI for real async delayed message handling.
A basic requirement is context propagation, and transaction creation when reading from queue.
Nice-to-have - span of the publishing event with details like queue name.

PR #650, opened 2019-05-30
Basic JMS Support

eyalkoren:
Closes elastic/apm-agent-java#466 

**Supports:** 
JMS 1.1 and 2, APIs: `send`, `receive`, `receiveNoWait`, `onMessage`.

**Testing:**
ActiveMQ (JMS 1.1), wrapped by a connection-pooling client and ActiveMQ Artemis (JMS 2) - the successor of HornetQ).

**Implementation details:**
`MessageProducer#send` traced and used for context propagation through Message String property. Since JMS 2 is an extension of JMS 1.1This is assumed to be enough to support `JMSProducer#send` (JMS 2 API) as it likely one makes use of the other (tested with one library).
`JMSConsumer#receive`, `JMSConsumer#receivereceiveNoWait` and `MessageListener#onMessage` are traced for receiver side, reading the context from the message property and using it for distributed tracing.

**Assumptions and decisions made:**
- `receive` API tracing can result either in a Transaction (if occurs on a non-traced thread) or a Span. If the latter, distributed tracing context will be ignored and span will remain child of its original parent.
- Until we have a dedicated filed in schema for that, destination (queue/topic) name will be part of the span/transaction name
- `onMessage` always results in a transaction. The assumption is that it is never expected to be executed on an already traced thread.

**Not within the scope of this PR:**
- JMS 2 async sends (`CompletionListener`) both in `JMSProducer` and `MessageProducer`
- [Reply-to/correlation-ID mechanisms](https://docs.oracle.com/cd/E13171_01/alsb/docs25/interopjms/MsgIDPatternforJMS.html) not tested


apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/Jms1InstrumentationIT.java
```diff
@@ -0,0 +1,127 @@
...
+    @Override
+    protected void send(Destination destination, Message message) throws JMSException {
+        session.createProducer(destination).send(message);
+    }
+
+    @Override
+    protected CompletableFuture<Message> registerListener(Destination destination) throws JMSException {
+        MessageConsumer consumer = session.createConsumer(destination);
+        final CompletableFuture<Message> incomingMessageFuture = new CompletableFuture<>();
+        //noinspection Convert2Lambda,Anonymous2MethodRef - we don't instrument lamdas or method references
```
felixbarny: We should probably wrap lambdas so that we can instrument the wrapper. Otherwise, we'd miss out on those.

docs/supported-technologies.asciidoc
```diff
@@ -224,6 +224,24 @@ This section lists all supported asynchronous frameworks.
...
+same trace.
+
+|===
+|Framework |Supported versions | Description | Since
+
+|JMS
+|1.1, 2.0
+|The agent propagates the context when using `javax.jms.MessageConsumer#send` and `javax.jms.JMSConsumer#send` methods
+through a `Message` property
+|1.7.0 - EXPERIMENTAL (off by default)
```
felixbarny: describe how to enable

docs/supported-technologies.asciidoc
```diff
@@ -224,6 +224,24 @@ This section lists all supported asynchronous frameworks.
...
+When using a messaging framework, sender context is propagated so that receiver events are correlated to the
+same trace.
+
+|===
+|Framework |Supported versions | Description | Since
+
+|JMS
+|1.1, 2.0
+|The agent propagates the context when using `javax.jms.MessageConsumer#send` and `javax.jms.JMSConsumer#send` methods
+through a `Message` property
```
felixbarny: describe for which APIs we create spans/transactions

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
  * Added support for AsyncHttpClient 2.x
  * Added [`global_labels`](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#global-labels) configuration option.
    This requires APM Server 7.2+.
+ * Added basic support for JMS- distributed tracing for basic scenarios of `send`, `receive`, `receiveNoWait` and 
+   `onMessage`. Both Queues and Topics are supported. Async `send` APIs are not supported in this version. 
+   NOTE: This feature is currently marked as "Experimental" and is disabled by default. In order to enable, it is 
+   required to set the `disable_instrumentations` configuration property to an empty string.
```
felixbarny:
```suggestion
   required to set the [`disable_instrumentations`](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#config-disable-instrumentations) configuration property to an empty string.
```

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
  * Added support for AsyncHttpClient 2.x
  * Added [`global_labels`](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#global-labels) configuration option.
    This requires APM Server 7.2+.
+ * Added basic support for JMS- distributed tracing for basic scenarios of `send`, `receive`, `receiveNoWait` and 
+   `onMessage`. Both Queues and Topics are supported. Async `send` APIs are not supported in this version. 
+   NOTE: This feature is currently marked as "Experimental" and is disabled by default. In order to enable, it is
```
felixbarny: marked as `incubating`?

felixbarny: Let's also discuss whether we can enable this by default but still having it marked as incubating. Do you think that the instrumentation and context handling is safe enough? What's missing that we can confidently remove the `incubating` tag?

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageConsumerInstrumentation.java
```diff
@@ -0,0 +1,157 @@
...
+                try {
+                    if (message != null) {
+                        if (abstractSpan instanceof Transaction) {
+                            try {
+                                String messageSenderContext = message.getStringProperty(JMS_TRACE_PARENT_HEADER);
+                                if (messageSenderContext != null) {
+                                    abstractSpan.getTraceContext().asChildOf(messageSenderContext);
+                                }
+                            } catch (JMSException e) {
+                                LoggerFactory.getLogger("JmsMessageConsumerInstrumentation")
```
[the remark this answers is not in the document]
eyalkoren: I thought it would be misleading to reference a field of the enclosing class, because in runtime, it is not really the enclosing class, just a random agent class...
However, using the `VisibleForAdvice` may be enough to indicate that. 
Will do.

apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/Jms1InstrumentationIT.java
eyalkoren: Good idea. I already have a helper in place for pooling wrappers 😉 
As far as I can see, APIs I will need to instrument for that are:
1. `MessageConsumer#setMessageListener` (following the same logic of supporting JMS 2 APIs through JMS 1.1 APIs, this should cover `JMSConsumer#setMessageListener` as well)
2. `Session#setMessageListener`

Will need to see further if one covers the other, so maybe I can get away with only one.

docs/supported-technologies.asciidoc
eyalkoren: Not too long for this table??

CHANGELOG.md
eyalkoren: I think the instrumentation is safe enough, however there are some stuff that we may want to change following the general messaging discussion, which will be kinda breaking changes, mainly the new `transaction.type` (I added `messaging`, but this may change).
In addition, `span.type` may change, as well as the span names (maybe we will want to remove the destination name from the span name once we add it to the schema).

Lastly, there is some UI handling required to fix some scenarios like very delayed receiver span, or a receiver span/transaction starting before (or a lot before) the sender span. As long as these are not implemented, they may cause a problematic user experience, so better start off with `incubating` I think.

docs/supported-technologies.asciidoc
felixbarny: I'd err towards being long but explicit. @bmorelli25 wdyt?

bmorelli25: +1 on long but explicit in this case. @eyalkoren if it's too long for the table I can come up with a better way to present the information.

apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/Jms1InstrumentationIT.java
eyalkoren: Looks like `Session#setMessageListener` is not an API to be used in plain JMS clients, more something for usage by app servers or the like. So, currently I won't instrument it.

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
```diff
@@ -86,4 +109,49 @@ public Span startJmsProducerSpan(Destination destination, Message message) {
...
+    }
+
+    private void recycle(MessageListenerWrapper messageListenerWrapper) {
+        messageListenerWrapperObjectPool.recycle(messageListenerWrapper);
+    }
+
+    public class MessageListenerWrapper implements MessageListener, Recyclable {
+
+        @Nullable
+        private MessageListener delegate;
```
felixbarny: does this need to be volatile?

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
```diff
@@ -86,4 +109,49 @@ public Span startJmsProducerSpan(Destination destination, Message message) {
...
+        }
+
+        @Override
+        public void onMessage(Message message) {
+            try {
+                if (delegate != null) {
+                    delegate.onMessage(message);
+                }
+            } finally {
+                JmsInstrumentationHelperImpl.this.recycle(this);
```
felixbarny: couldn't the same listener be invoked multiple times?

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageListenerInstrumentation.java
```diff
@@ -0,0 +1,140 @@
...
+    @SuppressWarnings("WeakerAccess")
+    public static final Logger logger = LoggerFactory.getLogger(JmsMessageListenerInstrumentation.class);
+
+    public JmsMessageListenerInstrumentation(ElasticApmTracer tracer) {
+        super(tracer);
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return not(isInterface()).and(hasSuperType(named("javax.jms.MessageListener")));
```
felixbarny: We can't do pre-filtering for these as it might be implemented as an anonymous inner class or lambda. How big is the impact in terms of startup performance?

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
eyalkoren: Of course it can 🤦‍♂ 
Good catch!!
The other option of pooling them is to instrument `MessageConsumer#close` as well, and in the `setMessageListener` being able to identify already used Listeners, which means I will need to cache them in a `MessageConsumer`-->`MessageListenerWrapper` map..
WDYT?

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageListenerInstrumentation.java
eyalkoren: Yeah, that's a given. There is an impact:
Screen Shot 2019-06-06 at 10 54 13
Hard to tell what real-world scenarios implications will be...
We do filter by ClassLoader, but this wouldn't help in any case JMS APIs are added in the app server level

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
eyalkoren: I mean- maybe the extra risk and complication doesn't worth the performance gain 🤔

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
eyalkoren: It better be!

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageListenerInstrumentation.java
eyalkoren: We decided to go with that, as it is.
Options discussed for relieving startup time overhead:
1. using `application_packages` filtering on type match- will capture all user's implementations of `MessageListener`s, but will miss framework listeners, like Message beans
2. add a mechanism for lazily match and instrument (`retransform`) already loaded classes on a background thread using some non-greedy strategy, forcing only specific instrumentations to be applied on runtime. For example- `MessageConsumer#setMessageListener` should be instrumented on startup because a consumer could be long-lived and we don't want to miss first invocation.

We can reconsider in the future. So far, we didn't hear any complaints about startup overhead/delays.

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
eyalkoren: Discussed and decided not to pool those atm. We don't know whether the common use case is long-lived or short-lived `MessageListener`s. We can revisit if we see it has a real impact in some use cases.

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
eyalkoren: Now that we don't pool them- irrelevant. Will just make final and initialize on construction.

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageConsumerInstrumentation.java
```diff
@@ -0,0 +1,199 @@
...
+
+    @VisibleForAdvice
+    @SuppressWarnings("WeakerAccess")
+    public static final Logger logger = LoggerFactory.getLogger(JmsMessageConsumerInstrumentation.class);
+
+    JmsMessageConsumerInstrumentation(ElasticApmTracer tracer) {
+        super(tracer);
+    }
+
+    // todo - check about this filter...
```
felixbarny: is this still relevant?

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageConsumerInstrumentation.java
```diff
@@ -0,0 +1,199 @@
...
+        @Override
+        public Class<?> getAdviceClass() {
+            return MessageConsumerAdvice.class;
+        }
+
+        public static class MessageConsumerAdvice {
+
+            @Advice.OnMethodEnter(suppress = Throwable.class)
+            @Nullable
+            public static AbstractSpan startSpan(@Advice.Origin Class<?> clazz) {
```
felixbarny:
```suggestion
            public static AbstractSpan beforeReceive(@Advice.Origin Class<?> clazz) {
```

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageConsumerInstrumentation.java
```diff
@@ -0,0 +1,199 @@
...
+                            .withAction("receive");
+                    }
+                    abstractSpan.setName("JMS RECEIVE");
+                    abstractSpan.activate();
+                }
+                return abstractSpan;
+            }
+
+            @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
+            public static void endSpan(@Advice.Enter @Nullable final AbstractSpan abstractSpan,
```
felixbarny:
```suggestion
            public static void afterReceive(@Advice.Enter @Nullable final AbstractSpan abstractSpan,
```

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageProducerInstrumentation.java
```diff
@@ -0,0 +1,173 @@
...
+
+    @VisibleForAdvice
+    @SuppressWarnings("WeakerAccess")
+    public static final Logger logger = LoggerFactory.getLogger(JmsMessageProducerInstrumentation.class);
+
+    JmsMessageProducerInstrumentation(ElasticApmTracer tracer) {
+        super(tracer);
+    }
+
+    // todo - check about this filter...
```
felixbarny: still relevant?

felixbarny: To reduce false negatives `nameContains("Message").or(nameContains("Producer"));`  may help?

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageConsumerInstrumentation.java
eyalkoren: I will remove and we'll wait and see if this produces false negatives

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageConsumerInstrumentation.java
eyalkoren: If you like it- you got it!

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageProducerInstrumentation.java
eyalkoren: Probably will reduce false negatives on the account of more false positives (so a bit decreased filtration efficiency)...
I'll apply this suggestion anyway, false negatives are more painful that false positives in this case

eyalkoren:
Jenkins test this please

PR #651, opened 2019-05-31
Allow log_file to be a relative path

felixbarny:


apm-agent-core/src/test/java/co/elastic/apm/agent/logging/LoggingConfigurationTest.java
```diff
@@ -48,9 +49,9 @@ void testSetLogFileInvalid() {
 
     @Test
     void testSetLogFile() {
-        final String logFile = "./apm.log";
-        LoggingConfiguration.init(Collections.singletonList(new SimpleSource().add("log_file", logFile)));
-        assertThat(System.getProperty(SimpleLogger.LOG_FILE_KEY)).isEqualTo(logFile);
+        final String logFile = "apm.log";
+        LoggingConfiguration.init(Collections.singletonList(new SimpleSource().add("log_file", LoggingConfiguration.getActualLogFile(null, logFile))));
```
eyalkoren: Wouldn't it be better tested like this:
```suggestion
        LoggingConfiguration.init(Collections.singletonList(new SimpleSource().add("log_file", logFile)));
```
?

PR #653, opened 2019-05-31
Disable metrics when active=false

felixbarny:


felixbarny:
I'm not really sure how to properly test that tbh. Not sure if the effort is worth it, especially as it's not a really critical component.

PR #655, opened 2019-06-05
feat(jenkins): new notifications and reports to Elasticsearch

kuisathaverat:
* new notifications and reports to Elasticsearch
Screenshot 2019-06-05 at 12 57 54

PR #656, opened 2019-06-05
feat(jenkins): new notifications and reports to Elasticsearch - 1.x

kuisathaverat:
see https://github.com/elastic/apm-agent-java/pull/655


felixbarny:
We probably don't need that for the 1.x branch, we only use that for pointing the docs to the current version but we typically release from master.

PR #657, opened 2019-06-10
[docs][7.2] update config-environment

bmorelli25:
**Do not merge yet. Blocked by: https://github.com/elastic/kibana/pull/38577**
Implements https://github.com/elastic/apm/issues/81

Link: `{kib}/filters.html#environment-selector[]`


felixbarny:
I think the HTML has precedence over the properties file. 

@bmorelli25 do you have insight into how often https://www.elastic.co/guide/en/apm/agent/java/current/config-reference-properties-file.html is requested, relative to the other config pages?

Slightly off topic: Rather than having the reference property file list all the config options, we could also think about only including the most relevant ones. Currently, it's a bit unwieldy.

eyalkoren:
I vote for using the most inclusive and elaborated properties file as reference.
The way I imagine its usage is copying it over and start playing with it. Getting back to it after a year and see the full description again and the original default is very useful.

eyalkoren:
@bmorelli25 @felixbarny Unless you have any objection, I can build (actually- test) and push (so that the automatic configuration-documentation processes will take effect). Let me know.

felixbarny:
go for it 🙂

PR #658, opened 2019-06-11
Fix implementation version discovery of encoded jar files

eyalkoren:
Fixing https://discuss.elastic.co/t/tomcat-elastic-apm-agent-and-application-version-in-context-path-fails-on-protectiondomain/184899


eyalkoren:
@felixbarny do you think we should change it so that matching based on implementation version will default to `true` if fails? False negatives mean we will not instrument. False positives mean we have a chance, but may break the application (depending on what this matching is used for).

felixbarny:
Hm, good question. It probably depends on a case-per-case-basis. But in the case of the `ApacheHttpAsyncClientRedirectInstrumentation` it makes sense to default to `true`, as the instrumentation is basically idempotent because of the `original.containsHeader` check.

eyalkoren:
Your'e right, but the only way to implement different behavior per-usage would be to throw the Exception from the matcher. Do we want that? I can add a dedicated declared Exception so that anyone that want to use it will have to think about it and handle it. WDYT?

eyalkoren:
Even a bit more complicated to apply differentiated behavior based on instrumentation case. I think this may over-complicate code.
Since we are only using in a single place, I will change the fallback to `true`, ie passing the matching when version discovery fails.

PR #660, opened 2019-06-12
Increase JMS tests timeouts to 3 seconds

eyalkoren:


eyalkoren:
I hope this would be enough...

PR #662, opened 2019-06-12
Handling recycling race and volatile read-writes in Apache Async client

eyalkoren:
Fixes the issue reported at https://discuss.elastic.co/t/a-null-pointer-exception-is-possible-when-performing-asynchronous-requests/185315


apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/HttpAsyncRequestProducerWrapper.java
```diff
@@ -118,14 +115,14 @@ public void resetRequest() throws IOException {
 
     @Override
     public void close() throws IOException {
-        helper.recycle(this);
```
felixbarny: the error was we recycled (which sets the delegate to null) before closing the delegate, right? why did this only fail under load?

eyalkoren: Not really load, the request producer may not have been closed in the test, or closed on a separate thread where the test is not affected 🤷‍♂

PR #672, opened 2019-06-17
Test manual flush

eyalkoren:
Just a draft PR to test manual flush in an AWS environment


eyalkoren:
Jenkins test this please

felixbarny:
The flush I had in mind for AWS would not initiate a new request each time but rather flush the OutputStream so that we can be sure all spans and the transaction have been sent to the APM Server.

felixbarny:
But as the API doesn't have to change for that, we can start with a "hard flush" and do the "soft flush" later.

eyalkoren:
> The flush I had in mind for AWS would not initiate a new request each time but rather flush the OutputStream so that we can be sure all spans and the transaction have been sent to the APM Server.

Of course, this is not a suggested solution, this is for https://discuss.elastic.co/t/apm-with-aws-lambda-java/185936. I just want him to have a quick build to test. It was a draft PR at first, but then Jenkins only creates docs and doesn't build...

What I thought is how we can still batch, being more efficient on compression and metadata, but of course there are other paths.

eyalkoren:
Jenkins test this please

PR #677, opened 2019-06-18
ci(jenkins): trigger async ITs and update github check for the pull requests

v1v:
fixes https://github.com/elastic/apm-integration-testing/issues/41

## Highlights
- It's triggered for each PR by default and disabled to any upstream branches.
- Trigger the Integration Tests, aka ITs, pipelines from https://github.com/elastic/apm-integration-testing based on PRs or on demand, aka manually.
- It does trigger only the `Java` stage in the ITs.
- ITs should be async.
- This new stage notifies the GitHub check with the status of the execution.
- Depends on https://github.com/elastic/apm-integration-testing/pull/470

## Tasks
- [x] Agree with the stakeholders
- [x] It does require to support ITs from commits rather than branches.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,15 +1,18 @@
 #!/usr/bin/env groovy
 
 @Library('apm@current') _
-  
+
 pipeline {
   agent any
   environment {
-    BASE_DIR="src/github.com/elastic/apm-agent-java"
+    REPO = 'apm-agent-java'
+    BASE_DIR = "src/github.com/elastic/${env.REPO}"
     NOTIFY_TO = credentials('notify-to')
     JOB_GCS_BUCKET = credentials('gcs-bucket')
     DOCKERHUB_SECRET = 'secret/apm-team/ci/elastic-observability-dockerhub'
     CODECOV_SECRET = 'secret/apm-team/ci/apm-agent-java-codecov'
+    GITHUB_CHECK_ITS_NAME = 'Integration Tests'
+    ITS_PIPELINE = 'apm-integration-tests-mbp/PR-470'
   }
   options {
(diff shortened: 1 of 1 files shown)


v1v:
Jenkins run the tests

eyalkoren:
Jenkins run the tests

felixbarny:
Is it possible to only trigger the Java integration tests instead of all?

v1v:
@felixbarny , IIUC, the java one cannot be triggered individually for two main reasons:
- It does care about the https://github.com/elastic/apm-integration-testing/blob/master/tests/versions/java.yml
- It's not supported yet.

I need to research whether we can provide this particular use case but it might take a bit of time, what do you think if we just leave as is in this iteration and work on the enhancement later on?

felixbarny:
> what do you think if we just leave as is in this iteration and work on the enhancement later on?

One question: what happens if the integration tests for a different agent fail? Does it fail the build for the Java agent as well?

If it doesn't fail the build for the Java agent, I'm good with doing that later.

v1v:
> what happens if the integration tests for a different agent fail? Does it fail the build for the Java agent as well?

As you said, the ITs pipeline fails and the github check status for the PR of the `apm-agent-java` would be notified with the ❌status and link to the ITs build URL.

v1v:
@felixbarny https://github.com/elastic/apm-agent-java/pull/677/commits/4022c41f84cc135a035f41e3eb871f840877692f is now in place to only trigger the ITs which are related to the `apm-agent-java`

v1v:
PoC is now in place, so a new GitHub check has been added, called `Integrations Tests`:

image

Please let me know what you think

PR #678, opened 2019-06-18
Jms receiveNoWait test

eyalkoren:
Maybe this has something to do with [the way ActiveMQ works relying on prefetch buffers](http://activemq.apache.org/what-is-the-prefetch-limit-for.html) 🤷‍♂ 
If this doesn't work, we should consider giving up the `receiveNoWait` tests for ActiveMQ, as they may be lacking the guarantees we need for unit tests.

### DIFF
--- a/apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/JmsInstrumentationIT.java
+++ b/apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/JmsInstrumentationIT.java
@@ -56,7 +56,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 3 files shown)


apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/JmsInstrumentationIT.java
```diff
@@ -114,25 +113,25 @@ public void endTransaction() throws Exception {
...
         final Queue queue = createQueue();
-        testQueueSendReceiveOnTracedThread(() -> brokerFacade.receive(queue), queue);
+        testQueueSendReceiveOnTracedThread(() -> brokerFacade.receive(queue, 2000), queue, false);
     }
 
     @Test
     public void testQueueSendReceiveNoWaitOnTracedThread() throws Exception {
         final Queue queue = createQueue();
-        testQueueSendReceiveOnTracedThread(() -> loopReceive(() -> brokerFacade.receiveNoWait(queue), 3000), queue);
+        testQueueSendReceiveOnTracedThread(() -> brokerFacade.receiveNoWait(queue), queue, true);
```
felixbarny: why did you get rid of the retry in favor of sleeping once? retrying with a longer timeout seems more robust but should not increase test durations

eyalkoren: Unfortunately it didn't add robustness, it kept on failing. Besides, it meant I would get a non-deterministic number of receive spans/transactions, as I instrument `receiveNoWait`, which I would like to invoke exactly once per test. 
I have a suspicion this has something to do with the way ActiveMQ works by pushing messages to consumers, so I added this voodoo sleep to get the broker a chance to execute internal async stuff, if it has such, before creating the consumer. One last attempt would be to add a sleep between consumer creation and invocation of `receive`.
If those solve the issue, I am more than fine with adding 200-400 ms to our tests. Otherwise, we will just give up this test in ActiveMQ. We will still test the instrumentation of `receiveNoWait` in Artemis. So far I don't recall failures there.

PR #684, opened 2019-06-21
Fix rendering of bootdelegation examples

felixbarny:
context: https://discuss.elastic.co/t/monitoring-jboss-fuse-using-apm/185758/8

Before:
Screen Shot 2019-06-21 at 10 02 54

After:
Screen Shot 2019-07-15 at 13 46 38

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -341,13 +341,17 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
             "If set with an empty string, nothing will be appended to the boot delegation system property.\n" +
             "Values to set in known environments:\n\n" +
             "Nexus:\n\n" +
-            "`boot_delegation_packages=com.sun.*, javax.transaction, javax.transaction.*, javax.xml.crypto, javax.xml.crypto.*, sun.*," +
-            "co.elastic.apm.agent.*`\n\n" +
-            "Pentaho:\n\n" +
-            "`boot_delegation_packages=org.apache.karaf.jaas.boot, org.apache.karaf.jaas.boot.principal, org.apache.karaf.management.boot, " +
+            "----\n" +
+            "boot_delegation_packages=com.sun.*, javax.transaction, javax.transaction.*, javax.xml.crypto, javax.xml.crypto.*, sun.*," +
+            "co.elastic.apm.agent.*\n" +
+            "----\n\n" +
+            "Pentaho and RedHat JBoss Fuse:\n\n" +
+            "----\n" +
+            "boot_delegation_packages=org.apache.karaf.jaas.boot, org.apache.karaf.jaas.boot.principal, org.apache.karaf.management.boot, " +
             "sun.*, com.sun.*, javax.transaction, javax.transaction.*, javax.xml.crypto, javax.xml.crypto.*, org.apache.xerces.jaxp.datatype, " +
             "org.apache.xerces.stax, org.apache.xerces.parsers, org.apache.xerces.jaxp, org.apache.xerces.jaxp.validation, " +
-            "org.apache.xerces.dom, co.elastic.apm.agent.*`")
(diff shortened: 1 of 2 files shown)


eyalkoren:
Ahh, asciidoctor treats the asterisks as bold...
Problem is that my original intention with this config description was that it would be transferred into the `elasticapm.properties` file reference in such a way that it is only required to uncomment in order to get to work. This will have the backtick in the commented area. I find it very common to use property file with ready-made configurations to just uncomment.
Is there a way to define different texts for the rendered description and the property file description?

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -340,10 +340,14 @@
...
             "If set with an empty string, nothing will be appended to the boot delegation system property.\n" +
             "Values to set in known environments:\n\n" +
-            "Nexus:\n" +
-            "boot_delegation_packages=com.sun.*,javax.transaction,javax.transaction.*,javax.xml.crypto,javax.xml.crypto.*,sun.*,co.elastic.apm.agent.*\n\n" +
-            "Pentaho:\n" +
-            "boot_delegation_packages=org.apache.karaf.jaas.boot,org.apache.karaf.jaas.boot.principal,org.apache.karaf.management.boot,sun.*,com.sun.*,javax.transaction,javax.transaction.*,javax.xml.crypto,javax.xml.crypto.*,org.apache.xerces.jaxp.datatype,org.apache.xerces.stax,org.apache.xerces.parsers,org.apache.xerces.jaxp,org.apache.xerces.jaxp.validation,org.apache.xerces.dom,co.elastic.apm.agent.*")
+            "Nexus:\n\n" +
+            "`boot_delegation_packages=com.sun.*, javax.transaction, javax.transaction.*, javax.xml.crypto, javax.xml.crypto.*, sun.*," +
+            "co.elastic.apm.agent.*`\n\n" +
+            "Pentaho:\n\n" +
```
eyalkoren:
```suggestion
            "Pentaho and RedHat JBoss Fuse:\n\n" +
```

eyalkoren:
Made a PR with suggestion fix for the property file comments - https://github.com/felixbarny/apm-agent-java/pull/2

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -340,10 +340,18 @@
         .description("A comma-separated list of packages to be appended to the boot delegation system property. \n" +
             "If set with an empty string, nothing will be appended to the boot delegation system property.\n" +
             "Values to set in known environments:\n\n" +
-            "Nexus:\n" +
-            "boot_delegation_packages=com.sun.*,javax.transaction,javax.transaction.*,javax.xml.crypto,javax.xml.crypto.*,sun.*,co.elastic.apm.agent.*\n\n" +
-            "Pentaho:\n" +
-            "boot_delegation_packages=org.apache.karaf.jaas.boot,org.apache.karaf.jaas.boot.principal,org.apache.karaf.management.boot,sun.*,com.sun.*,javax.transaction,javax.transaction.*,javax.xml.crypto,javax.xml.crypto.*,org.apache.xerces.jaxp.datatype,org.apache.xerces.stax,org.apache.xerces.parsers,org.apache.xerces.jaxp,org.apache.xerces.jaxp.validation,org.apache.xerces.dom,co.elastic.apm.agent.*")
+            "Nexus:\n\n" +
+            "----\n" +
```
eyalkoren: How does this fix rendering when transforming to HTML? Is this some kind of escape?

felixbarny: Yes, that's a [listing block](https://www.methods.co.nz/asciidoc/chunked/ch16.html). I've updated the screenshot to reflect the changes.

eyalkoren: Ahh, nice!

Issue #674, opened 2019-06-17
License maven plugin fails on build


theobisproject:
**Describe the bug**
I'm currently working on adding support for hibernate search to the apm plugin. While doing this I noticed the license plugin throws this error while building:
`Failed to execute goal org.codehaus.mojo:license-maven-plugin:1.19:update-file-header (first) on project apm-agent-parent: could not obtain the license repository: no licenses.properties found with url [file://<path-to-project>/apm-agent-java/licenses/licenses.properties] for resolver org.codehaus.mojo.license.model.LicenseRepository@626d2016 -> [Help 1]`

**To Reproduce**
1. Run `mvnw clean install` on current master

I'm using OpenJDK 11 on Windows 10 and got the error with the Maven wrapper and also with a local Maven installation.

**Expected behavior**
Build completes successfully and all added source files have the correct license header.

felixbarny:
> I'm currently working on adding support for hibernate search to the apm plugin.

Very cool! Let us know if we can support you. Are you planning to eventually create a PR for that?

Sounds like you have added a new plugin but forgot to add this to the pom.xml:


    <properties>
        <apm-agent-parent.base.dir>${project.basedir}/../..</apm-agent-parent.base.dir>
    </properties>

theobisproject:
> > I'm currently working on adding support for hibernate search to the apm plugin.
> 
> Very cool! Let us know if we can support you. Are you planning to eventually create a PR for that?

I plan to do a PR at the end of the week. Still need to do some real life application testing and cleanup.

I really missed the property you mentioned but the problem still persists. I even can reproduce this problem at the current master branch.

eyalkoren:
@theobisproject Thanks for doing this PR !! 
I assume you read through the [CONTRIBUTING page on GitHub](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md), but in case you didn't know, there is [A blog post for adding new plugins](https://www.elastic.co/blog/a-cookbook-for-contributing-a-plugin-to-the-elastic-apm-java-agent) as well.

Please verify that updating the pom as suggested above resolved this issue.

felixbarny:
> I really missed the property you mentioned but the problem still persists. I even can reproduce this problem at the current master branch.

Hmm, maybe it's a Windows-specific problem 🤔 
Could you post the exact error message when executing `mvnw clean package -DskipTests=true` from a fresh master?

theobisproject:
[...]

PR #685, opened 2019-06-22
Add hibernate search support

theobisproject:
Adds support for queries made by hibernate search. Supported versions are all 5.x versions and I also added support for the current alpha version of hibernate search 6.x.
Since I'm currently unable to execute the build with the license plugin the license header is missing in the files (See #674)

### DIFF
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/pom.xml
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/pom.xml
@@ -37,40 +37,9 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.hibernate</groupId>
-            <artifactId>hibernate-entitymanager</artifactId>
-            <version>5.4.1.Final</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
             <version>${version.h2}</version>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>javax.xml.bind</groupId>
-            <artifactId>jaxb-api</artifactId>
-            <version>2.2.11</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>com.sun.xml.bind</groupId>
-            <artifactId>jaxb-core</artifactId>
-            <version>2.2.11</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>com.sun.xml.bind</groupId>
-            <artifactId>jaxb-impl</artifactId>
-            <version>2.2.11</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>javax.activation</groupId>
-            <artifactId>activation</artifactId>
-            <version>1.1.1</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/pom.xml
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>apm-hibernate-search-plugin</artifactId>
         <groupId>co.elastic.apm</groupId>
-        <version>1.7.1-SNAPSHOT</version>
+        <version>1.8.1-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/pom.xml
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/pom.xml
@@ -1,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>apm-hibernate-search-plugin</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.7.1-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>apm-hibernate-search-plugin-5_x</artifactId>
+    <name>${project.groupId}:${project.artifactId}</name>
+
+    <properties>
+        <apm-agent-parent.base.dir>${project.basedir}/../../..</apm-agent-parent.base.dir>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>apm-hibernate-search-plugin-common</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-search-orm</artifactId>
+            <version>5.11.1.Final</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>apm-hibernate-search-plugin-common</artifactId>
+            <version>${project.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-entitymanager</artifactId>
+            <version>5.4.1.Final</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
+            <version>${version.h2}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+            <version>2.2.11</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.sun.xml.bind</groupId>
+            <artifactId>jaxb-core</artifactId>
+            <version>2.2.11</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.sun.xml.bind</groupId>
+            <artifactId>jaxb-impl</artifactId>
+            <version>2.2.11</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>javax.activation</groupId>
+            <artifactId>activation</artifactId>
+            <version>1.1.1</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+</project>
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
@@ -32,6 +32,7 @@ import javax.annotation.Nullable;
 import co.elastic.apm.agent.bci.ElasticApmInstrumentation;
 import co.elastic.apm.agent.bci.VisibleForAdvice;
 import co.elastic.apm.agent.hibernate.search.HibernateSearchConstants;
+import co.elastic.apm.agent.hibernate.search.HibernateSearchHelper;
 import co.elastic.apm.agent.impl.transaction.Span;
 import co.elastic.apm.agent.impl.transaction.TraceContextHolder;
 import net.bytebuddy.asm.Advice;
@@ -90,16 +91,7 @@ public class HibernateSearch5Instrumentation extends ElasticApmInstrumentation {
                     return;
                 }
 
-                final @Nullable String queryString = query.getQueryString();
-                span = active.createSpan().activate();
-
-                span.withType("db")
-                    .withSubtype(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
-                    .withAction("request");
-                span.getContext().getDb()
-                    .withType(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
-                    .withStatement(queryString);
-                span.setName(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_SPAN_NAME);
+                span = HibernateSearchHelper.createAndActivateSpan(active, query.getQueryString());
             }
         }
 
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
@@ -1,3 +1,27 @@
+/*-
+ * #%L
+ * Elastic APM Java agent
+ * %%
+ * Copyright (C) 2018 - 2019 Elastic and contributors
+ * %%
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * #L%
+ */
 package co.elastic.apm.agent.hibernate.search.v5_x;
 
 import java.util.Collection;
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
@@ -88,8 +88,9 @@ public class HibernateSearch5Instrumentation extends ElasticApmInstrumentation {
 
         @Advice.OnMethodEnter(suppress = Throwable.class)
         private static void onBeforeExecute(@Advice.This FullTextQueryImpl query,
-                                            @Advice.Local("span") Span span) {
-            span = HibernateSearchHelper.createAndActivateSpan(tracer, query.getQueryString());
+                                            @Advice.Local("span") Span span,
+                                            @Advice.Origin("#m") String methodName) {
+            span = HibernateSearchHelper.createAndActivateSpan(tracer, methodName, query.getQueryString());
         }
 
         @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
@@ -32,7 +32,7 @@ public class HibernateSearch5Instrumentation extends ElasticApmInstrumentation {
 
     @Override
     public ElementMatcher<? super MethodDescription> getMethodMatcher() {
-        return named("list");
+        return named("list").or(named("scroll")).or(named("iterate"));
     }
 
     @Override
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
@@ -11,9 +11,9 @@
  * the Apache License, Version 2.0 (the "License"); you may
  * not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -42,7 +42,9 @@ import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.matcher.ElementMatcher;
 import org.hibernate.search.query.hibernate.impl.FullTextQueryImpl;
 
+import static co.elastic.apm.agent.bci.bytebuddy.CustomElementMatchers.classLoaderCanLoadClass;
 import static net.bytebuddy.matcher.ElementMatchers.hasSuperType;
+import static net.bytebuddy.matcher.ElementMatchers.isBootstrapClassLoader;
 import static net.bytebuddy.matcher.ElementMatchers.isInterface;
 import static net.bytebuddy.matcher.ElementMatchers.nameContains;
 import static net.bytebuddy.matcher.ElementMatchers.named;
@@ -58,6 +60,12 @@ public class HibernateSearch5Instrumentation extends ElasticApmInstrumentation {
     @Override
     public ElementMatcher<? super NamedElement> getTypeMatcherPreFilter() {
         return nameContains("Query");
+    }
+
+    @Override
+    public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
+        return not(isBootstrapClassLoader())
+            .and(classLoaderCanLoadClass("org.hibernate.search.FullTextQuery"));
     }
 
     @Override
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
@@ -1,0 +1,82 @@
+package co.elastic.apm.agent.hibernate.search.v5_x;
+
+import java.util.Collection;
+import java.util.Collections;
+
+import javax.annotation.Nullable;
+
+import co.elastic.apm.agent.bci.ElasticApmInstrumentation;
+import co.elastic.apm.agent.bci.VisibleForAdvice;
+import co.elastic.apm.agent.hibernate.search.HibernateSearchConstants;
+import co.elastic.apm.agent.impl.transaction.Span;
+import co.elastic.apm.agent.impl.transaction.TraceContextHolder;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.hibernate.search.query.hibernate.impl.FullTextQueryImpl;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public class HibernateSearch5Instrumentation extends ElasticApmInstrumentation {
+
+    @Override
+    public Class<?> getAdviceClass() {
+        return HibernateSearch5ExecuteAdvice.class;
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return named("org.hibernate.search.query.hibernate.impl.FullTextQueryImpl");
+    }
+
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return named("list");
+    }
+
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Collections.singleton(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE);
+    }
+
+    @VisibleForAdvice
+    public static class HibernateSearch5ExecuteAdvice {
+
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        private static void onBeforeExecute(@Advice.This FullTextQueryImpl query,
+            @Advice.Local("span") Span span) {
+            if (tracer != null) {
+                TraceContextHolder<?> active = tracer.getActive();
+                if (active == null || active instanceof Span && HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE
+                    .equals(((Span) active).getSubtype())) {
+                    return;
+                }
+
+                final @Nullable String queryString = query.getQueryString();
+                span = active.createSpan().activate();
+
+                span.withType("db")
+                    .withSubtype(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
+                    .withAction("request");
+                span.getContext().getDb()
+                    .withType(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
+                    .withStatement(queryString);
+                span.setName(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_SPAN_NAME);
+            }
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
+        public static void onAfterExecute(@Advice.Local("span") @Nullable Span span,
+            @Advice.Thrown Throwable t) {
+            if (span != null) {
+                try {
+                    span.captureException(t);
+                } finally {
+                    span.end();
+                    span.deactivate();
+                }
+            }
+        }
+    }
+}
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
@@ -35,6 +35,7 @@ import co.elastic.apm.agent.hibernate.search.HibernateSearchConstants;
 import co.elastic.apm.agent.impl.transaction.Span;
 import co.elastic.apm.agent.impl.transaction.TraceContextHolder;
 import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.NamedElement;
 import net.bytebuddy.description.method.MethodDescription;
 import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.matcher.ElementMatcher;
@@ -42,6 +43,7 @@ import org.hibernate.search.query.hibernate.impl.FullTextQueryImpl;
 
 import static net.bytebuddy.matcher.ElementMatchers.hasSuperType;
 import static net.bytebuddy.matcher.ElementMatchers.isInterface;
+import static net.bytebuddy.matcher.ElementMatchers.nameContains;
 import static net.bytebuddy.matcher.ElementMatchers.named;
 import static net.bytebuddy.matcher.ElementMatchers.not;
 
@@ -50,6 +52,11 @@ public class HibernateSearch5Instrumentation extends ElasticApmInstrumentation {
     @Override
     public Class<?> getAdviceClass() {
         return HibernateSearch5ExecuteAdvice.class;
+    }
+
+    @Override
+    public ElementMatcher<? super NamedElement> getTypeMatcherPreFilter() {
+        return nameContains("Query");
     }
 
     @Override
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
@@ -16,7 +16,10 @@ import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.matcher.ElementMatcher;
 import org.hibernate.search.query.hibernate.impl.FullTextQueryImpl;
 
+import static net.bytebuddy.matcher.ElementMatchers.hasSuperType;
+import static net.bytebuddy.matcher.ElementMatchers.isInterface;
 import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.not;
 
 public class HibernateSearch5Instrumentation extends ElasticApmInstrumentation {
 
@@ -27,12 +30,15 @@ public class HibernateSearch5Instrumentation extends ElasticApmInstrumentation {
 
     @Override
     public ElementMatcher<? super TypeDescription> getTypeMatcher() {
-        return named("org.hibernate.search.query.hibernate.impl.FullTextQueryImpl");
+        return not(isInterface())
+            .and(hasSuperType(named("org.hibernate.search.FullTextQuery")));
     }
 
     @Override
     public ElementMatcher<? super MethodDescription> getMethodMatcher() {
-        return named("list").or(named("scroll")).or(named("iterate"));
+        return named("list")
+            .or(named("scroll"))
+            .or(named("iterate"));
     }
 
     @Override
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
@@ -30,11 +30,9 @@ import java.util.Collections;
 import javax.annotation.Nullable;
 
 import co.elastic.apm.agent.bci.ElasticApmInstrumentation;
-import co.elastic.apm.agent.bci.VisibleForAdvice;
 import co.elastic.apm.agent.hibernate.search.HibernateSearchConstants;
 import co.elastic.apm.agent.hibernate.search.HibernateSearchHelper;
 import co.elastic.apm.agent.impl.transaction.Span;
-import co.elastic.apm.agent.impl.transaction.TraceContextHolder;
 import net.bytebuddy.asm.Advice;
 import net.bytebuddy.description.NamedElement;
 import net.bytebuddy.description.method.MethodDescription;
@@ -86,21 +84,12 @@ public class HibernateSearch5Instrumentation extends ElasticApmInstrumentation {
         return Collections.singleton(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE);
     }
 
-    @VisibleForAdvice
     public static class HibernateSearch5ExecuteAdvice {
 
         @Advice.OnMethodEnter(suppress = Throwable.class)
         private static void onBeforeExecute(@Advice.This FullTextQueryImpl query,
-            @Advice.Local("span") Span span) {
-            if (tracer != null) {
-                TraceContextHolder<?> active = tracer.getActive();
-                if (active == null || active instanceof Span && HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE
-                    .equals(((Span) active).getSubtype())) {
-                    return;
-                }
-
-                span = HibernateSearchHelper.createAndActivateSpan(active, query.getQueryString());
-            }
+                                            @Advice.Local("span") Span span) {
+            span = HibernateSearchHelper.createAndActivateSpan(tracer, query.getQueryString());
         }
 
         @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/test/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5InstrumentationTest.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/test/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5InstrumentationTest.java
@@ -1,0 +1,112 @@
+package co.elastic.apm.agent.hibernate.search.v5_x;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+import javax.persistence.EntityManager;
+import javax.persistence.EntityManagerFactory;
+
+import co.elastic.apm.agent.AbstractInstrumentationTest;
+import co.elastic.apm.agent.hibernate.search.DeleteFileVisitor;
+import co.elastic.apm.agent.impl.transaction.TraceContext;
+import org.apache.lucene.search.Query;
+import org.hibernate.search.jpa.FullTextEntityManager;
+import org.hibernate.search.jpa.FullTextQuery;
+import org.hibernate.search.jpa.Search;
+import org.hibernate.search.query.dsl.BooleanJunction;
+import org.hibernate.search.query.dsl.QueryBuilder;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertAll;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class HibernateSearch5InstrumentationTest extends AbstractInstrumentationTest {
+
+    private Path tempDirectory;
+
+    private EntityManagerFactory entityManagerFactory;
+
+    private EntityManager entityManager;
+
+    @BeforeEach
+    void setUp() throws IOException {
+        tempDirectory = Files.createTempDirectory("HibernateSearch5InstrumentationTest");
+        entityManagerFactory = EntityManagerFactoryHelper.buildEntityManagerFactory(tempDirectory);
+        entityManager = entityManagerFactory.createEntityManager();
+        tracer.startTransaction(TraceContext.asRoot(), null, null).activate();
+    }
+
+    @AfterEach
+    void tearDown() throws IOException {
+        tracer.currentTransaction().deactivate().end();
+        entityManager.close();
+        entityManagerFactory.close();
+        Files.walkFileTree(tempDirectory, new DeleteFileVisitor());
+    }
+
+    @Test
+    public void performLuceneIndexSearch() {
+        saveDogsToIndex();
+
+        FullTextEntityManager fullTextSession = Search.getFullTextEntityManager(entityManager);
+        QueryBuilder builder = fullTextSession.getSearchFactory().buildQueryBuilder().forEntity(Dog.class).get();
+
+        BooleanJunction<BooleanJunction> bool = builder.bool();
+        bool.should(builder.keyword().onField("name").matching("dog1").createQuery());
+
+        Query query = bool.createQuery();
+
+        FullTextQuery ftq = fullTextSession.createFullTextQuery(query, Dog.class);
+
+        List<Dog> result = (List<Dog>) ftq.getResultList();
+
+        assertAll(() -> {
+            assertEquals(1, result.size(), "Query result is not 1");
+            assertEquals("dog1", result.get(0).getName(), "Result is not 'dog1'");
+            assertEquals(1, reporter.getSpans().size(), "Didn't find 1 span");
+            assertEquals("hibernate-search", reporter.getFirstSpan().getSubtype(),
+                "Subtype of span is not 'hibernate-search'");
+            assertEquals("name:dog1", reporter.getFirstSpan().getContext().getDb().getStatement(),
+                "Statement is not 'name:dog1'");
+        });
+    }
+
+    @Test
+    public void performMutiResultLuceneIndexSearch() {
+        saveDogsToIndex();
+
+        FullTextEntityManager fullTextSession = Search.getFullTextEntityManager(entityManager);
+        QueryBuilder builder = fullTextSession.getSearchFactory().buildQueryBuilder().forEntity(Dog.class).get();
+
+        BooleanJunction<BooleanJunction> bool = builder.bool();
+        bool.should(builder.keyword().wildcard().onField("name").matching("dog*").createQuery());
+
+        Query query = bool.createQuery();
+
+        FullTextQuery ftq = fullTextSession.createFullTextQuery(query, Dog.class);
+
+        List<Dog> result = (List<Dog>) ftq.getResultList();
+
+        assertAll(() -> {
+            assertEquals(2, result.size(), "Query result is not 2");
+            assertEquals(1, reporter.getSpans().size(), "Didn't find 1 span");
+            assertEquals("hibernate-search", reporter.getFirstSpan().getSubtype(),
+                "Subtype of span is not 'hibernate-search'");
+            assertEquals("name:dog*", reporter.getFirstSpan().getContext().getDb().getStatement(),
+                "Statement is not 'name:dog*'");
+        });
+    }
+
+    private void saveDogsToIndex() {
+        entityManager.getTransaction().begin();
+        Dog dog1 = new Dog("dog1");
+        Dog dog2 = new Dog("dog2");
+        entityManager.persist(dog1);
+        entityManager.persist(dog2);
+        entityManager.getTransaction().commit();
+    }
+}
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/test/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5InstrumentationTest.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/test/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5InstrumentationTest.java
@@ -158,6 +158,18 @@ class HibernateSearch5InstrumentationTest extends AbstractInstrumentationTest {
         });
     }
 
+    @Test
+    void performUniqueResultIndexSearch() {
+        Object uniqueResult = createNonJpaFullTextQuery(createQueryForDog1()).uniqueResult();
+
+        assertThat(uniqueResult).isNotNull();
+        final Dog dog = (Dog) uniqueResult;
+        assertAll(() -> {
+            assertThat(dog.getName()).isEqualTo("dog1");
+            assertApmSpanInformation(reporter, "name:dog1", "list");
+        });
+    }
+
     private Query createQueryForDog1() {
         BooleanQuery.Builder builder = new BooleanQuery.Builder();
         builder.add(new BooleanClause(new TermQuery(new Term("name", "dog1")), BooleanClause.Occur.SHOULD));
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/test/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5InstrumentationTest.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/test/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5InstrumentationTest.java
@@ -55,10 +55,8 @@ import java.util.Iterator;
 import java.util.List;
 
 import static co.elastic.apm.agent.hibernate.search.HibernateSearchAssertionHelper.assertApmSpanInformation;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertAll;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 class HibernateSearch5InstrumentationTest extends AbstractInstrumentationTest {
 
@@ -102,8 +100,8 @@ class HibernateSearch5InstrumentationTest extends AbstractInstrumentationTest {
         List<Dog> result = (List<Dog>) ftq.getResultList();
 
         assertAll(() -> {
-            assertEquals(1, result.size(), "Query result is not 1");
-            assertEquals("dog1", result.get(0).getName(), "Result is not 'dog1'");
+            assertThat(result.size()).isEqualTo(1);
+            assertThat(result.get(0).getName()).isEqualTo("dog1");
             assertApmSpanInformation(reporter, "name:dog1", "list");
         });
     }
@@ -123,7 +121,7 @@ class HibernateSearch5InstrumentationTest extends AbstractInstrumentationTest {
         List<Dog> result = (List<Dog>) ftq.getResultList();
 
         assertAll(() -> {
-            assertEquals(2, result.size(), "Query result is not 2");
+            assertThat(result.size()).isEqualTo(2);
             assertApmSpanInformation(reporter, "name:dog*", "list");
         });
     }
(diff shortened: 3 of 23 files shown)


apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return named("org.hibernate.search.query.hibernate.impl.FullTextQueryImpl");
+    }
+
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return named("list");
```
eyalkoren: What about other `org.hibernate.Query` relevant functions like `scroll` and `iterate`?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
```diff
@@ -0,0 +1,82 @@
...
+                TraceContextHolder<?> active = tracer.getActive();
+                if (active == null || active instanceof Span && HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE
+                    .equals(((Span) active).getSubtype())) {
+                    return;
+                }
+
+                final @Nullable String queryString = query.getQueryString();
+                span = active.createSpan().activate();
+
+                span.withType("db")
```
eyalkoren: I am not sure about this being a `db` type. If it is, maybe we should consider more basic hibernate stuff? Need to think a bit more about this 🤔

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return named("org.hibernate.search.backend.lucene.search.query.impl.LuceneSearchQueryImpl");
+    }
+
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return named("fetch");
```
eyalkoren: What about other query methods in `org.hibernate.search.engine.search.query.SearchFetchable`?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
```diff
@@ -0,0 +1,82 @@
...
+public class HibernateSearch6Instrumentation extends ElasticApmInstrumentation {
+
+    @Override
+    public Class<?> getAdviceClass() {
+        return Hibernate6ExecuteAdvice.class;
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return named("org.hibernate.search.backend.lucene.search.query.impl.LuceneSearchQueryImpl");
```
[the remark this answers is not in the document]
theobisproject: That's a good point. Maybe I was a bit too focused on Lucene 😄

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
theobisproject: We don't use those methods in our application because they are not available when you have a JPA FullTextQuery. But none the less a valid addition. I only wonder if the time can be correctly tracked because both methods only load the result when requested. Was this already done in any other plugin already.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
theobisproject: I also wasn't really sure about this 😉 The `db` type was the closest type already used by other plugins I found.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
theobisproject: That depends how far the support of the 6.x version should go since it still is in alpha and there could be still some changes made.

theobisproject:
Regarding a shared API between version 5 and 6. The hibernate team made a big rewrite of the Search DSL. So there is nothing shared we could instrument.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
eyalkoren: > I only wonder if the time can be correctly tracked because both methods only load the result when requested. Was this already done in any other plugin already.

Well, I believe this is the case with JDBC's ResultSet, which we do not instrument either (but only the query execution). We miss info about actual reads this way, but the alternative is having a LOT of spans traced, which are largely non-informative.

I assume these methods' duration account for the actual search action (without the results being fetched), as they require ranked results, so I think they should be informative. What do you think?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
eyalkoren: We _DO_ use `db` as type for the Elasticsearch client, so we can stick to it right now, but may reconsider later.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
eyalkoren: Is there a difference in expected stability between the `fetch` methods and the others? Is there a reason why they are preferable?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
theobisproject: I will check the JDBC implementation and will check if the tacing result is informative for the two methods.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
theobisproject: As I'm not involved in the development I personally would expect at the current state that everything could be changed. The `fetch` method is just the equivalent of the `list` method I traced for Hibernate 5. So no reason to prefer it over the other methods when adding tracing support.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
theobisproject: Your assumption was correct. The search happens in the `scroll` or `iterate` method and only the results are lazy loaded when requested.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
theobisproject: I now trace all methods from the `SearchFetchable` interface.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-entitymanager</artifactId>
+            <version>5.4.1.Final</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
```
eyalkoren: So this uses H2??

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/test/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5InstrumentationTest.java
```diff
@@ -0,0 +1,206 @@
...
+    void tearDown() throws IOException {
+        tracer.currentTransaction().deactivate().end();
+        entityManager.close();
+        entityManagerFactory.close();
+        Files.walkFileTree(tempDirectory, new DeleteFileVisitor());
+    }
+
+    @Test
+    public void performLuceneIndexSearch() {
+        saveDogsToIndex();
```
eyalkoren: Can this be done as a `@BeforeAll`, ie- once before all tests?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/test/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5InstrumentationTest.java
```diff
@@ -0,0 +1,206 @@
...
+        QueryBuilder builder = fullTextSession.getSearchFactory().buildQueryBuilder().forEntity(Dog.class).get();
+
+        BooleanJunction<BooleanJunction> bool = builder.bool();
+        bool.should(builder.keyword().onField("name").matching("dog1").createQuery());
+
+        Query query = bool.createQuery();
+
+        FullTextQuery ftq = fullTextSession.createFullTextQuery(query, Dog.class);
+
+        List<Dog> result = (List<Dog>) ftq.getResultList();
```
eyalkoren: Looks like lots of common stuff between tests. Consider adding stuff to `@BeforeEach` and `@BeforeAll` as much as possible

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/test/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5InstrumentationTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+        assertAll(() -> {
+            assertEquals(1, result.size(), "Query result is not 1");
+            assertEquals("dog1", result.get(0).getName(), "Result is not 'dog1'");
+            assertEquals(1, reporter.getSpans().size(), "Didn't find 1 span");
+            assertEquals("hibernate-search", reporter.getFirstSpan().getSubtype(),
+                "Subtype of span is not 'hibernate-search'");
+            assertEquals("name:dog1", reporter.getFirstSpan().getContext().getDb().getStatement(),
+                "Statement is not 'name:dog1'");
```
eyalkoren: Can you extract a common `validate` method for most of this?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/test/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6InstrumentationTest.java
```diff
@@ -0,0 +1,137 @@
...
+    void tearDown() throws IOException {
+        tracer.currentTransaction().deactivate().end();
+        entityManager.close();
+        entityManagerFactory.close();
+        Files.walkFileTree(tempDirectory, new DeleteFileVisitor());
+    }
+
+    @Test
+    public void performLuceneIndexSearch() {
+        saveDogsToIndex();
```
eyalkoren: Can probably fit `@BeforeAll` or `@BeforeEach`

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/test/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6InstrumentationTest.java
```diff
@@ -0,0 +1,137 @@
...
+            .fetchHits();
+
+        assertAll(() -> {
+            assertEquals(1, result.size(), "Query result is not 1");
+            assertEquals("dog1", result.get(0).getName(), "Result is not 'dog1'");
+            assertEquals(1, reporter.getSpans().size(), "Didn't find 1 span");
+            assertEquals("hibernate-search", reporter.getFirstSpan().getSubtype(),
+                "Subtype of span is not 'hibernate-search'");
+            assertEquals("+name:dog1 #__HSEARCH_type:main", reporter.getFirstSpan().getContext().getDb().getStatement(),
+                "Statement is not '+name:dog1 #__HSEARCH_type:main'");
```
eyalkoren: Try to reuse common code as much as possible, if possible

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/test/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6InstrumentationTest.java
```diff
@@ -0,0 +1,137 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertAll;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class HibernateSearch6InstrumentationTest extends AbstractInstrumentationTest {
+
```
eyalkoren: The more different `fetch` methods you can make invoked in these tests- the better

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
```diff
@@ -0,0 +1,112 @@
...
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.not;
+
+public class HibernateSearch5Instrumentation extends ElasticApmInstrumentation {
+
+    @Override
+    public Class<?> getAdviceClass() {
+        return HibernateSearch5ExecuteAdvice.class;
+    }
+
```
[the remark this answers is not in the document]
theobisproject: I added a prefilter with name `Query`.
**Before:**
| HibernateSearch5Instrumentation          |   1.377.372.500 |         781.000 |
**With Prefilter:**
| HibernateSearch5Instrumentation          |     117.124.600 |       1.107.700 |

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
```diff
@@ -0,0 +1,110 @@
...
+import static net.bytebuddy.matcher.ElementMatchers.nameStartsWith;
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.not;
+
+public class HibernateSearch6Instrumentation extends ElasticApmInstrumentation {
+
+    @Override
+    public Class<?> getAdviceClass() {
+        return Hibernate6ExecuteAdvice.class;
+    }
```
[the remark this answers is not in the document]
theobisproject: I added a prefilter with name `Search`.
**Before:**
| HibernateSearch6Instrumentation          |   1.512.267.400 |         515.000 |
**With Prefilter:**
| HibernateSearch6Instrumentation          |     109.494.600 |         558.200 |

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return nameStartsWith("fetch");
+    }
+
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Collections.singleton(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE);
```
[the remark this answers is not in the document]
theobisproject: Changed accordingly. I think it would be good to have an enumeration were all available groups are defined. Is there something like this planned?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
```diff
@@ -0,0 +1,110 @@
...
+import co.elastic.apm.agent.bci.ElasticApmInstrumentation;
+import co.elastic.apm.agent.bci.VisibleForAdvice;
+import co.elastic.apm.agent.hibernate.search.HibernateSearchConstants;
+import co.elastic.apm.agent.impl.transaction.Span;
+import co.elastic.apm.agent.impl.transaction.TraceContextHolder;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.hibernate.search.backend.lucene.search.query.impl.LuceneSearchQueryImpl;
```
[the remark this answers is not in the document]
theobisproject: The dependency is still needed for the unit tests but now no longer in the plugin

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-search-orm</artifactId>
+            <version>5.11.1.Final</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>apm-hibernate-search-plugin-common</artifactId>
```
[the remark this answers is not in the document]
theobisproject: Are you referring here to a specific dependency?

The jaxb dependencies are only neccessary for JDK >= 11 because of [JEP 320](https://openjdk.java.net/jeps/320)

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/pom.xml
theobisproject: Yes, the unit tests use H2 as database. Hibernate Search intercepts the entity lifecycle and updates the lucene/elasticsearch index when something in the database changes. The database connection is defined in the `META-INF/persistence.xml` file.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+
+        <dependency>
+            <groupId>org.hibernate.search</groupId>
+            <artifactId>hibernate-search-mapper-orm</artifactId>
+            <version>${version.hibernate.search}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate.search</groupId>
+            <artifactId>hibernate-search-backend-lucene</artifactId>
```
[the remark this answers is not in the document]
theobisproject: Moved dependency to test dependencies

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
```diff
@@ -0,0 +1,110 @@
...
+                final @Nullable String queryString = query.getQueryString();
+                span = active.createSpan().activate();
+
+                span.withType("db")
+                    .withSubtype(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
+                    .withAction("request");
+                span.getContext().getDb()
+                    .withType(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
+                    .withStatement(queryString);
+                span.setName(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_SPAN_NAME);
```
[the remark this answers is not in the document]
theobisproject: Refactored the span creation to helper class

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/test/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5InstrumentationTest.java
theobisproject: I refactored the whole test class and addressed all your comments

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/test/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6InstrumentationTest.java
theobisproject: I added a test for every method in the interface

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
```diff
@@ -0,0 +1,112 @@
...
+                final @Nullable String queryString = query.getQueryString();
+                span = active.createSpan().activate();
+
+                span.withType("db")
+                    .withSubtype(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
+                    .withAction("request");
+                span.getContext().getDb()
+                    .withType(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
+                    .withStatement(queryString);
+                span.setName(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_SPAN_NAME);
```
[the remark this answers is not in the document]
theobisproject: How is the correct way to get the method name? I tried `@SimpleMethodSignature` but only got the `FullTextQueryImpl#list` as signature ...

theobisproject:
Since I can't answer to your comment with the `JdbcHelperImpl#isAlreadyMonitored` I add a comment.
I personally don't think the information of the JDBC calls is relevant in this case because hibernate search simply loads all objects which are in the result from the database. This could add a lot of spans which provide nearly zero value.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
```diff
@@ -0,0 +1,111 @@
...
+            .or(named("scroll"))
+            .or(named("iterate"));
+    }
+
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Collections.singleton(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE);
+    }
+
+    @VisibleForAdvice
```
eyalkoren: No need for this annotation here

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
```diff
@@ -0,0 +1,109 @@
...
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return nameStartsWith("fetch");
+    }
+
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Arrays.asList(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE, "incubating");
+    }
+
+    @VisibleForAdvice
```
eyalkoren: No need for this annotation here

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
```diff
@@ -0,0 +1,49 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ * #L%
+ */
+package co.elastic.apm.agent.hibernate.search;
+
+import co.elastic.apm.agent.impl.transaction.Span;
+import co.elastic.apm.agent.impl.transaction.TraceContextHolder;
+
+public final class HibernateSearchHelper {
```
eyalkoren:
```suggestion
@VisibleForAdvice
public final class HibernateSearchHelper {
```

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private HibernateSearchHelper() {
+
+    }
+
+    public static Span createAndActivateSpan(final TraceContextHolder traceContextHolder, final String query) {
```
eyalkoren:
```suggestion
    @VisibleForAdvice
    public static Span createAndActivateSpan(final TraceContextHolder traceContextHolder, final String query) {
```

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
```diff
@@ -0,0 +1,111 @@
...
+    public static class HibernateSearch5ExecuteAdvice {
+
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        private static void onBeforeExecute(@Advice.This FullTextQueryImpl query,
```
eyalkoren: Regarding the question in https://github.com/elastic/apm-agent-java/pull/685#discussion_r298583471- the way to get the method name is by using the `@Advice.Origin` annotation, two ways to get the method name: the explicit way:
```
@Advice.Origin Method method
``` 
which gives you a `java.lang.reflect.Method` from which you can get the name through `getName`, and the utility shortcut:
```
@Advice.Origin("#m") String methodName
```

But maybe the `@SimpleMethodSignature` is actually better? What would you prefer as a user?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/test/java/co/elastic/apm/agent/hibernate/search/HibernateSearchAssertionHelper.java
```diff
@@ -0,0 +1,44 @@
...
+
+public final class HibernateSearchAssertionHelper {
+
+    private HibernateSearchAssertionHelper() {
+
+    }
+
+    public static void assertApmSpanInformation(final MockReporter reporter, final String expectedQuery) {
+        assertEquals(1, reporter.getSpans().size(), "Didn't find 1 span");
+        assertEquals("hibernate-search", reporter.getFirstSpan().getSubtype(),
```
eyalkoren:
```suggestion
        Span span = reporter.getFirstSpan();
        assertEquals("hibernate-search", span.getSubtype(),
```

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
```diff
@@ -0,0 +1,49 @@
...
+        Span span = traceContextHolder.createSpan().activate();
+
+        span.withType("db")
+            .withSubtype(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
+            .withAction("request");
```
eyalkoren: Why `request`? 
I see a few better options: 
1. Use your current subtype and the instrumented method name as the action.
2. Use `hibernate` as subtype and `search` as action.
3. If you can tell the underlying search engine (eg Lucene or Elasticsearch) - we can consider this info as well...
4. Use `search` as type and `hibernate` or underlying data store as subtype, method  name as action.

For now, let's go with option 2, ie `db.hibernate.search` and see what my colleague thinks about this. @felixbarny  WDYT?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/test/java/co/elastic/apm/agent/hibernate/search/HibernateSearchAssertionHelper.java
```diff
@@ -0,0 +1,44 @@
...
+            "Subtype of span is not 'hibernate-search'");
+        assertEquals(expectedQuery, reporter.getFirstSpan().getContext().getDb().getStatement(),
+            "Statement is not '" + expectedQuery + "'");
```
eyalkoren: validate the span type, action and name as well.

eyalkoren:
Regarding your comment https://github.com/elastic/apm-agent-java/pull/685#issuecomment-506725801 - I think you are better to judge. If you don't think this will add value for pinpointing latency issues and errors- no need to change. @felixbarny WDYT?

eyalkoren:
One more thing- after working so hard on supporting that, you should add it to the `supported-technologies.asciidoc` page 😄

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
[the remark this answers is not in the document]
theobisproject: The implementation class for the interface is called `FullTextQueryImpl`. So we will not miss something.

I have now overridden the `getClassLoaderMethod`.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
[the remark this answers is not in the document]
theobisproject: The implementation for Lucene is called `LuceneSearchQueryImpl` and the Elasticsearch is called `ElasticsearchSearchQueryImpl`.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/pom.xml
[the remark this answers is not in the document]
theobisproject: I rechecked the dependencies and removed those who are no longer needed.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
theobisproject: I would be interested in the names of my own methods. So speaking for the unit tests I would like to get the class and method name of the unit test class.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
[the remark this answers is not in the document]
eyalkoren: If there is only one implementation possible for this interface, no need for the name pre-matcher at all, and `getTypeMatcher` should be implemented as follows:
```
return named(org.hibernate.search.query.hibernate.impl.FullTextQueryImpl);
```

But are we looking at the right interface, or should we look at `org.hibernate.query.Query`?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
```diff
@@ -0,0 +1,119 @@
...
+    @Override
+    public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
+        return not(isBootstrapClassLoader())
+            .and(classLoaderCanLoadClass("org.hibernate.search.FullTextQuery"));
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return not(isInterface())
+            .and(hasSuperType(named("org.hibernate.search.FullTextQuery")));
```
eyalkoren: Enough, or do we need to look for implementations of `org.hibernate.query.Query`?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
```diff
@@ -0,0 +1,111 @@
...
+
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        private static void onBeforeExecute(@Advice.This FullTextQueryImpl query,
+            @Advice.Local("span") Span span) {
+            if (tracer != null) {
+                TraceContextHolder<?> active = tracer.getActive();
+                if (active == null || active instanceof Span && HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE
+                    .equals(((Span) active).getSubtype())) {
+                    return;
+                }
```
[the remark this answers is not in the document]
theobisproject: Moved the functionality to the helper. I sticked with the static method because the advice class is currently static.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
```diff
@@ -0,0 +1,109 @@
...
+
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        private static void onBeforeExecute(@Advice.This SearchQuery query,
+            @Advice.Local("span") Span span) {
+            if (tracer != null) {
+                TraceContextHolder<?> active = tracer.getActive();
+                if (active == null || active instanceof Span && HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE
+                    .equals(((Span) active).getSubtype())) {
+                    return;
+                }
```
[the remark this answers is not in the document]
theobisproject: Helper was adjusted.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
theobisproject: Removed

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
theobisproject: Removed

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
theobisproject: Added in own commit.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
theobisproject: Added in own commit.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/test/java/co/elastic/apm/agent/hibernate/search/HibernateSearchAssertionHelper.java
theobisproject: Added as own commit.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/test/java/co/elastic/apm/agent/hibernate/search/HibernateSearchAssertionHelper.java
theobisproject: Added the validation for the suggested information.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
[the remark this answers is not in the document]
theobisproject: In the latest version there is only one implementation left. As I remember there were at least 2 implementations in older 5.x versions which have the same class name but resided in different packages. Could check that again if we like to only match against those two.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
theobisproject: That is enough for the hibernate search functionality. The query interface would be interesting if someone would add support for some hibernate specific jdbc/jpa stuff.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
theobisproject: For now added span information from option 2.

I think option 3 would be possible with version 6. I know which provider my applicatoin is using so this would be no information I would need. But could be different in other scenarios.

theobisproject:
@eyalkoren Now everything should be addressed. I would appreciate if you could resolve the discussions which are resolved. Took me some time to figure out what still needs to be done and what I already fixed.
Will add the supported technology page later.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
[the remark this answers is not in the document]
eyalkoren: Well, it depends on what we say we support 🙂 
If we say we support 5.x, we need to know we do indeed. Otherwise, we will say we support 5.11.1. Obviously the former is better

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
eyalkoren: Gotcha. Then let's leave it out. JPA adds a lot of noise on top of JDBC which we support already.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
eyalkoren: This means going over the stack and search for those, not something we normally do. The way for a user to relate own methods to an instrumented framework is through the stack trace we capture for spans. If you set `application_packages` you would see which of your methods invoked the traced method.

For now, let's go with the method name appended as a postfix to the general prefix as follows: `Hibernate-Search list()` (with a dash in the prefix to separate from the method and with parentheses in the method to indicate this is the instrumented method). See how to get the method name in [this comment](https://github.com/elastic/apm-agent-java/pull/685#discussion_r298821590).

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
```diff
@@ -0,0 +1,62 @@
...
+            if (active == null || active instanceof Span && HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE
+                .equals(((Span) active).getSubtype())) {
+                return null;
+            }
+
+            span = active.createSpan().activate();
+
+            span.withType("db")
+                .withSubtype(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
+                .withAction("request");
```
eyalkoren:
```suggestion
                .withAction(HIBERNATE_SEARCH_ORM_ACTION);
```

We'll go with `db` for type, `hibernate` for subtype and `search` for action. Agreed?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchConstants.java
```diff
@@ -0,0 +1,33 @@
...
+ * under the License.
+ * #L%
+ */
+package co.elastic.apm.agent.hibernate.search;
+
+public final class HibernateSearchConstants {
+
+    public static final String HIBERNATE_SEARCH_ORM_TYPE = "hibernate-search";
+
+    public static final String HIBERNATE_SEARCH_ORM_SPAN_NAME = "Hibernate Search";
```
eyalkoren:
```suggestion
    public static final String HIBERNATE_SEARCH_ORM_SPAN_NAME = "Hibernate-Search";
```

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
```diff
@@ -0,0 +1,62 @@
...
+            span.getContext().getDb()
+                .withType(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
+                .withStatement(query);
+            span.setName(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_SPAN_NAME);
```
eyalkoren:
```suggestion
            span.withName(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_SPAN_NAME)
                .appendToName(" ").appendToName(methodName).appendToName("()");
```

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchConstants.java
```diff
@@ -0,0 +1,33 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * #L%
+ */
+package co.elastic.apm.agent.hibernate.search;
+
+public final class HibernateSearchConstants {
+
+    public static final String HIBERNATE_SEARCH_ORM_TYPE = "hibernate-search";
```
eyalkoren:
```suggestion
    public static final String HIBERNATE_SEARCH_ORM_TYPE = "hibernate";
    public static final String HIBERNATE_SEARCH_ORM_ACTION = "search";
```

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
theobisproject: Made the change yesterday and forgot to push it 🤦‍♂

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
theobisproject: I added the new name with a helper method to reuse it in the unit test.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
theobisproject: That is fine for me. Changed according to the suggestion.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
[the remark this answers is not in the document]
theobisproject: I checked 5.0.0.Final and there are 2 `FullTextQueryImpl` classes. So I would stay at the current implementation.

docs/supported-technologies.asciidoc
```diff
@@ -152,6 +152,10 @@ Other Servlet 3+ compliant servers will most likely work as well.
 |5.0.2+
 |The agent automatically creates Elasticsearch spans for queries done through the official REST client.
 
+|Hibernate Search
+|5.0.0+
```
eyalkoren:
```suggestion
|5.x (on by default), 6.x (off by default)
```

docs/supported-technologies.asciidoc
```diff
@@ -152,6 +152,10 @@ Other Servlet 3+ compliant servers will most likely work as well.
...
+|The agent automatically creates Hibernate Search spans for queries done through the Hibernate Search API.
```
eyalkoren:
```suggestion
|The agent automatically creates Hibernate Search spans for queries done through the Hibernate Search API.

 NOTE: this feature is marked as incubating for version 6.x, which means it is off by default. In order to enable,
 set the <<config-disable-instrumentations>> config option to an empty string
```

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/test/java/co/elastic/apm/agent/hibernate/search/HibernateSearchAssertionHelper.java
```diff
@@ -0,0 +1,47 @@
...
+public final class HibernateSearchAssertionHelper {
+
+    private HibernateSearchAssertionHelper() {
+
+    }
+
+    public static void assertApmSpanInformation(final MockReporter reporter, final String expectedQuery, final String searchMethod) {
+        assertEquals(1, reporter.getSpans().size(), "Didn't find 1 span");
+        final Span span = reporter.getFirstSpan();
+        assertEquals(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE, span.getSubtype(), "Subtype of span is not 'hibernate-search'");
```
eyalkoren:
```suggestion
        assertEquals(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE, span.getSubtype(), "Subtype of span is not 'hibernate'");
```

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
```diff
@@ -0,0 +1,67 @@
...
+
+            span = active.createSpan().activate();
+
+            span.withType("db")
+                .withSubtype(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
+                .withAction(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_ACTION);
+            span.getContext().getDb()
+                .withType(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE)
+                .withStatement(query);
+            span.setName(buildSpanName(methodName));
```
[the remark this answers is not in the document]
theobisproject: Ok, now I see why you made this suggestion.

theobisproject:
Here are screenshots from a search transaction
Hibernate Search Span
Hibernate Search Span Details

Thanks for your review input!

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @VisibleForAdvice
+    public static Span createAndActivateSpan(final ElasticApmTracer tracer, final String methodName,
+                                             final String query) {
+        Span span = null;
+
+        if (tracer != null) {
+            TraceContextHolder<?> active = tracer.getActive();
+            if (active == null || active instanceof Span && HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE
```
felixbarny: Add a comment what's the intent of this. If, and only if, my assumption is correct, just apply the suggested change.
```suggestion
            // avoid creating the same span twice for example, when an instrumented API is wrapped
            if (active == null || active instanceof Span && HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE
```

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return not(isInterface())
+            .and(hasSuperType(named("org.hibernate.search.FullTextQuery")));
+    }
+
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return named("list")
```
felixbarny: what about `uniqueResult`?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
```diff
@@ -0,0 +1,64 @@
...
+                .equals(((Span) active).getSubtype())) {
+                return null;
+            }
+
+            span = active.createSpan().activate();
+
+            span.withType("db")
```
felixbarny: We should discuss using more descriptive types here.

For the upcoming breakdown graphs it would use `db.hibernate` as the key which does not seem to informative. I'm not sure if `db` is the best type for this, as the query is not done on the `db` but rather with lucene, right? But we also don't want to be too specific on the underlying technology as there are other options to lucene. Is that correct?

Maybe something like `search.hibernate.list`. The action would be the method signature like `list`, `scroll` or `iterate`.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/test/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5InstrumentationTest.java
```diff
@@ -0,0 +1,183 @@
...
+import javax.persistence.EntityManager;
+import javax.persistence.EntityManagerFactory;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Iterator;
+import java.util.List;
+
+import static co.elastic.apm.agent.hibernate.search.HibernateSearchAssertionHelper.assertApmSpanInformation;
+import static org.junit.jupiter.api.Assertions.assertAll;
```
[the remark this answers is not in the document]
theobisproject: When the project is willing to use AssertJ instead of JUnit I'm fine with it. I didn't yet replaced the assertAll calls because the AssertJ `assertSoftly` would force me to pass the `SoftAssertions` object into the helper method. This feels a bit nasty.
What's your opinion about this?

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/src/main/java/co/elastic/apm/agent/hibernate/search/v5_x/HibernateSearch5Instrumentation.java
theobisproject: This is already covered by the `list` instrumentation because `org.hibernate.query.internal.AbstractProducedQuery` implements `uniqueResult` with a call to `list`.
I added a unit test for this case.

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
theobisproject: Yeah, this is basically one of the hardest things about this PR 😉

1. Query is done in Lucene/Elasticsearch **but** the result is fetched from the database. So it's a hybrid approach.
2. Yes, underlying technology can differ. Currently Lucene or Elasticsearch.

Do you mean `search.hibernate.list` as the whole type? I don't know if putting the methodname there is a good idea for grouping the results.

felixbarny: No, I meant `search` as the type, `hibernate` as the subtype and `list` as the action. @eyalkoren WDYT?

felixbarny: > Query is done in Lucene/Elasticsearch but the result is fetched from the database. So it's a hybrid approach.

Ah, I see. So maybe `type=db` `subtype=hibernate-search` and `action=list`?

felixbarny:
run the tests

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/src/main/java/co/elastic/apm/agent/hibernate/search/HibernateSearchHelper.java
theobisproject: I think this is good. Should the db type in the context then also changed to `hibernate-search`?

felixbarny: I guess it doesn't make sense to have the type and subtype set to the same value. But maybe `type=hibernate-search` `subtype=lucene/elasticsearch` and `action=list/scroll/iterate`? But I think I like `type=db` `subtype=hibernate-search` and `action=list/scroll/iterate` more.

theobisproject: I didn't meant the type of the span with my comment but the type set in line 58 for the context in the span.

felixbarny: Ah, gotcha. I think for the `db.type` `hibernate-serach` makes sense. Question: would there typically be a child span under a `hibernate-search` span for the JDBC query which is sent to the DB?

theobisproject: There are currently no spans on purpose. See [this comment](https://github.com/elastic/apm-agent-java/pull/685#issuecomment-506725801).

theobisproject: I adjusted the types to your suggestion

theobisproject:
@felixbarny @eyalkoren Is there still something I have to do to get this integrated?

felixbarny:
Sorry for the delay in getting this in. We are about to release a new version and were focusing on getting the necessary changes in. As this is quite a big new feature, we want to merge this in after the release. We plan to do releases more frequently so it won't take too long until this will be in an official release.

apm-agent-plugins/apm-hibernate-search-plugin/pom.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>apm-agent-plugins</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.7.1-SNAPSHOT</version>
```
eyalkoren:
```suggestion
        <version>1.8.1-SNAPSHOT</version>
```

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-5_x/pom.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>apm-hibernate-search-plugin</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.7.1-SNAPSHOT</version>
```
eyalkoren:
```suggestion
        <version>1.8.1-SNAPSHOT</version>
```

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/pom.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>apm-hibernate-search-plugin</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.7.1-SNAPSHOT</version>
```
eyalkoren:
```suggestion
        <version>1.8.1-SNAPSHOT</version>
```

apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-common/pom.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>apm-hibernate-search-plugin</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.7.1-SNAPSHOT</version>
```
eyalkoren:
```suggestion
        <version>1.8.1-SNAPSHOT</version>
```

theobisproject:
Ok. I will do a rebase when I have some time for it.

eyalkoren:
Thanks and sorry for the hassle. Version 1.8.0 was a big one and we wanted to make sure it is properly out with the 7.3.0 stack release.

theobisproject:
Branch is rebased to current master. Adjustment for changed apm version and apis are made.

eyalkoren:
Build and all tests passed!! 🎉 
@theobisproject thanks again for your effort! Highly appreciated!

theobisproject:
@eyalkoren @felixbarny Thanks for your help to get this integrated!

PR #688, opened 2019-06-24
Enabling build on Windows

### LINKED ISSUE #674: License maven plugin fails on build

eyalkoren:
Fixes elastic/apm-agent-java#674 

Changed the license file URI prefix from `file://` to `file:///`, which basically is a shortcut for a file on the localhost. In Unix-style paths this may be appended by a fourth slash for the root directory, but effectively the extra slashes should be collapsed.

Full solution would be to use `classpath://` for the license resolver, thus avoid using the
base dir path property in all modules - http://www.mojohaus.org/license-maven-plugin/examples/example-add-license.html, but out of the scope for now...

In addition, I changed the container ID parsing logic to not use file system APIs, as these are only strings and not real file system paths.

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/SystemInfo.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/SystemInfo.java
@@ -34,7 +34,6 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


eyalkoren:
Merging, as I want to rely on the container ID parsing logic. 
@felixbarny Please review anyway, we can revert still

PR #692, opened 2019-06-25
689 instrument quartz jobs

wolframhaussig:
WIP


wolframhaussig:
@eyalkoren Can you please review? It contains support for Quartz and the Spring Quartz Jobs now.

eyalkoren:
@wolframhaussig sure I will! May take a bit though as I have a lot going on at the moment

eyalkoren:
@wolframhaussig Please let me know when you feel this is ready for review

wolframhaussig:
@eyalkoren This can be reviewed: The last Bug I found was the Spring version conflict which I solved yesterday.

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,222 @@
...
+        resetCounter();
+        Scheduler scheduler=new StdSchedulerFactory().getScheduler();
+        scheduler.clear();
+        JobDetail job = JobBuilder.newJob(TestJob.class)
+    			.withIdentity("dummyJobName", "group1").build();
+        Trigger trigger = TriggerBuilder
+        		.newTrigger()
+        		.withIdentity("myTrigger")
+        		.withSchedule(
+        				SimpleScheduleBuilder.repeatSecondlyForTotalCount(2, 1))
```
eyalkoren: Not sure if we need to test this, as it tests the framework usage and not the instrumentation.

If you think this is important to test, it should be enough to do it only in one dedicated test called `testRepeatedInvocations`. I assume the other things tested shouldn't matter with the repetition, right?

The problem is that this makes the unit tests much slower. As we accumulate tests, it is important for us to keep them as quick as possible. 

Unless there is a good reason, please remove the repetitions from all the tests and use only one without a counter, but with an explicit expected number of transactions.

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,222 @@
...
+        		.build();
+        scheduler.scheduleJob(job, trigger);
+        scheduler.start();
+        Thread.sleep(2500);
```
eyalkoren:
```suggestion
        Thread.sleep(1500);
```
No need for such a long wait. It appears that the total would be 2 in this case, so waiting 1500 should be enough.

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,222 @@
...
+        scheduler.shutdown();
+        assertThat(reporter.getTransactions().size()).isEqualTo(getInvocationCount());
```
eyalkoren:
```suggestion
        assertThat(reporter.getTransactions().size()).isEqualTo(2);
```
Otherwise we don't know if this worked...

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,222 @@
...
+        assertThat(reporter.getTransactions().get(0).getContext().getCustom("scheduledTime")).asString().isNotEmpty();
```
eyalkoren:
```suggestion
        Transaction transaction = reporter.getTransactions().get(0);
        assertThat(transaction.getContext().getCustom("scheduledTime")).asString().isNotEmpty();
```

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,222 @@
...
+        assertThat(reporter.getTransactions().get(0).getContext().getCustom("trigger")).asString()
```
eyalkoren: This currently fails in all tests

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,222 @@
...
+        	.isEqualToIgnoringCase(String.format("%s.%s", trigger.getKey().getGroup(), trigger.getKey().getName()));
+        assertThat(reporter.getTransactions().get(0).getName())
+        	.isEqualToIgnoringCase(String.format("%s.%s", job.getKey().getGroup(), job.getKey().getName()));
```
eyalkoren: Add a method for all these common assertions, where transaction type is asserted as well

apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameAdvice.java
```diff
@@ -0,0 +1,82 @@
...
+    @VisibleForAdvice
+    public static final Logger logger = LoggerFactory.getLogger(JobTransactionNameInstrumentation.class);
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    private static void setTransactionName(@Advice.Argument(readOnly = true, value = 0) JobExecutionContext context,
+    		@SimpleMethodSignature String signature, @Advice.Origin Class<?> clazz, @Advice.Local("transaction") Transaction transaction) 
+    		throws InvocationTargetException, NoSuchMethodException {
+        if (ElasticApmInstrumentation.tracer != null) {
+            TraceContextHolder<?> active = ElasticApmInstrumentation.tracer.getActive();
+            if(context == null) {
+            	logger.warn("Cannot correctly name transaction for method {} because JobExecutionContext is null", signature);
```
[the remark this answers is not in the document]
wolframhaussig: This should not happen as Quartz always provides a JobExecutionContext but it may happen when the programmer calls the method manually. As this breaks the Transaction name recognition I added the warning.

apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameAdvice.java
```diff
@@ -0,0 +1,82 @@
...
+	                    .withName(signature)
+	                    .withType(JobTransactionNameInstrumentation.TRANSACTION_TYPE)
+	                    .activate();
+            }else if (active == null) {
+				transaction = ElasticApmInstrumentation.tracer.startTransaction(TraceContext.asRoot(), null, clazz.getClassLoader())
+	                    .withName(context.getJobDetail().getKey().toString())
+	                    .withType(JobTransactionNameInstrumentation.TRANSACTION_TYPE)
+	                    .activate();
+				transaction.addCustomContext("trigger", context.getTrigger().getKey().toString());
+				transaction.addCustomContext("scheduledTime", context.getScheduledFireTime().toString());
```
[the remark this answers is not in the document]
wolframhaussig: Quartz separates the logic of when a piec of code is being executed(trigger) from the code itself(job) so you may have multiple trigger executing the same job. In large applications it can be difficult to check why a job was wrongly executed, therefore I want to add the trigger name which triggered the execution of the job. Here, searching for the field might make sense.
As the quartz framework works with a thread pool it can occur that the the pool is empty as all threads are used so a job execution is delayed. In this case the timestamp of the transaction in APM would be the date and time when the job was executed and the field `scheduledTime` would indicate the date and time the job **should** have been executed.

apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameAdvice.java
```diff
@@ -0,0 +1,82 @@
...
+            } else {
+                logger.debug("Not creating transaction for method {} because there is already a transaction running ({})", signature, active);
```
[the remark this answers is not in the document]
wolframhaussig: I copied that from the other plugin - I let it stay on debug for now...

wolframhaussig:
I have updated the code

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+    @Test
+    void testJobWithGroup() throws SchedulerException, InterruptedException {
+        JobDetail job = JobBuilder.newJob(TestJob.class)
+    			.withIdentity("dummyJobName", "group1").build();
+        Trigger trigger = TriggerBuilder
+        		.newTrigger()
+        		.withIdentity("myTrigger")
+        		.withSchedule(
+        				SimpleScheduleBuilder.repeatSecondlyForTotalCount(1, 1))
```
eyalkoren: There is an option to use `SimpleScheduleBuilder.simpleSchedule().withIntervalInMilliseconds()`, which allows you test multiple executions quickly, and `SimpleScheduleBuilder#withMisfireHandlingInstructionFireNow` for a single one (?)

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,194 @@
...
+        		.build();
+        scheduler.scheduleJob(job, trigger);
+        scheduler.start();
+        Thread.sleep(250);
```
eyalkoren: I would really like to get rid of those sleeps as well.
What we can do is create a `Future` at the beginning of the test and give it to the Job, which will set it as completed and the test can wait on the Future. This will reduce the test time to minimum while still kept robust enough. 
For example, see the usage of `CompletableFuture` in [this test](https://github.com/elastic/apm-agent-java/blob/9498bee6d90f429fdae8c0408ed6963134d68daf/apm-agent-plugins/apm-apache-httpclient-plugin/src/test/java/co/elastic/apm/agent/httpclient/ApacheHttpAsyncClientInstrumentationTest.java#L57).

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,194 @@
...
+    @BeforeAll
+    static void setUpAll() {
+        reporter = new MockReporter();
+        tracer = new ElasticApmTracerBuilder()
+                .configurationRegistry(SpyConfiguration.createSpyConfig())
+                .reporter(reporter)
+                .build();
+        ElasticApmAgent.initInstrumentation(tracer, ByteBuddyAgent.install(),
+                Collections.singletonList(new JobTransactionNameInstrumentation(tracer)));
+    }
```
eyalkoren:
```suggestion
    }
    
```

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,194 @@
...
+    void assertTests(JobDetail job) {
+    	assertThat(reporter.getTransactions().size()).isEqualTo(1);
+        assertThat(reporter.getTransactions().get(0).getName())
+        	.isEqualToIgnoringCase(String.format("%s.%s", job.getKey().getGroup(), job.getKey().getName()));
+    }
```
eyalkoren:
```suggestion
    }
    
```

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,194 @@
...
+    static void setUpAll() {
+        reporter = new MockReporter();
+        tracer = new ElasticApmTracerBuilder()
+                .configurationRegistry(SpyConfiguration.createSpyConfig())
+                .reporter(reporter)
+                .build();
+        ElasticApmAgent.initInstrumentation(tracer, ByteBuddyAgent.install(),
+                Collections.singletonList(new JobTransactionNameInstrumentation(tracer)));
+    }
+    void assertTests(JobDetail job) {
```
eyalkoren:
```suggestion
    void verifyJobDetails(JobDetail job) {
```

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,194 @@
...
+    	assertThat(reporter.getTransactions().size()).isEqualTo(1);
+        assertThat(reporter.getTransactions().get(0).getName())
+        	.isEqualToIgnoringCase(String.format("%s.%s", job.getKey().getGroup(), job.getKey().getName()));
```
eyalkoren: assert that transaction type is correct

docs/supported-technologies.asciidoc
```diff
@@ -255,3 +255,25 @@ tracing.
 * The agent does currently not support running on JVMs with an enabled `SecurityManager`.
   You may see exceptions like this: `java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "getProtectionDomain")`.
   Try to grant `java.security.AllPermission` to the agent.
+
+[float]
+[[supported-scheduling-frameworks]]
+=== Scheduling frameworks
```
eyalkoren: 👍

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,194 @@
...
+        		.build();
+        scheduler.scheduleJob(job, trigger);
+        scheduler.start();
+        Thread.sleep(250);
+        scheduler.pauseAll();
+        assertTests(job);
+        assertThat(reporter.getTransactions().get(0).getResult()).isEqualTo("this is the result");
+    }
+    
+    private static AtomicInteger count = new AtomicInteger(0);
```
eyalkoren: Currently you assert for a single Transaction each test, do we need this counter?

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
```diff
@@ -0,0 +1,194 @@
...
+    	reporter.reset();
+    	count.set(0);
+    	scheduler=new StdSchedulerFactory().getScheduler();
+    }
+    @AfterEach
+    private void cleanup() throws SchedulerException {
+    	scheduler.shutdown();
+    }
+    public static class TestJob implements Job {
+		@Override
```
eyalkoren: please fix the indentation of these test classes

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
wolframhaussig: As the first execution fires instantly we do not need to change the code to millisecond interval. This only affects the interval between 2 executions.
withMisfireHandlingInstructionFireNow only affects the execution if the job was not executed(scheduler paused,...) so as soon the scheduler is available again the job will be executed instantly.

apm-agent-plugins/apm-quartz-job-plugin/src/test/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentationTest.java
wolframhaussig: I cannot get this to work: As soon as the Job sets the CompletableFuture to complete the asserts run but at this time the transaction is not yet closed and I get errors. I still have to sleep for a few milliseconds so the transaction is available to the assertions.

eyalkoren: Sorry, for that we have `MockReporter#getFirstTransaction(long)` which does a busy wait.
Should have mentioned that...

eyalkoren:
You got a leftover from the git merge of the test class (.java.orig) - please remove

eyalkoren:
Jenkins run the tests

docs/supported-technologies.asciidoc
```diff
@@ -255,3 +255,25 @@ tracing.
...
+When using a scheduling framework a transaction for every execution will be created.
+
+|===
+|Framework |Supported versions | Description | Since
+
+|Scheduling Annotation
+|
+|The agent instruments any method annotated with one of the following:
```
eyalkoren:
```suggestion
|The agent instruments any method defined in a package configured in <<config-application-packages>> and annotated with one of the following:  
```

docs/supported-technologies.asciidoc
```diff
@@ -255,3 +255,25 @@ tracing.
 * The agent does currently not support running on JVMs with an enabled `SecurityManager`.
   You may see exceptions like this: `java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "getProtectionDomain")`.
   Try to grant `java.security.AllPermission` to the agent.
+
```
eyalkoren: Please take this part to be above the "Caveats" section

docs/supported-technologies.asciidoc
```diff
@@ -255,3 +255,25 @@ tracing.
...
+|
+|The agent instruments any method annotated with one of the following:  
+`org.springframework.scheduling.annotation.Scheduled`  
+`org.springframework.scheduling.annotation.Schedules`  
+`javax.ejb.Schedule`  
+`javax.ejb.Schedules` in order to create a transaction with the type `scheduled`, representing the scheduled task execution
+|1.6.0
+|Quartz
+|
+|The agent instruments the `execute` method of any class implementing `org.quartz.Job`, as well as the `executeInternal` method of any class extending `org.springframework.scheduling.quartz.QuartzJobBean`, and creates a transaction with the type `quartz`, representing the job execution
```
eyalkoren:
```suggestion
|The agent instruments the `execute` method of any class implementing `org.quartz.Job`, as well as the `executeInternal` method of any class extending `org.springframework.scheduling.quartz.QuartzJobBean`, and creates a transaction with the type `quartz`, representing the job execution 

NOTE: only classes from packages configured in <<config-application-packages>>  will be instrumented.
```

eyalkoren:
Jenkins run the tests

inakipaz:
@wolframhaussig Just tested it on a dev tomcat. Works as expected. Captures quartz transactions on tomcat without requiring to specify specific methods.

By the way, @eyalkoren when will this be published on a 1.7.1 release?

eyalkoren:
@inakipaz hopefully it will be included in 1.8.0. I can't say exactly when it will be released, but shouldn't take too long. You can register to get release notifications in our repo.
In any case, you should be able to use the snapshot for now without a problem.

inakipaz:
@eyalkoren Ok thanks. Will do that. Im adding it to tomcats with quartz jobs.

apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameAdvice.java
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+    }
+    @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
+    public static void onMethodExitException(@Advice.Argument(value = 0) @Nullable JobExecutionContext context,
+    		@Advice.Local("transaction") @Nullable Transaction transaction, @Advice.Thrown Throwable t) {
+        if (transaction != null) {
+			if(transaction.getResult() == null && context != null && context.getResult() !=null) {
```
felixbarny: `transaction.getResult() == null`

Is that check needed because you want to avoid overriding a manually set result (`co.elastic.apm.agent.impl.transaction.Transaction#withResult`)?

apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentation.java
```diff
@@ -0,0 +1,75 @@
...
+import net.bytebuddy.matcher.ElementMatchers;
+
+import static co.elastic.apm.agent.bci.bytebuddy.CustomElementMatchers.isInAnyPackage;
+import static net.bytebuddy.matcher.ElementMatchers.declaresMethod;
+import static net.bytebuddy.matcher.ElementMatchers.hasSuperType;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public class JobTransactionNameInstrumentation extends ElasticApmInstrumentation {
+	public static final String TRANSACTION_TYPE = "quartz";
```
felixbarny: What do you think about using `"scheduled"`? That is also used for Spring and EJB scheduled jobs. Note that the transaction overview page shows the type `request` by default and when you want to see other types, you have to select them from the dropdown.

apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentation.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return named("execute").or(named("executeInternal")).and(takesArgument(0, named("org.quartz.JobExecutionContext")));
+    }
+
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Arrays.asList(TRANSACTION_TYPE);
```
felixbarny: I wouldn't use the same constant for `transaction.type` and the transaction group name.

apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameAdvice.java
wolframhaussig: Yes, I do not want to override anything the programmer has set explicitly

apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentation.java
wolframhaussig: I am fine with that - I thought that every plugin creates a custom type.

Issue #693, opened 2019-06-25
[apm-server-healthcheck] HealthChecker fails on cloud foundry garden


kmacpher67:
**Describe the bug**
The health checker for cloud foundry is unable to find guid -
apm.agent.impl.payload.SystemInfo - container ID is null

**To Reproduce**
Enable elk-apm and restage app 

**Expected behavior**

**Debug logs**
`
SystemInfo - Could not parse container ID from
SystemInfo - container ID is null
`

<details>
<summary> /**
     * Name of the system platform the agent is running on.
     */
    private final String platform;
</summary>

```
   2019-06-25T13:45:45.82-0400 [APP/PROC/WEB/0] OUT 2019-06-25 17:45:45.825 [main] DEBUG co.elastic.apm.agent.impl.payload.SystemInfo - Could not parse container ID from '/proc/self/cgroup' line: 12:memory:/system.slice/garden.service/garden/70eb4ce5-a065-4401-6990-88ed
   2019-06-25T13:45:45.82-0400 [APP/PROC/WEB/0] OUT 2019-06-25 17:45:45.826 [main] DEBUG co.elastic.apm.agent.impl.payload.SystemInfo - Could not parse container ID from '/proc/self/cgroup' line: 11:hugetlb:/garden/70eb4ce5-a065-4401-6990-88ed
   2019-06-25T13:45:45.82-0400 [APP/PROC/WEB/0] OUT 2019-06-25 17:45:45.826 [main] DEBUG co.elastic.apm.agent.impl.payload.SystemInfo - Could not parse container ID from '/proc/self/cgroup' line: 10:freezer:/garden/70eb4ce5-a065-4401-6990-88ed
   2019-06-25T13:45:45.82-0400 [APP/PROC/WEB/0] OUT 2019-06-25 17:45:45.827 [main] DEBUG co.elastic.apm.agent.impl.payload.SystemInfo - Could not parse container ID from '/proc/self/cgroup' line: 9:net_cls,net_prio:/garden/70eb4ce5-a065-4401-6990-88ed
   2019-06-25T13:45:45.82-0400 [APP/PROC/WEB/0] OUT 2019-06-25 17:45:45.827 [main] DEBUG co.elastic.apm.agent.impl.payload.SystemInfo - Could not parse container ID from '/proc/self/cgroup' line: 8:pids:/system.slice/garden.service/garden/70eb4ce5-a065-4401-6990-88ed
[...]

PR #695, opened 2019-06-26
Parsing container ID in cloud foundry garden

eyalkoren:
With regard to elastic/apm-agent-java#693


eyalkoren:
Updated algorithm according to https://github.com/elastic/apm/issues/105

Issue #676, opened 2019-06-18
further SQL parsing improvement


wolframhaussig:
**Is your feature request related to a problem? Please describe.**
We would like to further improve the Span naming for JDBC spans

- Calling DB procedures
  - SQL "{call demoSp(?, ?)}" gets the name "{call"
  - SQL "{ call demoSp(?, ?)}" gets the name "{"
- SQL merge
  - SQL "Merge into testTable using(select * from table2) src ..." gets the name "Merge"

**Describe the solution you'd like**
For the "{call demoSp(?, ?)}" I would like to get "call demoSp" as Span name.
For "Merge into testTable ..." I would like to get at least "Merge testTable", maybe it is even possible to get "Merge testTable with table2" as span name?

eyalkoren:
Implemented in [version 1.7.0](https://github.com/elastic/apm-agent-java/releases). 
Did you try it out?

wolframhaussig:
Oh, sorry that was a fact I wanted to add: I am already using version 1.7.0, sorry.

eyalkoren:
Ahh, thanks. Will look into it then.

eyalkoren:
Are those all the cases you found out not working as expected, or are the additional ones?
Please try to provide the most exhaustive list of examples, this can help us out.

eyalkoren:
🤦‍♂ Wrong button..

wolframhaussig:
We have an Oracle 12c DB running and Elastic Stack running on version 6.5.4. I upgraded today to apm-agent 1.7.0 and the following SQL namings works fine: UPDATE, SELECT, DELETE.

The following SQLs are not parsed correctly:
merge into APPLICATION_VERSION target using (select APPLICATION_ID, ? VERSION from APPLICATION where APPLICATION_NAME=? ) source on (target.APPLICATION_ID=source.APPLICATION_ID and target.VERSION=source.VERSION) when matched then update set STARTUP=sysdate, LAST_KEEP_ALIVE_SIGNAL=sysdate when not matched then insert(APPLICATION_ID,VERSION,STARTUP,FIRST_STARTUP,LAST_KEEP_ALIVE_SIGNAL) values(source.APPLICATION_ID, source.VERSION,sysdate,sysdate,sysdate)
This gets "merge" as Span name.

{ call TRACE(CONTEXT=>?, WHO=>?, WHAT=>?, TARGET=>?, TEXT=>?) }
This gets only "{" as Span name.

[...]

PR #696, opened 2019-06-26
further SQL parsing improvement (#676)

wolframhaussig:
support for CALL using JDBC escape syntax
support for MERGE

MERGE support also contains support for DB links


eyalkoren:
Closes #676

eyalkoren:
@wolframhaussig Thanks for your contribution!!
Please let me know when you feel this is ready for review

wolframhaussig:
@eyalkoren This can be reviewed: I have added support for DB links on the other types like UPDATE, SELECT now so this is all I can do for now.

eyalkoren:
@axw please take a look. In general, and specifically about the `@` stuff- we want to treat these as sensitive info?

axw:
Thanks @wolframhaussig!

MERGE looks good!

What's the reason for including the DB link in the span name? Is it necessary, given that you can still find the same information in the span context? Is it likely that you would query the same schema/table pair with multiple DB links, and _not_ want to aggregate them?

I don't think we need to consider the DB link info as sensitive. It's just a DB link name and optional user name, and we already record those kinds of things in span context.

Regarding the JDBC escape syntax: I think it might be tidier if we do this as a pre-processing step. This is (obviously) JDBC specific, so separating it out would allow us to keep the SQL-parsing code the same for all the agents excluding the pre-processing step. For the pre-processing step, I think we'd just skip over the leading "{" and possible "? =". What do you think?

wolframhaussig:
> What's the reason for including the DB link in the span name? Is it necessary, given that you can still find the same information in the span context? Is it likely that you would query the same schema/table pair with multiple DB links, and not want to aggregate them?

Our UseCase contains db links to a lot of different databases located in different locations(in germany but also worldwide). My reasoning is that accessing the same table from different locations might take longer. Also, not all locations have the same amount of data as every location has only data for their plant. So a performance issue might be not reproducable on another database which has different/less data. 

As far as I know the span context cannot be searched, right? So I could not search for access to the same table in the same db link. Can we come to a compromise? Are you concerned about the parsing performance or about the DB link in the span name? It would be fine for me if the information is in the Span but not in the name so I could - in theory - filter by tag/label. Of course I could add the data manually by using the public API but maybe we can add the data as label and if performance is an issue make it configurable for the user?

Regarding the JDBC escape syntax: I am not completely against the idea of taking the code out of the sql parsing as it is jdbc specific as you said. But I do not completely agree either: I know we currently need this syntax parser only for the sql CALL but there are more uses of it so it would be more future aware if it was in the sql parser itself. I guess the most interesting after the call syntax are the outer joins. I do not have the usecase so I didn't implement it but the jdbc plugin currently fails on the following JDBC escape sequence:
```
  {
    "input": "SELECT * FROM {oj Countries JOIN Cities ON (Countries.country_ISO_code=Cities.country_ISO_code)}",
    "output": "SELECT FROM Countries"
  }
```
My code currently gives "oj" as table name, your code would give "{oj" but the real table name is of course "Countries".

What is your opinion on that?

axw:
> As far as I know the span context cannot be searched, right? So I could not search for access to the same table in the same db link.

Right, good point. I take it then that you're performing aggregations over the span data, and not just using them in the APM UI?

> Are you concerned about the parsing performance or about the DB link in the span name? It would be fine for me if the information is in the Span but not in the name so I could - in theory - filter by tag/label. Of course I could add the data manually by using the public API but maybe we can add the data as label and if performance is an issue make it configurable for the user?

I'm not concerned about performance in this instance, but usability. Perhaps not everyone will want the link in the name, and having it in the name makes it more difficult to aggregate spans for multiple locations/whatever. Perhaps that's not that important. I'm not really opposed to including it in the span name, just wanted to consider our options. @roncohen WDYT?

> Regarding the JDBC escape syntax: I am not completely against the idea of taking the code out of the sql parsing as it is jdbc specific as you said. But I do not completely agree either: I know we currently need this syntax parser only for the sql CALL but there are more uses of it so it would be more future aware if it was in the sql parser itself. I guess the most interesting after the call syntax are the outer joins. I do not have the usecase so I didn't implement it but the jdbc plugin currently fails on the following JDBC escape sequence:

Ah yes, I was aware of the other cases but hadn't considered the `oj` case properly. I think we could just skip over the `{` and `oj` tokens, so this could be implemented as a filter on the token source rather than modifying the parser. We could deal with that if/when we handle that escape syntax, or we could just implement it like that now: wrap the Scanner with another type that is escape-syntax aware, which will keep track of the escape sequences and filter out those tokens.

wolframhaussig:
> Right, good point. I take it then that you're performing aggregations over the span data, and not just using them in the APM UI?

Currently, we only want the Span name to be searchable in case of a performance problem so we can see if it is a one time problem for this db link or not. But in future we also want to show the performance data from the APM agent grouped by destinations in an own dashboard.

> Ah yes, I was aware of the other cases but hadn't considered the oj case properly. I think we could just skip over the { and oj tokens, so this could be implemented as a filter on the token source rather than modifying the parser. We could deal with that if/when we handle that escape syntax, or we could just implement it like that now: wrap the Scanner with another type that is escape-syntax aware, which will keep track of the escape sequences and filter out those tokens.

I am fine with that. I will wait with the change until we have completed the discussion regarding the db link topic.

roncohen:
Maybe a good compromise would be to parse the SQL so the span names don't contain the DB LINK parts and then instead store that in something like the `destination` [fields](https://www.elastic.co/guide/en/ecs/current/ecs-destination.html#ecs-destination-nestings)?   They should get indexed and are searchable.

That should make it possible to create visualizations that shows whether a specific DB LINK location has problems for each SQL query or as a whole.

wolframhaussig:
> Maybe a good compromise would be to parse the SQL so the span names don't contain the DB LINK parts and then instead store that in something like the destination fields? They should get indexed and are searchable.

Sounds great but  how do I do that? I have removed the db link from the Span name but I am lost where I should store it instead. Do we need to add a field to the co.elastic.apm.agent.impl.transaction.Db class?

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
```diff
@@ -128,38 +130,62 @@ private void parse(String query, StringBuilder signature) {
...
                 signature.append("UPDATE");
                 // Scan for the table name
-                boolean hasPeriod = false, hasFirstPeriod = false;
+                boolean hasPeriod = false, hasFirstPeriod = false, inQuotes = false, inDbLink = false;
+        		StringBuilder targetBuilder=signature;
                 if (scanner.scanToken(IDENT)) {
                     signature.append(' ');
                     scanner.appendCurrentTokenText(signature);
-                    for (Scanner.Token t = scanner.scan(); t != EOF; t = scanner.scan()) {
+                    for (Scanner.Token t = scanner.scan(false); t != EOF; t = scanner.scan(false)) {
```
axw: I think we could make a couple of simplifications:

Don't tokenise the dblink string, reverting the additional parameter to `scan`, and just assume that whatever string follows the "@" is a dblink. Just look for Token.AT followed by Token.STRING; if we find this sequence immediately following the table name, then we record that as the db link.

Is there an issue with doing that that I've missed? Generally the philosophy of the parser is to not interpret anything unless we absolutely have to. If there's some other `Token.AT, Token.STRING` sequence that _doesn't_ relate to DB links, that would be an issue. 

Second: rather than looking for those tokens in this switch, look for the sequence after we break out of the loop, after finding the table name. Similar in appendIdentifiers. This would probably mean making a few small changes: break the loop rather than returning immediately in the default case, and widen the scope of `t`.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
```diff
@@ -57,9 +60,10 @@
     private final static ConcurrentMap<String, String> signatureCache = new ConcurrentHashMap<String, String>(DISABLE_CACHE_THRESHOLD, 0.5f, Runtime.getRuntime().availableProcessors());
 
     private final Scanner scanner = new Scanner();
+    private final StringBuilder dbLink = new StringBuilder();
 
     public void querySignature(String query, StringBuilder signature, boolean preparedStatement) {
-
+		dbLink.setLength(0);
```
axw: This probably should be another out parameter, like `signature`. Perhaps another method overload.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
```diff
@@ -54,55 +57,56 @@
...
      * That means that we are leaking Strings but as the size of the map is limited that should not be an issue.
      */
-    private final static ConcurrentMap<String, String> signatureCache = new ConcurrentHashMap<String, String>(DISABLE_CACHE_THRESHOLD, 0.5f, Runtime.getRuntime().availableProcessors());
+    private final static ConcurrentMap<String, String[]> signatureCache = new ConcurrentHashMap<String, String[]>(DISABLE_CACHE_THRESHOLD, 0.5f, Runtime.getRuntime().availableProcessors());
 
     private final Scanner scanner = new Scanner();
 
     public void querySignature(String query, StringBuilder signature, boolean preparedStatement) {
-
+    	querySignature(query, signature, new StringBuilder(), preparedStatement);
```
felixbarny: Can we make `StringBuilder dbLink` `@Nullable`, so that this doesn't allocate a `StringBuilder` when calling `querySignature(String query, StringBuilder signature, boolean preparedStatement)`?

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
```diff
@@ -54,55 +57,56 @@
...
+    }
```
felixbarny: Add new line

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/SignatureParser.java
```diff
@@ -54,55 +57,56 @@
...
-    private void parse(String query, StringBuilder signature) {
+    private void parse(String query, StringBuilder signature, StringBuilder dbLink) {
         final Scanner.Token firstToken = scanner.scanWhile(Scanner.Token.COMMENT);
         switch (firstToken) {
             case CALL:
                 signature.append("CALL");
-                if (scanner.scanUntil(Scanner.Token.IDENT)) {
-                    signature.append(' ');
-                    scanner.appendCurrentTokenText(signature);
+                if(scanner.scanUntil(Scanner.Token.IDENT)) {
```
felixbarny:
```suggestion
                if (scanner.scanUntil(Scanner.Token.IDENT)) {
```

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/filter/JdbcFilter.java
```diff
@@ -0,0 +1,96 @@
...
+		}
+		StringBuilder sb = cache.get();
+		sb.setLength(0);
+		boolean inQuote = false, inJdbcEscape = false;
+		char c;
+		int i = 0,j = 0;
+		while(i<sql.length()) {
+			c = sql.charAt(i);
+			switch(c) {
+				case '{':
```
felixbarny: can there be nested JDBC escapes?

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/filter/JdbcFilter.java
```diff
@@ -0,0 +1,96 @@
...
+					inQuote = !inQuote;
+					i++;
+					break;
+				default:
+					sb.append(c);
+					i++;
+					break;
+			}
+		}
+		return sb.toString();
```
felixbarny: That allocates memory when the SQL contains a JDBC escape. As it's more of an edge case that's probably fine but it's not the most efficient way. Maybe this could somehow plug in into the main parsing loop. Might be over engineered though.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/filter/JdbcFilter.java
wolframhaussig: not as far as I know

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/signature/filter/JdbcFilter.java
wolframhaussig: I don't really know what you mean by that - can you explain what you suggest?

felixbarny: It was hard to describe because I also didn't have a clear picture. I fiddled around a bit and I think I came up with a nice design which is both simpler and more performant as it does not cause allocations. I've created a PR to your PR here: https://github.com/wolframhaussig/apm-agent-java/pull/1. Let me know what you think about it 🙂

wolframhaussig: Looks good but I found an edge case which does not work with your implementation:
  {
    "input": "{ ? = call oj() }",
    "output": "CALL oj"
  }
Fails in your code because you do not check if oj is the first text after the opening bracket. I will merge your code into my branch and try to solve that.

felixbarny:
Could you format the code you edited? It's a mix of tabs and spaces (we use only spaces) and there are some spaces missing after keywords like `if(` instead of `if (`. Try not to format code you did not touch, so that git blame still works.

After that, we can merge 🙂

felixbarny:
Sorry to bother you with formatting again but `SignatureParser` and the other classes you have modified would need some formatting too. Just make sure you don't format lines you didn't work on. If it's too fiddly with your formatter it's no big deal. Just tell me and I'll do the formatting 🙂

wolframhaussig:
I think I fixed all places I modified now

felixbarny:
I found some more places. Can you merge https://github.com/wolframhaussig/apm-agent-java/pull/2?

felixbarny:
Thanks again, great work!

PR #706, opened 2019-07-07
Adding debug logging for reason not capturing request body

eyalkoren:


apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
```diff
@@ -137,6 +137,15 @@ private void startCaptureBody(Transaction transaction, String method, @Nullable
                 request.withBodyBuffer();
             } else {
                 request.redactBody();
+                if (webConfiguration.getCaptureBody() == OFF) {
```
felixbarny: wrap all in `logger.isDebugEnabled()` if statement

PR #709, opened 2019-07-08
Managing ApmScope active stack

### LINKED ISSUE #703: ApmScopeManager.active() is returning scope with finishSpanOnClose always set to false
**Describe the bug**
Based on the code: https://github.com/elastic/apm-agent-java/blob/master/apm-opentracing/src/main/java/co/elastic/apm/opentracing/ApmScopeManager.java#L58-L67, when the app gets the current active scope from APM tracer like this:
```
Scope scope = tracer.scopeManager().active();
```
the returned `scope` object always has `finishSpanOnClose` set to false, if we close the scope, the current span inside scope is not finished. 
This is causing spans created by [opentracing/java-spring-messaging](https://github.com/opentracing-contrib/java-spring-messaging/blob/master/opentracing-spring-messaging/src/main/java/io/opentracing/contrib/spring/integration/messaging/OpenTracingChannelInterceptor.java#L84-L96) being dropped.

**To Reproduce**
This issue can be simply reproduced by debugging the below snippet:
```java
Tracer tracer = new ElasticApmTracer();
tracer.buildSpan("test")
  .startActive(true);
Scope scope = tracer.scopeManager().active();
// verify that finishSpanOnClose is false.
scope.close();
```
Or it can be reproduced end-to-end by having a spring-cloud-stream app with [opentracing/java-spring-messaging](https://github.com/opentracing-contrib/java-spring-messaging). The spans produced by the library is not shipped to Elastics.

**Expected behaviour**
`tracer.scopeManager().active()` returns a `Scope` object that preserves the `finishSpanOnClose` config from current active scope.

eyalkoren:
Closes elastic/apm-agent-java#703 

Added an OpenTracing scope stack. It is used for the ability to return an OpenTracing library the same scope object it had created when using the `ScopeManager#active` API.
In case the internal stack state had progressed and someone calls `ScopeManager#active`, they will get a newly created Scope that is good for context propagation, but if the scope in such scenario will be attempted to `close`, we assume this is an error. In this case, removing the interfering instrumentation (eg changing the `instrument` config to false) will fix the error.

Additional changes:
- Added basic logging capabilities to the OT bridge
- changed `ApmSpan#dispatcher` to `volatile`
- resetting `TraceContext#transactionId` when recycled

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
@@ -238,6 +238,7 @@ public class TraceContext extends TraceContextHolder {
         traceId.resetState();
         id.resetState();
         parentId.resetState();
+        transactionId.resetState();
         outgoingHeader.setLength(0);
         flags = 0;
         discard = false;
--- a/apm-agent-plugins/apm-opentracing-plugin/src/main/java/co/elastic/apm/agent/opentracing/impl/ApmSpanInstrumentation.java
+++ b/apm-agent-plugins/apm-opentracing-plugin/src/main/java/co/elastic/apm/agent/opentracing/impl/ApmSpanInstrumentation.java
@@ -308,4 +308,36 @@ public class ApmSpanInstrumentation extends OpenTracingBridgeInstrumentation {
         }
     }
 
+    public static class CompareSpanInstrumentation extends ApmSpanInstrumentation {
+
+        public CompareSpanInstrumentation() {
+            super(named("compareSpan"));
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class)
+        public static void compareSpan(@Advice.Argument(value = 0, typing = Assigner.Typing.DYNAMIC) @Nullable TraceContext thisTraceContext,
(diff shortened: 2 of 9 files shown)


felixbarny:
The bug fix in `TraceContext` and making the `dispatcher` volatile makes sense obviously!

However, I don't particularly like that we now have to maintain two activation stacks. I think there should only be a single source of truth, which already is hard enough to manage and to debug in scenarios where unexpected things happen.

Also, I'd like the classes in the OT bridge module to contain as few logic as possible. We should try to have as much logic in the actual agent. That will make compatibility with mixed versions between the two modules much easier to handle.

eyalkoren:
I totally agree- there should be a single source of truth and that should be the agent's stack.
The OT stack is just an OT-scope-accounting one to be used as a hint. In 99% of the cases it tells us we don't have to create a new scope instance (and it's kind of weird that we do), but if something doesn't sit well with the stack, we fallback to the same thing we did before- create a new instance. When that happens, BTW, **it gives us quite a good indication that our agent is interfering with what the OT user is trying to do, as it means it tries to close something it thinks it created although it didn't**, so it leaks its own scopes and closes ours not when intended. This is why I added the logging capabilities and the warning.
Another reason why I think it is better this way is that even if it fixes a deprecated feature, it DOES fix it. So when someone is using the deprecated feature and we say: "you should upgrade", we would be right, but our main goal is not to be right, it is to provide the magical experience of installing the agent and see it works.

With that, the defense is done with the closing arguments 🙂

PR #714, opened 2019-07-09
Supporting SpringMapMessageListener

### LINKED ISSUE #710: Should add support for instrumenting the listeners with parameter as subTypes of javax.jms.Message (i.e. javax.jms.MapMessage)
JMS-Plugin

Existing JMS listeners are instrumented only if the parameter to onMessage is "javax.jms.Message". The "javax.jms.MapMessage" which is also frequently used subType isn't handled yet. We should consider supporting the suptypes of "javax.jms.Message"

SessionAwareMessageListeners from springframework are also widely used. Not having support for that can restrict the usage of jms plugin.

There are few messaging providers who have class names like "QuesSender" for example "TibjmsQueueSender.TibjmsQueueSender" from EMS which aren't supported in the current version.

eyalkoren:
closes elastic/apm-agent-java#710

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
@@ -199,6 +199,8 @@ public class TraceContext extends TraceContextHolder {
         } catch (IllegalArgumentException e) {
             logger.warn(e.getMessage());
             return false;
+        } finally {
+            onMutation();
         }
     }
 
--- a/apm-agent-plugins/apm-jms-plugin/pom.xml
+++ b/apm-agent-plugins/apm-jms-plugin/pom.xml
@@ -9,6 +9,9 @@
 
     <properties>
         <apm-agent-parent.base.dir>${project.basedir}/../..</apm-agent-parent.base.dir>
+        <spring-framework.version>5.1.8.RELEASE</spring-framework.version>
+        <activemq.version>5.15.9</activemq.version>
+        <artemis.version>2.8.1</artemis.version>
     </properties>
 
     <artifactId>apm-jms-plugin</artifactId>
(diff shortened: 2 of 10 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
```diff
@@ -199,6 +199,8 @@ public boolean asChildOf(String traceParentHeader) {
         } catch (IllegalArgumentException e) {
             logger.warn(e.getMessage());
             return false;
+        } finally {
+            onMutation();
```
eyalkoren: I think it is a bug without that, right?

felixbarny: Is that when re-using a TraceContext? Could you add a test case for this scenario?

apm-agent-plugins/apm-jms-plugin/pom.xml
```diff
@@ -24,21 +27,32 @@
...
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-jms</artifactId>
+            <version>${spring-framework.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-context</artifactId>
+            <version>${spring-framework.version}</version>
```
felixbarny: `<scope>provided</scope>`?

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageListenerInstrumentation.java
```diff
@@ -63,14 +63,15 @@ public JmsMessageListenerInstrumentation(ElasticApmTracer tracer) {
 
     @Override
     public ElementMatcher<? super TypeDescription> getTypeMatcher() {
-        return not(isInterface()).and(hasSuperType(named("javax.jms.MessageListener")));
+        return not(isInterface())
```
felixbarny: This formatting may make it a bit clearer than the unnecessary parenthesis? YMMV

```
return not(isInterface())
    .and(
        hasSuperType(named("javax.jms.MessageListener"))
            .or(hasSuperType(named("org.springframework.jms.listener.SessionAwareMessageListener")))
    );
```

apm-agent-plugins/apm-jms-plugin/pom.xml
eyalkoren: No, test! (only relies on JMS API, the Spring class is only used by name)
Thanks!

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageListenerInstrumentation.java
eyalkoren: Yes, there is an extra unnecessary pair

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
eyalkoren: No, it's when I set the Transaction a child of remote span after it is created. The tests pass without it, but I noticed in the logs that it was wrong. Will add a test.

eyalkoren:
@felixbarny Added to CHANGELOG, but I am not sure about adding to supported technologies.
I assume most users utilizing this Spring feature assume this is standard JMS..

PR #715, opened 2019-07-10
Update supported-technologies.asciidoc

ppf2:
Added reference to 32-bit and 64-bit for Java versions for clarification

Per discussion with @eyalkoren, as long as the supported Java version is used, both 32-bit and 64-bit Java should work.

### DIFF
--- a/docs/supported-technologies.asciidoc
+++ b/docs/supported-technologies.asciidoc
@@ -23,7 +23,7 @@ the agent does not capture transactions.
 
 [float]
 [[supported-java-versions]]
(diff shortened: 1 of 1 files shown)


eyalkoren:
Thanks @ppf2 !
We are a bit disinclined to add this, as we are not adding anything regarding the required OS.
Since we only work in the bytecode level, we are under the assumption that using a supported JVM is good enough, regardless of the OS. If we find out we have a specific issue with a specific OS, we will probably add that.

PR #719, opened 2019-07-15
Document load balancing and proxy configuration for server_urls

felixbarny:


apm-agent-core/src/main/java/co/elastic/apm/agent/report/ReporterConfiguration.java
```diff
@@ -61,7 +61,12 @@
         .label("The URLs for your APM Servers")
         .description("The URLs must be fully qualified, including protocol (http or https) and port.\n" +
             "\n" +
-            "NOTE: Providing multiple URLs only works if intake API v2 is enabled.")
+            "Fails over to the next APM Server URL in the event of connection errors.\n" +
+            "Achieves load-balancing by shuffling the list of configured URLs so that the fist URL will be randomly distributed.\n" +
```
axw:
```suggestion
            "Achieves load-balancing by shuffling the list of configured URLs so that the first URL will be randomly distributed.\n" +
```

When I hear load-balancing, I first imagined that a _single_ agent would be load-balancing requests across the set of servers.

felixbarny: That's true, a single agent would not load-balance. The assumption is that we only need to load balance if there is more than one agent active and that a single agent could not overwhelm one APM Server.

Issue #691, opened 2019-06-25
Add support for Ratpack


rbolkey:
I'd be interested in contributing support for Ratpack (https://ratpack.io) if it would be accepted and isn't already being worked on. If so, the CONTRIBUTING guide looks straight forward, and I think the code should be too.

Please let me know if I can proceed. Thanks!

eyalkoren:
Please go through [this blog post](https://www.elastic.co/blog/a-cookbook-for-contributing-a-plugin-to-the-elastic-apm-java-agent). It contains some basic guidelines to take into account when adding a plugin.
Good luck!

rbolkey:
Following up here. 

I'm hitting one small issue I think. Ratpack requires at least Java 8. I MIGHT be able to avoid touching their APIs that expose Java 8, but assuming that I can't, what's the best way to proceed given that the agent's build is targeting Java 7?

Otherwise, I think I have the code in OK shape for instrumenting the http request being handled by Ratpack as a Transaction. A code review would likely be beneficial at this point before I proceed further.

PR #722, opened 2019-07-15
Adding plugin for Ratpack (https://ratpack.io).

rbolkey:
Plugin currently instruments the HTTP request as a Transaction (see #691).

This is not feature complete at the moment and doesn't compile to Java 7, but I would appreciate a review as this is my first plugin.

Ratpack is based on Netty, and consequently highly asynchronous in processing. I had to take guidance from how The Async HTTP Client and the Async Servlet. 

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation


apm-agent-plugins/apm-ratpack-plugin/src/main/java/co/elastic/apm/agent/ratpack/AbstractDefaultExecutionInstrumentation.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        @SuppressWarnings("unused")
+        @Advice.OnMethodExit(suppress = Throwable.class)
+        public static void onAfterUnbind(@Advice.This Execution execution) {
+            activate(execution);
+        }
+
+        @VisibleForAdvice
+        public static void activate(Execution execution) {
```
rbolkey: Rename to `deactivate`.

apm-agent-plugins/apm-ratpack-plugin/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+        <groupId>co.elastic.apm</groupId>
+        <version>1.7.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>apm-ratpack-plugin</artifactId>
+    <name>${project.groupId}:${project.artifactId}</name>
+
+    <properties>
+        <!-- Ratpack requires at least Java 8 -->
+        <maven.compiler.target>8</maven.compiler.target>
```
rbolkey: My attempt to compile the Ratpack plugin at Java 8. Need guidance as to appropriate approach.

felixbarny: The compiler target should still be 7. Ratpack will still be capable of running Java 7 bytecode on a Java 8 JVM. As long as you are not using Java 8 bytecode instructions (for example lambdas) you can still compile it with bytecode level targeting Java 7. Obviously, you'll need at least a JDK 8 to compile which has, for example, `java.util.function.Consumer` in the classpath. But we already require that in other places.

apm-agent-plugins/apm-ratpack-plugin/src/main/java/co/elastic/apm/agent/ratpack/AbstractDefaultExecutionInstrumentation.java
```diff
@@ -0,0 +1,121 @@
...
+            execution.maybeGet(Transaction.class).ifPresent(new Consumer<Transaction>() {
+                @Override
+                public void accept(Transaction transaction) {
+
+                    if (tracer == null) {
+                        return;
+                    }
+
+                    if (tracer.currentTransaction() == null) {
+                        transaction.activateInScope();
```
felixbarny: as you are not using the scope, just call `activate()` which does not allocate a scope object

apm-agent-plugins/apm-ratpack-plugin/src/main/java/co/elastic/apm/agent/ratpack/AbstractExecStarterInstrumentation.java
```diff
@@ -0,0 +1,288 @@
...
+            //noinspection ConstantConditions
+            if (coreConfiguration != null
+                && coreConfiguration.isActive()
+                && tracer.currentTransaction() == null) {
+
+
+                // Enhancement: check for excluded urls
+                // @see co.elastic.apm.agent.servlet.ServletTransactionHelper.isExcluded
+
+                transaction = tracer.startTransaction(TraceContext.fromTraceparentHeader(), null, null);
```
felixbarny: The second argument should be the `elastic-apm-traceparent` header. The third argument would be the class loader of the class which initiates the transaction. You can inject the current class via `@Advice.This Class<?> clazz` and call `clazz.getClassLoader()` to get the application class loader.

apm-agent-plugins/apm-ratpack-plugin/src/main/java/co/elastic/apm/agent/ratpack/AbstractRatpackInstrumentation.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package co.elastic.apm.agent.ratpack;
+
+import co.elastic.apm.agent.bci.ElasticApmInstrumentation;
+
+import java.util.Collection;
+import java.util.Collections;
+
+public abstract class AbstractRatpackInstrumentation extends ElasticApmInstrumentation {
+
```
felixbarny: Add a classloader matcher so that the instrumentation only gets applied for applications which actually have Ratpack on the classpath.

Example:
https://github.com/elastic/apm-agent-java/blob/ed3480abd5de58c9d12fc2500b1b06389c1dafec/apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java#L65-L68

apm-agent-plugins/apm-ratpack-plugin/src/test/java/co/elastic/apm/agent/ratpack/RatpackHandlerInstrumentationTest.java
```diff
@@ -0,0 +1,128 @@
...
+        assertHasOneTransaction("/fail/" + i, body -> true, 500);
+    }
+
+    private void assertHasOneTransaction(String path, Predicate<String> bodyPredicate, int status) throws Exception {
+        final TestHttpClient client = applicationUnderTest.getHttpClient();
+
+        assertThat(client.get(path).getBody().getText()).matches(bodyPredicate);
+        assertThat(reporter.getFirstTransaction(500)).isNotNull();
+        assertThat(reporter.getTransactions()).hasSize(1);
+        assertThat(reporter.getFirstTransaction().getName().toString()).isEqualTo("GET " + path);
```
felixbarny: The expected name should be `GET /happy/:iteration`, for example

apm-agent-plugins/apm-ratpack-plugin/src/main/java/co/elastic/apm/agent/ratpack/AbstractExecStarterInstrumentation.java
```diff
@@ -0,0 +1,288 @@
...
+
+                request.getSocket()
+                    .withRemoteAddress(ClientIpUtils.getRealIp(request.getHeaders(), context.getRequest().getRemoteAddress().toString()));
+
+                request.getUrl()
+                    .withPort(serverConfig.getPort())
+                    .withPathname(context.getRequest().getUri())
+                    .withSearch(context.getRequest().getQuery());
+
+                request.redactBody();
```
felixbarny: add a TODO to record the body according to `capture_body` and `capture_body_content_types`

apm-agent-plugins/apm-ratpack-plugin/src/main/java/co/elastic/apm/agent/ratpack/AbstractExecStarterInstrumentation.java
```diff
@@ -0,0 +1,288 @@
...
+                transaction.withResult(context.getResponse().getStatus().toString());
+
+                fillTransactionName(transaction, context);
+                fillRequestContext(transaction.getContext().getRequest(), context);
+                fillResponseContext(transaction.getContext().getResponse(), context);
+            }
+
+            private void fillTransactionName(Transaction transaction, Context context) {
+
+                transaction.getName()
```
felixbarny: The transaction name should not be the path which may contain path parameters. Rather use the url template like `GET /foo/:bar`. This makes for a better group name.

apm-agent-plugins/apm-ratpack-plugin/pom.xml
rbolkey: OK. I take what you're saying to mean that I should be able to use `java.util.Optional` then, but as long as I use anonymous classes in place of lambdas? However, the animal-sniffer plugin is complaining when I do. Is there a way to resolve this?

[INFO] --- animal-sniffer-maven-plugin:1.17:check (signature-check) @ apm-ratpack-plugin ---
[INFO] Checking unresolved references to org.codehaus.mojo.signature:java17:1.0
[ERROR] Undefined reference: java.util.Optional
[ERROR] Undefined reference: void java.util.Optional.ifPresent(java.util.function.Consumer)
[ERROR] Undefined reference: boolean java.util.Optional.isPresent()

rbolkey:
@felixbarny Thank you for the review. I was able to apply all the feedback. Please confirm. In addition to  getting the build working. The next steps I have.

1. Be able to build using Maven (resolve uses of Java 8 classes).
1. Run a snapshot locally to validate performance and functionality.
1. [Enhancement] Recognize excluded URLs
1. [Enhancement] Allow for capturing of request body.
1. [Enhancement] Create spans for each Ratpack Handler.

I believe the last item would be a useful abstraction to instrument, but likely not required. Handlers are a key building block in Ratpack. Kind of a "unit of work" while handling a Request. But I also understand that creating additional Spans can hurt performance. Is there a guiding philosophy here?

apm-agent-plugins/apm-ratpack-plugin/pom.xml
rbolkey: I ran across the `import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement` annotation, which appears to have done the trick (I'm assuming that's the correct approach). Maven build appears to be working.

felixbarny:
> Be able to build using Maven (resolve uses of Java 8 classes).

I would propose to use distinct classes for the advices. For an example of this, see https://github.com/elastic/apm-agent-java/blob/9498bee6d90f429fdae8c0408ed6963134d68daf/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletInstrumentation.java#L81-L83 

You can then annotate these classes with `@IgnoreJRERequirement` to make animal sniffer ignore those classes.

rbolkey:
Ok. I broke them apart into distinct classes, and moved the `@IgnoreJRERequirement` annotation from the method level to the class level on the Advice.

apm-agent-plugins/apm-ratpack-plugin/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+        <artifactId>apm-agent-plugins</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.7.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>apm-ratpack-plugin</artifactId>
+    <name>${project.groupId}:${project.artifactId}</name>
+
+    <properties>
+        <!-- Ratpack requires at least Java 8 -->
```
felixbarny: remove comment?

rbolkey:
@felixbarny I have a question about "sampling" and capturing the request body? When I look at the servlet plugin it looks like a lot of the request and response details (query params, cookies, etc) are only captured when the transaction is supposed to be sampled. However, the body appears to be always captured, but redacted under certain EventTypes. Is that correct?

felixbarny:
> However, the body appears to be always captured, but redacted under certain EventTypes. Is that correct?

That's an oversight. The transaction context is cleared for non-sampled transactions so capturing the body and then discarding it is obviously wasteful. We'll happily accept a PR which fixes that in the servlet implementation.

rbolkey:
Thanks. An additional question. In the servlet plugin, it appears that query parameters are only captured when the `content-type` is `application/x-www-form-urlencoded`. I don't understand the rational for that, and am unsure if I should mirror that functionality.

felixbarny:
Jenkins run the tests please

apm-agent-plugins/apm-ratpack-plugin/pom.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>apm-agent-plugins</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.7.0</version>
```
felixbarny:
```suggestion
        <version>1.8.1-SNAPSHOT</version>
```

rbolkey: I merged in v1.8.0 for the moment. I'm testing in production, so wanted to make sure I'm on a stable release. I'll rebase to master and apply the suggestion once I sort a couple things out. Hopefully this week.

felixbarny:
Which features are still missing from your perspective?
As Ratpack is based on Netty, have you thought about instrumenting Netty itself? That might be a more generic approach. WDYT?

rbolkey:
@felixbarny I don't think there are any features left. I've been running it in production, and getting good results. My only remaining concern is tracking down what might be a concurrency bug that is showing up in a handful of requests. I'm busy doing that right now.

I did start down the Netty route initially, and I like that idea. However, it was easier for me to get started following the abstraction that Ratpack provides.

felixbarny:
What’s the symptom of that potential concurrency bug?

We'll probably want to add Netty support down the line anyway as it’s the base of more frameworks. 

I’m also wondering if the instrumentation could be made with fewer allocations when instrumenting Netty directly. Instrumenting Ratpack requires to allocate quite a few object allocations for anonymous inner classes and optionals. 

With the other instrumentations we’re normally trying to minimize or even cause no allocations at all, if possible.

rbolkey:
@felixbarny We use Elastic's HLRC in our application, so the symptom that I'm seeing in the APM dashboard is a sample that has an Elasticsearch query span that is unrelated to the request recorded by the transaction. My hunch is there's something around how I'm interacting with the ExecutorService managed by Ratpack in combination with the existing instrumentation that the agent has for Executors, but need to dig in further.

As for Netty, would it be sufficient to just focus on Netty's support for HTTP? One of my hesitations at the beginning was that a transaction seems highly dependent on your choice of application layer protocol, and Netty supports many or lets you just operate at the transport layer.

Also, trying to think outside the box on allocations, if I targeted the anonymous classes directly instead of the services that produce and consume them, would that be better? I'm not sure if I can make it work, but would reducing the number of allocations at the cost of instrumentation matching many more instances of classes be a good trade-off?

eyalkoren:
Marking as `blocked` for now, maybe go back to this once we do Netty and see if can leverage that.

PR #725, opened 2019-07-16
Make server_urls dynamically reloadable

### LINKED ISSUE #723: Make server_urls dynamically reloadable
Context https://discuss.elastic.co/t/intranet-solution-java-agent-send-metric-to-cloud-apm-server-via-proxy-server/190153/7

wolframhaussig:
closes #723

PropertyFileConfigurationSource now reloads changes done to the config file(currently every minute).

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
    NOTE: This feature is currently marked as "Incubating" and is disabled by default. In order to enable, it is 
    required to set the [`disable_instrumentations`](https://www.elastic.co/guide/en/apm/agent/java/1.x/config-core.html#config-disable-instrumentations) 
    configuration property to an empty string.
+ * Property File will now be reloaded every minute(works only for files not on the classpath)
 
 ## Bug Fixes
  * ClassCastException related to async instrumentation of Pilotfish Executor causing thread hang (applied workaround)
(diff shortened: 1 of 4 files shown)


felixbarny:
Hi and thanks for the PR!

The agent already supports reloading the configuration file:

https://github.com/elastic/apm-agent-java/blob/35f13a95fef02f3daf8d9c1c2bf26b7cc8027218/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java#L119

So refreshing the values for `server_urls` is not the problem here. It's rather that after initialization, it's not read from the configuration anymore. One of the reasons is that we shuffle the list in order to achieve load balancing. Another reason is that we have explicitly set the `dynamic` flag to false which signals users (this is also gets rendered in the [configuration docs](https://www.elastic.co/guide/en/apm/agent/java/current/config-reporter.html#config-server-urls)) that this is an option which is not reloadable at runtime:

https://github.com/elastic/apm-agent-java/blob/35f13a95fef02f3daf8d9c1c2bf26b7cc8027218/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ReporterConfiguration.java#L57-L72

In a PR which is currently in review, I have also refactored a lot in that area and created an [`ApmServerClient`](https://github.com/elastic/apm-agent-java/pull/640/files#diff-5613edf2f5fe8e90ede6aadcb346b6c8).

So it makes sense to wait for that PR to land until we tackle the challenge of reading the updated configuration. We can either do this inside of the `ApmServerClient` or we can swap out the references to the client for all users with a new instance of the client when the configuration changes. I haven't really made up my mind yet on which option is better. Probably the first is a bit easier. The configuration option allows us to add a change listener via `org.stagemonitor.configuration.ConfigurationOption#addChangeListener`.

felixbarny:
sorry for not providing more detail earlier 🙈

Issue #730, opened 2019-07-18
Transaction Service Name for Jetty without display name


Argannor:
**Describe the bug**
When no display name is set in Jetty, it will return the context path instead. When the context path is "/" this will result in a service name of "-" for that transaction, since "/" gets escaped.

This is especially problematic when using other middleware like KumuluzEE (an Eclipse MicroProfile Server) where you cannot set the display name.

**To Reproduce**
1. Use Jetty with no Display name configured and APM-agent attached
2. Make some request
3. Find a service named "-" in Kibana with a transaction for this request

**Expected behavior**
In that case the service name should fall back to the application wide service name (e.g. config value, jar-name, ...).

**Debug logs**
n/a

Argannor:
This is fixed in #731
Thanks for merging!

PR #731, opened 2019-07-18
Transaction service name: exception for jetty

Argannor:
Jetty returns context path if no display name was set. In that case it can be "/" for the root context path, which gets escaped to "-" as the transaction name.

This PR resets the displayName from "/" to null, so that normal logic for context path based naming applies. 

This fixes issue #730 

### Checklist

- [x] Implement code


Argannor:
CLA is now signed.

felixbarny:
run the tests

Argannor:
Thank you for the very fast merge!
Looking forward to the new release :)

Issue #726, opened 2019-07-16
[docs] Clarify use of application_packages config settings


roncohen:
We're getting a lot of question about what the application packages setting does and it seems users are having trouble understanding the purpose of it. Clarifying the purpose and adding examples should help reduce the confusion.

https://www.elastic.co/guide/en/apm/agent/java/current/config-stacktrace.html#config-application-packages

cc @bmorelli25

felixbarny:
See also my answer on https://discuss.elastic.co/t/apm-java-jboss-doubt-related-application-packages/190707. Maybe we can use a Brandonified version of that.

However, ideally, we should also think about options on how to automatically infer that so that users don't need to configure it at all or that we don't need to recommend and promote setting the option anymore.

For example, we could detect the application class loader and treat classes loaded from there as application classes. We need to find a way to distinguish them from libraries which may be loaded from the same class loader but mostly from a different jar file.

bmorelli25:
I will attempt to Brondonify™ this and open a PR

PR #733, opened 2019-07-18
[docs] Clarify application_packages

bmorelli25:
Closes: https://github.com/elastic/apm-agent-java/issues/726.

Adds more detail and an example for `application_packages`. Thanks to Felix's discuss response for clarifying things.

I still need to sit down with one of you and discuss how the Java Agent asciidoc files are generated. Do I need to update this in `StacktraceConfiguration.java` and run something to update this asciidoc file? Somewhere else?

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/stacktrace/StacktraceConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/stacktrace/StacktraceConfiguration.java
@@ -40,11 +40,23 @@ public class StacktraceConfiguration extends ConfigurationOptionProvider {
         .key(APPLICATION_PACKAGES)
         .configurationCategory(STACKTRACE_CATEGORY)
         .description("Used to determine whether a stack trace frame is an 'in-app frame' or a 'library frame'.\n" +
-            "Multiple packages can be set as a comma-separated list.\n" +
-            "Setting this option can also improve the startup time.\n" +
+            "This allows the UI to collapse the stack frames of library code,\n" +
+            "and highlight the stack frames that originate from your application.\n" +
+            "Multiple root packages can be set as a comma-separated list;\n" +
+            "there's no need to configure sub-packages.\n" +
+            "Because this setting helps determine which classes to scan on startup,\n" +
+            "setting this option can also improve startup time.\n" +
             "\n" +
-            "In order to be able to use the API annotations @CaptureTransaction and @CaptureSpan,\n" +
-            "it is required to set these options.")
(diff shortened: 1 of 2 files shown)


felixbarny:
Yes, this has to be updated in `StacktraceConfiguration.java`. The asciidoc is auto-generated. You can generate it via `./mvnw clean test`.

docs/configuration.asciidoc
```diff
@@ -1097,12 +1097,14 @@ Prepending an element with `(?-i)` makes the matching case sensitive.
...
-In order to be able to use the API annotations @CaptureTransaction and @CaptureSpan,
-it is required to set these options.
+This allows the UI to collapse the stack frames of library code,
+and highlight the stack frames that originate from your application.
+Multiple root packages can be set as a comma-separated list;
+there's no need to configure sub-packages.
+Because this setting helps determine which classes to scan on startup,
+setting this option can also improve startup time.
 
+You must set this option in order to use the API annotations @CaptureTransaction and @CaptureSpan.
```
felixbarny:
```suggestion
You must set this option in order to use the API annotations `@CaptureTransaction` and `@CaptureSpan`.
```

bmorelli25:
Thanks Felix! I updated `StacktraceConfiguration.java`.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/stacktrace/StacktraceConfiguration.java
```diff
@@ -40,11 +40,24 @@
...
+            "\n" +
+            "**Example**\n" +
+            "\n" +
+            "Most Java projects have a root package, e.g. `com.myproject`.\n" +
+            "Here's how you can set that application package using Java system properties:\n" +
+            "\n" +
+            "[source,bash]\n" +
+            "----\n" +
+            "-Delastic.apm.application_packages=com.myproject\n" +
+            "----\n")
```
eyalkoren: The problem is that this goes also to the elasticapm.properties file, where it doesn't make sense.
I'd say get rid of these HTML hints. I think using backticks for that would be enough, as in:
`-Delastic.apm.application_packages=com.myproject`.

eyalkoren: For additional clarity, I would add something like:
```suggestion
            "And if you are only interested in specific subpackages, you can set something like:\n" + 
            `-Delastic.apm.application_packages=com.myproject.api,com.myproject.impl`\n")
```

bmorelli25:
Thanks for the fix @eyalkoren. I've update this PR per your suggestions.

Issue #732, opened 2019-07-18
Add database link to span context


felixbarny:
Continuation of https://github.com/elastic/apm-agent-java/issues/676: add the db link to the span context

apm server issue: https://github.com/elastic/apm-server/issues/2478

@wolframhaussig are you interested in implementing this?

wolframhaussig:
@felixbarny I can try. But currently, the discussion in elastic/apm#107 is not yet completed so I would want to wait for that. And can you confirm what I should implement exactly as the field has to be transferred to the APM server and I am not sure if you would want to implement that youself. Just so I do not implement more than required...

PR #735, opened 2019-07-19
Add database link to span context(#732)

wolframhaussig:
closes elastic/apm-agent-java#732

### Checklist

- [x] Implement code
- [x] Add tests


apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
```diff
@@ -723,7 +723,8 @@ private void serializeDbContext(final Db db) {
                 }
             }
             writeField("type", db.getType());
-            writeLastField("user", db.getUser());
+            writeField("user", db.getUser());
+            writeLastField("link", db.getDbLink());
```
eyalkoren: `writeLastField` always adds a field to the serialized JSON, even if the value is null, so such a field (I assume not often used) is not the best fit. Better switch the order and make the `type` the last field.

eyalkoren:
@wolframhaussig please merge with master and push. There is an error in master currently that won't let us run the tests.

felixbarny:
run the tests

PR #736, opened 2019-07-22
Adding support for legacy ApacheHttpClient

eyalkoren:
When Axis2 makes use of ApacheHttpClient transport, it uses deprecated APIs, probably because it doesn't make use of the client builders.
Adding support for the deprecated API seems to be the quickest way to support that.

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/LegacyApacheHttpClientInstrumentation.java
+++ b/apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/LegacyApacheHttpClientInstrumentation.java
@@ -11,9 +11,9 @@
  * the Apache License, Version 2.0 (the "License"); you may
  * not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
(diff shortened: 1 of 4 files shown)


apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/LegacyApacheHttpClientInstrumentation.java
```diff
@@ -0,0 +1,127 @@
...
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        private static void onBeforeExecute(@Advice.Argument(0) HttpHost host,
+                                            @Advice.Argument(1) HttpRequest request,
+                                            @Advice.Local("span") Span span) {
+            if (tracer == null || tracer.getActive() == null) {
+                return;
+            }
+            final TraceContextHolder<?> parent = tracer.getActive();
+            String method;
+            if (request instanceof HttpUriRequest) {
```
eyalkoren: This may prove to be a limitation, we can extend support to get `org.apache.http.message.BasicHttpRequest#method` and `org.apache.http.message.BasicHttpRequest#uri` through reflection if required

eyalkoren:
Added [a comment](https://github.com/elastic/apm-agent-java/pull/736/files#r305702399) - this is basic implementation, maybe we will need to expand.

PR #737, opened 2019-07-22
Only track breakdown metrics if enabled

felixbarny:


felixbarny:
reverted because duration metrics still have to be tracked

PR #739, opened 2019-07-22
Add feature flag for remote config

felixbarny:
Update 404 handling

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ApmServerConfigurationSource.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ApmServerConfigurationSource.java
@@ -151,6 +151,10 @@ public class ApmServerConfigurationSource extends AbstractConfigurationSource im
      */
     @Nullable
     String fetchConfig(final ConfigurationRegistry configurationRegistry) {
+        if (!configurationRegistry.getConfig(CoreConfiguration.class).isCentralConfigEnabled()) {
+            logger.debug("Remote configuration is disabled");
+            return null;
+        }
         try {
             return apmServerClient.execute("/config/v1/agents", new ApmServerClient.ConnectionHandler<String>() {
                 @Override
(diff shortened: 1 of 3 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ApmServerConfigurationSource.java
```diff
@@ -202,16 +206,9 @@ private String tryFetchConfig(ConfigurationRegistry configurationRegistry, HttpU
...
-                etag = null;
                 // means that there either is no configuration for this agent
                 // or that this is an APM Server < 7.3 which does not have the config endpoint
-                logger.debug("No remote config found for this agent");
-                // makes sure to remove the configuration if all configs are deleted for this agent in Kibana
-                if (!config.isEmpty()) {
-                    logger.info("Received new configuration from APM Server: <empty>");
-                }
-                config = Collections.emptyMap();
-                configurationRegistry.reloadDynamicConfigurationOptions();
```
eyalkoren: Does this means a user can't cancel configuration (ie restore defaults) by just removing it, and is instead required to actively set everything to specified values?

felixbarny: Not sure I completely understand. The user can delete the central configuration in Kibana. The APM Server would then respond with a status code 200 and a `{}` body. That basically sets the `config` map to an empty map.

eyalkoren: The comment says:
> // means that there either is no configuration for this agent or...

Sorry, I don't remember the spec, but I read that as getting 404 in such case

felixbarny: The spec has recently been changed. The APM Server will now only return with 404 if it's pre 7.3: https://github.com/elastic/apm/pull/118

felixbarny: removed the outdated comment

PR #742, opened 2019-07-23
Make log correlation GA

### LINKED ISSUE #720: Stop using Thread context class loader for log correlation
Following up on #718, the eventual fix should be stop using the Thread context class loader to look for the slf4j MDC class. 
We agreed to use the transaction class loader, which can be stored in the transaction when created through the `ElasticApmTracer#startTransaction(... ClassLoader)` API.
This means we won't have the class loader ref when activating a `TraceContext`, but this currently only occurs in the OT bridge, where, if we really decide we must, we can send the class loader reference through the `activate` API.

### LINKED ISSUE #724: Log correlation docs
Improve the docs on how to enable log correlation. Also add a section to supported-technologies that slf4j is supported.

felixbarny:
- Load the MDC with the application class loader, as opposed to the
  Thread's context classloader (fixes #720)
- Remove span.id MDC entry
  This causes the most allocations as in contrast to transaction.id and
  trace.id, the String manifestation can't be reused.
  Other agents also don't add the span.id so it makes sense to align on that
- Docs improvements 
- Mention slf4j in supported technologies page

closes #724

- [x] Implement code
- [x] Add tests
- [x] Update documentation
- [x] Update [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
@@ -101,6 +101,8 @@ public class TraceContext extends TraceContextHolder {
     private final StringBuilder outgoingHeader = new StringBuilder(TRACE_PARENT_LENGTH);
     private byte flags;
     private boolean discard;
+    @Nullable
+    private ClassLoader applicationClassLoader;
 
     /**
      * Avoids clock drifts within a transaction.
@@ -227,6 +229,7 @@ public class TraceContext extends TraceContextHolder {
         id.setToRandomValue();
         clock.init(parent.clock);
         serviceName = parent.serviceName;
+        applicationClassLoader = parent.applicationClassLoader;
         onMutation();
     }
 
@@ -245,6 +248,7 @@ public class TraceContext extends TraceContextHolder {
         discard = false;
         clock.resetState();
         serviceName = null;
+        applicationClassLoader = null;
     }
 
     /**
@@ -361,6 +365,7 @@ public class TraceContext extends TraceContextHolder {
         discard = other.discard;
         clock.init(other.clock);
         serviceName = other.serviceName;
+        applicationClassLoader = other.applicationClassLoader;
         onMutation();
     }
 
@@ -404,6 +409,15 @@ public class TraceContext extends TraceContextHolder {
     @Override
     public Span createSpan(long epochMicros) {
         return tracer.startSpan(fromParent(), this, epochMicros);
+    }
+
+    void setApplicationClassLoader(@Nullable ClassLoader classLoader) {
+        this.applicationClassLoader = classLoader;
+    }
+
+    @Nullable
+    public ClassLoader getApplicationClassLoader() {
+        return applicationClassLoader;
     }
 
     public interface ChildContextCreator<T> {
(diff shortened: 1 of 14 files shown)


felixbarny:
run the tests

apm-agent-plugins/apm-slf4j-plugin/src/test/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListenerTest.java
```diff
@@ -134,7 +134,30 @@ void testWithNullContextClassLoader() throws Exception {
...
+        assertThat(result.get().booleanValue()).isTrue();
+        assertMdcIsEmpty();
+        thread.join();
+        transaction.end();
+    }
+
+    @Test
+    void testWithNullApplicationClassLoaderFallbackPlatformCL() throws Exception {
+        when(loggingConfiguration.isLogCorrelationEnabled()).thenReturn(true);
+        Transaction transaction = tracer.startTransaction(TraceContext.asRoot(), null, null).withType("request").withName("test");
```
eyalkoren: What about a positive test for the fallback class loader (ie when the context class loader is properly set)?
It can be the same test, but without setting the Platform CL, then the the context CL would be System CL

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
```diff
@@ -101,6 +101,8 @@ public boolean asChildOf(TraceContext child, Object ignore) {
     private final StringBuilder outgoingHeader = new StringBuilder(TRACE_PARENT_LENGTH);
     private byte flags;
     private boolean discard;
+    @Nullable
+    private ClassLoader applicationClassLoader;
```
eyalkoren: As we discussed- let's avoid hard reference, so that span leaks are not a big issue if they occur (ie limited to the amount of spans stored in stacks of JVM threads, but not holding entire applications alive).
And as you suggested- in order to avoid reallocating WeakReference, use a WeakConcurrentMap from WeakReference to a WeakReference 🙂

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
```diff
@@ -100,17 +100,16 @@ public void onChange(ConfigurationOption<?> configurationOption, Level oldValue,
     private final ConfigurationOption<Boolean> logCorrelationEnabled = ConfigurationOption.booleanOption()
         .key("enable_log_correlation")
         .configurationCategory(LOGGING_CATEGORY)
-        .description("A boolean specifying if the agent should integrate into SLF4J's MDC to enable trace-log correlation.\n" +
-            "If set to `true`, the agent will set the `spanId` and `traceId` for the currently active spans and transactions to the MDC.\n" +
+        .description("A boolean specifying if the agent should integrate into SLF4J's https://www.slf4j.org/api/org/slf4j/MDC.html[MDC] to enable trace-log correlation.\n" +
+            "If set to `true`, the agent will set the `trace.id` and `transaction.id` for the currently active spans and transactions to the MDC.\n" +
             "You can then use the pattern format of your logging implementation to write the MDC values to your log file.\n" +
+            "If you are using logback or log4j, add `%X` to the format to log all MDC values or `%X{trace.id}` to only log the trace id." +
```
eyalkoren:
```suggestion
            "If you are using logback or log4j, add `%X` to the format to log all MDC values or `%X{trace.id}` to only log the trace id.\n" +
```

PR #745, opened 2019-07-24
ci: remove the documentation stage

kuisathaverat:
Since the documentation team has elastic-ci/docs build in place for all our PRs, our Documentation stage lose sense, we are doing the same and we have to maintain it. Because of that, we will start to remove this stage from all projects.


felixbarny:
run the tests

Issue #687, opened 2019-06-24
check server version on startup


wolframhaussig:
As discussed in elastic support ticket 00352713 I open this feature request.

**Is your feature request related to a problem? Please describe.**
Last week I fell into the trap of addLabel for numbers and booleans requiring APM server 6.7 where we currently use 6.5.4. I find it unfortunate as the [comatibility page](https://www.elastic.co/guide/en/apm/get-started/current/agent-server-compatibility.html) says that the agent requires 6.5 or higher. In my mind compatible means that the server can import any data received from the agent(even if he ignores some fields). This is not the case as the server rejects any document from the agent if addLabel with numbers or booleans are used:
2019-06-18 14:19:07.606 [apm-reporter] WARN co.elastic.apm.agent.report.IntakeV2ReportingEventHandler - {"accepted":155,"errors":[{"message":"Problem validating JSON document against schema: I[#] S[#] doesn't validate with \"transaction#\"\n I[#] S[#/allOf/0] allOf failed\n  I[#/context/tags/JOBS_FOUND] S[#/allOf/0/properties/context/properties/tags/patternProperties/%5E%5B%5E.%2A%22%5D%2A$/type] expected string or null, but got number","document":...

**Describe the solution you'd like**
It would be great if the agent could check the server version on start and make the data compatible. In my example: Instead of sending the labels as numbers and booleans convert them to strings first.

Maybe you can even write a warning once into the log file?

**Describe alternatives you've considered**
Modifying the code would fix this problem, but this would mean that the new features cannot be used in environments with the newer APM server version.

eyalkoren:
Thanks for reporting.

>Instead of sending the labels as numbers and booleans convert them to strings first

The idea of Labels is providing a way to store data that is indexed, therefore searchable and aggregate-able. If you would think this is what you achieve by sending a Number label, and we would instead index it as text, I don't think that would be a good experience for you once you try to query them. 

So, I believe it is better to get an error somewhere, either on server or agent. Currently you get the error on the server, which was useful enough to make you understand what the problem is.
I agree it would be nicer get a more informative logging message on the agent side. We will think whether we want to get into this versioning logic.

wolframhaussig:
I would even agree with dropping the labels completely. My point is: currently the server cannot even import the data so all Spans with those labels are not visible in Kibana APM UI.

PR #746, opened 2019-07-24
Drop non-string labels if connected to older APM Server

felixbarny:
- [x] Implement code
- [x] Add tests
- [ ] ~Update documentation~

fixes elastic/apm-agent-java#687

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
@@ -24,6 +24,7 @@
  */
 package co.elastic.apm.agent.report;
 
+import co.elastic.apm.agent.util.Version;
 import co.elastic.apm.agent.util.VersionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -42,6 +43,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 
 /**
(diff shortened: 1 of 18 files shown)


felixbarny:
run the tests

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
```diff
@@ -64,34 +66,44 @@
...
-                    errorCount.set(0);
+                    setServerUrls(shuffleUrls(newValue));
                 }
             }
         });
-        this.serverUrls = Collections.unmodifiableList(shuffledUrls);
+        setServerUrls(Collections.unmodifiableList(shuffledUrls));
+    }
+
+    private void setServerUrls(List<URL> serverUrls) {
```
eyalkoren:
```suggestion
    private void initialize(List<URL> serverUrls) {
```

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerHealthChecker.java
```diff
@@ -24,36 +24,49 @@
...
 
+    @Nullable
     @Override
-    public void run() {
-        apmServerClient.executeForAllUrls("/", new ApmServerClient.ConnectionHandler<Void>() {
+    public Version call() {
+        List<Version> versions = apmServerClient.executeForAllUrls("/", new ApmServerClient.ConnectionHandler<Version>() {
             @Override
-            public Void withConnection(HttpURLConnection connection) {
+            public Version withConnection(HttpURLConnection connection) {
```
eyalkoren: What do you say about always returning a version from this method (eg version 0.0.0 on failures), so always having a list of versions at the size of configured server URLs? Then, we don't have to limit ourselves to look only at the minimal version, we can use the `errorCount` state to always know **the version of the currently connected server**.
We only need to make sure we use shuffle the URL list once and use the same shuffle here and for the other users.

eyalkoren: Forget about it- not THAT required: if we have a bunch of APM servers with different versions (shouldn't be a typical case unless during an APM server cluster upgrade), we would probably want our agents to behave the same with all of them, so the minimal version is actually probably what we are looking for

PR #748, opened 2019-07-25
Introduce priorities for transaction name

### LINKED ISSUE #673: Override transaction name based on priorities
Transaction names should have assigned priorities so that they can only be overridden with a transaction name which has a higher priority.

Priorities from highest to lowest:
- Transaction names set by the public API should have the highest priority
- High-level web frameworks like Spring MVC or JAX-RS (`Controller#soSomething`)
- The URL path, according to `use_path_as_transaction_name`
- Low-level web frameworks, like the Servlet API (`Servlet#doGet`)

See https://discuss.elastic.co/t/transaction-name-is-wrong/186013

felixbarny:
Now uses a transaction name according to use_path_as_transaction_name
rather than ServletClass#doGet.
But if a name can be determined from a high level framework,
like Spring MVC, that takes precedence.
User-supplied names from the API always take precedence over any others.

fixes elastic/apm-agent-java#673

- [x] Implement code
- [x] Add tests
- ~[ ] Update documentation~
- ~[ ] Update [CHANGELOG.md](CHANGELOG.md)~
- ~[ ] Update [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)~


eyalkoren:
I just realized that I reviewed this through #751.
JSF tests seem to fail

PR #754, opened 2019-07-28
Add example for context delegation through BlockingQueue using API

eyalkoren:


eyalkoren:
Jenkins run the tests

PR #763, opened 2019-07-31
Add central config to config source list

felixbarny:


docs/configuration.asciidoc
```diff
@@ -11,6 +11,7 @@ which have different naming conventions for the property key.
 The first configuration sources override the configuration values of over the latter sources.
 
 [arabic]
+. {kibana-ref}/agent-configuration.html[Central configuration]
```
eyalkoren: This change looks good, but I have trouble understanding the English in the sentence above:
> The first configuration sources override the configuration values of over the latter sources.

Is that proper?

felixbarny: @bmorelli25 will know. I'm sure there's a better way to express that even if it was technically correct. But as that has not been changed in this PR, I'll merge this one and will do a follow-up once Brandon is back next week.

PR #764, opened 2019-08-01
Avoid loading libattach twice

felixbarny:
context: https://discuss.elastic.co/t/not-able-to-add-java-apm-agent-as-dependency-in-spring-boot-project/193224

### DIFF
--- a/apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
+++ b/apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
@@ -46,7 +46,8 @@ public class ElasticApmAttacher {
     /**
      * Attaches the Elastic Apm agent to the current JVM.
      * <p>
-     * This method may only be invoked once.
+     * This method should only be invoked once.
+     * Subsequent calls will be ignored.
      * </p>
      *
      * @throws IllegalStateException if there was a problem while attaching the agent to this VM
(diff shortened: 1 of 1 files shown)


apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
```diff
@@ -74,14 +75,23 @@ public static void attach() {
...
+     * This method should only be invoked once.
+     * Subsequent calls will be ignored.
      * </p>
      *
      * @param configuration the agent configuration
      * @throws IllegalStateException if there was a problem while attaching the agent to this VM
      */
     public static void attach(Map<String, String> configuration) {
-        ByteBuddyAgent.attach(AgentJarFileHolder.INSTANCE.agentJarFile, ByteBuddyAgent.ProcessProvider.ForCurrentVm.INSTANCE, toAgentArgs(configuration));
+        synchronized (ElasticApmAttacher.class) {
```
eyalkoren: It's not really thread safe because this system property is set elsewhere.
We can kind of assume that this wouldn't be called again before `co.elastic.apm.agent.bci.AgentMain#init` is invoked, but if that's the case- no need to make it `synchronized`.

I suggest just in order to stay on the safe side and avoid over-thinking this, you can just use a different property here and there...

felixbarny: It's set in a method which is `synchronized` as well. I don't think there's a need for `attach` to be called concurrently. It seems safer to enforce this to be handled by one thread at a given time. I'd also like to avoid have two system properties for essentially the same use-case.

eyalkoren: This block is synchronized on `ElasticApmAttacher.class` and the other is synchronzed on `AgentMain.class`. Maybe just sync them on the same monitor?

eyalkoren:
Closed in favor of #775

felixbarny:
Actually, this is still relevant. Even if the attachment provider and thus the `com.sun.tools.attach.VirtualMachine` gets cached, users might try to attach from two distinct web applications. Checking the system property before trying to resolve the `VirtualMachine` from the `tools.jar` avoids it from being loaded by two distinct class loaders.

However, now that we are leveraging the emulated attach, it's far less likely that the `tools.jar`-based attachment provider will be used at all. We can even think about removing it and only rely on the emulated attachment provider.

However, checking the system property does not do any harm so I would just merge this tool

apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
felixbarny: I don't think there is a need to do that. There's also no cyclic dependency so that there's no danger of deadlocks.

eyalkoren:
> users might try to attach from two distinct web applications

For this scenario, the current synchronization provides no thread safety. Two web applications can still invoke `ByteBuddyAgent.attach` because the `ElasticApm.attached` is set elsewhere.

felixbarny:
> For this scenario, the current synchronization provides no thread-safety. Two web applications can still invoke ByteBuddyAgent.attach because the ElasticApm.attached is set elsewhere.

This is not so much about handling race conditions. The check in `co.elastic.apm.agent.bci.AgentMain#init` is responsible for that. This is mainly to not even attempt the attachment if the agent is already attached. Are you suggesting that we should get rid of the synchronized block or that we should fix it by synchronizing on the `"ElasticApm.attached".intern()` monitor, for example?

eyalkoren:
If `AgentMain#init` is always called when `ElasticApmAttacher#attach` is blocking on it (ie is lower on the call stack), then this actually safe enough...

PR #766, opened 2019-08-01
Added cdi-plugin for automated cdi span creation

Theoderich:
This PR adds a new cdi-plugin that creates spans for public method calls on CDI-Annotated beans.
All classes inside the application_packages that are annotated with a CDI scope annotation (@ApplicationScoped, @ViewScoped, etc.) are instrumented to create spans on public method calls. To avoid creating to many spans, short running methods may be filtered out via the cdi_methods_duration_threshold configuration option.

We have been using this plugin for a while in our ~200K LOC production application and it produces very good traces with basically zero configuration effort.

The pull request is complete (I hope) with integration tests and documentation.
The only problem is a failing integration test on Websphere 8.5.5 (and only that version) where the name of my cdi test application is not present in the resulting trace. I could not determine what the problem is here.


felixbarny:
Hi and thanks for this PR 🙂

While this is a really useful functionality I feel like it's not quite generic enough. What if I want to monitor spring services? What if I only want to monitor services in a specific sub-package? It's also almost duplicating the `trace_methods_duration_threshold`.

What do you think about extending the `trace_methods` functionality so that you can for example set `public @javax.enterprise.context.*Scoped com.example.*`? It seems like that would cover your use-case as well, right?

eyalkoren:
I agree with @felixbarny .
Extending the `trace_methods` mechanism to use annotations will be extremely useful to other users, as it will support much more use cases.
If you really need to add syntax for meta annotations, you can use something like: `public @@javax.inject.Scope co.example.*`, where `@@` representing that.

The great tests you wrote can still be used for validating the implementation.

Theoderich:
Hi,
I see where you are coming from and why you would like a more generic approach.

I do think there is an argument to be made to use more narrow focused plugins like this. When we first evaluated different instrumentation solutions we almost decided against using APM because out of the box the tracing was pretty coarse grained. We only saw the entry points and SQL queries, but no spans in between. Other products automagically found appropriate spans. Manually defining what to trace was out of the question because of the size of our codebase. We ultimately decided to use APM because it is open source and we were sure we could easily add the required features. As you can see, we did just that :)

So I think there would be benefits to offer "zero-configuration" automatic span creation for widely used frameworks like CDI. But I do see the downsides, especially the increased maintenance for multiple specialised plugins.

I will look into implementing annotations for trace_methods when I have some free time. For now we will maintain the fork.

felixbarny:
We’re currently investigating several ways to provide more method-level insight. 

One approach would be to have a continuously running sampling profiler. You can then analyze the cpu time in a flame graph. You could select a time frame for the flame graph and filter by transaction name and potentially by specific transaction id. 

Another idea would be to automatically trace methods which take a lot of time to execute.

eyalkoren:
> Other products automagically found appropriate spans

Do you mean stuff that fall into some framework category, like with the case of this PR, or something like @felixbarny mentioned- finding arbitrary "interesting" spans?

> So I think there would be benefits to offer "zero-configuration" automatic span creation for widely used frameworks like CDI.

I definitely agree, we need to find the right balance. I assume that if your instrumentation was doing some more framework-specific stuff (eg collecting and setting specific information in spans), we wouldn't think this is too fine-grained. 

But we can try and find a middle way, like using the `trace_methods` mechanism, with zero or minimal configuration. For example, the `trace_methods` default value doesn't have to be empty. It can contain known stuff like this. The problem with that is that it will increase startup time as all classes will be matched, but as long as there is no inheritance inference involved in the matching, this shouldn't be so bad. In this specific case, only application classes are being matched (if configured) so we can also add a flag to the syntax to indicate that.
Another option is to have a preconfigured commented-out value for the `trace_methods` config, like we have for [the `boot_delegation_packages` config option](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#config-boot-delegation-packages). This is not zero-configuration, but it is minimal configuration.

Theoderich:
> > Other products automagically found appropriate spans
> 
> Do you mean stuff that fall into some framework category, like with the case of this PR, or something like @felixbarny mentioned- finding arbitrary "interesting" spans?

I thought it was like this PR but I just reinstalled Perfino (easiest to acquire) to check.
Seems I was wrong. Perfino does samples on of all method calls in a transaction and also finds hotspots. So more or less what @felixbarny said. No framework specific spans.

eyalkoren:
@Theoderich all tests are covered in #771? Can we close this one?

Theoderich:
Sure

PR #771, opened 2019-08-02
Support annotation matching for TraceMethodInstrumentation

Theoderich:
As suggested in [PR 766](https://github.com/elastic/apm-agent-java/pull/766) I have made a PoC implementation for supporting annotations for the TraceMethodInstrumentation.

What do you think? Should I pursue this further?


felixbarny:
Looks great! I’d enhance it so that you can use wildcard marchers for annotation names. Otherwise, the functionality seems pretty complete.

Theoderich:
Ok I have added wildcard matching, documentation and my integration test from the other PR. Let me know what you think.

felixbarny:
Looks great!
I'm not really sure if the integration tests are needed and provide enough value to warrant a higher test execution time and complexity. They don't seem to test application server specifics like their (often strange) interpretation of the Servlet API. WDYT?

Theoderich:
The difference in test execution time is 8' 56''  vs. 9'44'' on my machine™. Not big, but noticable.
I agree that the test does not add a lot of value now that the implementation is very generic.

Theoderich:
> What about method annotations? Is there more than adding the same AnnotationMatcher to the method matching and some tests?

Just adding the AnnotationMatcher to method matching does not work. A matching Annotation would have to be present on both the class and the method to pass the check.

eyalkoren:
@Theoderich can you do this:
> adding the ready "recipe" for what you originally wanted to get to the documentation, so that users can only uncomment to get working. See what we did with this OSGi related configuration documentation.

?

And please merge with master so we can run the tests and merge into master.

Theoderich:
> 
> 
> @Theoderich can you do this:
> 
> > adding the ready "recipe" for what you originally wanted to get to the documentation, so that users can only uncomment to get working. See what we did with this OSGi related configuration documentation.
> 
> ?
> 
> And please merge with master so we can run the tests and merge into master.

I think I will get to it tomorrow

Theoderich:
Done

eyalkoren:
Jenkins run the tests

integration-tests/cdi-app/cdi-app-standalone/pom.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>jsf-app</artifactId>
```
eyalkoren: @Theoderich the build failed with:
```
[2019-08-14T06:14:27.054Z] [ERROR] [ERROR] Some problems were encountered while processing the POMs:

[2019-08-14T06:14:27.054Z] [FATAL] Non-resolvable parent POM for co.elastic.apm:cdi-app-standalone:[unknown-version]: Could not find artifact co.elastic.apm:jsf-app:pom:1.8.1-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 5, column 13
```
Is this parent reference valid with the path of the module and pom?

felixbarny: It should probably be
```suggestion
        <artifactId>cdi-app</artifactId>
```

Theoderich: yeah right. For some reason it built on my machine but its defninitly wrong

felixbarny:
Could you merge with master? That should fix the `No service name set. Expected 'CDI App'` error.

Theoderich:
I think its finally ready :)

PR #654, opened 2019-06-03
Support for opentracing 0.32.0 in Java

Miuler:
Upgrade of the version the opentracing from  0.31.0 to  0.32.0 in java

https://github.com/elastic/apm-agent-java/issues/634

PR #773, opened 2019-08-04
Support open tracing 0.33

### LINKED ISSUE #634: Support for opentracing 0.32.0
Please add support for opentracing 0.32.0, that version brings improvements, specifically the ability to obtain the traceId and the spantId using the SpanContext Interface, example:

```java
io.opentracing.Span span = tracer.activeSpan().context();
span.toTraceId();
span.toSpanId();
```

**APM Agent language and version:**

apm-opentracing: 1.6.1
elastic-apm-agent: 1.6.1
apm-agent-attach: 1.6.1

eyalkoren:
Closes elastic/apm-agent-java#634 
Closes elastic/apm-agent-java#654 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/apm-opentracing/src/main/java/co/elastic/apm/opentracing/ApmScopeManager.java
+++ b/apm-opentracing/src/main/java/co/elastic/apm/opentracing/ApmScopeManager.java
@@ -33,7 +33,7 @@ import javax.annotation.Nullable;
 
 class ApmScopeManager implements ScopeManager {
 
-    @Override
+    @Deprecated
     public ApmScope activate(@Nonnull Span span, boolean finishSpanOnClose) {
         final ApmSpan apmSpan = (ApmSpan) span;
         final Object traceContext = apmSpan.context().getTraceContext();
@@ -56,22 +56,10 @@ class ApmScopeManager implements ScopeManager {
 
     @Override
     public Scope activate(Span span) {
-        final ApmSpan apmSpan = (ApmSpan) span;
-        final Object traceContext = apmSpan.context().getTraceContext();
-        if (traceContext != null) {
-            // prevents other threads from concurrently setting ApmSpan.dispatcher to null
-            // we can't synchronize on the internal span object, as it might be finished already
-            // we can't synchronize on the ApmSpan, as ApmScopeManager.active() returns a different instance than ApmScopeManager.activate(Span)
-            synchronized (traceContext) {
-                // apmSpan.getSpan() has to be called within the synchronized block to avoid race conditions,
-                // so we can't do the synchronization in ScopeManagerInstrumentation
-                doActivate(apmSpan.getSpan(), traceContext);
-            }
-        }
-        return new ApmScope(true, apmSpan);
+        return activate(span, false);
     }
 
-    @Override
+    @Deprecated
     @Nullable
     public ApmScope active() {
         final ApmSpan apmSpan = activeApmSpan();
@@ -96,15 +84,13 @@ class ApmScopeManager implements ScopeManager {
 
     private ApmSpan activeApmSpan() {
         final Object span = getCurrentSpan();
-        final ApmSpan apmSpan;
+        ApmSpan apmSpan = null;
         if (span != null) {
             apmSpan = new ApmSpan(span);
         } else {
             final Object traceContext = getCurrentTraceContext();
             if (traceContext != null) {
                 apmSpan = new ApmSpan(new TraceContextSpanContext(traceContext));
-            } else {
-                apmSpan = null;
             }
         }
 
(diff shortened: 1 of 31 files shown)


apm-opentracing/src/main/java/co/elastic/apm/opentracing/ApmScopeManager.java
```diff
@@ -54,18 +55,46 @@ private void doActivate(@Nullable Object span, Object traceContext) {
...
+
+    @Deprecated
     @Nullable
     public ApmScope active() {
+        final ApmSpan apmSpan = activeApmSpan();
+        final ApmScope apmScope;
+        if (apmSpan != null) {
+            apmScope = new ApmScope(false, apmSpan);
+        } else {
+            final Object traceContext = getCurrentTraceContext();
```
felixbarny: Isn't that redundant and already handled in the `else` block of `activeApmSpan`?

apm-opentracing/src/main/java/co/elastic/apm/opentracing/ElasticApmTracer.java
```diff
@@ -77,4 +80,10 @@ public SpanBuilder buildSpan(String operationName) {
         }
         return null;
     }
+
+    @Override
+    public void close() {
+        scopeManager().active().close();
+        // co.elastic.apm.agent.impl.ElasticApmTracer#stop ???
```
felixbarny: makes sense I think

docs/opentracing.asciidoc
```diff
@@ -6,6 +6,8 @@ endif::[]
 [[opentracing-bridge]]
 == Elastic APM OpenTracing bridge
 
+NOTE: Latest supported OpenTracing version: 0.33 (as of agent and OpenTracing-bridge version 1.9.0)
```
felixbarny: 0.31.0 and 0.32.0 are still supported, right?

apm-opentracing/src/main/java/co/elastic/apm/opentracing/ExternalProcessSpanContext.java
```diff
@@ -30,16 +30,36 @@
 import java.util.Map;
 
 class ExternalProcessSpanContext implements ApmSpanContext {
+
+    // co.elastic.apm.agent.impl.transaction.TraceContext - only for the purpose of holding the parsed IDs
+    @Nullable
+    private Object childTraceContext;
```
felixbarny: Just an idea: instead of eagerly allocating the TraceContext object, could we instead validate the header and just parse out the trace id and span id on demand?

apm-opentracing/src/main/java/co/elastic/apm/opentracing/ApmScopeManager.java
eyalkoren: Yes, seems so...

apm-opentracing/src/main/java/co/elastic/apm/opentracing/ElasticApmTracer.java
eyalkoren: Yes! missed that..

docs/opentracing.asciidoc
eyalkoren: Yes, does this imply it is not? How would you put it?

apm-opentracing/src/main/java/co/elastic/apm/opentracing/ExternalProcessSpanContext.java
eyalkoren: Thought about all kind of stuff in order to reduce the allocations here, but I felt we can live with the overhead in this case on the favor of simplification.

felixbarny: Yeah, we definitely can. But we don't have to.

More often than not, the `toTraceId`/`toSpanId` methods won't be called (especially `ExternalProcessSpanContext` and allocating a TraceContext just in case seems a bit wasteful.

It wouldn't be any more complex I think. IMHO it's ok to just allocate a TraceContext each time `toTraceId`/`toSpanId` is called (it even says on the javadoc that it may cause allocations). So we don't even need a field for it.

We could then later further optimize it by not allocating a TraceContext object but to parse just the trace id/span id out of the header, without any changes in the bridge classes.

I don't insist on that though - I just think it would be a nice and simple improvement.

eyalkoren: Ahh, I misunderstood you. I thought you meant adding code to parse the ID on demand from the raw text map. I will just use this field and set it lazily on demand only if null, then we get one allocation max.

felixbarny: > I thought you meant adding code to parse the ID on demand from the raw text map.

That's actually what I meant 😄

eyalkoren: Well, I would really like to avoid duplicating the traceparent header parsing code for this.
I now only parse and create a trace context lazily if the getId APIs are used, and only once.
I believe that addresses your allocation concern, right?

PR #782, opened 2019-08-06
Decrease JUnit version to 5.4.2

eyalkoren:
Tests did not run apparently with 5.5.1. Not sure yet which dependency is missing and if we can fix that while still have the JUnit 4 dependency, so for now just a quick fix to restore tests


felixbarny:
So the tests did not run at all? Only on CI or also locally? Only JUnit 5-based tests or also the JUnit 4 bases ones?

eyalkoren:
Lots of tests did not run, locally as well. I assume this is restricted to the JUnit 5 ones, but not sure.

@bmorelli25 please merge master and retry `mvn clean test`.

PR #785, opened 2019-08-06
[docs] Configuration updates

bmorelli25:
My suggested updates based on the conversation here: https://github.com/elastic/apm-agent-java/pull/763/files#r309547992.

## Current docs:
Screen Shot 2019-08-06 at 11 26 02 AM

## This PR:
Screen Shot 2019-08-06 at 11 26 10 AM

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -9,7 +9,7 @@ You can configure the Elastic APM Agent using different configuration sources,
 which have different naming conventions for the property key.
 The Agent gives precedence to the configuration sources as follows,
 where Central configuration has the highest priority,
-and `elasticapm.properties` has the lowest.
+and `elasticapm.properties` has the lowest:
 
 . {kibana-ref}/agent-configuration.html[Central configuration]
 . Java system properties -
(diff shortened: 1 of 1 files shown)


felixbarny:
Higher overrides lower May be misleading as the lower numbers in the numbered list actually override the higher numbers.

bmorelli25:
Huh, I wasn't thinking about it in a number sense, just a high vs. low sense. Interesting, and a very good point! 

What about something like:
```asciidoc
The Agent gives precedence to the configuration sources as follows,
where Central configuration has the highest priority,
and `elasticapm.properties` has the lowest.
```

felixbarny:
SGTM

docs/configuration.asciidoc
```diff
@@ -5,21 +5,21 @@ Please only make changes in configuration.asciidoc.ftl
...
 [[configuration]]
 == Configuration
-To adapt the Elastic APM agent to your needs,
-you can configure it using different configuration sources,
+You can configure the Elastic APM Agent using different configuration sources,
 which have different naming conventions for the property key.
-The first configuration sources override the configuration values of over the latter sources.
+The Agent gives precedence to the configuration sources as follows,
+where Central configuration has the highest priority,
+and `elasticapm.properties` has the lowest.
```
eyalkoren:
```suggestion
and `elasticapm.properties` has the lowest:
```

bmorelli25:
Test failure looks unrelated, but I'll leave this for one of you two to merge.

PR #787, opened 2019-08-08
Disabling Log correlation when agent is inactive

eyalkoren:
- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## Bug Fixes
  * A warning in logs saying APM server is not available when using 1.8 with APM server 6.x
  * `ApacheHttpAsyncClientInstrumentation` matching increases startup time considerably
+ * Log correlation feature is active when `active==false`
 
 # 1.8.0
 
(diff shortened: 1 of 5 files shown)


felixbarny:
What if active is set to false after the activation and before the deactivation? We should have a transaction-scoped flag (can we use the noop flag?) which is consistent for a whole transaction. If that is false, call no scope listeners.

eyalkoren:
> What if active is set to false after the activation and before the deactivation?

I only check in adding to MDC but always removing from MDC.

apm-agent-plugins/apm-slf4j-plugin/src/test/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListenerTest.java
```diff
@@ -76,6 +79,33 @@ void testMdcIntegration() {
         transaction.end();
     }
 
+    @Test
+    void testDisabledWhenInactive() {
+        when(loggingConfiguration.isLogCorrelationEnabled()).thenReturn(true);
```
felixbarny: this seems like two tests in one

eyalkoren: It's a positive testing proving that the following negative testing works.
You prefer I remove the positive test?

felixbarny: Either remove it completely if it's already tested by other tests or make a separate test out of it.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ## Bug Fixes
  * A warning in logs saying APM server is not available when using 1.8 with APM server 6.x
  * `ApacheHttpAsyncClientInstrumentation` matching increases startup time considerably
+ * Log correlation feature is active when `active==false`
```
felixbarny:
```suggestion
 * Disable log correlation feature when `active=false`
```

eyalkoren: The `Bug Fixes` section usually describes the bugs and not the fixes

PR #792, opened 2019-08-15
ci(jenkins): MAVEN_CONFIG at the top level and also per group of stages

v1v:
## Highlights
- Ensure the `'-Dmaven.repo.local=.m2'` cannot be overridden even when triggering the build manually with the parameter `MAVEN_CONFIG`
- Ensure all the top-level stages, aka `stage('Tests')` and `stage('Initializing')` append the `MAVEN_CONFIG` env variable in conjunction with the `MAVEN_CONFIG` parameter.
- It does solve the very first build issues for new PRs, see https://github.com/elastic/apm-agent-java/pull/792#issuecomment-521601006

## Some other thoughts
- I initially thought to use a profile but unfortunately the maven.repo.local is not supported at that level.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,6 +13,7 @@ pipeline {
     CODECOV_SECRET = 'secret/apm-team/ci/apm-agent-java-codecov'
     GITHUB_CHECK_ITS_NAME = 'Integration Tests'
     ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
+    MAVEN_CONFIG = '-Dmaven.repo.local=.m2'
   }
   options {
     timeout(time: 1, unit: 'HOURS')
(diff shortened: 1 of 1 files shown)


v1v:
I actually found an interesting behavior with Jenkins and the creation of new PRs:
- The very first build won't contain the parameters section :(
- See [build](https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/view/change-requests/job/PR-793/1/) failed as the .m2 cache folder was not created in the workspace. But [build #2](https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/view/change-requests/job/PR-793/2/parameters/) does inject the parameters, therefore, it does inject the .m2 cache folder as expected.

In fact, [JENKINS-41929](https://issues.jenkins-ci.org/browse/JENKINS-41929) was reported and Ivan provided further details ... as it's still an issue, I'd rather say this particular PR will actually fix that behavior even though It was something I didn't think about it though.

PR #799, opened 2019-08-19
Setting JDBC statement map cleaner thread ccl to null

eyalkoren:
Fixing the problem reported in https://discuss.elastic.co/t/elastic-apm-agent-jdbchelper-seems-to-use-a-lot-of-memory/195295/8.

According to [the WebappClassLoderBase code](https://github.com/apache/tomcat/blob/master//java/org/apache/catalina/loader/WebappClassLoaderBase.java#L1722), setting the Context class loader of the cleaner thread to the system or bootstrap classloader should solve the problem.

### Checklist

- [x] Implement code
- [x] Update documentation (comment in code)

### DIFF
--- a/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/JdbcHelper.java
+++ b/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/JdbcHelper.java
@@ -35,7 +35,12 @@ import java.sql.Connection;
 public abstract class JdbcHelper {
     @SuppressWarnings("WeakerAccess")
     @VisibleForAdvice
-    public static final WeakConcurrentMap<Object, String> statementSqlMap = new WeakConcurrentMap<>(true);
+    public static final WeakConcurrentMap<Object, String> statementSqlMap;
+
+    static {
+        statementSqlMap = new WeakConcurrentMap<>(true);
+        statementSqlMap.getCleanerThread().setContextClassLoader(null);
+    }
 
     /**
      * Maps the provided sql to the provided Statement object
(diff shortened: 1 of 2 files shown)


felixbarny:
Do we have to do similar things when initializing thread pools? Currently probably not because they are initialized eagerly, right? Because of that, the context class loader will be the Attachment Listener thread's context class loader. But because the JdbcHelper is loaded in the context of a request processing thread, it's context class loader is the WebApplicationClassLoader, which creates a leak.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/JdbcHelper.java
```diff
@@ -35,7 +35,12 @@
 public abstract class JdbcHelper {
     @SuppressWarnings("WeakerAccess")
     @VisibleForAdvice
-    public static final WeakConcurrentMap<Object, String> statementSqlMap = new WeakConcurrentMap<>(true);
+    public static final WeakConcurrentMap<Object, String> statementSqlMap;
+
+    static {
+        statementSqlMap = new WeakConcurrentMap<>(true);
+        statementSqlMap.getCleanerThread().setContextClassLoader(null);
```
felixbarny:
```suggestion
        // This class may be loaded in the context of a request processing thread
        // whose context class loader is the WebApplicationClassLoader.
        // This leaks the WebApplicationClassLoader when the application is undeployed.
        // Tomcat will then stop the thread because it thinks it was created by the web application.
        // That means that the map will never be cleared, creating a severe memory leak.
        statementSqlMap.getCleanerThread().setContextClassLoader(null);
```

eyalkoren:
Merging without waiting for a verification on the fix, as this may cause an application class loader leak regardless.

PR #806, opened 2019-08-22
ci: packer cache script should fail never to help to cache dependencies

### LINKED ISSUE #805: packer_cache is failing
It looks like your packer_cache script is currently failing:
```
Failed to execute goal org.apache.maven.plugins:maven-failsafe-plugin:2.19.1:verify (default) on project apm-jdbc-plugin: There are test failures.
```

Is there any chance you could not run tests in that script? That doesn't seem like something that makes sense to do while building the nodes that'll run your CI. When they fail they make a huge, huge mess of my CI reports. Just red logs everywhere.

v1v:
Closed #805 

## Highlights
- The responsibility to validate the build should not be addressed within the packer cache script.
- This will help to fetch and cache all the maven dependencies and also the docker images independently of the maven build status.
- `-Dmaven.test.failure.ignore=true` could be an option but I'd rather prefer to avoid failing the maven build instead.

### DIFF
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 source /usr/local/bin/bash_standard_lib.sh
 
-JAVA_HOME=$HOME/.java/java10 ./mvnw clean verify
+JAVA_HOME=$HOME/.java/java10 ./mvnw clean verify --fail-never
 
 (retry 2 docker pull docker.elastic.co/observability-ci/weblogic:12.2.1.3-dev) \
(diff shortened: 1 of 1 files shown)


felixbarny:
+1 for `-Dmaven.test.failure.ignore=true`

v1v:
I'd rather say `-Dmaven.test.failure.ignore=true` will be more prescriptive in the hypothetical case the master branch got a compilation error, when it should not happen, and therefore this script will be fail. That's the main reason for using `fail-never` instead. What do you think @felixbarny ?

felixbarny:
I misunderstood the meaning of `--never-fail`. I thought it would behave as normal but always exit with exit code 0. But it seems like it continues when tests fail 👍

Issue #664, opened 2019-06-12
Enhance JMS support


eyalkoren:
Following #650, cover some more use cases.
This should be used as a meta issue, for the following enhancements (ordered according to prioritization):
- [x] Reverse polling APIs transaction creation logic (start transaction on method exit and end transaction on method enter) and avoid creating polling transactions when returned message is null
- [x] Capture destination name (ie queue/topic name) - requires schema change
- [x] Add special care for `javax.jms.TemporaryQueue/Topic` and Tibco temp queues/topic (starting with `$TMP$`)
- [x] Capture message bodies of text Messages (we can rely on the `javax.jms.Message#isBodyAssignableTo` API for this purpose)
- [x] Avoid creating Transactions in `onMessage` and `receive` when `active==false`
- [x] Disable instrumentation (message tagging) for specific queues/topics as suggested in elastic/apm-agent-java#710 - requires schema change
- [x] Capture message headers/properties - requires schema change
- [ ] JMS 2 async send APIs (`CompletionListener`) both in `JMSProducer` and `MessageProducer`
- [ ] [Reply-to/correlation-ID mechanisms](https://docs.oracle.com/cd/E13171_01/alsb/docs25/interopjms/MsgIDPatternforJMS.html) not tested - at least cover in tests, see if requires additional implementation
- [ ] Capture message bodies of additional message types (optional)
- [ ] Capture broker name (optional)

eyalkoren:
Jenkins test this please

eyalkoren:
Closing this meta issue, as we got to a point we prefer waiting for input before implementing the remaining tasks.
I will open a separate issue only for supporting JMS 2 async send APIs.

PR #816, opened 2019-08-25
JMS changing polling logic

eyalkoren:
Implementing first bullet of elastic/apm-agent-java#664 

- [x] Implement code
- [x] Add tests
- [ ] ~~Update documentation~~ will be done with #876 


eyalkoren:
With the new logic, Spring JMS test passes when the MessageListener instrumentation is disabled, validating that this logic is viable at least in this case (Spring actually uses polling APIs under the hood and applies logic to invoke a JMS-like `onMessage` API).

eyalkoren:
Jenkins test this please

eyalkoren:
Following a discussion, required changes:
- Use only a single type of transaction: `messaging`.
- Don't create polling transactions by default, but only handling ones.
- Try to find a way to reduce risk of leaking unclosed transactions, for example- instrument (or wrap) all Runnables and Callables so that at the end of `run`/`call` - we end any active transaction. This must not interfere with the Executors' plugin and beware of added related overhead.

apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/JmsInstrumentationIT.java
```diff
@@ -112,87 +123,211 @@ public void endTransaction() throws Exception {
...
-        Message incomingMessage = receiveOnNonTracedThread(receiveMethod).get(3, TimeUnit.SECONDS);
+        Message incomingMessage = resultQ.poll(2, TimeUnit.SECONDS);
+        assertThat(incomingMessage).isNotNull();
         verifyMessage(message, incomingMessage);
-        verifySendReceiveOnNonTracedThread(queue.getQueueName(), 1);
+        verifySendReceiveOnNonTracedThread(queue.getQueueName());
     }
 
-    private void verifySendReceiveOnNonTracedThread(String destinationName, int expectedReadTransactions) {
+    private void verifySendListenOnNonTracedThread(String destinationName, int expectedReadTransactions) {
```
felixbarny: Could we use [awaitility](https://github.com/awaitility/awaitility) in order to get rid of some `Thread.sleep`s? 

```java
await().atMost(500, MILLISECONDS).until(() -> !reporter.getSpans().isEmpty());
```

Sleeps are not only a source of flakiness but also make it hard to follow on which condition we are waiting for.

PR #823, opened 2019-08-29
Issue 819 disconnect apm server client

### LINKED ISSUE #819: Investigate APM Server connection reuse
In ApmServerClient.execute (and executeForAllUrls), we call `HttpURLConnection.disconnect` after executing each request:

https://github.com/elastic/apm-agent-java/blob/620a47fe45b0db7e71e97d4cbccee50f308deff7/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java#L217

The [java.net.HttpURLConnection docs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/HttpURLConnection.html) say:

> Calling the disconnect() method may close the underlying socket if a persistent connection is otherwise idle at that time

We should probably not be calling disconnect, but rather ensure that the input stream is consumed and and closed, so that the underlying socket can be reused.

kananindzya:
### Checklist

closes elastic/apm-agent-java#819
- [x] Implement code


apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
```diff
@@ -234,14 +233,31 @@ void onConnectionError() {
...
+            try {
+                int responseCode = connection.getResponseCode();
+                InputStream responseInputStream = null;
+                if (responseCode < 400) {
+                    responseInputStream = connection.getInputStream();
+                } else if (responseCode >= 400) {
+                    responseInputStream = connection.getErrorStream();
+                }
+                if (responseInputStream != null) {
+                    responseInputStream.close();
```
felixbarny: We not only have to close the input stream but also have to make sure to consume the stream. Use `org.stagemonitor.util.IOUtils#consumeAndClose` for that.
However, I'm not sure what happens if trying to consume an already closed stream.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
```diff
@@ -234,14 +233,31 @@ void onConnectionError() {
...
         }
         return results;
     }
 
+    private void closeInputStream(@Nullable HttpURLConnection connection) {
+        if (connection != null) {
+            try {
+                int responseCode = connection.getResponseCode();
+                InputStream responseInputStream = null;
+                if (responseCode < 400) {
```
felixbarny: rather than checking for the response code, check if `getErrorStream` returns a non-null result.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
```diff
@@ -234,14 +233,31 @@ void onConnectionError() {
...
-                if (connection != null) {
-                    connection.disconnect();
-                }
+                closeInputStream(connection);
             }
         }
         return results;
     }
 
+    private void closeInputStream(@Nullable HttpURLConnection connection) {
```
felixbarny: also close the stream in `co.elastic.apm.agent.report.IntakeV2ReportingEventHandler`

kananindzya: added to HttpUtils as static method

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
kananindzya: changed

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
kananindzya: Inside  ```IOUtils#consumeAndClose``` exist exception handling:
```
	public static void consumeAndClose(InputStream is) {
		if (is == null) {
			return;
		}
		try {
			while (is.read() != EOF) {}
		} catch (IOException e) {
			logger.warn(e.getMessage(), e);
		} finally {
			closeQuietly(is);
		}
	}

```

felixbarny: Do you see an issue with that?

kananindzya: No, there are no issues.

felixbarny:
As suspected, reading from an already closed stream is not a good idea

```
java.io.IOException: stream is closed
  at sun.net.www.protocol.http.HttpURLConnection$HttpInputStream.ensureOpen(HttpURLConnection.java:3423)
  at sun.net.www.protocol.http.HttpURLConnection$HttpInputStream.read(HttpURLConnection.java:3428)
  at co.elastic.apm.agent.shaded.stagemonitor.util.IOUtils.consumeAndClose(IOUtils.java:66)
  at co.elastic.apm.agent.report.HttpUtils.closeInputStream(HttpUtils.java:84)
  at co.elastic.apm.agent.report.IntakeV2ReportingEventHandler.flush(IntakeV2ReportingEventHandler.java:238)
  at co.elastic.apm.agent.report.IntakeV2ReportingEventHandler.handleEvent(IntakeV2ReportingEventHandler.java:128)
  at co.elastic.apm.agent.report.IntakeV2ReportingEventHandler.onEvent(IntakeV2ReportingEventHandler.java:116)
  at co.elastic.apm.agent.report.IntakeV2ReportingEventHandler.onEvent(IntakeV2ReportingEventHandler.java:50)
  at co.elastic.apm.agent.shaded.lmax.disruptor.BatchEventProcessor.processEvents(BatchEventProcessor.java:168)
  at co.elastic.apm.agent.shaded.lmax.disruptor.BatchEventProcessor.run(BatchEventProcessor.java:125)
```

felixbarny:
I suggest removing all other `close()` calls and let the `ApmServerClient` handle the responsibility of closing the stream.

kananindzya:
> I suggest removing all other `close()` calls and let the `ApmServerClient` handle the responsibility of closing the stream.

removed. In IntakeV2ReportingEventHandler already exists methods for closing input stream: onFlushError, onFlushSuccess.

felixbarny:
The last commits were not quite what I meant. I pushed some commits to this PR. What do you think?

kananindzya:
> The last commits were not quite what I meant. I pushed some commits to this PR. What do you think?

LGTM

apm-agent-core/src/main/java/co/elastic/apm/agent/report/HttpUtils.java
```diff
@@ -38,37 +40,44 @@
...
+     * In order to be able to reuse the underlying TCP connections,
+     * the input stream must be consumed and closed
+     * see also https://docs.oracle.com/javase/8/docs/technotes/guides/net/http-keepalive.html
+     *
+     * @param connection the connection
+     */
+    public static void consumeAndClose(@Nullable HttpURLConnection connection) {
+        if (connection != null) {
+            try {
+                InputStream responseInputStream = connection.getErrorStream();
```
eyalkoren: If error stream is not null- would be useful to log an error or at least a warning

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerHealthChecker.java
```diff
@@ -85,7 +85,7 @@ public Version withConnection(HttpURLConnection connection) {
                     } else {
                         try {
                             // prints out the version info of the APM Server
-                            String body = HttpUtils.getBody(connection);
+                            String body = HttpUtils.readToString(connection.getInputStream());
```
eyalkoren: @felixbarny we may be missing valuable info from the error stream, but we can log that when consuming and closing (see below). WDYT?

felixbarny: > we may be missing valuable info from the error stream

You mean you would like to log the APM Server response?

If there is an error stream, `connection.getResponseCode()` would have thrown an exception. But we could check in the outer catch if there's an error and log it. I just didn't want to change the logic in this PR

apm-agent-core/src/main/java/co/elastic/apm/agent/report/HttpUtils.java
felixbarny: What would you like to log? Just that there is an error stream or the response from APM Server? I would say that's the responsibility of the connection handler. In this method, we don't know whether the contents of the error stream are binary or whether they can be converted to a string.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerHealthChecker.java
eyalkoren: I raised this because this was the logic before this PR (through `getBody`) 🙂

apm-agent-core/src/main/java/co/elastic/apm/agent/report/HttpUtils.java
eyalkoren: Yeah, you're right, not the best fit here in terms of encapsulation, but a convenient location. Let's leave that to the handlers.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerHealthChecker.java
felixbarny: Got it. I removed it because that code path could never be reached and even if, it would not make much sense.

If there was an error stream, we would not even be in this branch, as `connection.getResponseCode()` (called on line 77) would have thrown an exception. There's also another if branch for status codes >= 300 so in this branch, there are only successful responses.

But even if there was an error stream, it would not really make sense to log in the next line `logger.info("Elastic APM Server is available: {}", body);` as an error stream would indicate a non-available APM Server.

PR #855, opened 2019-09-19
Enabling Mule4 Instrumentation

eyalkoren:
This PR contains two fixes required for Mule4 instrumentation enablement:
1. Adding a fallback to ByteBuddy class resolving mechanism to search the Boot class loader for `java.lang` and `java.lang.invoke` packages
2. Handling the real class loading visibility issue, where Mule's class loaders do not delegate to the basic class loaders, by adding a plugin and overriding this class loading decision.

This is mainly a POC, so we know that we know what the problems are and how to tackle them. Not sure we want to add a plugin that is meaningless without the mule agent. Options to make it better:
- Instead of explicitly modifying ByteBuddy configuration this way with hard coded packages, expose an attach config option to specify packages for appending to boot classpath search
- Using this plugin is not a big issue, it is just weird to have it in the agent where it is not functional without the mule agent. I found a couple of Mule issues ([10194](https://www.mulesoft.org/jira/browse/MULE-10194) and [9524](https://www.mulesoft.org/jira/browse/MULE-9524)) that talk about adding a config to specify packages that will be loaded from bootstrap class loader, but I couldn't find such config. All I found is a hard coded list (`org.mule.runtime.container.internal.ContainerClassLoaderFactory#BOOT_PACKAGES`). If there is a configuration to do that- much better.

@michaelhyatt you can [download the artefacts](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/PR-855/1/artifacts) or pull this PR and build/install locally for testing.
Note that it seems you still have work to do, as it doesn't look very good now, but at least the core agent instruments properly:
Screen Shot 2019-09-19 at 19 34 01


eyalkoren:
With our API reachable through the `java.lang` package, the real visibility issue is likely to be solved. We will still need to set the Boot CL class locator as a fallback to ByteBuddy class resolving mechanism with some changes - adding the a locator for the API (`Dispatcher` or other) package and removing the agent Jar locator fallback (won't be needed anymore).

Don't you prefer using a config option for the attacher API to setup packages to be searched with the Boot loader, instead of hard coding these two specific packages? It just seems arbitrary and very specific to this problem.

For the plugin- I agree, not a big deal to keep in, unless we find a configuration that does that instead.

PR #856, opened 2019-09-20
Set stacktrace.abs_path

felixbarny:
To make the integration with the Code app easier, they need the `abs_path` of the stack frame. In Java it's impossible to determine the exact path within the root of the git repository, like `apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java` but based on the package name and file name, we can get something like `co/elastic/apm/agent/report/serialize/DslJsonSerializer.java`.

The question is if it's ok to put a relative path in `abs_path`.

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] ~Update documentation~

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
@@ -682,10 +682,21 @@ public class DslJsonSerializer implements PayloadSerializer, MetricRegistry.Metr
         writeField("lineno", stacktrace.getLineNumber());
         int lastDotIndex = className.lastIndexOf('.');
         if (lastDotIndex > 0) {
-            writeField("abs_path", className.substring(0, lastDotIndex).replace('.', '/') + "/" + stacktrace.getFileName());
+            StringBuilder replaceBuilder = getReplaceBuilder();
+            replaceBuilder.append(className, 0, lastDotIndex);
+            replace(replaceBuilder, ".", "/", 0);
+            replaceBuilder.append("/");
+            replaceBuilder.append(stacktrace.getFileName());
+            writeField("abs_path", replaceBuilder);
         }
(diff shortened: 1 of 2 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
```diff
@@ -677,9 +677,14 @@ private void serializeStackTraceElement(StackTraceElement stacktrace) {
...
         writeField("filename", stacktrace.getFileName());
         writeField("function", stacktrace.getMethodName());
-        writeField("library_frame", isLibraryFrame(stacktrace.getClassName()));
+        String className = stacktrace.getClassName();
+        writeField("library_frame", isLibraryFrame(className));
         writeField("lineno", stacktrace.getLineNumber());
-        serializeStackFrameModule(stacktrace.getClassName());
+        int lastDotIndex = className.lastIndexOf('.');
+        if (lastDotIndex > 0) {
+            writeField("abs_path", className.substring(0, lastDotIndex).replace('.', '/') + "/" + stacktrace.getFileName());
```
eyalkoren: This allocates two String objects (`substring()` and `replace()`) for each stack frame of each span. Can we use a char-copy to a StringBuilder that includes the char replacement as well?
If so, can it also be used for handling inner classes, lambdas and the like or that using `getFileName()` will always be safer?

felixbarny:
We decided in today's meeting that it's not a good idea to put a relative path in a field called `abs_path`. For now, the code search integration should rely on just the file name, even if it may cause ambiguities.

We discovered discrepancies between the agent implementations as some put a relative path in the `filename` and some only put the base path in there.

Going forward, we should investigate alignment with ECS for stack traces. There are discussions going on at https://github.com/elastic/ecs/issues/154 and https://github.com/elastic/ecs/pull/563.

PR #858, opened 2019-09-23
Reduce allocations when serializing stack traces

felixbarny:


apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
```diff
@@ -106,6 +109,18 @@
     @Nullable
     private OutputStream os;
 
+    static {
+        Method getOurStackTrace;
+        try {
+            getOurStackTrace = Throwable.class.getDeclaredMethod("getOurStackTrace");
+            getOurStackTrace.setAccessible(true);
```
eyalkoren: I think this comes with a warning it is not going to be allowed in future versions, maybe since 9.

felixbarny:
Closing for now because it would introduce this warning:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by co.elastic.apm.agent.report.serialize.DslJsonSerializer (file:/Users/felixbarnsteiner/projects/github/elastic/apm-agent-java/apm-agent-core/target/classes/) to method java.lang.Throwable.getOurStackTrace()
WARNING: Please consider reporting this to the maintainers of co.elastic.apm.agent.report.serialize.DslJsonSerializer
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

PR #859, opened 2019-09-24
Avoid creating a temp agent file for each attachment

felixbarny:
Currently, every time the attacher is executed, it extracts a temp jar file for the agent. This PR attaches a MD5 checksum to the agent's file name. If a file with the same checksum is found, it's used instead of extracting the file again.

This avoids filling the disk with lots of agent jars.

### DIFF
--- a/apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
+++ b/apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
@@ -167,7 +167,8 @@ public class ElasticApmAttacher {
         try (InputStream agentJar = resourceAsStream) {
             MessageDigest md = MessageDigest.getInstance("MD5");
             byte[] buffer = new byte[1024];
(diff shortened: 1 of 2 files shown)


apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
```diff
@@ -138,19 +142,30 @@ private static File getAgentJarFile() {
...
             }
         }
+
+    }
+
+    static String md5Hash(InputStream resourceAsStream) throws IOException, NoSuchAlgorithmException {
+        try (InputStream agentJar = resourceAsStream) {
+            MessageDigest md = MessageDigest.getInstance("MD5");
+            byte[] buffer = new byte[1024];
+            while (new DigestInputStream(agentJar, md).read(buffer) != -1) {}
```
eyalkoren: Can't the `DigestInputStream` be reused?

felixbarny: Why would you want to reuse it?

PR #869, opened 2019-10-02
Missing version tag for hostname

OrangeDog:
### Checklist

- [x] Update documentation

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -227,7 +227,7 @@ A version string for the currently deployed version of the service. If you donâ€
 
 [float]
 [[config-hostname]]
-==== `hostname`
+==== `hostname` (added[1.10.0])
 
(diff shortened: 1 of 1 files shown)


OrangeDog:
Hmm, I did sign the CLA. Perhaps I did not wait long enough.

OrangeDog:
> This is a generated file.

In which case the "Edit" links on [the documentation](https://www.elastic.co/guide/en/apm/agent/java/master/config-core.html#config-hostname) should not have directed me to editing it.
I do not want to deal with the tests.

felixbarny:
This link is generically generated via the docs build and I don’t think it’s possible to omit it for this file.

But I think the link still has value. It’s a way for users to tell us if there are any typos or missing bits in the docs. As mentioned, I’m happy to take over at this point.

You did nothing wrong and thanks again for letting us know about the missing version tag!

PR #872, opened 2019-10-03
Add service.node.configured_name

### LINKED ISSUE #843: Adding ELASTIC_APM_SERVICE_NODE_NAME config
Related to https://github.com/elastic/apm/issues/141: add an `ELASTIC_APM_SERVICE_NODE_NAME` config option and send its value in `service.node.configured_name` metadata field

eyalkoren:
Closes elastic/apm-agent-java#843 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation
- [x] Manually test with server that `service.node.name` gets populated as expected

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/Service.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/Service.java
@@ -50,6 +50,11 @@ public class Service {
     @Nullable
     private Language language;
     /**
+     * Representation of a service node
+     */
+    @Nullable
+    private Node node;
+    /**
      * Immutable name of the service emitting this event
      * (Required)
      */
@@ -118,6 +123,22 @@ public class Service {
      */
     public Service withLanguage(Language language) {
         this.language = language;
+        return this;
+    }
+
+    /**
+     * Representation of a service node
+     */
+    @Nullable
+    public Node getNode() {
+        return node;
+    }
+
+    /**
+     * Representation of a service node
+     */
+    public Service withNode(Node node) {
+        this.node = node;
         return this;
     }
 
--- a/docs/metrics.asciidoc
+++ b/docs/metrics.asciidoc
@@ -14,6 +14,12 @@ You can adjust the interval with the setting <<config-metrics-interval>>.
 
 The metrics will be stored in the `apm-*` index and have the `processor.event` property set to `metric`.
 
+NOTE: Dedicated JVM metrics views are available since Elastic stack version 7.1.
+Since 7.5, metrics are aggregated separately for each JVM, relying on the ID of the underlying system- either container ID (where applicable) or hostname.
+Since Java agent version 1.11.0, it is possible to manually configure a unique name for each service node/JVM through
+<<config-service-node-name>>.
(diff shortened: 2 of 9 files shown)


eyalkoren:
Links in docs render properly, but linked configurations that has the since-version in title, get strange parenthesis:
Screen Shot 2019-10-03 at 18 41 11

Also - verify that the JVM release version is indeed 7.1

docs/metrics.asciidoc
```diff
@@ -14,6 +14,12 @@ You can adjust the interval with the setting <<config-metrics-interval>>.
 
 The metrics will be stored in the `apm-*` index and have the `processor.event` property set to `metric`.
 
+NOTE: Dedicated JVM metrics views are available since Elastic stack version 7.1.
+Since 7.5, metrics are aggregated separately for each JVM, relying on the ID of the underlying system- either container ID (where applicable) or hostname.
+Since Java agent version 1.11.0, it is possible to manually configure a unique name for each service node/JVM through
+<<config-service-node-name>>.
```
bmorelli25: I think the funky parenthesis are occurring because of the "added" tag in the title:
```
==== `service_node_name` (added[1.11.0])
```

You may need to change this line to manually specify the link text:
```suggestion
<<config-service-node-name,`service_node_name`>>.
```

docs/metrics.asciidoc
```diff
@@ -14,6 +14,12 @@ You can adjust the interval with the setting <<config-metrics-interval>>.
 
 The metrics will be stored in the `apm-*` index and have the `processor.event` property set to `metric`.
 
+NOTE: Dedicated JVM metrics views are available since Elastic stack version 7.1.
+Since 7.5, metrics are aggregated separately for each JVM, relying on the ID of the underlying system- either container ID (where applicable) or hostname.
```
bmorelli25: Should this be "starting in"?
```suggestion
Starting in 7.5, metrics are aggregated separately for each JVM, relying on the ID of the underlying system- either container ID (where applicable) or hostname.
```

bmorelli25: This should be an em dash instead of en dash:
```suggestion
Since 7.5, metrics are aggregated separately for each JVM, relying on the ID of the underlying system -- either container ID (where applicable) or hostname.
```

bmorelli25: And a third recommendation in case you like both suggestions:

```suggestion
Starting in 7.5, metrics are aggregated separately for each JVM, relying on the ID of the underlying system -- either container ID (where applicable) or hostname.
```

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/Service.java
```diff
@@ -49,6 +49,11 @@
      */
     @Nullable
     private Language language;
+    /**
+     * Representation of a service node
+     */
+    @Nullable
+    private Node node;
```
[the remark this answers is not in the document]
eyalkoren: True, although it's true for everything else as well. Even if not required for the JSON creation, it reflects the notion of the Node as a sub-entity of a Service in an OO way.

Issue #762, opened 2019-07-31
Allow certain errors to be ignored or excluded


nehaduggal:
There might be certain errors that an application throws which a user might want to ignore. We should have an API that allows for ignoring those errors/exceptions. The API should allow users to ignore errors based on either status codes, error message or error class.

felixbarny:
Related: https://discuss.elastic.co/t/apm-java-agent-ignore-exception/193005

felixbarny:
Another mention: https://discuss.elastic.co/t/filtering-out-exceptions-by-class-when-tracing/196864

felixbarny:
Should add a `ConfigurationOption<List<WildcardMatcher>> ignoreExceptions` to `CoreConfiguration`

cdalexndr:
Using Spring Boot, all `@Scheduled` methods are recorded. I have a method that fires regularly and I want to ignore it.
An option to ignore should be added, like `@IgnoreTransaction`.

PR #873, opened 2019-10-03
Configurable ignored exceptions

SylvainJuge:
closes elastic/apm-agent-java#762

Adds a new configuration parameter `ignore_exceptions` that contains a list of wildcard patterns of exceptions that should be ignored by the agent and not considered as errors.

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation
- [ ] ~~Update [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)~~

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -233,8 +233,13 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
         .description("A list of exceptions that should be ignored and not reported as errors.\n" +
             "This allows to ignore exceptions thrown in regular control flow that are not actual errors\n" +
             "\n" +
-            "This list supports wildcards patterns, thus using exception class qualified name is not required\n" +
-            "For example `com.mycompany.ExceptionToIgnore` could be shortened to `*ExceptionToIgnore`." +
+            WildcardMatcher.DOCUMENTATION + "\n" +
+            "\n" +
+            "Examples, all of them are equivalent:\n" +
+            "\n" +
+            " - `com.mycompany.ExceptionToIgnore`: using fully qualified name\n" +
+            " - `*ExceptionToIgnore`: using wildcard to avoid package name\n" +
+            " - `*exceptiontoignore`: case-insensitive by default\n" +
             "\n" +
             "NOTE: Exception inheritance is not supported, thus you have to explicitly list all the thrown exception types"
         )
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -233,7 +233,7 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
         .description("A list of exceptions that should be ignored and not reported as errors.\n" +
             "This allows to ignore exceptions thrown in regular control flow that are not actual errors")
         .dynamic(true)
-        .buildWithDefault(Collections.emptyList());
+        .buildWithDefault(Collections.<WildcardMatcher>emptyList());
 
     private final ConfigurationOption<Map<String, String>> globalLabels = ConfigurationOption
         .builder(new MapValueConverter<String, String>(StringValueConverter.INSTANCE, StringValueConverter.INSTANCE, "=", ","), Map.class)
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -56,8 +56,9 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
     public static final String SERVICE_NAME = "service_name";
     public static final String SAMPLE_RATE = "transaction_sample_rate";
     private static final String CORE_CATEGORY = "Core";
-    public static final String DEFAULT_CONFIG_FILE = AGENT_HOME_PLACEHOLDER + "/elasticapm.properties";
+    private static final String DEFAULT_CONFIG_FILE = AGENT_HOME_PLACEHOLDER + "/elasticapm.properties";
     public static final String CONFIG_FILE = "config_file";
+
     private final ConfigurationOption<Boolean> active = ConfigurationOption.booleanOption()
         .key(ACTIVE)
         .configurationCategory(CORE_CATEGORY)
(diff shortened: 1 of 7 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -357,7 +358,7 @@ public void captureException(long epochMicros, @Nullable Throwable e, TraceConte
     }
 
     public void captureException(long epochMicros, @Nullable Throwable e, @Nullable TraceContextHolder<?> parent, @Nullable ClassLoader initiatingClassLoader) {
-        if (e != null) {
+        if (e != null && !WildcardMatcher.isAnyMatch(coreConfiguration.getIgnoreExceptions(), e.getClass().getName())) {
```
eyalkoren: Because this may be called frequently in some scenarios, what do you say we cache the result for each Exception class name and save some CPU? Then, if we want to apply this logic for inheritance as well, performance is not a concern anymore.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -224,6 +225,16 @@
         .dynamic(true)
         .buildWithDefault(Collections.singletonList(WildcardMatcher.valueOf("(?-i)*Nested*Exception")));
 
+    private final ConfigurationOption<List<WildcardMatcher>> ignoreExceptions = ConfigurationOption
+        .builder(new ListValueConverter<>(new WildcardMatcherValueConverter()), List.class)
+        .key("ignore_exceptions")
+        .tags("added[1.11.0]")
+        .configurationCategory(CORE_CATEGORY)
+        .description("A list of exceptions that should be ignored and not reported as errors.\n" +
```
eyalkoren: We must document that we allow patterns and not only exact names, including examples. See what we do in `sanitize_field_names` for example, and how it looks like [in the documentation](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#config-sanitize-field-names)

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ElasticApmTracerTest.java
```diff
@@ -174,6 +179,44 @@ void testRecordException() {
         assertThat(error.getTransactionInfo().isSampled()).isFalse();
     }
 
+    @Test
+    void testDoesNotIgnoreExceptionsWithEmptyWildcards(){
+        DummyException1 exception = new DummyException1();
+
+        when(config.getConfig(CoreConfiguration.class).getIgnoreExceptions())
```
eyalkoren: No need, as this is the default

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ElasticApmTracerTest.java
```diff
@@ -174,6 +179,44 @@ void testRecordException() {
         assertThat(error.getTransactionInfo().isSampled()).isFalse();
     }
 
+    @Test
+    void testDoesNotIgnoreExceptionsWithEmptyWildcards(){
```
eyalkoren: Basically, this is the same as `testRecordException`, right?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
felixbarny: Let's add the cache once we want to match on the hierarchy. For now, I don't think the cache is worth it.

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ElasticApmTracerTest.java
```diff
@@ -174,6 +179,44 @@ void testRecordException() {
...
+        assertThat(reporter.getFirstError().getException()).isEqualTo(exception);
+    }
+
+    @Test
+    void testDoesNotRecordIgnoredExceptions() {
+        List<WildcardMatcher> wildcardList = Stream.of(
+            "co.elastic.apm.agent.impl.ElasticApmTracerTest$DummyException1",
+            "*DummyException2")
+            .map(WildcardMatcher::valueOf)
+            .collect(Collectors.toList());
```
felixbarny: Use `List.of`

felixbarny: I missed the `.map(WildcardMatcher::valueOf)` part, please ignore

SylvainJuge: using `.map(WildcardMatcher::valueOf)` thus not relevant to switch to `List.of(...)`.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
SylvainJuge: good point, I'll do that.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
SylvainJuge: I've added a comment not to forget this when (and if) we need to implement this.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -242,6 +243,27 @@
...
+            "This allows to ignore exceptions thrown in regular control flow that are not actual errors\n" +
+            "\n" +
+            WildcardMatcher.DOCUMENTATION + "\n" +
+            "\n" +
+            "Examples, all of them are equivalent:\n" +
```
eyalkoren: They will all cause the `com.mycompany.ExceptionToIgnore` to be ignored, but they are not equivalent

SylvainJuge: Good catch, I forgot to re-generate doc after changing this. I've removed the 'all of them are equivalent'.

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ElasticApmTracerTest.java
SylvainJuge: test removed to avoid duplication

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
SylvainJuge: fixed in next commit.

PR #874, opened 2019-10-06
Tracing not-excluded URLs

eyalkoren:


apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
```diff
@@ -288,7 +288,12 @@ private boolean isExcluded(String servletPath, @Nullable String pathInfo, @Nulla
             logger.debug("Not tracing this request as the User-Agent {} is ignored by the matcher {}",
                 userAgentHeader, excludeAgentMatcher);
         }
-        return excludeUrlMatcher != null || excludeAgentMatcher != null;
+        boolean isExcluded = excludeUrlMatcher != null || excludeAgentMatcher != null;
+        if (!isExcluded) {
```
felixbarny:
```suggestion
        if (!isExcluded && logger.isTraceEnabled()) {
```

eyalkoren: That is done implicitly. What's the benefit?

felixbarny: It avoids allocating an `Object[]` for the `argument` var-args parameter.

eyalkoren: Right, will change that.
Why is it different from the `debug` one above (the one with the three args)? If there is a match, it may be called very frequently as well.

felixbarny: Makes sense to add it there as well. But I think for two arguments, there is a special method which does not use varargs.

eyalkoren: Yes there is, which is why I wrote "the one with the three args", but probably should make it clearer 🙂

PR #879, opened 2019-10-09
Fix JMX collection/Support registering MBeans which are added after agent startup

felixbarny:
See https://discuss.elastic.co/t/apm-java-agent-jmx-metrics-for-springboot-application-capture-jmx-metrics/202700

- [x] Implement code
- [x] Add tests
- [x] ~Update documentation~
- [x] ~Update [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)~
- [x] ~Added an API method or config option? Document in which version this will be introduced.~


apm-agent-plugins/apm-jmx-plugin/src/main/java/co/elastic/apm/agent/jmx/JmxMetricTracker.java
```diff
@@ -83,50 +96,131 @@ public void onChange(ConfigurationOption<?> configurationOption, List<JmxMetric>
...
+                    registerListener(jmxMetric);
+                }
             }
         });
-        for (JmxMetricRegistration registration : compileJmxMetricRegistrations(jmxConfiguration.getCaptureJmxMetrics().get())) {
+        register(jmxConfiguration.getCaptureJmxMetrics().get());
+    }
+
+    private void register(List<JmxMetric> jmxMetrics) {
+        ArrayList<JmxMetric> notFound = new ArrayList<>();
```
SylvainJuge: might change variable type to List

apm-agent-plugins/apm-jmx-plugin/src/main/java/co/elastic/apm/agent/jmx/JmxMetricTracker.java
```diff
@@ -83,50 +96,131 @@ public void onChange(ConfigurationOption<?> configurationOption, List<JmxMetric>
...
+                                logger.warn("Can't create metric '{}' because composite value '{}' is not a number: '{}'", jmxMetric, key, value);
+                            }
                         }
+                    } else {
+                        logger.warn("Can't create metric '{}' because attribute '{}' is not a number: '{}'", jmxMetric, attribute.getJmxAttributeName(), value);
                     }
-                } else {
-                    logger.warn("Can't create metric '{}' because attribute '{}' is not a number: '{}'", jmxMetric, attribute.getJmxAttributeName(), value);
+                } catch (AttributeNotFoundException e) {
+                    logger.warn("Can't create metric '{}' because attribute '{}' could not be found", jmxMetric, attribute.getJmxAttributeName());
```
SylvainJuge: why not return 'false' here when there is an error ? Would that make retry forever in case of error ?

apm-agent-plugins/apm-jmx-plugin/src/test/java/co/elastic/apm/agent/jmx/JmxMetricTrackerTest.java
```diff
@@ -111,13 +116,54 @@ void testRemoveMetric() throws Exception {
         }
     }
 
+    @Test
+    void testMBeanAddedLater() throws Exception {
+        setConfig(JmxMetric.valueOf("object_name[foo:type=Bar] attribute[Baz]"));
+        assertThat(jmxMetricTracker.getRegisteredListeners()).hasSize(1);
+        ObjectName objectName = new ObjectName("foo:type=Bar");
+        ManagementFactory.getPlatformMBeanServer().registerMBean(new TestMetric(), objectName);
```
SylvainJuge: Maybe making MBeanServer instance injectable would definitely help testing without side effects on global state.

apm-agent-plugins/apm-jmx-plugin/src/test/java/co/elastic/apm/agent/jmx/JmxMetricTrackerTest.java
```diff
@@ -111,13 +116,54 @@ void testRemoveMetric() throws Exception {
...
+        try {
+            assertThat(jmxMetricTracker.getRegisteredListeners()).isEmpty();
+            assertThat(metricRegistry.getGauge("jvm.jmx.Baz", Labels.Mutable.of("type", "Bar"))).isEqualTo(42);
+        } finally {
+            ManagementFactory.getPlatformMBeanServer().unregisterMBean(objectName);
+        }
```
SylvainJuge: we probably miss an assertion on the number of listeners, which should be equal to zero when the bean is available. Also, what happens if a bean is registered only from time to time (thus multiple register/unregister cycles) ?

apm-agent-plugins/apm-jmx-plugin/src/main/java/co/elastic/apm/agent/jmx/JmxMetricTracker.java
felixbarny: It means that the MBean itself was found but that the attribute was not. This is likely a misspelling of the attribute name.

> Would that make retry forever in case of error ?

Yes, that's what I think

apm-agent-plugins/apm-jmx-plugin/src/test/java/co/elastic/apm/agent/jmx/JmxMetricTrackerTest.java
felixbarny: > we probably miss an assertion on the number of listeners, which should be equal to zero when the bean is available.

Isn't that done in line 126?

> Also, what happens if a bean is registered only from time to time

not quite sure what you mean with that

apm-agent-plugins/apm-jmx-plugin/src/test/java/co/elastic/apm/agent/jmx/JmxMetricTrackerTest.java
felixbarny: I agree that this would make it a bit cleaner. But the test setup would also be more complex as the other tests rely on the standard MBeans added in `java.lang.management.ManagementFactory#getPlatformMBeanServer`. Unfortunately, `java.lang.management.ManagementFactory#platformComponents` is private. That makes it impossible to create new `MBeanServer` for each test method which also contains the standard MBeans.

apm-agent-plugins/apm-jmx-plugin/src/main/java/co/elastic/apm/agent/jmx/JmxMetricTracker.java
```diff
@@ -83,50 +96,131 @@ public void onChange(ConfigurationOption<?> configurationOption, List<JmxMetric>
...
                 logger.error("Failed to register JMX metric {}", jmxMetric.toString(), e);
             }
         }
         return registrations;
     }
 
-    private void addJmxMetricRegistration(final JmxMetric jmxMetric, List<JmxMetricRegistration> registrations) throws JMException {
-        for (ObjectInstance mbean : server.queryMBeans(jmxMetric.getObjectName(), null)) {
+    private void registerListener(final JmxMetric jmxMetric) {
+        logger.info("Could not find mbeans for {}. Adding a listener in case the mbean is registered at a later point.", jmxMetric.getObjectName());
```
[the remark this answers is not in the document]
felixbarny: renaming to `registerListenerForNotFoundMBean`

apm-agent-plugins/apm-jmx-plugin/src/main/java/co/elastic/apm/agent/jmx/JmxMetricTracker.java
```diff
@@ -180,7 +241,7 @@ public int hashCode() {
     }
 
     @Override
-    public void stop() {
-
+    public void stop() throws Exception {
+        server.removeNotificationListener(MBeanServerDelegate.DELEGATE_NAME, listener);
```
SylvainJuge: probably better to avoid having `throws Exception` in the signature and handle it with a try/catch block here,
as expecting the method calling `stop()` to do it explicitly is not obvious as it is not part of the `LifecycleListener` interface definition.

apm-agent-core/src/main/java/co/elastic/apm/agent/context/LifecycleListener.java
```diff
@@ -32,14 +32,19 @@
...
  */
 public interface LifecycleListener {
 
     /**
      * Callback for when the {@link ElasticApmTracer} starts.
      *
      * @param tracer The tracer.
+     * @deprecated in favor of a constructor with an {@link ElasticApmTracer} argument
      */
+    @Deprecated
```
SylvainJuge: After thinking a bit more about this, I'm not sure that deprecating the `start` method and replace with constructor arguments is the perfect strategy here. As I might be close to bike-shedding I'll elaborate a bit:
- on the positive side, it allows to keep a final reference to `ElasticApmTracer` instance instead of having a mutable field
- on the other side, there is no enforcement at compile time for implementing any constructor contract, thus it's likely to produce errors at runtime when creating instances of such classes.
- if listener needs to start a thread (or do any kind of expensive operation) when starting, doing it in the constructor is generally not the recommended option, and keeping a `start` would probably provide better semantics in this case.

felixbarny: I agree. Let's not deprecate it but still make it possible to inject the tracer into the constructor so implementations can choose which is more appropriate for them. For the `JmxMetricTracker` not only the tracer instance variable can't be `final` but because the `ChangeListener` is invoked from another thread, it would also have to be `volatile`. Thinking about it, we must not register the listener within the constructor as there's a chance that the listener gets invoked before the constructor finishes. Another reason to keep both the constructor tracer argument and the `start` method.

eyalkoren:
Closing in favour of #900

PR #884, opened 2019-10-15
Fix setting start timestamp

eyalkoren:
Related to [this discussion](https://discuss.elastic.co/t/setting-start-timestamp-on-transaction-to-an-earlier-time-is-always-added-to-app-type/203615)

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
@@ -73,15 +73,18 @@ public abstract class AbstractSpan<T extends AbstractSpan> extends TraceContextH
         private AtomicLong start = new AtomicLong();
         private AtomicLong duration = new AtomicLong();
 
-        /**
-         * Starts the timer if it has not been started already.
-         *
-         * @param startTimestamp
-         */
-        void onChildStart(long startTimestamp) {
-            if (activeChildren.incrementAndGet() == 1) {
+        void overrideTimestamp(long startTimestamp) {
+            if (activeChildren.get() == 1) {
                 start.set(startTimestamp);
             }
+        }
+
+        /**
+         * Increments active children count.
+         *
+         */
+        void onChildStart() {
+            activeChildren.incrementAndGet();
         }
 
         /**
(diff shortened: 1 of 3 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -73,15 +73,18 @@ public int getReferenceCount() {
         private AtomicLong start = new AtomicLong();
         private AtomicLong duration = new AtomicLong();
 
+        void overrideTimestamp(long startTimestamp) {
+            if (activeChildren.get() == 1) {
```
felixbarny: Could you add a comment on why that is? I don't get it and it seems a bit arbitrary to only do it if there's one active child. Why does it not work with no children or with 2? What about race conditions? Should we log if the timestamp can't be overridden?

eyalkoren: The main problem is: if the parent span is already using the timer to count "non-self-duration" due to some child span- then we can't let it set it to **a later time**. This means we need a way to know if a span modifying the timestamp is the active child or another. If we allow both 0 and 1- we can't really tell that. Therefore, I made sure the child counter is increased when a child is started **before starting the counter**, so 0 becomes invalid and 1 means setting the timestamp is made by the active span.
As for >1, this is trickier...
We can say: always allow overriding the timestamp if child count is 1, otherwise allow overriding only if the given timestamp is older than the one already used. The problem is that when starting a span, we use the wall clock by default.
One option to deal with that is: start known async spans with no timestamp (eg -1) and be aware that it is expected to be updated later. Then when it does- override with it only if it is the only child or it is older than the already-set timestamp. 

@felixbarny WDYT?

felixbarny: Maybe we should just start all spans with a -1 timestamp and only determine it when the span ends. If it's still -1, it means that noone set an explicit start timestamp, thus use the one from the default clock.

eyalkoren: The clock only holds a offset to Epoch in nanos. You mean keeping a tmp timestamp on start and not use it until the span ends?

felixbarny: yes, that's what I mean

PR #890, opened 2019-10-17
Add warning message when transaction_max_spans is exceeded

### LINKED ISSUE #849: Log when tranasction_max_spans reached
Just testing elastic apm with java agent, with a long running transaction I've noticed no spans are recorded after a certain amount of time.
I have inspected the logs (my app & apm-server) with no trace of possible causes.
After more digging and time spent debugging, I've found the following configuration [transaction_max_spans](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#config-transaction-max-spans), that was causing spans to be ignored after 500 count.

A message should be logged when `transaction_max_spans` has been reached for a transaction, ex:
`Transaction max spans 500 has been reached for id c82e1e0bb6d86f12. Further spans will be ignored. See config param transaction_max_spans.`
The message should be logged with default configuration, so a level of INFO or WARN is appropriate.

zendern:
Simply as it says above. Fixes #849 

### Checklist
- [x] Implement code
- [x] Update documentation


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -317,6 +317,7 @@ public Span startSpan(AbstractSpan<?> parent, long epochMicros) {
         Transaction transaction = currentTransaction();
         if (transaction != null) {
             if (isTransactionSpanLimitReached(transaction)) {
+                logger.warn("Max spans ({}) for a transaction has been reached for transaction {}. Further spans will be ignored. See config param 'transaction_max_spans'.", coreConfiguration.getTransactionMaxSpans(), transaction);
```
eyalkoren: We can't log it at every span creation attempt. If the max is 500 and transactions have 1000 on average, we will get 500 log messages **for each transaction**. Maybe the simplest is log once (at most) per time frame, say- 5 minutes. WDYT?

zendern:
> Thanks for the PR!
> Please see the comment.
> Also, in order for us to be able to merge this PR, please sign the CLA.

Eh I've tried twice now. Originally I added `@` in my github username and figured since it didn't pick right up I should have maybe not done that and did it a second time without it. Any thoughts or should I just wait longer for the bot to pick it up??

eyalkoren:
We're good now 👍

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
eyalkoren: BTW, this method is invoked with a timestamp, so you can use it for that- will spare a couple of CPU cycles for not querying the system timing APIs 🤓

zendern: Actually what might make more sense is to just move the check and log into `ElasticAPMTracer.endTransaction()` of the transaction. That way at least when the transaction is finished we can report out that the spans were not sent due to the max being met.

eyalkoren: We don't want to log for each transaction instance and even making sure we log for each transaction type feels overkill. All we want is make the user aware that there is a situation of losing data due to exceeding a maximum. Since such messages can get quickly lost over time as logs accumulate, I suggested logging every 5 minutes. Unfortunately, this implies we need to maintain a state. I don't like that too much as well, but I think it is fine in this case.

As long as transactions are shorter than 5 minutes, this suggested code have the same issue. The diff between the span timestamp and the transaction timestamp is irrelevant.

So I suggest you just introduce this new `lastSpanMaxWarningTimestamp` state and use it.

Two notes:
- Why modulus and not `if (epochMicros - lastSpanMaxWarningTimestamp > MAX_LOG_INTERVAL_MICRO_SECS`)?
- No need for `logger.isWarnEnabled()` - in this case it will cost more than it saves

zendern: Thanks for clarifying. I'll implement it that way. 

Re: Why modulus.... no reason really. Should have probably been `>` you're right.

zendern: Should be fixed based on fb95082

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -317,6 +321,10 @@ public Span startSpan(AbstractSpan<?> parent, long epochMicros) {
         Transaction transaction = currentTransaction();
         if (transaction != null) {
             if (isTransactionSpanLimitReached(transaction)) {
+                if(lastSpanMaxWarningTimestamp == 0L || (epochMicros - lastSpanMaxWarningTimestamp > MAX_LOG_INTERVAL_MICRO_SECS)) {
```
felixbarny:
```suggestion
                if (lastSpanMaxWarningTimestamp == 0L || (epochMicros - lastSpanMaxWarningTimestamp > MAX_LOG_INTERVAL_MICRO_SECS)) {
```

docs/configuration.asciidoc
```diff
@@ -333,6 +333,7 @@ This is helpful in cases where a transaction creates a very high amount of spans
 
 Setting an upper limit will prevent overloading the agent and the APM server with too much work for such edge cases.
 
+A message will be logged when the limit has been exceeded.
```
felixbarny: This is a generated file. Please add this to `CoreConfiguration` and execute the `ConfigurationExporterTest`.

zendern: Fixed in c8c5900

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -317,6 +321,10 @@ public Span startSpan(AbstractSpan<?> parent, long epochMicros) {
         Transaction transaction = currentTransaction();
         if (transaction != null) {
             if (isTransactionSpanLimitReached(transaction)) {
+                if (lastSpanMaxWarningTimestamp == 0L || (epochMicros - lastSpanMaxWarningTimestamp > MAX_LOG_INTERVAL_MICRO_SECS)) {
```
eyalkoren:
```suggestion
                if (epochMicros - lastSpanMaxWarningTimestamp > MAX_LOG_INTERVAL_MICRO_SECS) {
```
The `lastSpanMaxWarningTimestamp == 0L` check is redundant.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -317,6 +321,10 @@ public Span startSpan(AbstractSpan<?> parent, long epochMicros) {
...
+                    lastSpanMaxWarningTimestamp = epochMicros;
+                    logger.warn("Max spans ({}) for a transaction has been reached. For this transaction and possibly others, further spans will be dropped. See config param 'transaction_max_spans'.", coreConfiguration.getTransactionMaxSpans());
```
eyalkoren: The transaction name would be a useful hint even if it's only an example:
```suggestion
                    logger.warn("Max spans ({}) for transaction {} has been reached. For this transaction and possibly others, further spans will be dropped. See config param 'transaction_max_spans'.", coreConfiguration.getTransactionMaxSpans(), transaction.getName());
```

zendern: Fixed in e910d86. `transaction.getName()` wasn't a thing and the `transaction.getNameAsString()` mentions only to use in test. So just defaulted to the toString() on transaction. I also updated the docs to call out the log statement is only after a period of time since adding the transaction in there made it feel like it needed called out.

eyalkoren: The usage warning is intended to avoid the String object allocation, but since this happens once every 5 minutes at most, this is not an issue. 
Let's merge it as is though.
Thanks a lot!

PR #897, opened 2019-10-21
Add a comment on every configuration option clarifying that configuration.asciidoc is auto-generated

felixbarny:


felixbarny:
No, we currently don't. But mostly, it would be vice-versa - for example, we update the description in `CoreConfiguration.java` but forget to generate the asciidoc file. In that case, nothing bad happens. The next time someone executes the tests and creates a PR, the file will be updated. So it will basically be eventually consistent. It's not perfect and maybe a bit annoying but probably does not warrant to put too much brain power into it. Having said that, if you have a good idea of how to tackle that in a simple way I'd be glad for such a check.

SylvainJuge:
In fact, if the test that updates documentation is run on every build, we just have to check that file has not been modified by doing a kind of "git diff".

If I remember correctly, this is exactly what the maven-release plugin does at `release:prepare` phase, so we might already covered, even if failing at release time might feel a bit late (but would be definitely easy to fix).

felixbarny:
Great idea. We could also make the test fail if the file is not up-to-date and create a main method that has to be executed to update the file. The test should give a hint on fail that the main method should be executed. That way the tests on a PR would fail if the file was not updated.

PR #911, opened 2019-10-31
JMS - some more enhancements

eyalkoren:
Continue implementation of #664, addressing the followings:
- Add special care for `javax.jms.TemporaryQueue/Topic` and Tibco temp queues/topic
- Capture message bodies of text Messages
    - Rely on the existing `ELASTIC_APM_CAPTURE_BODY` agent config option.
    - Send as `context.message.body`
    - Limit size to 10000 characters. If longer than this size, trim to 9999 and append with ellipsis
- Disable instrumentation (message tagging) for specific queues/topics as suggested in elastic/apm-agent-java#710 
- Capture predefined message headers and all properties
    - Rely on the existing `ELASTIC_APM_CAPTURE_HEADERS` agent config option.
    - Send as `context.message.headers`
    - Sanitize sensitive headers/properties based on the `sanitize_field_names` config option

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation
- [ ] ~~Update [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)~~ (no need)


apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
```diff
@@ -110,20 +137,74 @@ public void onMessage(Message message) {
...
+
+            // Currently only capturing body of TextMessages. The javax.jms.Message#getBody() API is since 2.0, so,
+            // if we are supporting JMS 1.1, it makes no sense to rely on isAssignableFrom.
+            if (coreConfiguration.getCaptureBody() != CoreConfiguration.EventType.OFF && message instanceof TextMessage) {
+                messageContext.withBody(((TextMessage) message).getText());
+            }
+
+            // Addition of non-String headers/properties will cause String instance allocations
+            if (coreConfiguration.isCaptureHeaders()) {
+                messageContext.addHeader(JMS_MESSAGE_ID_HEADER, message.getJMSMessageID());
```
felixbarny: This only gets some headers, why not get all via `getPropertyNames`/`getObjectProperty`?

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
```diff
@@ -110,20 +137,74 @@ public void onMessage(Message message) {
...
+                    destinationName = ((Topic) destination).getTopicName();
+                }
+            } catch (JMSException e) {
+                logger.error("Failed to obtain destination name", e);
+            }
+        }
+        return destinationName;
+    }
+
+    private boolean isTempDestination(Destination destination, @Nullable String extractedDestinationName) {
```
felixbarny: Could you describe some high-level concepts around temp queues in `co/elastic/apm/agent/jms/package-info.java`? For example, what are temp queues used for, are they only used in specific implementations ( is `$TEMP$` a convention?),  and why do we need to treat them specially?

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
```diff
@@ -74,11 +96,16 @@ public Span startJmsSendSpan(Destination destination, Message message) {
...
         try {
             if (span.isSampled()) {
-                span.withName("JMS SEND to ");
-                addDestinationDetails(destination, span);
+                message.setStringProperty(JMS_TRACE_PARENT_PROPERTY, span.getTraceContext().getOutgoingTraceParentHeader().toString());
+                if (destinationName != null) {
+                    span.withName("JMS SEND to ");
+                    addDestinationDetails(null, destination, destinationName, span);
+                    if (isDestinationNameComputed) {
+                        message.setStringProperty(JMS_DESTINATION_NAME_PROPERTY, destinationName);
```
felixbarny: why do we have to propagate that downstream and why only for temp queues?

eyalkoren:
I totally agree with everything 🙂 - it should have been few small PRs. Because they are dependent on each other, I didn't want to block on waiting for review/approval, but probably the wrong call.
I will add more documentation in code. I hoped the PR main comment would provide the context. I assume everything in this comment will go into changelog (which I always do last, to reduce merges 😊 )

apm-agent-core/src/main/java/co/elastic/apm/agent/util/NoRandomAccessMap.java
```diff
@@ -0,0 +1,142 @@
...
+
+        public K getKey() {
+            if (key == null) {
+                throw new IllegalStateException("Key shouldn't be null");
+            }
+            return key;
+        }
+
+        public V getValue() {
+            if (value == null) {
```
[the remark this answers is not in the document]
eyalkoren: It could serve as a good indication that there are errors with iteration/reset/addition (either a bug, or a misuse due to concurrency). However, I need to enforce that on insertion as well (currently it is only enforced through an IDE warning, not enough)

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
eyalkoren: I will add to the code. In general, those are short-lived ad-hoc queues, for example- a reply queue. This is why they require different handling. Non-temp queues are globally known, very stable entities, like databases. You want those in your service map, but tmp queues will cause a disaster.

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
eyalkoren: The JMS `Destination` APIs (`Queue` and `Topic`) provide a `getName()` method. When we create a tmp queue explicitly through the `createTmpQueue()` API on the sender side, its type will be indicative of that, and we know not to use `getName()` (as it will create high cardinality in transaction names and in the service map, once we have such). However, on the receiver side, the queue/topic type is not guaranteed to indicate that it is temp, as it doesn't necessarily know that it is a tmp queue, so it may use what it gets from `getName`, which will be incompatible with the sender span.

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
eyalkoren: In JMS, headers are predefined and properties are custom. 
For headers- we don't have an API to get by name, only through dedicated APIs. So I chose the three I thought would be most interesting. Once another is requested, it is now super-easy to add.

As for properties- the `javax.jms.Message` only offers `getXXXProperty()` APIs, where `XXX` is an explicit type... So, the only way to get all properties is to iterate over the `Enumeration` returned from `getPropertyNames()` and try all types one-by-one....
I decided it is too much for now and we can reconsider if we see demand.

eyalkoren: Ohh, you're right! I completely overlooked it- the `getObjectProperty` is actually what I need. Thanks!!

apm-agent-core/src/main/java/co/elastic/apm/agent/util/NoRandomAccessMap.java
```diff
@@ -47,7 +47,16 @@
     private List<V> values = new ArrayList<>();
     private NoGarbageIterator iterator = new NoGarbageIterator();
 
-    public void add(K key, V value) {
+    /**
+     * @param key   can't be null
+     * @param value can't be null
```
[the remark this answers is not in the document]
eyalkoren: Currently it serves me well to limit - see https://github.com/elastic/apm-agent-java/pull/911#discussion_r347422847

PR #912, opened 2019-11-01
Exclude Jps from discovered JVMs

felixbarny:
- [x] Implement code
- [x] Add tests


felixbarny:
Does not work because `Process#pid` has only been introduced in Java 9

PR #914, opened 2019-11-03
Avoid using agent args when attaching

felixbarny:
Because they are limited in size

### DIFF
--- a/apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
+++ b/apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
@@ -117,19 +117,7 @@ public class ElasticApmAttacher {
      * @throws IllegalStateException if there was a problem while attaching the agent to this VM
      */
     public static void attach(Map<String, String> configuration) {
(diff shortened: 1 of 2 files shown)


apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
```diff
@@ -168,15 +156,29 @@ static String toAgentArgs(Map<String, String> configuration) {
...
     }
 
     /**
      * Attaches the agent to a remote JVM
      *
      * @param pid       the PID of the JVM the agent should be attached on
      * @param agentArgs the agent arguments
+     * @deprecated use {@link #attach(String, Map)}
      */
+    @Deprecated
```
eyalkoren: Why deprecating and not removing?

felixbarny: Because it's a public API

eyalkoren: With PID? It's only relevant to remote attach, isn't it?

PR #919, opened 2019-11-09
Added error instrumentation for log4j loggers.

petermylemans:
Add error instrumentation for log4j (2.x.x) loggers. 

Support for slf4j loggers already existed, so this PR extends it to the lo4j2 API. With this PR calls to 
`org.apache.logging.log4j.Logger#error(java.lang.String, java.lang.Throwable)` are captured as errors.

### Checklist

- [x] Implement code
- [X] Add tests
- [X] Update documentation
- [X] Update [CHANGELOG.md](CHANGELOG.md)
- [X] Update [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)


petermylemans:
Correct, good catch. The two bridge implementations seem to be using each others API. 

I'll have a look on how to exclude these from both the slf4j and log4j error capture code.

elasticcla:
Hi @petermylemans, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/apm-agent-java/pull/919.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

petermylemans:
@eyalkoren  it accidentally went right, because the bridge implementations use another low level method to delegate the log messages and not the signature that was included in the pointcut. 

But we should probably bullet proof the implementation, so I've added:
 * mutual exclusion to the type filter match to prevent accidental double activation of the advise
 * added a test for the log4j > slf4j bridge. I can't test both in the same maven project, because introducing both bridge dependencies would introduce an endless loop.

The case that remains is delegation pattern as implemented by the log4j2-slf4j-impl:
https://github.com/apache/logging-log4j2/blob/master/log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jLogger.java

   @Override
    public void error(final String format, final Throwable t) {
        **logger.logIfEnabled**(FQCN, Level.ERROR, null, format, t);
    }

eyalkoren:
I see what you did, but not sure this the most bulletproof approach. I didn't look at specific bridges, but I can imagine an implementation of a bridge where two different classes implement each a different interface and they are bridged by delegation.
Besides, we may want a more generic approach to add some kind of awareness to slf4j, so we can reuse also for #922.

eyalkoren:
@petermylemans You just hit an issue we want to address in general, where we want to trace the actual logging implementation wherever slf4j is not used, but still trace slf4j where used without duplicating events. 
We thought of introducing a state to the error-capturing logic, relying on the fact that when a logging implementation is used with slf4j, its calls will be nested within the slf4j calls. A bit too difficult to describe in writing, so I will make a PR on your PR suggesting this approach. It make take a couple of days, though, as I need to address some other stuff. I hope this is OK.

petermylemans:
That's perfect!

Just as an additional note about double capturing. I'm not familiar with the entire code base, but what would happen if e.g. the TraceMethod instrumentation records an exception that is later catched and logged as an error resulting in a capture by the Logging instrumentation. 

A typical example would be an application that has a trace method annotation in the service layer and has an error handler + logger in the rest layer.

My point being: this could be an issue cross-instrumentation, not only the interference of slf4j and log4j.

eyalkoren:
Yes, you're right and we discussed that as well. If an Exception is caught, logged and re-thrown - we would have duplications. Even worse- if someone uses the `trace_methods` config with a base package and we therefore instrument a big stack and an Exception is thrown all across this stack depths- each will report an error...
We are aware and will defer the handling of those cases to some other PR.
Thanks for raising this!

eyalkoren:
@petermylemans Sorry for the delay, was a bit occupied with other stuff...
Please see my suggestion in the PR made on this branch.

petermylemans:
@eyalkoren merged and conflicts should be resolved. Let's see how this goes.

eyalkoren:
@petermylemans please run `mvn clean test` locally. This should update the `configuration.asciidoc` file with the added instrumentation groups. Then push the updated file, so the build can pass.
Thanks.

petermylemans:
Done. I've had some issues with platform encoding on a clean mac os x: see https://github.com/elastic/apm-agent-java/pull/947 for a proposed resolution.

Build should be fixed now. Do you want to squash the commits? Makes more sense that you are the author as you did most of the work? (Although it was very cool to learn a bit about your project that way! Awesome work btw!)

eyalkoren:
Awesome work! Thanks for this PR and the other as well!
We will squash as we do with all PRs for a cleaner history, but without regard to the importance of your contribution.
Let me just run the tests first.

SylvainJuge:
@petermylemans the only requirement before merging is to merge conflicts `CHANGELOG.md`, I can handle adding the extra comments before merging.

petermylemans:
@SylvainJuge I was on holiday, hence the small delay. Conflicts and tests are fixed.

I'll leave adding the extra comments to you?

PR #920, opened 2019-11-11
Replace underscore with hyphen for OKD pod UID

eyalkoren:
Fixes the OKD pod UID discovery error reported through [Discuss](https://discuss.elastic.co/t/infrastructure-ui-kubernetes-apm-traces-integration/207096).

Related to https://github.com/elastic/apm/pull/165

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/SystemInfo.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/SystemInfo.java
@@ -11,9 +11,9 @@
  * the Apache License, Version 2.0 (the "License"); you may
  * not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -94,7 +94,7 @@ public class SystemInfo {
         }
         return new SystemInfo(System.getProperty("os.arch"), reportedHostname, System.getProperty("os.name")).findContainerDetails();
     }
-    
+
     public static SystemInfo create() {
         return create(null);
     }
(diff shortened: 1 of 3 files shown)


eyalkoren:
Yes, that's at least a relevant thing to let others know. 
The thing is that I am not yet sure whether the right approach is what I did (ie - assume that a pod UID should always have hyphens and never underscores) or look for this specific pattern (and leave the option of using an underscore ID with other K8s systems).
Once I have an opinion on that, I will open that issue.

felixbarny:
Actually, this is also relevant for beats. Maybe there's even something in https://github.com/elastic/beats/blob/master/libbeat/common/kubernetes/metadata.go already?

eyalkoren:
@elastic/beats can someone shed some light here- how do you obtain the Kubernetes pod ID? metricbeat seem to be doing it right. Not sure about libbeat. 
We are parsing the `/proc/self/cgroup` file lines for this purpose, but it seems that in some cases the pod UID we get from this source has underscores replacing the hyphens.

urso:
@exekias Can you comment? In some places beats query the pid and then try to parse `/proc/<pid>` instead of reading from `/proc/self`. But not sure about cgroup info here.

felixbarny:
When making changes to the parsing logic, make sure to update https://github.com/elastic/apm/blob/master/docs/agent-development.md#containerkubernetes-metadata

@exekias can you review that spec and make sure it's in sync with what beats does?

exekias:
:wave: As far as I know we don't use cgroup parsing to retrieve Kubernetes Pod UIDs. Both autodiscover and metadata processor rely on Kubernetes API to retrieve the Pod info, so this transformation is not really needed there.

As per the spec, again, we don't do that, but it looks good to me

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/SystemInfo.java
```diff
@@ -191,6 +191,7 @@ SystemInfo parseContainerId(String line) {
                         for (int i = 1; i <= matcher.groupCount(); i++) {
                             String podUid = matcher.group(i);
                             if (podUid != null && !podUid.isEmpty()) {
+                                podUid = podUid.replace('_', '-');
```
axw: The hyphens are only escaped when using the systemd cgroup driver:

https://github.com/kubernetes/kubernetes/blob/6d0994fa66e34db874e4e26c86b55ba34e75d1c1/pkg/kubelet/cm/cgroup_manager_linux.go#L86

I think we should make this conditional on the path ending with ".scope"

eyalkoren: Sure thing. See my latest commit.

PR #927, opened 2019-11-12
Refactor plugin versions & cleanup pom.xml

SylvainJuge:
## Refactor `pom.xml`
- plugin versions defined at parent module level through `pluginManagement`
- remove default packaging `jar` by default
- remove default `groupId` for plugins `org.apache.maven.plugins`
- dependencies/plugin version properties kept only when used more than once or needs to be kept in sync (spring version, animal sniffer plugin/annotations, ...)
- try to enforce a consistent order of elements in `pom.xml`

### Benefits
- enhanced readeability
- fewer properties
- remove duplication and default values
- same plugin versions used across all modules (might be overloaded locally to a module if needed)

### Risks
- some modules are now built with different plugin versions
- integration tests should ensure that agent is properly packaged and usable

### Checklist

- [x] Implement code


felixbarny:
Thx for cleaning this up, LGTM

PR #931, opened 2019-11-14
Improve HttpUtils test coverage (initially 'remote config verbose log)

SylvainJuge:
### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation
- [ ] Added an API method or config option? Document in which version this will be introduced.


apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ApmServerConfigurationSource.java
```diff
@@ -191,7 +191,7 @@ private String tryFetchConfig(ConfigurationRegistry configurationRegistry, HttpU
                 reader.startObject();
                 config = MapConverter.deserialize(reader);
                 configurationRegistry.reloadDynamicConfigurationOptions();
-                logger.info("Received new configuration from APM Server: {}", config);
+                logger.debug("Received new configuration from APM Server: {}", config);
```
felixbarny: Why did the server not respond with `304 Not Modified`? Is there something wrong with the etag?

SylvainJuge:
That's a very good question, I checked locally with 7.5 version.
I will check again what is the response code from the server and drop this
PR if that's actually a server bug.

SylvainJuge:
After checking again with 7.5 server locally, it appears that the Etag value is always the same (but looks empty or only contain `""` which definitely seems to be the root cause.

felixbarny:
I'm ok with this change but I still think that the canonical place to fix this is the APM Server. If the server responds with a 200 after making a conditional get request with an Etag, the assumption is that there have been changes.

Currently, the APM Server responds with an Etag of `""` if there is no configuration (`{}`). When the agent makes a subsequent conditional get (`If-None-Match: ""`), the APM Server responds with `200 {}` with the same Etag value of `""` instead of responding with `304 Not Modified`.

The problem is that we log and reload the configuration if we get a 200 back as we assume there are changes in the configuration. With the current setup, we spam the log every `30s` with `Received new configuration from APM Server: {}` log messages. Also, we trigger a reload of the dynamic configuration options. This is not too costly but seems like unnecessary work if nothing has actually changed.

/cc @jalvz

SylvainJuge:
I agree on this @felixbarny, and that also means that implement this kind of detail will be required for all agents. I would definitely be happy to get rid of this extra logic and just follow HTTP Etag, the fact that this Etag value comes from Kibana is an implementation detail and should not impact agents.

felixbarny:
Relates to https://github.com/elastic/apm-server/issues/2916

SylvainJuge:
server issue has been fixed, will likely be shipped in 7.5.1 version.
Thus we can drop the "detecting changes part of this PR".

jalvz:
> server issue has been fixed, will likely be shipped in 7.5.1 version.

7.5.0 actually!

SylvainJuge:
Even better than expected, :+1:  @jalvz for the update :-)

PR #962, opened 2019-12-08
Work around JULI deadlock

### LINKED ISSUE #954: Capture jmx metrics not working on Tomcat
hi, I have problem with capture JMX metrics. when I specify capture_jmx_metrics in APM configuration file, my Tomcat web server can't start. my configuration is: 
service_name=QC
application_packages=com.somepackage
server_urls=http://apm-url:8200
capture_jmx_metrics= object_name[java.lang:type=Memory,name=*] attribute[HeapMemoryUsage:metric_name=heap]
secret_token=mySecretToken

and Tomcat log:
[tomcat@myhost bin]$ ./catalina.sh run
Using CATALINA_BASE:   /Tomcat
Using CATALINA_HOME:   /Tomcat
Using CATALINA_TMPDIR: /Tomcat/temp
Using JRE_HOME:        /Tomcat/jdk1.8.0_152
Using CLASSPATH:       /Tomcat/bin/bootstrap.jar:/Tomcat/bin/tomcat-juli.jar
2019-12-04 10:48:29.416 [main] INFO co.elastic.apm.agent.configuration.StartupInfo - Starting Elastic APM 1.12.0 as QC on Java 1.8.0_152 (Oracle Corporation) Linux 3.10.0-862.11.6.el7.x86_64

and then I have to kill -f tomcat process. when I comment capture_jmx_metrics everything is OK and my web server starts normally.

eyalkoren:
Fixes #954 

Not the most elegant or generic way to work around this deadlock.
There are multiple ways to work around it, but I wouldn't want to synchronize the healthcheck with the JMX tracker intialization, as this could block the main thread waiting for the healthcheck to be done, which defeats the purpose of doing it on a separate thread. 
If we want to invest more time on it, we can move the JMX tracker initialization to another thread and then sync with the healthcheck, but this requires a bit of effort due to listening on the JMX configuration.
So, in order to provide a quick fix without investing too much effort- this should be good enough.

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
@@ -79,6 +79,7 @@ public class ApmServerClient {
     }
 
     public ApmServerClient(ReporterConfiguration reporterConfiguration, List<URL> shuffledUrls) {
+        initHttpUrlConnectionClass();
         this.reporterConfiguration = reporterConfiguration;
         this.healthChecker = new ApmServerHealthChecker(this);
         this.reporterConfiguration.getServerUrlsOption().addChangeListener(new ConfigurationOption.ChangeListener<List<URL>>() {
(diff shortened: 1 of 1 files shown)


SylvainJuge:
Fine for a quick workaround, what would be the best option to avoid this in a more generic way ?
- fix the JDK implementation, maybe by instrumenting `java.util.logging.LogManager.getLogManager(LogManager.java:378)` and adding an extra lock
- use another HTTP client that does not relies on JUL

However, we still have JUL init being triggered by our JMX thread, thus maybe having a way to detect "when is JUL properly initialized" would be better.

PR #967, opened 2019-12-11
improve objectpool

SylvainJuge:
### Checklist

- [x] Implement code
- [x] Add tests


apm-agent-core/src/main/java/co/elastic/apm/agent/objectpool/ObjectPool.java
```diff
@@ -27,19 +27,50 @@
...
     T tryCreateInstance();
 
+    /**
+     * Tries to reuse any existing instance if pool has any, otherwise creates a new un-pooled instance
+     *
+     * @return object instance, either from pool or freshly allocated
+     */
     T createInstance();
 
     void fillFromOtherPool(ObjectPool<T> otherPool, int maxElements);
```
SylvainJuge: @felixbarny do you have any suggestion for this method javadoc ?

apm-agent-core/src/main/java/co/elastic/apm/agent/objectpool/impl/AbstractObjectPool.java
```diff
@@ -26,45 +26,60 @@
...
         return garbageCreated.longValue();
     }
 
+    /**
+     * Pushes object reference back into the available pooled instances
+     *
+     * @param obj recycled object to return to pool
+     * @return true if object has been returned to pool, false if pool is already full
+     */
+    abstract protected boolean returnToAvailablePool(T obj);
```
felixbarny:
```suggestion
    abstract protected boolean returnToPool(T obj);
```

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
```diff
@@ -313,12 +315,44 @@ public void decrementReferences() {
...
+            assertThat(hasEmptyTraceContext.test(s))
+                .describedAs("should have empty trace context : %s", s)
+                .isTrue();
+            assertThat(s.isReferenced())
+                .describedAs("should not have any reference left, but has %d : %s", s.getReferenceCount(), s)
+                .isFalse();
+        });
+
+        // errors are recycled directly because they have no reference counter
+        errors.forEach(ErrorCapture::recycle);
```
felixbarny: Call `checkAllPooledObjectsHaveBeenRecycled` here?

apm-agent-core/src/test/java/co/elastic/apm/agent/MockTracer.java
```diff
@@ -57,9 +59,27 @@ public static ElasticApmTracer createRealTracer(Reporter reporter) {
...
+            .withLifecycleListener(new LifecycleListener() {
+                @Override
+                public void start(ElasticApmTracer tracer) {
+
+                }
+
+                @Override
+                public void stop() {
+                    // checking proper object pool usage using tracer lifecycle events
+                    objectPoolFactory.checkAllPooledObjectsHaveBeenRecycled();
```
felixbarny: Call `MockReporter#assertRecycledAfterDecrementingReferences` here? Otherwise, we're calling `checkAllPooledObjectsHaveBeenRecycled` before the objects have been recycled.

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
SylvainJuge: Since this class does not have any reference to the ObjectPool implementation with bookkeeping, I'd rather keep that as two method calls like in `ElasticApmTracerTest.java`

```
    @AfterEach
    void cleanupAndCheck() {
        reporter.assertRecycledAfterDecrementingReferences();
        objectPoolFactory.checkAllPooledObjectsHaveBeenRecycled();
    }
```

felixbarny: got it, makes sense

apm-agent-core/src/test/java/co/elastic/apm/agent/MockTracer.java
SylvainJuge: As in this case `tracer` is not always an instance of `MockTracer`, I'd rather keep it as is, there is now at least a couple of ways to assert for this:
1. by calling `MockReporter#assertRecycledAfterDecrementingReferences` then `TestObjectPoolFactory#checkAllPooledObjectsHaveBeenRecycled`
2. by calling `MockReporter#assertRecycledAfterDecrementingReferences` then `MockTracer#stop()`, which uses tracer lifecycle listener.

The nice thing here is that when we do call things in the wrong order, we make most of the test fail, which makes it quite obvious.

felixbarny: But currently, calling `MockTracer#stop()` would fail when not calling `MockReporter#assertRecycledAfterDecrementingReferences`, wouldn't it?

what about this:
```suggestion
                    if (reporter instanceof MockReporter) {
                        ((MockReporter) reporter).assertRecycledAfterDecrementingReferences();
                        objectPoolFactory.checkAllPooledObjectsHaveBeenRecycled();
                    }
```

SylvainJuge: That's definitely a good idea ! Then we just need to make sure `MockTracer#stop()` is properly called in tests, which could also be asserted, maybe with a JVM shutdown hook (but that might be overkill).

apm-agent-core/src/test/java/co/elastic/apm/agent/MockTracer.java
```diff
@@ -57,9 +59,25 @@ public static ElasticApmTracer createRealTracer(Reporter reporter) {
...
+            // is left behind
+            .withObjectPoolFactory(objectPoolFactory)
+            .withLifecycleListener(LifecycleListener.ClosableAdapter.of(() -> {
+
+                if (reporter instanceof MockReporter) {
+                    ((MockReporter) reporter).assertRecycledAfterDecrementingReferences();
+                }
+
+                // checking proper object pool usage using tracer lifecycle events
+                objectPoolFactory.checkAllPooledObjectsHaveBeenRecycled();
```
felixbarny: I think it makes sense to move this into the `if` scope. If the reporter is just a `Mockito.mock(Reporter.class)`, the references are not decremented and `objectPoolFactory.checkAllPooledObjectsHaveBeenRecycled();` would fail.

SylvainJuge: I'm not sure about this one, as there are also some items that are pooled but do not have any references counter (like errors), I think we might change this if we have a case that fails. Right now it seems that only a few places use MockTracer.

PR #972, opened 2019-12-16
Profiler-inferred spans

### LINKED ISSUE #971: Inferring spans representing method executions based on a statistical profiler
**Motivation**
Our agents mostly capture I/O events like incoming and outgoing HTTP requests. So when most of the time s spend there, it’s easy to troubleshoot latency issues. However, if the application is slow because of inefficient code, users have to manually instrument the application. But in some cases, it’s not feasible to do that, for example, if the code base is huge and it’s unclear which methods cause the slowdown.

The [`trace_methods`](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#config-trace-methods) configuration option is often used to match a large portion, if not all methods within a codebase. The documentation warns about the fact that this can significantly increase the overhead and degrade the application's performance.

The slowdown is inherent to the way `trace_methods` works. All methods which match the `trace_methods` expression are instrumented so that a timer is started at the start and stopped at the end of the method. If the execution time was significant enough ([`trace_methods_duration_threshold`](https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#config-trace-methods)), a span is created. When frequently executing methods are instrumented, it can significantly slow down the application by adding just a little overhead to each invocation. Instrumenting the method can also hinder optimizations the JIT could normally do, like inlining.

**The new approach**
This issue is about adding an alternative to `trace_methods` which does not require to instrument any methods. Instead, a sampling aka statistical profiler would be used as a foundation. These profilers work by gathering the stack trace of the application at frequent intervals, like every 20ms.

By correlating the stack traces with when which span has been active on which thread, we can create a call tree based on the stack traces, correlate them to a span and create spans for it. As the UI is concerned, those spans look just like regular spans so no changes are required in the UI and the APM Server. At a later stage, we could make the UI aware of the profiler-inferred spans and display them with a special icon or color.

**The tradeoffs**
The duration of the spans won’t be as accurate as we’re not exactly measuring the execution time but rather estimate the duration based on the number of consecutive stack traces a method has been present.

To reduce the overhead, the profiler won’t be active all the time. Instead, it’s active for the first 10 seconds of every minute, by default. Only transactions that happen within a profiling session will have profiler-inferred spans.

**Q&A**
* Will `trace_methods` be removed?
  There are no plans to remove that option as it can still be useful in combination or instead of profiler-inferred spans.
* How can I try this out?
  Use a snapshot form https://github.com/elastic/apm-agent-java/pull/972 or build the https://github.com/felixbarny/apm-agent-java/tree/inferred-spans branch.
* How does this relate to https://github.com/elastic/apm/issues/121?
  * The CPU profiling proposal is about having a macro-level view of what the service is doing. This is great to find out about the hotspots of an application. Optimizing those can have a big overall effect on the application. However, it's less useful to troubleshoot latency if the application is mostly idle waiting for I/O.
  * This issue is about creating regular spans for long executing methods. It's mostly used to troubleshoot latency for a specific instance of a transaction.
  * There's no intention to deprecate one in favor of the other. Both are very useful tools to have at your belt to optimize an application.
  * Both can use the same underlying sampling profiler. For the CPU profiling, the stack trace is only considered for threads in a `RUNNABLE` thread state. Also, for CPU profiling, there will be only one flattened data structure, representing a flame graph for the whole profiling session.
* Which settings are there to configure the profiler? See the [docs preview](http://apm-agent-java_983.docs-preview.app.elstc.co/guide/en/apm/agent/java/master/config-profiling.html)

**Screenshot**
Screen Shot 2019-12-04 at 16 07 52

felixbarny:
Closes #971 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/SamplingProfiler.java
```diff
@@ -0,0 +1,287 @@
...
+            }
+        }
+    }
+
+    private void takeThreadSnapshot() {
+        List<WildcardMatcher> excludedClasses = config.getExcludedClasses();
+        List<WildcardMatcher> includedClasses = config.getIncludedClasses();
+        long nanoTime = System.nanoTime();
+        processActivationEventsUpTo(nanoTime);
+        long[] profiledThreadIds = getProfiledThreadIds();
```
eyalkoren: What if, instead of only thread IDs, you would use a different method that returns a map between thread ID and the active span. Then, right after you get the stack traces, you check again and you only use the stack trace of a thread if the current active span is still the same as before the capturing and if not- you discard it altogether (which is not so bad, as if there was only one such it is not interesting anyway). Can this fix the problem of wrong child-parent inference?

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/SamplingProfiler.java
```diff
@@ -0,0 +1,287 @@
...
+        ThreadInfo[] threadInfos = threadMXBean.getThreadInfo(profiledThreadIds, Integer.MAX_VALUE);
+        if (profiledThreadIds.length == 0) {
+            return;
+        }
+        logger.trace("Taking snapshot of threads {} timestamp {}", profiledThreadIds, nanoTime);
+
+        List<StackFrame> stackTraces = new ArrayList<>(256);
+        for (int i = 0; i < profiledThreadIds.length; i++) {
+            long threadId = profiledThreadIds[i];
```
eyalkoren: So here, if active span is not the same as you saw before taking the stack trace- discard this stack trace.
You just need to make sure the Root#activeSpan is volatile

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/SamplingProfiler.java
felixbarny: Are you suggesting to add inferred spans to the activation stack? That doesn't work if we want to integrate with async-profiler, for example. That's because we'd record all activity within a profiling session and only afterwards process activation events and stack traces. Also, we don't know which CallTree nodes to convert to a span until they have ended.

> To reduce the number of spans we create, we only create spans for CallTree nodes which have appeared in at least two consecutive samples (count >= 2). Also, if a node has only one child and that child has the same count we don’t create a span for it but instead add it to the span.stacktrace of the leaf node.

eyalkoren: > Are you suggesting to add inferred spans to the activation stack?

No, I am suggesting a way to make sure you know which is the actual parent of the inferred span. If the same (real) span is active before and after you capture the stack trace, you can tell it is a real parent of anything you infer from this stack trace. If the active span changes while you took the stack trace, you can decide to discard this sample.

felixbarny:
snapshot build: [elastic-apm-agent-1.12.1-SNAPSHOT.jar.zip](https://github.com/elastic/apm-agent-java/files/3978434/elastic-apm-agent-1.12.1-SNAPSHOT.jar.zip)

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/SamplingProfiler.java
felixbarny: I think I see what you mean but that's not what causes the parent/child relationship mismatches. I've updated the `Caveats` session of https://github.com/elastic/apm-agent-java/issues/971#issuecomment-566054290 to explain the scenarios which leat to a mismatch.

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/SamplingProfiler.java
felixbarny: Updating the `CallTree` and taking thread dumps is deliberately done on the same thread so that there can't be any concurrency issues. There's a mechanism to process `ActivationEvents` up to a certain nano timestamp from the ring buffer. So there can't be any race conditions from concurrent activations as they are just put in the queue but they don't modify the `CallTree`.

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/SamplingProfiler.java
eyalkoren: Ahh, I see.

felixbarny:
superseded by 983

PR #976, opened 2019-12-26
Adding Destination fields

### LINKED ISSUE #934: Add destination.* fields
See https://github.com/elastic/apm/issues/115

### LINKED ISSUE #957: Add service-map-required fields
Implement https://github.com/elastic/apm/issues/180

### LINKED ISSUE #969: Add DB2 Integration Tests
A followup on elastic/apm-agent-java#926 

We can rely on the [DB2 test container](https://www.testcontainers.org/modules/databases/db2/) after receiving legal team's approval to accept the licence, as long as this is for test purposes only and we're not distributing. 
The licence can be accepted programatically or by adding a file to the classpath root (see the "EULA Acceptance" on the DB testcontainer link above).

[This is the relevant license](https://github.com/elastic/apm-agent-java/files/3964695/IBM.License.for.Unwarranted.Programs.pdf).

eyalkoren:
Closes elastic/apm-agent-java#934  
Closes elastic/apm-agent-java#957
Closes elastic/apm-agent-java#969 

Some technology-specific info:
- Hibernate search is not supported as it seems a low ROI, especially since only relevant for usage with remote data stores.
- Databases support is largely based on online research. It is not documented whether the URL we would get from the actual connection is exactly the same as the one used to configure the data source. I tested where and what possible, but since our integration tests rely on testcontainers, we cannot test all syntaxes.
For example- the [Oracle](https://docs.oracle.com/cd/B28359_01/java.111/b31224/urls.htm#BEIJFHHB) address list parenthesis syntax or [MySQL's](https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-jdbc-url-format.html) multiple host configuration. If those are required, we can add them in the future.
- JMS doesn't offer a standard API to retrieve the broker address, so this needs to be implemented for each JMS client. Since the address and port are not required for the service map, I excluded it from this already big enough PR. We can implement that in a later time. Examples for getting the address in the two clients we have integration tests for:
   - ActiveMQ: `((ActiveMQMessageProducer) producer).session.getConnection().getTransport().getRemoteAddress()` returns a String that we can use to create a URI, from which host and port can be extracted.
   - ActiveMQ Artemis:`((TopologyMember)((ActiveMQMessageProducer)producer).connection.getSessionFactory().getServerLocator().getTopology().getMembers().toArray()[0]).toURI()` returns a String that we can use to create a URI, from which host and port **of one broker node** can be extracted. If multiple destinations are configured, this is not enough to get the actual address used for a specific connection.
- Redis (requires @felixbarny's input):
   - Jedis: it seems like the Sharded client instrumentation is redundant, as this client first finds a shard and then calls the already instrumented Jedis. The integration tests work without those. Am I missing something here?
   - Lettuce: I couldn't find an elegant way to locate the address and port. We can instrument the Client construction and keep mapping from Connections to the Client, then instrument the Connections and map created Commands to Connections and use those eventually (although there are differences between versions). Based on your experience, can you think of a better way? Unless you can think of something, we will keep Lettuce unsupported and if required create a separate issue.

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update [CHANGELOG.asciidoc](CHANGELOG.asciidoc)

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
@@ -1,0 +1,96 @@
+package co.elastic.apm.agent.impl.context;
+
+import co.elastic.apm.agent.objectpool.Recyclable;
+
+import javax.annotation.Nullable;
+
+/**
+ * Context information about a destination of outgoing calls.
+ */
+public class Destination implements Recyclable {
+    /**
+     * Information about the service related to this destination.
+     */
+    private final Service service = new Service();
+
+    public Service getService() {
+        return service;
+    }
+
+    public boolean hasContent() {
+        return service.hasContent();
+    }
+
+    @Override
+    public void resetState() {
+        service.resetState();
+    }
+
+    /**
+     * Context information required for service maps.
+     */
+    public static class Service implements Recyclable {
+        /**
+         * Used for detecting unique destinations from each service.
+         * For HTTP, this is the address, with the port (even when it's the default port), without any scheme.
+         * For other types of connections, it's just the {@code span.subtype} (kafka, elasticsearch etc.).
+         * For messaging, we additionally add the queue name (eg jms/myQueue).
+         */
+        private final StringBuilder resource = new StringBuilder();
+
+        /**
+         * Used for detecting “sameness” of services and then the display name of a service in the Service Map.
+         * In other words, the {@link Service#resource} is used to query data for ALL destinations. However,
+         * some `resources` may be nodes of the same cluster, in which case we also want to be aware.
+         * Eventually, we may decide to actively fetch a cluster name or similar and we could use that to detect "sameness".
+         * For now, for HTTP we use scheme, host, and non-default port. For anything else, we use {@code span.subtype}
+         * (for example- postgresql, elasticsearch).
+         */
+        private final StringBuilder name = new StringBuilder();
+
+        /**
+         * For displaying icons or similar. Currently, this should be equal to the {@code span.type}.
+         */
+        @Nullable
+        private String type;
+
+        public Service withResource(String resource) {
+            this.resource.append(resource);
+            return this;
+        }
+
+        public StringBuilder getResource() {
+            return resource;
+        }
+
+        public Service withName(String name) {
+            this.name.append(name);
+            return this;
+        }
+
+        public StringBuilder getName() {
+            return name;
+        }
+
+        public Service withType(String type) {
+            this.type = type;
+            return this;
+        }
+
+        @Nullable
+        public String getType() {
+            return type;
+        }
+
+        public boolean hasContent() {
+            return resource.length() > 0 || name.length() > 0 || type != null;
+        }
+
+        @Override
+        public void resetState() {
+            resource.setLength(0);
+            name.setLength(0);
+            type = null;
+        }
+    }
+}
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
@@ -43,8 +43,8 @@ public class Destination implements Recyclable {
      */
     private int port;
 
-    public Destination withAddress(CharSequence address) {
-        if (address.length() > 0) {
+    public Destination withAddress(@Nullable CharSequence address) {
+        if (address != null && address.length() > 0) {
             // remove square brackets for IPv6 addresses
             int startIndex = 0;
             if (address.charAt(0) == '[') {
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
@@ -32,6 +32,46 @@ import javax.annotation.Nullable;
  * Context information about a destination of outgoing calls.
  */
 public class Destination implements Recyclable {
+
+    /**
+     * An IP (v4 or v6) or a host/domain name.
+     */
+    private final StringBuilder address = new StringBuilder();
+
+    /**
+     * The destination's port used within this context.
+     */
+    private int port;
+
+    public Destination withAddress(CharSequence address) {
+        if (address.length() > 0) {
+            // remove square brackets for IPv6 addresses
+            int startIndex = 0;
+            if (address.charAt(0) == '[') {
+                startIndex = 1;
+            }
+            int endIndex = address.length();
+            if (address.charAt(endIndex - 1) == ']') {
+                endIndex--;
+            }
+            this.address.append(address, startIndex, endIndex);
+        }
+        return this;
+    }
+
+    public StringBuilder getAddress() {
+        return address;
+    }
+
+    public Destination withPort(int port) {
+        this.port = port;
+        return this;
+    }
+
+    public int getPort() {
+        return port;
+    }
+
     /**
      * Information about the service related to this destination.
      */
@@ -42,11 +82,13 @@ public class Destination implements Recyclable {
     }
 
     public boolean hasContent() {
-        return service.hasContent();
+        return address.length() > 0 || port > 0 || service.hasContent();
     }
 
     @Override
     public void resetState() {
+        address.setLength(0);
+        port = -1;
         service.resetState();
     }
 
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
@@ -1,3 +1,27 @@
+/*-
+ * #%L
+ * Elastic APM Java agent
+ * %%
+ * Copyright (C) 2018 - 2019 Elastic and contributors
+ * %%
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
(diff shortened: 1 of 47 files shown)


felixbarny:
Some general questions:
- Can this be a post-processing step done by APM Server to avoid duplication across agents?
- Is there a spec besides the GH issues?
- Are there common test cases/acceptance tests?
- What is the consequence of not supporting a particular RDBMS or JMS impl?
- When communicating with a cluster consisting of multiple nodes, like with Elasticsearch, we send different destination fields based on which node we happen to talk to. Is that intentional? What is the effect of that on the service map?

eyalkoren:
> Can this be a post-processing step done by APM Server to avoid duplication across agents?

May be only relevant for HTTP URLs, but maybe there are things agents need to take care differently anyway in this sense (eg IPv6 with/without square brackets ?). For DB- those are JDBC-specific URLs AFAIU.

> Is there a spec besides the GH issues? 

What's missing from the GH issues?

> What is the consequence of not supporting a particular RDBMS or JMS impl?

> When communicating with a cluster consisting of multiple nodes, like with Elasticsearch, we send different destination fields based on which node we happen to talk to. Is that intentional? What is the effect of that on the service map?

The `destination.XXX` fields are mostly for SIEM atm, although I can see how they can be useful down the road for service maps to realize clusters, integrate with metrics etc.
However, for now, the service maps will rely on the `destination.service` fields, where this is not relevant. If you look into these issues, you will see there is a discussion we started about how to deal with clustered services, but we worked around it for the first service map iteration. In fact, data stores are probably not going to be included for now.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
```diff
@@ -0,0 +1,434 @@
...
+        return "ConnectionMetaData{" +
+            "dbVendor='" + dbVendor + '\'' +
+            ", host='" + host + '\'' +
+            ", port=" + port +
+            ", user='" + user + '\'' +
+            '}';
+    }
+
+    @SuppressWarnings("unused")
+    private enum ConnectionUrlParser {
```
felixbarny: Oof, I wish there was a standard for those URLs or a library we could use to offload the parsing. @axw are you aware of anything of that sort by any chance?

apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/JdbcDbIT.java
```diff
@@ -49,6 +49,8 @@ public JdbcDbIT(String url, String expectedDbVendor) throws Exception {
             {"jdbc:tc:postgresql:10://hostname/databasename", "postgresql"},
             {"jdbc:tc:mariadb:10://hostname/databasename", "mariadb"},
             {"jdbc:tc:sqlserver:2017-CU12://hostname/databasename", "sqlserver"},
+            {"jdbc:tc:db2:11.5.0.0a://hostname/databasename", "db2"},
+//            {"jdbc:tc:oracle://hostname/databasename", "oracle"},
```
felixbarny: what about this one? a leftover?

apm-agent-plugins/apm-mongoclient-plugin/src/main/java/co/elastic/apm/agent/mongoclient/ConnectionAdvice.java
```diff
@@ -62,16 +66,22 @@ public static Span onEnter(@Advice.Argument(0) Object databaseOrMongoNamespace,
...
+        span.getContext().getDestination().getService()
+            .withName("mongodb").withResource("mongodb").withType("db");
+        ServerAddress serverAddress = thiz.getDescription().getServerAddress();
+        span.getContext().getDestination()
+            .withAddress(serverAddress.getHost())
+            .withPort(serverAddress.getPort());
         try {
             String cmd =
                 // try to determine main commands in a garbage free way
-                command.containsKey("find")    ? "find"    :
```
felixbarny: revert formatting

apm-agent-plugins/apm-redis-plugin/apm-jedis-plugin/src/main/java/co/elastic/apm/agent/redis/jedis/JedisInstrumentation.java
```diff
@@ -63,9 +70,7 @@ private static void afterSendCommand(@Nullable @Advice.Local("span") Span span,
     @Override
     public ElementMatcher<? super TypeDescription> getTypeMatcher() {
         return named("redis.clients.jedis.Jedis")
-            .or(named("redis.clients.jedis.BinaryJedis"))
-            .or(named("redis.clients.jedis.ShardedJedis"))
-            .or(named("redis.clients.jedis.BinaryShardedJedis"));
+            .or(named("redis.clients.jedis.BinaryJedis"));
```
felixbarny: Maybe, instead of excluding the Jedis clients from instrumentation which don't have the `getClient()` method, check `if (thiz instanceof BinaryJedis)` and only then set the destination fields.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
```diff
@@ -0,0 +1,162 @@
...
+     * Context information required for service maps.
+     */
+    public static class Service implements Recyclable {
+        /**
+         * Used for detecting unique destinations from each service.
+         * For HTTP, this is the address, with the port (even when it's the default port), without any scheme.
+         * For other types of connections, it's just the {@code span.subtype} (kafka, elasticsearch etc.).
+         * For messaging, we additionally add the queue name (eg jms/myQueue).
+         */
+        private final StringBuilder resource = new StringBuilder();
```
felixbarny: What I'd like to see is documentation about which purpose all of these `destination.*` fields exactly serve. Like, where in the UI are they shown and how they are used to produce the service map. Probably the Javadoc is not the best place, but the spec should IMHO explain that.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
axw: Nope, sorry. For the Go agent I have driver-specific logic which either uses some code from the database driver library to parse, or vendors internal code from the driver library when there's no exported parser function.

eyalkoren:
> check the presence of those fields in `MockReporter.reportSpan` if it's an external span. 

Good idea, will do.

> Overall a bit more complex than what I was expecting/hoping for but seems to be inherent complexity as opposed to accidental. 

I did expect some complexity due to the reasons you listed, as well as the fact that you need to look in different places in different frameworks. In addition, trying to avoid allocations or expensive lookups, you need to be specific on where you get the info, for example in some cases you it's cheapest to use a ready URL and in others it's best to get the host and port. But certainly more complex than I wanted it to be...

> This is also a thing to always keep in mind when adding new integrations

True, but your suggestion to fail tests when not setting these fields should remind us.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
eyalkoren: Imagine how much I wished for such 😅. Couldn't find anything useful. I also thought of using driver-internal APIs, but I figured the URL syntax is expected to be much more stable than such APIs, so went for parsing. I hope it won't bite us too much.

apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/JdbcDbIT.java
eyalkoren: Yes, I tried it, but it requires building an image (or using some unofficial image) so I decided to keep out. I will remove the comment.

apm-agent-plugins/apm-redis-plugin/apm-jedis-plugin/src/main/java/co/elastic/apm/agent/redis/jedis/JedisInstrumentation.java
eyalkoren: Of course, if it is required then I will not remove it. I removed it because it seemed redundant (as the sharded client internally calls a non-sharded one- see my question in the top issue comment).

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
eyalkoren: This one is not shown in the UI, it is only used for the query. The `name` is the one shown in the UI, as I wrote in its javadoc. I tried to extract the essence from the GitHub issue and add it here. Once we have a spec, as you suggested, we can refer to it from the javadoc directly.

apm-agent-plugins/apm-redis-plugin/apm-jedis-plugin/src/main/java/co/elastic/apm/agent/redis/jedis/JedisInstrumentation.java
felixbarny: Ah, didn't know this but seems you are right 👍

PR #981, opened 2020-01-02
Add Kafka support

### LINKED ISSUE #789: Instrument Kafka clients
This is a placeholder for something we will have to do sooner or later.
Added it to refer to this Discuss topic that provides some useful info - https://discuss.elastic.co/t/spring-kafka-instrumentation-kafkalistenerendpointcontainer-mono-thread/193810/5. 
Mostly, the usage of `org.springframework.kafka.annotation.KafkaListener` seems like an interesting option of converting the Kafka consumer's polling APIs into much more useful listener APIs.

This is about the instrumentation of the clients for span creation and distributed tracing support. In addition, elastic/apm-agent-java#318 and elastic/apm-agent-java#469 talk about adding the capability to monitor Kafka specific metrics through JMX.

eyalkoren:
Closes elastic/apm-agent-java#789 

Fully supports Kafka from version 0.11.0 (clients and server), since this is when record headers were added, which we rely on for context propagation. Older client and/or broker versions are partially supported- no distributed tracing. 

Kafka streams are not traced atm. APM traces are not expected to provide useful info for the natural Kafka streams usage. If we get feedback that this is required (e.g. the Kafka stream APIs are used for messaging purposes) - we can add later.

### WIP- tasks left:
- [x] Add support for additional options of record iteration after poll
- [x] Add message body tracing
- [x] Apply messaging configurations- headers, body
- [x] Add tests for header sanitation
- [x] Add tests for disabled tracing for specific topics
- [x] Implement destination fields - elastic/apm-agent-java#976 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update [CHANGELOG.asciidoc](CHANGELOG.asciidoc)
- [x] Update [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)
- [x] Update spans type/subtype/action cross-agent alignment document

### Compatibility and tests
There is a clear cut for brokers (and clients) in the 0.11.0 version. 
Continuously tested:
- 2.3.0 broker with 2.4.0 client
- 2.3.0 broker with 0.10.0.0 client

Disabled test:
- Legacy (0.10.2.2) broker with 2.4.0 client. This test passes locally but fails on CI, maybe due to parallelism.

Manually tested:
- 0.11.0.0 client version behaves the same as 2.4.0.
- full end-to-end tests (using a test app on a Servlet container), verifying that clients 0.10.0.0, 0.11.0.0 and 2.4.0 behave as expected.
- 0.9.0.0 client version behaves the same as 0.10.0.0

### Screenshots
Screen Shot 2020-01-21 at 6 18 37
Screen Shot 2020-01-21 at 6 19 21
Screen Shot 2020-01-21 at 6 19 44


eyalkoren:
Jenkins run the tests

apm-agent-plugins/apm-kafka-plugin/src/main/java/co/elastic/apm/agent/kafka/helper/ConsumerRecordsIteratorWrapper.java
```diff
@@ -0,0 +1,132 @@
...
+                if (record.timestampType() == TimestampType.CREATE_TIME) {
+                    message.withAge(System.currentTimeMillis() - record.timestamp());
+                }
+                // todo - add destination fields
+
+                if (coreConfiguration.isCaptureHeaders()) {
+                    for (Header header : record.headers()) {
+                        String key = header.key();
+                        if (!TraceContext.TRACE_PARENT_HEADER.equals(key) &&
+                            WildcardMatcher.anyMatch(coreConfiguration.getSanitizeFieldNames(), key) == null) {
```
felixbarny: Maybe sanitize centrally with a `co.elastic.apm.agent.report.processor.Processor`?

apm-agent-core/src/main/java/co/elastic/apm/agent/util/BinaryHeaderMap.java
```diff
@@ -0,0 +1,223 @@
...
+        return keys.size();
+    }
+
+    public boolean isEmpty() {
+        return size() == 0;
+    }
+
+    public boolean add(String key, byte[] value) throws InsufficientCapacityException {
+        int valuesPos = valueBuffer.position();
+        CoderResult coderResult = IOUtils.decodeUtf8Bytes(value, valueBuffer);
```
felixbarny: I'm not sure it's a valid assumption that all or even most header values are UTF-8 encoded strings. Maybe fall back to hex-encoding if it's not UTF-8 data?

apm-agent-core/src/main/java/co/elastic/apm/agent/util/BinaryHeaderMap.java
```diff
@@ -0,0 +1,223 @@
...
+            }
+            valueLengths[size] = valueBuffer.position() - valuesPos;
+            result = true;
+        }
+        return result;
+    }
+
+    private void enlargeBuffer() throws InsufficientCapacityException {
+        if (valueBuffer.capacity() == MAXIMUM_HEADER_BUFFER_SIZE) {
+            throw new InsufficientCapacityException();
```
felixbarny: Avoid creating a stack trace with the singleton exception pattern. See also `co.elastic.apm.agent.bci.bytebuddy.MinimumClassFileVersionValidator.UnsupportedClassFileVersionException`.

felixbarny: Another approach would be to return a boolean

apm-agent-core/src/main/java/co/elastic/apm/agent/util/BinaryHeaderMap.java
```diff
@@ -0,0 +1,223 @@
...
+
+    public int size() {
+        return keys.size();
+    }
+
+    public boolean isEmpty() {
+        return size() == 0;
+    }
+
+    public boolean add(String key, byte[] value) throws InsufficientCapacityException {
```
felixbarny: Maybe instead of throwing an `InsufficientCapacityException`, just return `false`?

apm-agent-plugins/apm-kafka-plugin/src/main/java/co/elastic/apm/agent/kafka/package-info.java
```diff
@@ -0,0 +1,52 @@
...
+ * Consumption is a bit trickier as it relies on the {@link org.apache.kafka.clients.consumer.KafkaConsumer#poll(java.time.Duration)}
+ * API. The challenge with polling APIs is that their invocation event is typically arbitrary (meaning their
+ * duration is not of interest) and we really want to trace what happens AFTER they exit with a Message. Since Kafka
+ * records are only consumed in {@link java.lang.Iterable} batches, we try to trace the actions executed between
+ * iterations by starting/activating and ending/deactivating a transaction for each record during iteration.
+ * Such transactions will have the {@code messaging} type. This may result in non-desirable traces, for example if the
+ * iterator is held and not immediately iterated all through, but we assume this is somewhat of an edge case.
+ * As long as we can make it safe from memory leaks, this is acceptable.
+ * In case the polling is executed from within a traced transaction, we won't create a transaction-per-record. Instead,
+ * we will trace the polling action itself and create a span for it.
```
felixbarny: In this case, are you creating Spans when iterating over the list of messages returned by `poll`?

apm-agent-plugins/apm-kafka-plugin/src/test/java/co/elastic/apm/agent/kafka/KafkaClientVersionsIT.java
```diff
@@ -0,0 +1,62 @@
...
+            KafkaIT.class, KafkaIT.Consumer.class, KafkaIT.RecordIterationMode.class, KafkaIT.TestScenario.class,
+            KafkaIT.ConsumerRecordConsumer.class);
+    }
+
+    @Parameterized.Parameters(name = "{0}")
+    public static Iterable<Object[]> data() {
+        return Arrays.asList(new Object[][]{
+            // No real need to run both versions, the APIs are the same. Running both requires a separate module
+            // because within a module that declares the dependency, the Kafka client is loaded by the app class loader
+            // IN ADDITION to the test class loader. This works fine only when the tested version is the same as the
```
felixbarny: What causes the Kafka client to be loaded by the system class loader?

> probably due to some static initializations done by the first one loaded

Where does that initialization happen? Do you have a stack trace for this?

apm-agent-plugins/apm-kafka-plugin/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+        <artifactId>apm-agent-plugins</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.12.1-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <properties>
+        <apm-agent-parent.base.dir>${project.basedir}/../..</apm-agent-parent.base.dir>
+        <version.testcontainers>1.12.4</version.testcontainers>
+        <version.kafka>2.4.0</version.kafka>
```
felixbarny: To automatically test new versions as they are releases:
```suggestion
        <version.kafka>[2.4.0,)</version.kafka>
```

apm-agent-core/src/test/java/co/elastic/apm/agent/util/BinaryHeaderMapTest.java
```diff
@@ -0,0 +1,162 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+class BinaryHeaderMapTest {
+    private BinaryHeaderMap headerMap = new BinaryHeaderMap();
+
+    @AfterEach
+    void reset() {
+        headerMap.resetState();
+        assertThat(headerMap.isEmpty()).isTrue();
+    }
```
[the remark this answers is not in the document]
eyalkoren: Right, but it tests `resetState` and `isEmpty`

apm-agent-plugins/apm-kafka-plugin/src/main/java/co/elastic/apm/agent/kafka/KafkaProducerInstrumentation.java
```diff
@@ -0,0 +1,183 @@
...
+            }
+
+            final TraceContextHolder<?> activeSpan = tracer.getActive();
+            if (activeSpan == null || !activeSpan.isSampled()) {
+                return null;
+            }
+
+            if (activeSpan instanceof Span && "kafka".equals(((Span) activeSpan).getSubtype())) {
+                return null;
+            }
```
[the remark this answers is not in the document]
eyalkoren: It IS redundant. Thanks!

apm-agent-plugins/apm-kafka-plugin/src/main/java/co/elastic/apm/agent/kafka/helper/ConsumerRecordsIteratorWrapper.java
```diff
@@ -0,0 +1,132 @@
...
+                            try {
+                                message.addHeader(key, header.value());
```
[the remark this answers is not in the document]
eyalkoren: Yes it does. I could potentially access the underlying `org.apache.kafka.common.header.internals.RecordHeader#valueBuffer`, but this is implementation detail, not API, so I am not sure it is best, especially assuming that normally if headers are sent they are also read. WDYT?

apm-agent-plugins/apm-kafka-plugin/src/main/java/co/elastic/apm/agent/kafka/helper/ConsumerRecordsIteratorWrapper.java
eyalkoren: Why? Especially given the comment above- better avoid this decoding if possible.

apm-agent-core/src/main/java/co/elastic/apm/agent/util/BinaryHeaderMap.java
eyalkoren: It is certainly not the assumption. This is only to support recording of textual headers (as Kafka records only support binary). I prefer not to capture anything if it is not textual, and currently this is the indication I have to whether it is or not. 
The problem with hex-decoding, is that it always works, even if the value was not hex-encoded, so it would make no sense.

apm-agent-plugins/apm-kafka-plugin/src/main/java/co/elastic/apm/agent/kafka/package-info.java
eyalkoren: No. You think I should? 
The `poll` is the actual **interesting** action. The thread is traced already, so anything the application will do after reading the record on this thread will be traced as well. Maybe when we support multiple span parents (or span references) this may make more sense.

apm-agent-plugins/apm-kafka-plugin/src/test/java/co/elastic/apm/agent/kafka/KafkaClientVersionsIT.java
eyalkoren: Didn't look too much into it...
Bottom line is that I don't need to have both versions tested every time we build as they use the same instrumentation. I only want to see once that it works for both, so I do it manually.

apm-agent-plugins/apm-kafka-plugin/src/main/java/co/elastic/apm/agent/kafka/helper/ConsumerRecordsIteratorWrapper.java
felixbarny: So that we don't forget it on other messaging frameworks. What do you think about doing both maybe?

apm-agent-plugins/apm-kafka-plugin/src/main/java/co/elastic/apm/agent/kafka/helper/ConsumerRecordsIteratorWrapper.java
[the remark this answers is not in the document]
felixbarny: Yeah, let's leave it as-is for now.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Headers.java
```diff
@@ -0,0 +1,146 @@
...
+            if (textHeadersIterator.hasNext()) {
+                NoRandomAccessMap.Entry<String, String> textHeader = textHeadersIterator.next();
+                header.key = textHeader.getKey();
+                header.value = textHeader.getValue();
+            } else if (binaryHeadersIterator.hasNext()) {
+                BinaryHeaderMap.Entry binaryHeader = binaryHeadersIterator.next();
+                header.key = binaryHeader.getKey();
+                header.value = binaryHeader.getValue();
+            } else {
+                throw new IllegalStateException("next() called on a depleted iterator");
```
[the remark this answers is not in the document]
felixbarny: TIL!

apm-agent-plugins/apm-kafka-plugin/src/main/java/co/elastic/apm/agent/kafka/package-info.java
felixbarny: Not sure yet, it just seems a bit inconsistent to create a transaction for each element in the iterator but if there's already a transaction, only a single span for the poll gets created.

eyalkoren: I don't think consistency is a consideration here. The `poll` action is interesting like any other outreaching request (HTTP, DB, ES etc). The iteration is not as interesting- think about it as instrumenting the iteration over JDBC `ResultSet`. Besides, don't forget we create the transaction **outside all instrumented methods**. We do that so we can support distributed tracing. What would we create a span for? If we create a span for the `next` operation itself, it would still be inconsistent with what we do for transactions, and if we create a span between `next` invocations- not sure what that represents.

apm-agent-plugins/apm-kafka-plugin/src/main/java/co/elastic/apm/agent/kafka/helper/ConsumerRecordsIteratorWrapper.java
eyalkoren: I think it is a lot of overhead to pay in order to be future-proof, and requires special implementation for the new `Headers` type I am using. I rather not do that, at least at this point.

Yogi-SAP:
hi, please let us know if Kafka version 1.10 is supported. I know it is mentioned above that only version 1.11 and above are supported but we have version 1.10 and I had requested the Product owner in a call if version 1.10 is supported as there is not much of a difference in the high level producer and consumer of both versions.

eyalkoren:
@Yogi-SAP the problem is with the broker, not the clients, and it is due to the fact that record headers were introduced in 1.11.0. Since our distributed tracing through Kafka is going to rely on record headers, it will not work. We can trace spans for sends and polls, but we will not be able to stitch the producer and consumer transactions into one trace.
Do you have a way of transferring metadata alongside the Kafka records?

eyalkoren:
I tested manually also with 0.9.0.0. 
0.8.2.0 and is really very old, so I think not a good ROI at this point. APIs look the same.

Yogi-SAP:
> @Yogi-SAP the problem is with the broker, not the clients, and it is due to the fact that record headers were introduced in 1.11.0. Since our distributed tracing through Kafka is going to rely on record headers, it will not work. We can trace spans for sends and polls, but we will not be able to stitch the producer and consumer transactions into one trace.
> Do you have a way of transferring metadata alongside the Kafka records?

hi, this makes sense, thanks for clarifying. This was solved in versions prior by having the trace id as part of the message body (AVRO/JSON field etc). If we do this through the message body then we will need a mechanism to override the trace ID on the consumer end to stitch together. Is it possible to override the trace ID?

eyalkoren:
> If we do this through the message body then we will need a mechanism to override the trace ID on the consumer end to stitch together. Is it possible to override the trace ID?

No need to override the trace ID. If your Kafka messages carry textual body, you can use our public API: use [`injectTraceHeaders`](https://www.elastic.co/guide/en/apm/agent/java/current/public-api.html#api-transaction-inject-trace-headers) on the sender side, where you would implement the `HeaderInjector` interface to add the traceparent header (as key and value) to the message body. An interceptor would be a good candidate to doing that. 
On the consumer side, you need to identify the proper location in your code that would best represent the message-handling transaction (for example- the method that encapsulates the message-handling logic) and call [`startTransactionWithRemoteParent`](https://www.elastic.co/guide/en/apm/agent/java/current/public-api.html#api-start-transaction-with-remote-parent-header), where you would provide your `HeaderExtractor` implementation that knows how to read the value of a given field from the message body.
When using these APIs, the agent will treat these transactions like it does in any other distributed  traces.

Yogi-SAP:
> > If we do this through the message body then we will need a mechanism to override the trace ID on the consumer end to stitch together. Is it possible to override the trace ID?
> 
> No need to override the trace ID. If your Kafka messages carry textual body, you can use our public API: use [`injectTraceHeaders`](https://www.elastic.co/guide/en/apm/agent/java/current/public-api.html#api-transaction-inject-trace-headers) on the sender side, where you would implement the `HeaderInjector` interface to add the traceparent header (as key and value) to the message body. An interceptor would be a good candidate to doing that.
> On the consumer side, you need to identify the proper location in your code that would best represent the message-handling transaction (for example- the method that encapsulates the message-handling logic) and call [`startTransactionWithRemoteParent`](https://www.elastic.co/guide/en/apm/agent/java/current/public-api.html#api-start-transaction-with-remote-parent-header), where you would provide your `HeaderExtractor` implementation that knows how to read the value of a given field from the message body.
> When using these APIs, the agent will treat these transactions like it does in any other distributed traces.

Makes perfect sense. This then is the solution to Kafka versions prior to Version 1.11.

eyalkoren:
@Yogi-SAP I just now see that the entire discussion above uses wrong versions- we are talking about up to (excluding) and from (including) **0.11.0** Kafka version (edited my comment accordingly). I assume it is a typo, but just in case.

PR #984, opened 2020-01-07
Set classname instead of filename when connected to APM Server 7.6

felixbarny:
See https://github.com/elastic/kibana/issues/49467

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] ~Update documentation~
- [ ] Update [CHANGELOG.asciidoc](CHANGELOG.asciidoc). Not sure if that's worth mentioning?

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
@@ -693,11 +693,8 @@ public class DslJsonSerializer implements PayloadSerializer, MetricRegistry.Metr
 
     private void serializeStackTraceElement(StackTraceElement stacktrace) {
         jw.writeByte(OBJECT_START);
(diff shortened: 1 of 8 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
```diff
@@ -693,7 +693,11 @@ private static boolean isExcluded(StackTraceElement stackTraceElement) {
 
     private void serializeStackTraceElement(StackTraceElement stacktrace) {
         jw.writeByte(OBJECT_START);
-        writeField("filename", stacktrace.getFileName());
+        if (apmServerClient.supportsStacktraceClassname()) {
+            writeField("classname", stacktrace.getClassName());
+        } else {
+            writeField("filename", stacktrace.getFileName());
+        }
```
eyalkoren: As per https://github.com/elastic/kibana/issues/49467#issuecomment-571738211 - let's send both for a while.

PR #985, opened 2020-01-08
Reduce startup costs for AsyncHttpClient instrumentation

### LINKED ISSUE #932: Instrument implementations of interfaces more efficiently
I have an idea to efficiently match all implementations of an interface, such as `org.asynchttpclient.AsyncHandler`. The problem is that those may be implemented as an anonymous inner class so we can’t pre-filter on the name, like `endsWith("Handler")` which means we have to check the whole inheritance hierarchy of every class in order to determine whether a class implements AsyncHandler. Instead of eagerly searching for all implementations, we could just check in `DefaultAsyncHttpClient#executeRequest(Request, AsyncHandler)` whether we have already instrumented the provided `AsyncHandler`. If we did not, we can just retransform the class and apply the corresponding instrumentation to it. We currently don’t have a facility to do that but should not be too complex to do™. 

The benefit is faster startups of the application.

felixbarny:
Instead of the expensive `hasSuperType(named("org.asynchttpclient.AsyncHandler"))` matcher which matches the inheritance hierarchy of each and every loaded class, we just re-instrument those `AsyncHandler`s which are provided to the `org.asynchttpclient.DefaultAsyncHttpClient#execute(Request request, final AsyncHandler<T> asyncHandler)` method.

Closes elastic/apm-agent-java#932

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] ~Update documentation~


apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java
```diff
@@ -123,7 +132,7 @@ protected AbstractAsyncHandlerInstrumentation(ElementMatcher<? super MethodDescr
 
         @Override
         public ElementMatcher<? super TypeDescription> getTypeMatcher() {
-            return hasSuperType(named("org.asynchttpclient.AsyncHandler"));
+            return any()/*.and(ElasticApmAgent.getTypeMatcher(asyncHandler.getClass()))*/;
```
SylvainJuge: maybe add a comment why it's safe to do with `any()` here

PR #986, opened 2020-01-08
Make instrumentation options dynamic

felixbarny:
### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation
- [ ] Added an API method or config option? Document in which version this will be introduced.


apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -348,7 +396,7 @@ private static AgentBuilder getAgentBuilder(final ByteBuddy byteBuddy, final Cor
 
         return new AgentBuilder.Default(byteBuddy)
             .with(AgentBuilder.RedefinitionStrategy.RETRANSFORMATION)
```
eyalkoren:
```suggestion
            .with(AgentBuilder.RedefinitionStrategy.RETRANSFORMATION)
            .with(new AgentBuilder.RedefinitionStrategy.BatchAllocator() {
                @Override
                public Iterable<? extends List<Class<?>>> batch(List<Class<?>> types) {
                    // if this is on startup, do the same as BatchAllocator.ForTotal, otherwise use batches
                }
            })
```
Not exactly this, but this `BatchAllocator` can be handy for retransformations...

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -516,6 +519,10 @@ public boolean isInstrument() {
         return instrument.get();
     }
 
+    public List<ConfigurationOption<?>> getInstrumentationOptions() {
+        return Arrays.asList(instrument, traceMethods, disabledInstrumentations);
```
eyalkoren: `traceMethods` and `disabledInstrumentations` are tested, but seems that dynamically changing `instrument` is not tested.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -141,12 +148,54 @@ public void run() {
...
+            });
+        } finally {
+            executor.shutdown();
+        }
+    }
+
+    static synchronized void doReInitInstrumentation(Iterable<ElasticApmInstrumentation> instrumentations) {
+        final Logger logger = LoggerFactory.getLogger(ElasticApmAgent.class);
+        logger.info("Re initializing instrumentation");
+        // HYBRID to speed up the matchers, assuming dependant classes have already loaded so we don't interfere with the ordering of initialization
```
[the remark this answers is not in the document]
felixbarny: > Can this happen during startup if the remote config overrides local settings?

yes, that's possible

> the usage of getDeclaredMethods surfaced this incompatibility as it caused an attempt to resolve all methods

What is the consequence of that? Can't we just ignore the linkage errors?

felixbarny:
> Maybe batching can somewhat address the retransformation overhead concern.

It's not the retransformation I'm concerned about but the type matching. When using `POOL_ONLY`, we probably have to load lots of type definitions from disk again.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
felixbarny: Byte Buddy is batching by default (via `RedefinitionStrategy.BatchAllocator.ForTotal.INSTANCE`)

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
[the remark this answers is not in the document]
eyalkoren: Good question, I am not sure how the JVM would treat a class that is already loaded and partly linked and suddenly fails linkage...

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
eyalkoren: `RedefinitionStrategy.BatchAllocator.ForTotal.INSTANCE` means it is not batching, it includes all in one batch

eyalkoren:
> It's not the retransformation I'm concerned about but the type matching. When using `POOL_ONLY`, we probably have to load lots of type definitions from disk again.

Right, wrong wording. Nevertheless, batching may be a way to make that less noticeable.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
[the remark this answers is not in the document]
eyalkoren: >  > Can this happen during startup if the remote config overrides local settings?
>
> yes, that's possible

In this case reflection may be used on startup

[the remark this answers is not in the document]
felixbarny: I guess it should be relatively safe to ignore them. However, we'd lose the ability to instrument those classes. Changing to `POOL_ONLY` for now.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
felixbarny: I see what you mean now. Which batch size would you propose? Should we sleep between batches? Or should we just start with a single batch?

eyalkoren: I would say it makes sense to use a single batch on startup when the instrumentations are applied through `premain` or programmatic attach and use batches when applied through `agentmain` or when re-instrumenting. 

> should we just start with a single batch?

The reason I am suggesting it in this PR is because of the introduction of re-instrumentation, but if you have a reason to merge this quickly, it can be done in a separate PR. I think that if it is a matter of small effort- it worth it. Your call.

> Which batch size would you propose? 

Well, as you imagine, it would just be random anyways 🙂 . The main idea is not to cause a long degradation in app performance. From the top of my head- 100 per batch.

> Should we sleep between batches?

Certainly, . We can use `net.bytebuddy.agent.builder.AgentBuilder.RedefinitionStrategy.Listener.Pausing` for that.

PR #999, opened 2020-01-22
JMS fix: transfer header on non-sampled traces

eyalkoren:
## What does this PR do?

Our distributed tracing sampling relies on context propagation. Currently, context propagation doesn't work in traces using JMS. This PR fixes that.

## Checklist

### DIFF
--- a/apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
+++ b/apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
@@ -71,7 +71,7 @@ public class JmsInstrumentationHelperImpl implements JmsInstrumentationHelper<De
     public Span startJmsSendSpan(Destination destination, Message message) {
 
         final TraceContextHolder<?> activeSpan = tracer.getActive();
-        if (activeSpan == null || !activeSpan.isSampled()) {
+        if (activeSpan == null) {
             return null;
         }
 
@@ -97,8 +97,8 @@ public class JmsInstrumentationHelperImpl implements JmsInstrumentationHelper<De
             .activate();
 
         try {
+            message.setStringProperty(JMS_TRACE_PARENT_PROPERTY, span.getTraceContext().getOutgoingTraceParentHeader().toString());
             if (span.isSampled()) {
-                message.setStringProperty(JMS_TRACE_PARENT_PROPERTY, span.getTraceContext().getOutgoingTraceParentHeader().toString());
                 span.getContext().getDestination().getService()
                     .withName("jms")
                     .withResource("jms")
--- a/apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/JmsInstrumentationIT.java
+++ b/apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/JmsInstrumentationIT.java
@@ -209,6 +209,20 @@ public class JmsInstrumentationIT extends AbstractInstrumentationTest {
     }
 
     @Test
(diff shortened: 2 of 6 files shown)


apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/JmsInstrumentationIT.java
```diff
@@ -208,6 +219,25 @@ public void testQueueSendReceiveOnNonTracedThread() throws Exception {
...
     }
 
+    @SuppressWarnings("ConstantConditions")
+    @Test
+    public void testQueueSendReceiveOnNonTracedThread_NonSampledTransaction() throws Exception {
+        final Queue queue = createTestQueue();
+
+        // End current transaction and start a non-sampled one
+        tracer.currentTransaction().deactivate().end();
+        reporter.reset();
```
SylvainJuge: why do we need to reset the reporter here ?

apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/JmsInstrumentationIT.java
```diff
@@ -386,7 +416,10 @@ private void verifyQueueSendReceiveOnNonTracedThread(Queue queue)
         Message incomingMessage = resultQ.poll(2, TimeUnit.SECONDS);
         assertThat(incomingMessage).isNotNull();
         verifyMessage(message, incomingMessage);
-        verifySendReceiveOnNonTracedThread(queue.getQueueName(), outgoingMessage);
+        //noinspection ConstantConditions
+        if (tracer.currentTransaction().isSampled()) {
```
SylvainJuge: what about having the annotation directly on `isSampled` method in order to avoid duplicates of `//noinspection ConstantConditions` ?

apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/JmsInstrumentationIT.java
eyalkoren: Because we `end` a transaction just above, which means it is reported, but soon after I use `reporter.getFirstTransaction()`, which I want to return my tested transaction and not this one

apm-agent-plugins/apm-jms-plugin/src/test/java/co/elastic/apm/agent/jms/JmsInstrumentationIT.java
eyalkoren: Not sure what you mean. This removes the warning about `tracer.currentTransaction()` potentially returning `null`

apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java
```diff
@@ -97,8 +97,8 @@ public Span startJmsSendSpan(Destination destination, Message message) {
             .activate();
 
         try {
+            message.setStringProperty(JMS_TRACE_PARENT_PROPERTY, span.getTraceContext().getOutgoingTraceParentHeader().toString());
```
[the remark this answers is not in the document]
eyalkoren: YES! If we don't, it is a bug as we are expected to (which is what this PR is about). This is what we use the flags part of the traceparent header- it is how we tell downstream transactions whether they should be sampled-in or out.

eyalkoren:
`KafkaIT` is disabled because it is ran through `KafkaClientVersionsIT`. The reason is that `KafkaIT` doesn't provide a full test setup for the instrumentations, as the pom dependencies mean that even `provided` dependencies are loaded by the system class loader, which does not reflect real application scenarios.

Issue #997, opened 2020-01-20
Official Docker images


cachedout:
## Credit

This request was originally raised by @eyalkoren and credit for the suggestion goes to him. :)

## Overview

There are cases in which it makes sense to have an official Docker image for the APM Java agent. This is a proposal for the creation of such an image and for its deployment as a part of the build and release process for the APM Java Agent.

## Use cases

One use case which has been well-outlined by @eyalkoren in a separate document is for a standalone agent container as a part of a K8S cluster. As a pod starts, an agent container can be configured as an [init container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) for the pod. In doing so, the APM agent would start up prior to the application(s) in the pod and be available for instrumentation by application containers through the use of the `JAVA_TOOL_OPTIONS` environment variable. 

I'm sure there are likely other use cases for this or the above case can be expanded upon, but I'll let @eyalkoren extend what I've written above as he sees fit. :)

## Current Build Process

Currently, the build process for the Java APM agent [is outlined here](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#releasing).

As one can see, there is currently no mention of an official container being built, nor does the repo currently contain a `Dockerfile` with which to build one.

## Proposed Changes

This request is a collection of a few concrete changes:

### 1. Official Dockerfile

Obviously, the build process will need a Dockerfile in the repository. @eyalkoren has kindly [provided a basic sample](https://gist.github.com/eyalkoren/7036f777954ac4f1918b97cffe9bd727) which could get us started by checking it directly into the root of the repository.

#### Alternatives

The above is simple enough that I assume we can just check it in directly instead of making it a part of the build process to create it dynamically. However, I'll mention that as an alternative here just in case the team thinks this might be a worthwhile area of exploration.

### 2. Official Docker image

Right now, there is no `observability` namespace in our Docker repository, which we would likely want in order to host these images. If we agree that creating Docker images for the APM Java Agent is a worthwhile endeavor, I will reach out to the appropriate folks to make this happen.

### 3. Build process

Right now, we use an internal CI server as a part of the release process. We propose modifying that pipeline to create an image as an artifact which can be automatically pushed up to Docker Hub once the requirements in step 2 are in place.

### Snapshots?

We could, in theory, create nightly image snapshots as Docker images and make them available as well. My guess is that with a simple Docker file, anybody who wants an image could just clone the repository and build one themselves but if this is something people are interested in, it's worth discussing.

[...]

PR #1001, opened 2020-01-22
Add Docker build and push to release stage of build

cachedout:
## What does this PR do?
This PR modifies the release pipeline for the APM Java agent project and instructs it to create and publish a Docker image on each release to the public repository. For the time being, this behavior happens if-and-only-if the `push_docker` parameter is set when the job is run. In this PR, the default value for `push_docker` is set to `false`.

## Checklist

## Related issues

Relates #997

## Screenshots
<!-- _(Optional)_

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -258,7 +258,6 @@ If you have access to make releases, the process is as follows:
    1. Tags this new commit with the version name, e.g. `v1.1.0`
    1. Advances the version for next development iteration and makes a commit
    1. Uploads artifacts to maven central
-   1. (Optional) If the `push_docker` parameter is set, build and push a Docker image for this release to the docker.elastic.co repo.
 1. Login to https://oss.sonatype.org, go to Staging Repositories, close and release the staged artifacts.
 1. Fetch and checkout the latest tag e.g. `git fetch origin`
 1. If this was a major release,
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ Talk about what you would like to do.
 It may be that somebody is already working on it,
 or that there are particular issues that you should know about before implementing the change.
 
-Once you are all set to go, [this "cookbook recipe" blog post](https://www.elastic.co/blog/a-cookbook-for-contributing-a-plugin-to-the-elastic-apm-java-agent) can guide you through. 
+Once you are all set to go, [this "cookbook recipe" blog post](https://www.elastic.co/blog/a-cookbook-for-contributing-a-plugin-to-the-elastic-apm-java-agent) can guide you through.
 
 ### Testing
 
@@ -53,13 +53,13 @@ they will be executed by a CI server.
    there are a few things which need to be configured.
    Open the preferences,
    search for nullable and enable the inspections `Constant conditions & exceptions` and `@NotNull/@Nullable problems`
-   
+
    <img width="684" alt="inspections" src="https://user-images.githubusercontent.com/2163464/43443888-87e3ffd2-94a2-11e8-8cc6-263f408479e3.png">
 1. In the same window,
    select `Constant conditions & exceptions`
    click on `Configure annotations`,
    and select the `javax.annotation` annotations.
-    
+
    <img width="382" alt="configure annotations" src="https://user-images.githubusercontent.com/2163464/43444414-f1e5ce5a-94a3-11e8-8fa4-9f048c50ccc0.png">
 
 ##### Useful Live Templates
@@ -173,7 +173,7 @@ we should think about whether they bring us closer to or further away from those
     document why you did something that way,
     rather than how you did it.
     The latter should be expressed by readable code.
-    Documenting the considered alternatives can also help others a lot to understand the problem space. 
+    Documenting the considered alternatives can also help others a lot to understand the problem space.
 * Diagnosability
   * Most issues should be resolvable when the user provides their debug logs.
     This decreases the back-and-forth to gather additional information.
@@ -251,15 +251,16 @@ and follow [those instructions](https://github.com/elastic/docs#for-a-local-repo
 If you have access to make releases, the process is as follows:
 
 1. Check if sonatype is up: https://status.maven.org
-1. Review project version. The release version will be `${project.version}` without the `-SNAPSHOT`. 
+1. Review project version. The release version will be `${project.version}` without the `-SNAPSHOT`.
    1. In case you want to update the version, execute `mvn release:update-versions`
 1. Execute the release Jenkins job on the internal ci server. This job is same as the snapshot-build job, but it also:
    1. Removes `-SNAPSHOT` from all `${project.version}` occurrences and makes a commit before build
    1. Tags this new commit with the version name, e.g. `v1.1.0`
    1. Advances the version for next development iteration and makes a commit
    1. Uploads artifacts to maven central
+   1. (Optional) If the `push_docker` parameter is set, build and push a Docker image for this release to the docker.elastic.co repo.
 1. Login to https://oss.sonatype.org, go to Staging Repositories, close and release the staged artifacts.
-1. Fetch and checkout the latest tag e.g. `git fetch origin` 
+1. Fetch and checkout the latest tag e.g. `git fetch origin`
 1. If this was a major release,
    create a new branch for the major
    1. For example `git checkout -b 2.x && git push -u origin`
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,4 @@
+FROM alpine:3.9
+RUN mkdir /usr/agent
+ARG JAR_FILE
+COPY ${JAR_FILE} /usr/agent/elastic-apm-agent.jar
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,7 @@
-FROM alpine:3.9
+# Pin to Alpine 3.9
+# For a complete list of hashes, see:
+# https://github.com/docker-library/repo-info/tree/master/repos/alpine/remote
+FROM alpine@sha256:115731bab0862031b44766733890091c17924f9b7781b79997f5f163be262178
 RUN mkdir /usr/agent
 ARG JAR_FILE
 COPY ${JAR_FILE} /usr/agent/elastic-apm-agent.jar
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,7 +36,7 @@ pipeline {
     booleanParam(name: 'test_ci', defaultValue: true, description: 'Enable test')
     booleanParam(name: 'smoketests_ci', defaultValue: true, description: 'Enable Smoke tests')
     booleanParam(name: 'bench_ci', defaultValue: true, description: 'Enable benchmarks')
-    booleanParam(name: 'push_docker', defaultValut: false, description 'Push Docker image during release stage')
+    booleanParam(name: 'push_docker', defaultValut: false, description: 'Push Docker image during release stage')
   }
 
   stages {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -296,7 +296,7 @@ pipeline {
           steps {
             sh(label: "Build Docker image", script: "scripts/jenkins/build_docker.sh")
             // Get Docker registry credentials
-            dockerLogin(secret: ${ELASTIC_DOCKER_SECRET}, registry: 'docker.elastic.co')
+            dockerLogin(secret: "${ELASTIC_DOCKER_SECRET}", registry: 'docker.elastic.co')
             sh(label: "Push Docker image", script: "scripts/jenkins/push_docker.sh")
           }
         }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -297,7 +297,7 @@ pipeline {
             sh(label: "Build Docker image", script: "scripts/jenkins/build_docker.sh")
             // Get Docker registry credentials
             dockerLogin(secret: ${ELASTIC_DOCKER_SECRET}, registry: 'docker.elastic.co')
-            sh(label: "Push Docker image"), script: "scripts/jenkins/push_docker.sh"
+            sh(label: "Push Docker image"), script: "scripts/jenkins/push_docker.sh")
           }
         }
         stage('Opbeans') {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,6 +10,7 @@ pipeline {
     NOTIFY_TO = credentials('notify-to')
     JOB_GCS_BUCKET = credentials('gcs-bucket')
     DOCKERHUB_SECRET = 'secret/apm-team/ci/elastic-observability-dockerhub'
+    ELASTIC_DOCKER_SECRET = 'secret/apm-team/ci/docker-registry/prod'
     CODECOV_SECRET = 'secret/apm-team/ci/apm-agent-java-codecov'
     GITHUB_CHECK_ITS_NAME = 'Integration Tests'
     ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
@@ -35,6 +36,7 @@ pipeline {
     booleanParam(name: 'test_ci', defaultValue: true, description: 'Enable test')
     booleanParam(name: 'smoketests_ci', defaultValue: true, description: 'Enable Smoke tests')
     booleanParam(name: 'bench_ci', defaultValue: true, description: 'Enable benchmarks')
+    booleanParam(name: 'push_docker', defaultValut: false, description 'Push Docker image during release stage')
   }
 
   stages {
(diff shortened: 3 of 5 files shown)


cachedout:
@eyalkoren Just so there's no confusion, the scripts to generate and push the Docker image are written to be run independently if needed, so this does _not_ need to be merged prior to us pushing an image, [as discussed here](https://github.com/elastic/apm-agent-java/issues/997#issuecomment-576228501).

scripts/jenkins/build_docker.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+set -euxo pipefail
+
+if ! which docker
```
kuisathaverat: NIT https://stackoverflow.com/questions/37056192/which-vs-command-v-in-bash

```suggestion
if ! command -v docker
```

scripts/jenkins/build_docker.sh
```diff
@@ -0,0 +1,39 @@
...
+then
+  echo "ERROR: Building Docker image requires Docker binary to be installed" && exit 1
+elif ! docker version
+then
+  echo "ERROR: Building Docker image requires Docker daemon to be running" && exit 1
+elif ! which curl
```
kuisathaverat:
```suggestion
elif ! command -v curl
```

scripts/jenkins/build_docker.sh
```diff
@@ -0,0 +1,39 @@
...
+then
+  echo "ERROR: Building Docker image requires cURL to be installed" && exit 1
+fi
+
+echo "INFO: Fetching latest tag"
+#TODO Move this into apm-pipeline-library
+readonly GIT_TAG=`git describe --abbrev=0|sed s/^v//`
```
kuisathaverat: Allow passing the GIT_TAG on the environment, by default use the latest TAG

```suggestion
GIT_TAG_DEFAULT=$(git describe --abbrev=0|sed s/^v//)
readonly GIT_TAG=${GIT_TAG:-$GIT_TAG_DEFAULT}
```

scripts/jenkins/push_docker.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# This script is intended to be run from a CI job and will not work if run in
+# standalone manner unless certain envrionment variables are set.
+
+# 1. Grab the tag we are working with
+
+#FIXME DRY between Docker scripts
+echo "INFO: Fetching latest tag"
+#TODO Move this into apm-pipeline-library
+readonly CUR_TAG=`git describe --abbrev=0|sed s/^v//`
```
kuisathaverat:
```suggestion
CUR_TAG_DEFAULT=$(git describe --abbrev=0|sed s/^v//)
readonly CUR_TAG=${CUR_TAG:-$CUR_TAG_DEFAULT}
```

CONTRIBUTING.md
```diff
@@ -251,15 +251,16 @@ and follow [those instructions](https://github.com/elastic/docs#for-a-local-repo
...
 1. Check if sonatype is up: https://status.maven.org
-1. Review project version. The release version will be `${project.version}` without the `-SNAPSHOT`. 
+1. Review project version. The release version will be `${project.version}` without the `-SNAPSHOT`.
    1. In case you want to update the version, execute `mvn release:update-versions`
 1. Execute the release Jenkins job on the internal ci server. This job is same as the snapshot-build job, but it also:
    1. Removes `-SNAPSHOT` from all `${project.version}` occurrences and makes a commit before build
    1. Tags this new commit with the version name, e.g. `v1.1.0`
    1. Advances the version for next development iteration and makes a commit
    1. Uploads artifacts to maven central
+   1. (Optional) If the `push_docker` parameter is set, build and push a Docker image for this release to the docker.elastic.co repo.
```
v1v: IIUC, this step is not for the https://github.com/elastic/apm-agent-java/pull/1001/files#diff-6a3371457528722a734f3c51d9238c13R256 but for another Jenkins instance

scripts/jenkins/build_docker.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
v1v: Cosmetic change :)

```suggestion
#!/usr/bin/env bash
```

scripts/jenkins/build_docker.sh
```diff
@@ -0,0 +1,40 @@
...
+elif ! docker version
+then
+  echo "ERROR: Building Docker image requires Docker daemon to be running" && exit 1
+elif ! command -v curl
+then
+  echo "ERROR: Building Docker image requires cURL to be installed" && exit 1
+fi
+
+echo "INFO: Fetching latest tag"
+#TODO Move this into apm-pipeline-library
```
v1v: There are different env variable to know the tag name:
- TAG_NAME
- BRANCH_NAME
- GIT_BRANCH

Some references:
- https://apm-ci.elastic.co/pipeline-syntax/globals#env
- https://github.com/jenkins-infra/jenkins.io/blob/master/content/doc/book/pipeline/syntax.adoc#when

scripts/jenkins/build_docker.sh
```diff
@@ -0,0 +1,40 @@
...
+GIT_TAG_DEFAULT=$(git describe --abbrev=0|sed s/^v//)
+readonly GIT_TAG=${GIT_TAG:-$GIT_TAG_DEFAULT}
+
+echo "INFO: Downloading artifact from Sonatype Nexus repository for version $GIT_TAG"
+curl -L  -o apm-agent-java.jar "http://repository.sonatype.org/service/local/artifact/maven/redirect?r=central-proxy&g=co.elastic.apm&a=elastic-apm-agent&v=$GIT_TAG"
```
v1v: Minor suggestion to avoid the download output

```suggestion
curl -L -s -o apm-agent-java.jar "http://repository.sonatype.org/service/local/artifact/maven/redirect?r=central-proxy&g=co.elastic.apm&a=elastic-apm-agent&v=$GIT_TAG"
```

eyalkoren: Do we know this runs only when the artifact is available on Sonatype? 
Why not using the local artifact?

scripts/jenkins/push_docker.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+set -euxo pipefail
+
+# This script is intended to work in conjunction with the build_docker.#!/bin/sh
```
v1v:
```suggestion
# This script is intended to work in conjunction with the build_docker 
```

scripts/jenkins/push_docker.sh
```diff
@@ -0,0 +1,30 @@
...
+readonly CUR_TAG=${CUR_TAG:-$CUR_TAG_DEFAULT}
+
+# 2. Construct the image:tag that we are working with
+# This is roughly <repo>/<namespace>/image
+readonly DOCKER_PUSH_IMAGE="docker.elastic.co/apm/apm-agent-java:$CUR_TAG"
+
+# 3. Proceed with pushing to the registry
+readonly DOCKER_REGISTRY_URL=`echo $DOCKER_PUSH_IMAGE|cut -f1 -d/`
+echo "INFO: Pushing image $DOCKER_PUSH_IMAGE to $DOCKER_REGISTRY_URL"
+docker push $DOCKER_PUSH_IMAGE || echo "Push failed. When running in " \
```
v1v: Sometimes the docker registry can be offline or got some glitches, a retry with asleep could help?  Something similar to:
- https://github.com/elastic/apm-agent-java/blob/master/.ci/packer_cache.sh#L6

scripts/jenkins/push_docker.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
v1v: Cosmetic preference :)
```suggestion
#!/usr/bin/env bash
```

CONTRIBUTING.md
cachedout: That's true for the time being. We are hoping that soon we can migrate everything over here which is why I left this in but after looking at this, I think it could mislead people. I'll remove it. Thanks!

Jenkinsfile
```diff
@@ -286,6 +288,18 @@ pipeline {
         tag pattern: 'v\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
       }
       stages {
+        stage('Docker build') {
```
v1v: If the docker registry instance is only for internal usage/access, what do you think to change this stage with 'Generate staging docker image' or something similar?

cachedout: The goal that I believe @eyalkoren is hoping for is to have a Docker image that can be consumed by the public and is not just only for internal usage.

eyalkoren: Yes, it needs to be public. We are going to publish a blog post next week telling users how it can be used as an init container in Kubernetes.

Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM alpine:3.9
```
v1v: NIT: could it be more accurate with the SHA reference? Just wondering if the `alpine:3.9` image might be no deterministic

Jenkinsfile
v1v: got it, what about 'Docker image' or `Docker push` as it does consist of two steps, `docker build` and `docker push`, and the aim of this stage is to publish the Docker image,?

scripts/jenkins/push_docker.sh
cachedout: Good call! I wasn't aware of the `bash_standard_lib.sh` inclusion in other places so this is good to know! Thanks!

Jenkinsfile
cachedout: I changed it to `Docker push` per your latest suggestion. Thanks!

Dockerfile
cachedout: Good idea! I have pinned this to ths SHA of Alpine 3.9. Thank you!

scripts/jenkins/push_docker.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+# This script is present on workers but may not be present in a development
+# environment.
+
+if [ ${WORKERS+x} ]  # We are on a CI worker
```
v1v: I'm not familiar with this env variable, where is it coming from?

cachedout: Ugh. This is what happens when I don't get enough sleep. This is supposed to be `WORKSPACE`. I'll fix.

scripts/jenkins/build_docker.sh
cachedout: I went ahead and made using the local artifact the default. It is not present, the script presents instructions on how to re-run it with a flag set to use the Sonatype artifact.

scripts/jenkins/build_docker.sh
cachedout: I modified it to use `TAG_NAME` if it is present in the environment. Thanks!

scripts/jenkins/build_docker.sh
```diff
@@ -0,0 +1,68 @@
...
+readonly SCRIPT_PATH="$( cd "$(dirname "$0")" ; pwd -P )"
+readonly PROJECT_ROOT=$SCRIPT_PATH/../../
+readonly NAMESPACE="observability"
+
+if [ "$(ls -A  $SCRIPT_PATH/../../elastic-apm-agent/target)" ]
+then
+  # We have build files to use
+  echo "INFO: Found local build artifact. Using locally built for Docker build"
+  find -E $PROJECT_ROOT/elastic-apm-agent/target -regex '.*/elastic-apm-agent-[0-9]+.[0-9]+.[0-9]+(-SNAPSHOT)?.jar' -exec cp {} $PROJECT_ROOT/apm-agent-java.jar \; || echo "INFO: No locally built image found"
+  # cp $SCRIPT_PATH/../../elastic-apm-agent/target/elastic-apm-agent-.*!('-javadoc'|'-sources').jar $SCRIPT_PATH/../../apm-agent-java.jar
```
v1v: Is this required?

scripts/jenkins/build_docker.sh
```diff
@@ -0,0 +1,68 @@
...
+  readonly GIT_TAG_DEFAULT=$(git describe --abbrev=0|sed s/^v//)
+fi
+
+readonly GIT_TAG=${GIT_TAG:-$GIT_TAG_DEFAULT}
+
+readonly SCRIPT_PATH="$( cd "$(dirname "$0")" ; pwd -P )"
+readonly PROJECT_ROOT=$SCRIPT_PATH/../../
+readonly NAMESPACE="observability"
+
+if [ "$(ls -A  $SCRIPT_PATH/../../elastic-apm-agent/target)" ]
```
v1v:
```suggestion
if [ "$(ls -A  ${PROJECT_ROOT}elastic-apm-agent/target)" ]
```

scripts/jenkins/build_docker.sh
```diff
@@ -0,0 +1,68 @@
...
+then
+  # We have build files to use
+  echo "INFO: Found local build artifact. Using locally built for Docker build"
+  find -E $PROJECT_ROOT/elastic-apm-agent/target -regex '.*/elastic-apm-agent-[0-9]+.[0-9]+.[0-9]+(-SNAPSHOT)?.jar' -exec cp {} $PROJECT_ROOT/apm-agent-java.jar \; || echo "INFO: No locally built image found"
```
v1v: PROJECT_ROOT ends with `/`

```suggestion
  find -E ${PROJECT_ROOT}elastic-apm-agent/target -regex '.*/elastic-apm-agent-[0-9]+.[0-9]+.[0-9]+(-SNAPSHOT)?.jar' -exec cp {} ${PROJECT_ROOT}apm-agent-java.jar \; || echo "INFO: No locally built image found"
```

scripts/jenkins/build_docker.sh
```diff
@@ -0,0 +1,68 @@
...
+  curl -L -s -o apm-agent-java.jar \
+    "http://repository.sonatype.org/service/local/artifact/maven/redirect?r=central-proxy&g=co.elastic.apm&a=elastic-apm-agent&v=$GIT_TAG"
+  else
+    echo "ERROR: No suitable build artifact was found. Re-running this script with the SONATYPE_FALLBACK variable set to true will try to use the Sonatype artifact for the latest tag"
+    exit 1
+fi
+
+echo "INFO: Starting Docker build for version $GIT_TAG"
+
+docker build . -t docker.elastic.co/$NAMESPACE/apm-agent-java:$GIT_TAG \
```
v1v: Cosmetic change, what do you think to `.` at the very end of the docker build call ?

```
docker build -t docker.elastic.co/$NAMESPACE/apm-agent-java:$GIT_TAG \
             --build-arg JAR_FILE=apm-agent-java.jar .

```

scripts/jenkins/build_docker.sh
```diff
@@ -0,0 +1,68 @@
...
+  --build-arg JAR_FILE=apm-agent-java.jar
+
+if [ $? -eq 0 ]
+then
+  echo "INFO: Docker image built succesfully"
+else
+  echo "ERROR: Problem building Docker image!"
+fi
+
+function finish {
```
v1v: It might not be required when using the CI ephemeral workers

scripts/jenkins/push_docker.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# This script is intended to be run from a CI job and will not work if run in
+# standalone manner unless certain envrionment variables are set.
+
+# 1. Grab the tag we are working with
+
+#FIXME DRY between Docker scripts
+echo "INFO: Fetching latest tag"
+#TODO Move this into apm-pipeline-library
+CUR_TAG_DEFAULT=$(git describe --abbrev=0|sed s/^v//)
```
v1v: Should it use the same approach as the one for building?

scripts/jenkins/build_docker.sh
cachedout: Not needed. It is handled in L35 as part of the `find` command with the `-exec` flag. I will remove it for the sake of clarity.

scripts/jenkins/build_docker.sh
cachedout: When writing this, I tried to ensure that it could be cleanly run from a workstation as well. I agree that it is not needed on the workers but shouldn't cause any harm to leave it in and it ensures that a local developer's environment remains tidy if executing this script locally.

scripts/jenkins/build_docker.sh
cachedout: I checked the Docker documentation and it looks like they recommend the same form that you do. I will change this to stick with that style. Thanks!

scripts/jenkins/push_docker.sh
cachedout: Fixed.

cachedout:
@eyalkoren Could you please take a final look at this? Thanks!

cachedout:
Jenkins test this please

PR #1005, opened 2020-01-23
Fix ObjectPool PR merge error

eyalkoren:
@SylvainJuge for some reason this invalid import was introduced by merging elastic/apm-agent-java#967 🤷‍♂ 
Not sure what I missed here, but please take a look if this is now as expected.


eyalkoren:
Right, this shouldn't be the case, we rely on these tests for merging...
Your last commit passed all tests.
Maybe it got corrupted when commits were squashed? I couldn't find evidence for that.

felixbarny:
That can happen when relocating files, in this case `Resetter`. Eyal, you introduced a new usage of `Resetter` for the message body `StringBuilder`. Sylvain moved the interface to a different package. Both PRs are consistent in itself and tests pass. But when both are merged at the same time, without rebasing one after the first has been merged, it can result in compile errors on master.

eyalkoren:
It appears that the Jetty 9.4 container was updated lately, now behaving similar to 9.2 and 9.3 with regard to where to expect stack traces capture 😫

eyalkoren:
@felixbarny that's right! I actually DID look at the merges int #967, but read them wrong thus assuming the Kafka PR was merged... 😫 X 2

Issue #1002, opened 2020-01-22
Duplicated smoke tests


cachedout:
### Credit

Credit goes to @eyalkoren for the discovery of this issue

### Overview

It appears that certain tests are duplicated across sets of smoke tests. For example, take the `greetingShouldReturnDefaultMessage` test. This test is located in the `integrations/test` directory, and can [be seen here](https://github.com/elastic/apm-agent-java/blob/master/integration-tests/spring-boot-1-5/src/test/java/co/elastic/apm/spring/boot/SpringBoot1_5IT.java#L87).

If one examines a test run by [opening a run and going to the Blue Ocean page](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/master/218/tests), they can see that the same test appears to have run as a part of both the `Smoke Tests 01` group and the `Smoke Tests 02` group. 

### Details

Tests are segmented through the use of simple bash scripts. The scripts which provide this segmentation are below:

[Smoke Tests 01](https://github.com/elastic/apm-agent-java/blob/master/scripts/jenkins/smoketests-01.sh)
Smoke Tests 02(https://github.com/elastic/apm-agent-java/blob/master/scripts/jenkins/smoketests-02.sh)

Each script contains a preparatory step to generate the modules to be tested. Below is the output of each:

#### smoketest01 
`find apm-agent-plugins -maxdepth 1 -mindepth 1 -type d|grep -v "target"`

```
apm-agent-plugins/apm-mongoclient-plugin
apm-agent-plugins/apm-redis-plugin
apm-agent-plugins/apm-spring-webmvc-plugin
apm-agent-plugins/apm-quartz-job-plugin
apm-agent-plugins/apm-jaxrs-plugin
apm-agent-plugins/apm-jsf-plugin
apm-agent-plugins/apm-httpclient-core
apm-agent-plugins/apm-kafka-plugin
apm-agent-plugins/apm-java-concurrent-plugin
apm-agent-plugins/apm-jdbc-plugin
apm-agent-plugins/apm-jmx-plugin
apm-agent-plugins/apm-servlet-plugin
apm-agent-plugins/apm-es-restclient-plugin
apm-agent-plugins/apm-opentracing-plugin
apm-agent-plugins/apm-scheduled-annotation-plugin
apm-agent-plugins/apm-mule4-plugin
apm-agent-plugins/apm-jaxws-plugin
apm-agent-plugins/apm-api-plugin
[...]

PR #1014, opened 2020-01-29
[DO NOT MERGE] Remove parallelism from smoke tests for debugging

cachedout:
This is a test PR to observe the CI system running smoke tests sequentially instead of in parallel. It may then be easier to determine how tests are being duplicated.

Refs https://github.com/elastic/apm-agent-java/issues/1002

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -85,6 +85,70 @@ pipeline {
         }
       }
     }
+
+// TMP move to watch tests run sequentially 
+      /**
+        Run smoke tests for different servers and databases.
+      */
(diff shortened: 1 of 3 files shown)


cachedout:
This didn't turn out to yield any useful information. We don't get individual log entries for tests as they are executed.

cachedout:
Let's try this again without the `quiet` flag.

cachedout:
I have what I need from this for the time being. Closing so that it doesn't linger in anybody's workflow. :)

cachedout:
This was only a temporary PR designed just to test one thing in the CI. It should remain closed.

Issue #813, opened 2019-08-23
Return error id from Span#captureException


felixbarny:
Applications can then show that error ID to their users. This makes it easier to analyze user-reported errors.

kananindzya:
Hi @felixbarny , 
Can you review PR ? 
Is it correct that we should replace returning string instead of void?

PR #1015, opened 2020-01-29
return error id with captureException in public api

kananindzya:
## Checklist

## Related issues

Closes elastic/apm-agent-java#813 
## Screenshots
<!-- _(Optional)_

### DIFF
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
@@ -11,9 +11,9 @@
  * the Apache License, Version 2.0 (the "License"); you may
  * not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -58,7 +58,7 @@ public interface Span {
     /**
      * NOTE: THIS METHOD IS DEPRECATED AND WILL BE REMOVED IN VERSION 2.0.
      * Instead, setting the span type can be done when starting a new span through {@link #startSpan(String, String, String)}.
-     * 
+     *
      * @param type the type of the span
      */
     @Nonnull
@@ -79,9 +79,9 @@ public interface Span {
      * Multiple labels can be defined with different key-value pairs.
      * </p>
      * <ul>
-     *     <li>Indexed: Yes</li>
-     *     <li>Elasticsearch type: <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html">object</a></li>
-     *     <li>Elasticsearch field: {@code labels} (previously {@code context.tags} in stack version {@code < 7.0})</li>
+     *      <li>Indexed: Yes</li>
+     *      <li>Elasticsearch type: <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html">object</a></li>
+     *      <li>Elasticsearch field: {@code labels} (previously {@code context.tags} in stack version {@code < 7.0})</li>
      * </ul>
      * <p>
      * Label values can be a string, boolean, or number.
@@ -108,9 +108,9 @@ public interface Span {
      * Multiple labels can be defined with different key-value pairs.
      * </p>
      * <ul>
-     *     <li>Indexed: Yes</li>
-     *     <li>Elasticsearch type: <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html">object</a></li>
-     *     <li>Elasticsearch field: {@code labels} (previously {@code context.tags} in stack version {@code < 7.0})</li>
+     *      <li>Indexed: Yes</li>
+     *      <li>Elasticsearch type: <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html">object</a></li>
+     *      <li>Elasticsearch field: {@code labels} (previously {@code context.tags} in stack version {@code < 7.0})</li>
      * </ul>
      * <p>
      * Label values can be a string, boolean, or number.
@@ -141,9 +141,9 @@ public interface Span {
      * Multiple labels can be defined with different key-value pairs.
      * </p>
      * <ul>
-     *     <li>Indexed: Yes</li>
-     *     <li>Elasticsearch type: <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html">object</a></li>
-     *     <li>Elasticsearch field: {@code labels} (previously {@code context.tags} in stack version {@code < 7.0})</li>
+     *      <li>Indexed: Yes</li>
+     *      <li>Elasticsearch type: <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html">object</a></li>
+     *      <li>Elasticsearch field: {@code labels} (previously {@code context.tags} in stack version {@code < 7.0})</li>
      * </ul>
      * <p>
      * Label values can be a string, boolean, or number.
@@ -281,7 +281,15 @@ public interface Span {
      *
      * @param throwable the exception to report
      */
-    String captureException(Throwable throwable);
+    void captureException(Throwable throwable);
+
+    /**
+     * Captures an exception and reports it to the APM server.
+     *
+     * @param throwable the exception to report
+     * @return the id of reported error
+     */
+    String captureExceptionAndReturnErrorId(Throwable throwable);
 
     /**
      * Returns the id of this span (never {@code null})
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
@@ -79,9 +79,9 @@ public interface Span {
      * Multiple labels can be defined with different key-value pairs.
      * </p>
      * <ul>
-     *      <li>Indexed: Yes</li>
-     *      <li>Elasticsearch type: <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html">object</a></li>
-     *      <li>Elasticsearch field: {@code labels} (previously {@code context.tags} in stack version {@code < 7.0})</li>
+     *     <li>Indexed: Yes</li>
+     *     <li>Elasticsearch type: <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/object.html">object</a></li>
+     *     <li>Elasticsearch field: {@code labels} (previously {@code context.tags} in stack version {@code < 7.0})</li>
      * </ul>
      * <p>
      * Label values can be a string, boolean, or number.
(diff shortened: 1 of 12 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -320,7 +321,14 @@ public void captureException(long epochMicros, @Nullable Throwable e, TraceConte
         captureException(epochMicros, e, parent, null);
     }
 
-    public void captureException(long epochMicros, @Nullable Throwable e, @Nullable TraceContextHolder<?> parent, @Nullable ClassLoader initiatingClassLoader) {
+    @Nullable
+    public String captureExceptionAndGetErrorId(long epochMicros, @Nullable Throwable e, @Nullable TraceContextHolder<?> parent) {
```
eyalkoren: I don't think we need this API, it doesn't save much to the caller (only extract the ID). The one returning the `ErrorCapture` is very useful as is. 
I can easily imaging how a caller may want to do lots of other stuff with the `ErrorCapture` and we wouldn't add an API for each usage.

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ElasticApmTracerTest.java
```diff
@@ -489,4 +490,17 @@ void testOverrideServiceNameExplicitServiceName() {
         tracer.startTransaction(TraceContext.asRoot(), null, getClass().getClassLoader()).end();
         assertThat(reporter.getFirstTransaction().getTraceContext().getServiceName()).isNull();
     }
+
+    @Test
+    void testCaptureExceptionAndGetErrorId() {
+        Transaction transaction = tracerImpl.startTransaction(TraceContext.asRoot(), null, getClass().getClassLoader());
+        String errorId = tracerImpl.captureExceptionAndGetErrorId(System.currentTimeMillis() * 1000, new Exception("test"), transaction);
```
eyalkoren: Better call `transaction.captureException`, then you test both added methods.

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/plugin/api/AbstractSpanInstrumentation.java
```diff
@@ -178,8 +178,9 @@ public CaptureExceptionInstrumentation() {
         @VisibleForAdvice
         @Advice.OnMethodExit(suppress = Throwable.class)
         public static void captureException(@Advice.FieldValue(value = "span", typing = Assigner.Typing.DYNAMIC) TraceContextHolder<?> context,
-                                            @Advice.Argument(0) Throwable t) {
-            context.captureException(t);
+                                            @Advice.Argument(0) Throwable t,
+                                            @Advice.Return(readOnly = false) String errorId) {
```
eyalkoren: This probably won't work when a user upgrades the agent and not the API (a very valid scenario)- the agent will instrument the `captureException` API that returns void and the instrumentation will expect a String being returned.

What you need to do is change the constructor of this class, so that the method matcher is more specific on the return type, probably something like: `named("captureException").and(takesArguments(Throwable.class)).and(returns(TypeDescription.VOID))` and add another `CaptureExceptionWithReturnIdInstrumentation` class that uses a method matcher like: `named("captureException").and(takesArguments(Throwable.class)).and(returns(TypeDescription.STRING))` and does what you did here. 

Eventually, you must test your branch's agent with an older API jar, and you should test your branch's API jar with an older agent. We don't have automatic tests for that, but should be easy to do manually.

apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
```diff
@@ -281,7 +281,7 @@
      *
      * @param throwable the exception to report
      */
-    void captureException(Throwable throwable);
+    String captureException(Throwable throwable);
```
felixbarny: Technically, this is a breaking change in binary compatibility. When not re-compiling the application and updating the dependency, it would result in an exception when calling the method. That is because methods are linked also based on their return type.
But still think we can leave it as-is as most likely, someone would re-compile the application again after updating the dependency to the agent API. 

A more conservative option would be to create a method with a different name, like `captureExceptionAndReturnErrorId` and leave `void captureException(Throwable throwable)` as-is.

eyalkoren: Good point! 
I agree it's best to leave as is (meaning - just change the method signature) and not to complicate the API for this corner case.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
kananindzya: fixed

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ElasticApmTracerTest.java
kananindzya: fixed

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/plugin/api/AbstractSpanInstrumentation.java
kananindzya: added new method captureExceptionAndReturnErrorId

apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
kananindzya: added new method

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
kananindzya: added documentation

kananindzya:
with example:
```
Thread captureExceptionViaSpan = new Thread(() -> {
            Transaction transaction = ElasticApm.startTransaction();
            try {
                transaction.setName("Thread2#run");
                transaction.setType(Transaction.TYPE_REQUEST);

                Span span = transaction.startSpan("db", "mysql", "query");
                try {
                    span.setName("SELECT FROM customer");
                    throw new RuntimeException("Thread-2 span exception");
                    // do your thing...
                } catch (Exception e) {
                    String spanErrorId = span.captureExceptionAndReturnErrorId(e);
                    System.out.println("spanErrorId = " + spanErrorId);
                } finally {
                    span.end();
                }
            } catch (Exception e) {
                transaction.captureExceptionAndReturnErrorId(e);
            } finally {
                transaction.end();
            }
        });
```
```
2020-02-03 12:06:11.484 [Thread-10] DEBUG co.elastic.apm.agent.impl.transaction.AbstractSpan - increment references to 'Thread2#run' 00-73a67e878c6d82fdcab2365b40729884-f74234b2c9f1b1de-01 (6bd054e1) (3)
spanErrorId = d17fb71297fab7ec453c749796a25069
2020-02-03 12:06:11.485 [Thread-10] DEBUG co.elastic.apm.agent.impl.transaction.Span - endSpan 'SELECT FROM customer' 00-73a67e878c6d82fdcab2365b40729884-579076be63f9af4a-01 (7cc225da)
2020-02-03 12:06:11.485 [Thread-11] DEBUG co.elastic.apm.agent.impl.transaction.Span - endSpan 'SELECT FROM customer' 00-c6834ca9640b8717c1b923337655414f-ff237a4914ef7822-01 (4aa58328)
2020-02-03 12:06:11.485 [Thread-10] DEBUG co.elastic.apm.agent.impl.transaction.AbstractSpan - decrement references to 'Thread2#run' 00-73a67e878c6d82fdcab2365b40729884-f74234b2c9f1b1de-01 (6bd054e1) (2)
2020-02-03 12:06:11.485 [Thread-11] DEBUG co.elastic.apm.agent.impl.transaction.AbstractSpan - decrement references to 'Thread3#run' 00-c6834ca9640b8717c1b923337655414f-3243eabeb90b9e9a-01 (1318dfe8) (2)
2020-02-03 12:06:11.
```
and screenshot from kibana, with errorId
image

apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
felixbarny: I think you got us wrong. While there is a corner case for a potential breaking change, just leave `captureException` as the method name to not clutter the API too much.

kananindzya:
> After changing `captureExceptionAndReturnErrorId` back to `captureException` this should be good to merge. Thanks a lot heart

I fixed this.

SylvainJuge:
@kananindzya did you pushed your changes already ? I don't see any new commits.

kananindzya:
> @kananindzya did you pushed your changes already ? I don't see any new commits.

hah, sorry ) pushed

SylvainJuge:
Compatibility with previous versions have been added and tested (manually)

The following cases have been tested:
- previous API in application + updated agent --> "Legacy" instrumentation is being used, thus behaves as usual
- updated API in application + previous agent --> behaves as expected, but `""` (empty string) is returned as error ID.
- updated API in application + updated agent --> returns the error ID.

Thus should be good to merge :tada:

felixbarny:
Changelog is still missing

docs/public-api.asciidoc
```diff
@@ -382,8 +382,8 @@ transaction.setUser(user.getId(), user.getEmail(), user.getUsername());
 
 [float]
 [[api-transaction-capture-exception]]
-==== `void captureException(Exception e)`
-Captures an exception and reports it to the APM server.
+==== `String captureException(Exception e)`
+Captures an exception and reports it to the APM server. Returns the id of reported error.
```
eyalkoren:
```suggestion
Captures an exception and reports it to the APM server. Since version 1.14.0 - returns the id of reported error.
```

docs/public-api.asciidoc
```diff
@@ -666,8 +666,8 @@ span.addLabel("foo", "bar");
 
 [float]
 [[api-span-capture-exception]]
-==== `void captureException(Exception e)`
-Captures an exception and reports it to the APM server.
+==== `String captureException(Exception e)`
+Captures an exception and reports it to the APM server. Returns the id of reported error.
```
eyalkoren:
```suggestion
Captures an exception and reports it to the APM server. Since version 1.14.0 - returns the id of reported error.
```

kananindzya:
> Changelog is still missing

added

SylvainJuge:
seems that you forgot to get a proper coffee this morning, you forgot to push again :-)

kananindzya:
> seems that you forgot to get a proper coffee this morning, you forgot to push again :-)
bugaga :)
```
nindzya@pc:~/Documents/PROJECTS/apm-agent-java$ git push origin issue-813-return-span-error-id 
Enumerating objects: 9, done.
Counting objects: 100% (8/8), done.
Delta compression using up to 8 threads
Compressing objects: 100% (5/5), done.
Writing objects: 100% (5/5), 648 bytes | 648.00 KiB/s, done.
Total 5 (delta 3), reused 0 (delta 0)
remote: Resolving deltas: 100% (3/3), completed with 2 local objects.
^C
nindzya@pc:~/Documents/PROJECTS/apm-agent-java$ git push origin issue-813-return-span-error-id 
Enumerating objects: 9, done.
Counting objects: 100% (8/8), done.
Delta compression using up to 8 threads
Compressing objects: 100% (5/5), done.
Writing objects: 100% (5/5), 648 bytes | 648.00 KiB/s, done.
Total 5 (delta 3), reused 0 (delta 0)
remote: Resolving deltas: 100% (3/3), completed with 2 local objects.
To github.com:kananindzya/apm-agent-java.git
   8934870b..db9662ad  issue-813-return-span-error-id -> issue-813-return-span-error-id

```

SylvainJuge:
Thanks a lot @kananindzya !

PR #1017, opened 2020-02-02
Small change in main configuration doc

eyalkoren:
Something didn't sit well with this sentence for me. Would it be better this way?

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -8,8 +8,7 @@ Please only make changes in configuration.asciidoc.ftl
 To adapt the Elastic APM agent to your needs,
 you can configure it using different configuration sources,
 which have different naming conventions for the property key.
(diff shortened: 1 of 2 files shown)


felixbarny:
This sentence is indeed a bit strange. But not sure if it's clearer now. Maybe @bmorelli25 can help.

bmorelli25:
I agree the original line is confusing. I also agree with Felix that I'm not sure this edit clears it up. What about using _precedence?_
```diff
- The first configuration sources override the configuration values of over the latter sources.
+ In descending order of precedence:
```

We could go a bit further:
```diff
- The first configuration sources override the configuration values of over the latter sources.
+ In descending order of precedence (smallest number wins).
```

eyalkoren:
Thanks for the feedback. Changed accordingly. Approve to verify this looks OK now.

PR #1020, opened 2020-02-04
Add log correlation for log4j and log4j2

### LINKED ISSUE #922: Log correlation: add support for log4j2 and log4j
Currently, only slf4j is supported. So when using plain log4j without slf4j, log correlation does not work.

Nowadays, it's not necessarily recommended or required to use slf4j in order to swap out implementations. Log4j2 also has an API and can bridge to Logback, for example.

felixbarny:
## What does this PR do?

closes elastic/apm-agent-java#922 

## Checklist

~- [ ] Added an API method or config option? Document in which version this will be introduced~
~- [ ] Added an instrumentation plugin? How did you make sure that old, non-supported versions are not instrumented by accident?~


felixbarny:
Checks whether the class loader which initiated the transaction can load slf4j, log4j or log4j2. For each logging framework, it populates the MDC. So even if slf4j is wired to use log4j2, we will set the MDC twice. But that shouldn't be a problem as `put` and `remove` are idempotent. 
This approach also works in the cases, where slf4j is used with Logback, and log4j and log4j2 are, for some reason, also used in the same application.

apm-agent-plugins/apm-log-correlation-plugin/src/main/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListener.java
```diff
@@ -41,44 +41,101 @@
...
+    private static final String SLF4J_MDC = "org!slf4j!MDC".replace('!', '.');
+    private static final String LOG4J_MDC = "org.apache.log4j.MDC";
+    private static final String LOG4J2_MDC = "org.apache.logging.log4j.ThreadContext";
+
     private static final String TRACE_ID = "trace.id";
     private static final String TRANSACTION_ID = "transaction.id";
     private static final Logger logger = LoggerFactory.getLogger(Slf4JMdcActivationListener.class);
 
     // Never invoked- only used for caching ClassLoaders that can't load the slf4j MDC class
     private static final MethodHandle NOOP = MethodHandles.constant(String.class, "ClassLoader cannot load slf4j API");
```
eyalkoren: The returned message is not really used, but change it to something generic so that it won't be misleading (same for the comment)

apm-agent-plugins/apm-log-correlation-plugin/src/test/java/co/elastic/apm/agent/slf4j/Slf4JMdcActivationListenerTest.java
```diff
@@ -49,12 +53,25 @@
...
     private LoggingConfiguration loggingConfiguration;
     private CoreConfiguration coreConfiguration;
+    private Boolean log4jMdcWorking;
+    private ExecutorService executorService = Executors.newSingleThreadExecutor();
 
     @BeforeEach
-    void setUp() {
+    void setUp() throws Exception {
+        org.apache.log4j.MDC.put("test", true);
+        log4jMdcWorking = (Boolean) org.apache.log4j.MDC.get("test");
```
eyalkoren: What is this?

felixbarny: Log4j thinks that Java 11 is Java 1 and disables `ThreadLocal`s

felixbarny:
> Package and class names still contain slf4j.

What do you mean with that?

Issue #1037, opened 2020-02-14
java.lang.IllegalStateException: Cannot resolve type description for 


WassimDhib:
## Describe the bug
I have an issue when starting Jboss server

## Steps to reproduce
Whe have on this linux server an other commercial apm which is deactivated and stopped, but it is still have a part loaded with linux LD_PRELOAD

## Debug logs
```
12/02/2020 12:41:52-239 | INFO  | stdout | 2020-02-12 12:41:52.239 [ServerService Thread Pool -- 41] INFO co.elastic.apm.agent.bci.bytebuddy.ErrorLoggingListener - com.another.commercial.apm.agent.introspection.uem.impl.CacheHookingRequestWrapper refers to a missing class
12/02/2020 12:41:52-240 | INFO  | stdout | java.lang.IllegalStateException: Cannot resolve type description for com.another.commercial.apm.agent.introspection.uem.apiobjects.UemServletRequestWrapperInterface
12/02/2020 12:41:52-241 | INFO  | stdout |      at co.elastic.apm.agent.shaded.bytebuddy.pool.TypePool$Resolution$Illegal.resolve(TypePool.java:159)
12/02/2020 12:41:52-242 | INFO  | stdout |      at co.elastic.apm.agent.shaded.bytebuddy.pool.TypePool$Default$WithLazyResolution$LazyTypeDescription.delegate(TypePool.java:914)
12/02/2020 12:41:52-242 | INFO  | stdout |      at co.elastic.apm.agent.shaded.bytebuddy.description.type.TypeDescription$AbstractBase$OfSimpleType$WithDelegation.getModifiers(TypeDescription.java:8122)
12/02/2020 12:41:52-242 | INFO  | stdout |      at co.elastic.apm.agent.shaded.bytebuddy.description.ModifierReviewable$AbstractBase.matchesMask(ModifierReviewable.java:618)
12/02/2020 12:41:52-243 | INFO  | stdout |      at co.elastic.apm.agent.shaded.bytebuddy.description.ModifierReviewable$AbstractBase.isPublic(ModifierReviewable.java:336)
12/02/2020 12:41:52-243 | INFO  | stdout |      at co.elastic.apm.agent.shaded.bytebuddy.description.type.TypeDescription$AbstractBase.isVisibleTo(TypeDescription.java:7631)
[...]

PR #1038, opened 2020-02-14
Add instrumentation excludes for common APM vendors

felixbarny:


SylvainJuge:
should also add `com.compuware` as suggested in #1037

SylvainJuge:
And also `io.sqreen` might also make sense, even if it's not an APM, it is more likely to be used with another agent.

WassimDhib:
also com.appdynamics.

SylvainJuge:
Also, "com.singularity" which used to be the package used previously by
AppDynamics, there are still some references to it on their support forum:
https://community.appdynamics.com/t5/Java-Java-Agent-Installation-JVM/java-lang-NoClassDefFoundError-com-singularity-ee-agent-appagent/td-p/29468

felixbarny:
I have already added `com.singularity`, I don't think there's a `com.appdynamics` package.

PR #1040, opened 2020-02-16
Adding pause/resume functionality to reduce agent overhead on stress

### LINKED ISSUE #1029: Temporarily disable agent when the application is at risk of running out of memory
Java implementation of https://github.com/elastic/apm/issues/201

## Motivation

When the application is under high load and at risk to run into a `OutOfMemoryError`, the agent should deactivate itself and free any memory it currently holds. 

## Detection of an overloaded application

When the heap utilization reaches a certain threshold, like 90% the agent waits for a GC. If the heap utilization then still exceeds the threshold, the agent deactivates itself.

## Resuming

If the heap utilization is under the threshold, the agent should resume monitoring.

## What this is not
This does not limit the overhead of the agent but is a mechanism to ensure that the agent is not the straw that breaks the camel’s back if the application is about to crash. We’re not looking at the overhead the agent imposes on the application as the agent runs within the application. It would be very hard and expensive (in terms of performance) to calculate just the CPU time that the agent itself spends. However, the agent is designed to have very little overhead (unless tracing too many methods via `trace_methods`).

Disabling the agent will also not tune down the amount of instrumentation. When the agent deactivates itself, it does not remove the instrumentation but just short-circuits the tracing. 

## Impact on `trace_methods`:
If lots of methods are instrumented via `trace_methods`, temporarily disabling the agent does not significantly reduce the overhead. 

## Impact on inferred spans via async-profiler:
When the agent is temporarily disabled, async-profiler can be completely disabled and any overhead related to it will be zero. This is because async-profiler does not instrument methods but rather takes thread dumps/stack traces at regular intervals.

eyalkoren:
## What does this PR do?
This PR introduces a tracer state, including currently the following states:
- `RUNNING` - agent is fully functional
- `PAUSED` - agent resource consumption (more specifically- CPU*, memory allocations and IO) is at minimum, but can quickly fully resume to normal working state
- `STOPPED` - agent is non functional

\* If agent is configured to heavily instrument methods (for example, if `trace_methods` config option is abused), CPU overhead may still be considerable.

In addition, it extends the `LifecycleListener` interface to include `pause` and `resume` methods, to be implemented by any heavy resource users. 
A new `LifecycleListener` implementation runs a background thread that polls system/process stress monitors on preset intervals. If the agent in in `RUNNING` state and ANY of the monitors indicates that the tracer should be pause, it will invoke this state change. If the agent in in `PAUSED` state and ALL of the monitors indicate that the tracer should resume, it will invoke this state change.  

This PR is currently in Draft mode, mainly used for allowing design review.

## Checklist

## Author's Checklist

- [x] Consolidate the `active` configuration with the circuit breaker events into a single tracer's state
- [x] Make monitors' settings configurable and properly documented
- [x] Add a GC stress monitor (test manually)
- [x] Add a CPU stress monitor (test manually)
- [x] Test compatibility with HotSpot and J9
- [x] In potentially massive instrumentations, when agent is `PAUSED`, instrumented code being executed should be minimal - we do not create spans in `trace_methods` spans, but we cannot change the API spans currently - changing the API to return `null` on span creation is a breaking change and using an immutable NOOP object means we cannot support context propagation.
- [x] Implement `pause` and `resume` callbacks in Async Profiler `LifecycleListener`s - @felixbarny I'll leave this one to you
- [x] Verify that break down metrics are not collected when `PAUSED`
- [x] Align plugins to behave the same with regard to context propagation when the agent is `PAUSED` (spans should not be created).
- [x] Look into clearing object pools when pausing without increasing usage overhead - decided to defer this one for now - only clearing the pool as a `pause` callback doesn't make a lot of sense as it won't include all in-flight objects. Adding a timestamp on objects and return only objects allocated after the `pause` means concurrency (visibility) overhead and complication. I think the best option is using a thread that will periodically clear specific pools (probably only spans and transactions) when the agent is paused. Still, this should come with a safety check that we are actually not keeping allocating objects, thus increasing overhead...

## Related issues

Closes #1029 

### GC monitoring
When garbage collector fails to recover minimal portion of the heap (configurable by percentage), agent should switch to `PUASED` state. When garbage collector succeeds to recover sufficient heap space (configurable by percentage), agent should resume to `RUNNING` state.

### CPU monitoring
If the **system** CPU load crosses a configured threshold for a consecutive (configured) number of measurements- the agent will pause. By default, that would happen after a minute (12 consecutive measurements done in 5 second intervals) with system CPU usage over 95%.
When the system CPU load is measured below another threshold for the same period of time, the agent will resume. By default, that would happen after one minute of CPU under 80%.

#### Agent log
```
2020-02-20 18:14:23.950 [elastic-apm-circuit-breaker] INFO co.elastic.apm.agent.impl.circuitbreaker.GCStressMonitor - Heap G1 Old Gen pool usage after the last GC has crossed the configured threshold 0.8: 133650656/157286400 (used/max)
2020-02-20 18:14:23.950 [elastic-apm-circuit-breaker] INFO co.elastic.apm.agent.impl.ElasticApmTracer - Tracer switched to PAUSED state
...
2020-02-20 18:15:08.974 [elastic-apm-circuit-breaker] DEBUG co.elastic.apm.agent.impl.circuitbreaker.GCStressMonitor - Heap G1 Old Gen pool usage after the last GC is over the threshold of 0.6: 133650656/157286400 (used/max)
...
2020-02-20 18:19:34.086 [elastic-apm-circuit-breaker] DEBUG co.elastic.apm.agent.impl.circuitbreaker.GCStressMonitor - Heap G1 Old Gen pool usage after the last GC is below the threshold of 0.6: 91855984/157286400 (used/max)
2020-02-20 18:19:34.086 [elastic-apm-circuit-breaker] INFO co.elastic.apm.agent.impl.ElasticApmTracer - Tracer switched to RUNNING state
```

#### Metrics view
Screen Shot 2020-02-20 at 18 22 55

#### Transaction view
Screen Shot 2020-02-20 at 18 22 40

#### Agent log
```
020-02-25 08:54:09.075 [elastic-apm-circuit-breaker] DEBUG co.elastic.apm.agent.impl.circuitbreaker.SystemCpuStressMonitor - System CPU measurement: 0.8467177607464179
2020-02-25 08:54:09.075 [elastic-apm-circuit-breaker] INFO co.elastic.apm.agent.impl.circuitbreaker.CircuitBreaker - Stress detected by co.elastic.apm.agent.impl.circuitbreaker.SystemCpuStressMonitor: Latest system CPU load value measured is 0.8467177607464179. This is the 12th consecutive measurement that crossed the configured stress threshold - 0.8, which indicates this host is under CPU stress.
2020-02-25 08:54:09.075 [elastic-apm-circuit-breaker] INFO co.elastic.apm.agent.impl.ElasticApmTracer - Tracer switched to PAUSED state
...
2020-02-25 08:58:24.170 [elastic-apm-circuit-breaker] DEBUG co.elastic.apm.agent.impl.circuitbreaker.SystemCpuStressMonitor - System CPU measurement: 0.12389675270607826
2020-02-25 08:58:24.170 [elastic-apm-circuit-breaker] INFO co.elastic.apm.agent.impl.circuitbreaker.SystemCpuStressMonitor - Latest system CPU load value measured is 0.12389675270607826. This is 12th consecutive measurement that is below the configured relief threshold - 0.6
2020-02-25 08:58:24.170 [elastic-apm-circuit-breaker] INFO co.elastic.apm.agent.impl.circuitbreaker.CircuitBreaker - All registered stress monitors indicate that the stress has been relieved
2020-02-25 08:58:24.170 [elastic-apm-circuit-breaker] INFO co.elastic.apm.agent.impl.ElasticApmTracer - Tracer switched to RUNNING state
```

#### Metrics view
Screen Shot 2020-02-25 at 9 03 30

#### Transaction view
Screen Shot 2020-02-25 at 9 02 57


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -555,11 +561,57 @@ public void registerSpanListener(ActivationListener activationListener) {
...
+    public boolean isRunning() {
+        return state == TracerState.RUNNING;
+    }
+
+    public TracerState getState() {
+        return state;
+    }
+
+    private synchronized TracerState setState(TracerState newState) {
+        state = newState;
```
eyalkoren: Change to a simple volatile (benchmark to test)

apm-agent-plugins/apm-servlet-plugin/src/test/java/co/elastic/apm/agent/servlet/ApmFilterTest.java
```diff
@@ -79,9 +79,10 @@ void testEndsTransaction() throws IOException, ServletException {
 
     @Test
     void testDisabled() throws IOException, ServletException {
-        when(tracer.getConfig(CoreConfiguration.class).isActive()).thenReturn(false);
+        tracer.pause();
         filterChain.doFilter(new MockHttpServletRequest(), new MockHttpServletResponse());
         assertThat(reporter.getTransactions()).hasSize(0);
+        tracer.resume();
```
felixbarny: The state should be reset after each test. If there's an exception in this methods, the state should not leak into other tests. Probably best to call `tracer.resume()` in `co.elastic.apm.agent.AbstractInstrumentationTest#afterAll`

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreakerConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+            "If ANY of the monitors detects a stress indication, the agent will become inactive, as if the \n" +
+            "<<config-active>> configuration option has been set to `false`, thus reducing resource consumption to a minimum. \n" +
+            "When inactive, the agent continues polling the same monitors in order to detect whether the stress state \n" +
+            "has been relieved. If ALL monitors approve that the system/process/JVM is not under stress anymore, the \n" +
+            "agent will resume and become fully functional.")
+        .dynamic(true)
+        .buildWithDefault(false);
+
+    private final ConfigurationOption<TimeDuration> stressMonitoringInterval = TimeDurationValueConverter.durationOption("s")
+        .key("stress_monitoring_interval")
```
felixbarny: This is quite a low-level config option. Maybe make it `internal`?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreakerConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+        .configurationCategory(CIRCUIT_BREAKER_CATEGORY)
+        .description("A boolean specifying whether the circuit breaker should be enabled or not. \n" +
+            "When enabled, the agent periodically polls stress monitors to detect system/process/JVM stress state. \n" +
+            "If ANY of the monitors detects a stress indication, the agent will become inactive, as if the \n" +
+            "<<config-active>> configuration option has been set to `false`, thus reducing resource consumption to a minimum. \n" +
+            "When inactive, the agent continues polling the same monitors in order to detect whether the stress state \n" +
+            "has been relieved. If ALL monitors approve that the system/process/JVM is not under stress anymore, the \n" +
+            "agent will resume and become fully functional.")
+        .dynamic(true)
+        .buildWithDefault(false);
```
felixbarny: I think we should enable the circuit breaker by default. If the stress monitors have config options, maybe we don't need this global switch?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
```diff
@@ -0,0 +1,123 @@
...
+            try {
+                if (circuitBreakerConfiguration.isCircuitBreakerEnabled()) {
+                    if (isCurrentlyUnderStress) {
+                        boolean stressRelieved = true;
+                        for (StressMonitor stressMonitor : stressMonitors) {
+                            stressRelieved &= stressMonitor.isStressRelieved();
+                        }
+                        if (stressRelieved) {
+                            isCurrentlyUnderStress = false;
+                            tracer.stressRelieved();
```
felixbarny: Probably not for the first version but I think it makes sense to wait until the stress monitors didn't detect stress for n consecutive iterations. Otherwise, we might quickly switch between off and on.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
```diff
@@ -0,0 +1,123 @@
...
+                        }
+                    } else {
+                        for (StressMonitor stressMonitor : stressMonitors) {
+                            if (stressMonitor.isUnderStress()) {
+                                isCurrentlyUnderStress = true;
+                                tracer.stressDetected();
```
felixbarny: Log which stress monitor has detected stress and what the current values are (for example CPU% or heap%)

apm-agent-core/src/main/java/co/elastic/apm/agent/context/ClosableLifecycleListenerAdapter.java
```diff
@@ -0,0 +1,45 @@
...
+package co.elastic.apm.agent.context;
+
+import java.io.Closeable;
+
+public class ClosableLifecycleListenerAdapter extends AbstractLifecycleListener {
+
+    private final Closeable closeable;
+
+    public static LifecycleListener of(Closeable closeable) {
+        return new co.elastic.apm.agent.context.ClosableLifecycleListenerAdapter(closeable);
```
SylvainJuge: remove FQN

apm-agent-core/src/main/java/co/elastic/apm/agent/context/LifecycleListener.java
```diff
@@ -47,6 +47,34 @@
...
     void start(ElasticApmTracer tracer);
 
+    /**
+     * Callback for when {@link ElasticApmTracer#pause()} has been called.
+     * <p>
+     * Typically, this method is used to reduce overhead on the application to a minimum. This can be done by cleaning
+     * up resources like object pools, as well as by avoiding tracing-related overhead.
+     * </p>
+     * <p>
+     * Exceptions thrown from this method are caught and handled so that they don't prevent further cleanup actions.
```
SylvainJuge: maybe replace "are caught and handled" maybe "should be caught and handled", because we document how users should use this method.

apm-agent-core/src/main/java/co/elastic/apm/agent/context/LifecycleListener.java
```diff
@@ -47,6 +47,34 @@
...
+    void pause() throws Exception;
+
+    /**
+     * Callback for when {@link ElasticApmTracer#resume()} has been called.
+     * <p>
+     * Typically, used in order to revert the actions taken by the {@link LifecycleListener#pause()} method, allowing
+     * the agent to restore all tracing capabilities
+     * </p>
+     * <p>
+     * Exceptions thrown from this method are caught and handled so that they don't prevent further cleanup actions.
```
SylvainJuge: same as above, + reword it so we don't mention any cleanup (as it's a resume operation).

apm-agent-core/src/main/java/co/elastic/apm/agent/context/LifecycleListener.java
```diff
@@ -47,6 +47,34 @@
...
+     * </p>
+     *
+     * @throws Exception When something goes wrong performing the cleanup.
```
SylvainJuge: cleanup > resume

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -641,4 +751,10 @@ private String getServiceName(@Nullable ClassLoader initiatingClassLoader) {
     public void resetServiceNameOverrides() {
         serviceNameByClassLoader.clear();
     }
+
+    public enum TracerState {
+        STOPPED,
```
SylvainJuge: maybe describe a bit what are agent states and the main transitions between them.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+
+    private void pollStressMonitors() {
+        while (!Thread.currentThread().isInterrupted()) {
+            try {
+                if (circuitBreakerConfiguration.isCircuitBreakerEnabled()) {
+                    if (isCurrentlyUnderStress) {
+                        boolean stressRelieved = true;
+                        for (StressMonitor stressMonitor : stressMonitors) {
```
SylvainJuge: this is a bit hard to read due to lots of conditionals & try/catch, maybe splitting in two smaller methods `isStressRelieved` and `isUnderStress` would make the intent a bit easier to grasp.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
```diff
@@ -0,0 +1,144 @@
...
+                pollStressMonitors();
+            }
+        });
+    }
+
+    private void loadGCStressMonitor(ElasticApmTracer tracer) {
+        try {
+            stressMonitors.add(new GCStressMonitor(tracer));
+        } catch (Throwable throwable) {
+            logger.error("Failed to load the GC stress monitor. Circuit breaker will not be triggered based on GC events.", throwable);
```
SylvainJuge: what kind of failures do we expect here ? shouldn't they be managed at monitor implementation level ?
if there are missing JMX attributes for example, we should rather have a monitor that does never calls pause/resume than an exception thrown from constructor.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
```diff
@@ -0,0 +1,144 @@
...
+                                }
+                            }
+                        }
+                    }
+                } else if (isCurrentlyUnderStress) {
+                    isCurrentlyUnderStress = false;
+                    tracer.stressRelieved();
+                }
+            } catch (Throwable throwable) {
+                logger.error("Error occurred during Circuit Breaker polling", throwable);
```
SylvainJuge: this catch block should never be called, because all potential exceptions from stress monitor are already caught.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreakerConfiguration.java
SylvainJuge: +1 on keeping it active by default, if config options have "disabled values" that make equivalent behavior that will be one less parameter to document and to explain.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreakerConfiguration.java
eyalkoren: The main reason I want to expose it is the CPU monitor I am about to add- I think it is a delicate monitor and I certainly don't want to make a decision to pause the agent because CPU usage was over some threshold for 5 seconds or so. I am thinking of adding a CPU-monitor-config option that we define how many consecutive measurements should be considered before making a decision. I think having both together makes sense.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreakerConfiguration.java
eyalkoren: We are thinking of making it on by default in the near future, in which case, in order to turn it off, users will have to change multiple configurations if we do that. 
I am thinking about someone getting sporadic pauses with their agents because of some specific JVM or other environmental setups, being required to do that. Moreover, I made sure it is dynamic as well so users can turn off while it's running, without being required to figure out which monitor settings need to be changed.
I already documented and explained, so currently the additional work is to remove.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
eyalkoren: The stress is considered relieved by looking on a separate threshold, so this should be addressed.
Still, deciding **that there is stress** based on CPU usage will probably make more sense if based on n consecutive iterations...

apm-agent-core/src/main/java/co/elastic/apm/agent/context/LifecycleListener.java
eyalkoren: > because we document how users should use this method.

Not really... It's a listener interface so we document how users should implement this method. The tracer is using it in a single place. This is basically encourage implementors to throw informative Exceptions.

apm-agent-core/src/main/java/co/elastic/apm/agent/context/LifecycleListener.java
eyalkoren: Same as above, reworded the `cleanup` to match resume, thanks for noticing!

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
eyalkoren: > what kind of failures do we expect here ? 

Not expecting, just watching out from

> shouldn't they be managed at monitor implementation level ?

Yes, unless the class itself fails to load, eg because of missing dependencies. Since it is JMX, I was cautious even though we ended up using a generic API. I can see why it seems overly cautious, but if there's no reason to remove, let's leave it.

> if there are missing JMX attributes for example, we should rather have a monitor that does never calls pause/resume than an exception thrown from constructor.

I thought more about failure to load the class, rather than an exception in the constructor. This guard is not about the concern of pausing/resuming, it is in order to make sure we do not fail to register the next monitor if this one cannot be loaded.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
eyalkoren: All potential `Exception`s but not `Throwable`s. 
If, by accident, we get an `OutOfMemoryError` in our monitor, this is probably the only agent functionality I want to keep alive 🙂 .
This is the thread's main loop, it makes sense to not let the thread die from unnatural causes.

apm-agent-plugins/apm-servlet-plugin/src/test/java/co/elastic/apm/agent/servlet/ApmFilterTest.java
eyalkoren: Add to be done `@AfterEach`

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
eyalkoren: I did it through the GC monitor, but now added it to the API so it can be done centrally here

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
SylvainJuge: Ok then probably makes sense to at least add a comment why we need to catch everything here.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
```diff
@@ -108,6 +115,41 @@ private void pollStressMonitors() {
         }
     }
 
+    private void checkIfUnderStress() {
```
felixbarny: It seems a bit strange that a method named `check*` modifies state. Could we just return a boolean here and handle based on that in `pollStressMonitors`?

```
if (isUnderStress()) {
    isCurrentlyUnderStress = true;
    tracer.stressDetected();
}
```

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
```diff
@@ -108,6 +115,41 @@ private void pollStressMonitors() {
...
+                if (logger.isDebugEnabled()) {
+                    logger.debug("Failed to poll " + stressMonitor.getClass().getName(), e);
+                }
+            }
+        }
+        if (stressRelieved) {
+            logger.info("All registered stress monitors indicate that the stress has been relieved");
+            isCurrentlyUnderStress = false;
+            tracer.stressRelieved();
+        }
```
felixbarny: Similar story: return boolean and extract this part to the caller.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreakerConfiguration.java
```diff
@@ -51,17 +53,99 @@
...
+        .tags("performance")
+        .description("The threshold used by the GC monitor to rely on for identifying when the heap is not under stress .\n" +
+            "If `stress_monitor_gc_stress_threshold` has been crossed, the agent will consider it a heap-stress state. \n" +
+            "In order to determine that the stress state is over, percentage of occupied memory in ALL heap pools should \n" +
+            "be lower than this threshold. The GC monitor relies only on memory consumption measured after a recent GC.")
+        .dynamic(true)
+        .addValidator(isInRange(0d, 1d))
+        .buildWithDefault(0.75);
+
+    private final ConfigurationOption<Integer> cpuConsecutiveMeasurements = ConfigurationOption.integerOption()
```
felixbarny: Can we make this a `TimeDuration` option which calculates the number of measurements based on `stress_monitoring_interval`?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreaker.java
eyalkoren: Yes we can

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreakerConfiguration.java
eyalkoren: Yes we can!

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -560,11 +584,84 @@ public void registerSpanListener(ActivationListener activationListener) {
...
         }
+        if (activeConfigOptionSet) {
+            tracerState = TracerState.RUNNING;
+            logger.info("Tracer switched to RUNNING state");
+        } else {
+            pause();
+        }
+    }
+
+    public synchronized void stressDetected() {
```
SylvainJuge: Naming, maybe `onStressDetected` would be more explicit, same for `onStressRelieved` below.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -560,11 +584,84 @@ public void registerSpanListener(ActivationListener activationListener) {
...
+        //      if current state is PAUSED and currentlyUnderStress==false - then resume
+        if (wasActive && !shouldBeActive && tracerState == TracerState.RUNNING) {
+            pause();
+        } else if (!wasActive && shouldBeActive && tracerState == TracerState.PAUSED && !currentlyUnderStress) {
+            resume();
+        }
+        activeConfigOptionSet = shouldBeActive;
+    }
+
+    synchronized void pause() {
```
SylvainJuge: why do we have `pause` and `resume` as package-private ? if only for testing, doing it through `stressDetected` and `stressRelieved` methods above should be enough.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -119,7 +120,11 @@ protected Boolean initialValue() {
     private final MetricRegistry metricRegistry;
     private Sampler sampler;
     boolean assertionsEnabled = false;
-    private static final WeakConcurrentMap<ClassLoader, String> serviceNameByClassLoader = new WeakConcurrentMap.WithInlinedExpunction<>();
+    private long lastSpanMaxWarningTimestamp;
+
+    private volatile TracerState tracerState = TracerState.STOPPED;
```
SylvainJuge: Maybe add a small comment to explain why those variables are `volatile` and which object is used to synchronize access to them.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -646,4 +743,26 @@ private String getServiceName(@Nullable ClassLoader initiatingClassLoader) {
...
     }
+
+    /**
+     * An enumeration used to represent the current tracer state.
+     */
+    public enum TracerState {
+        /**
+         * Agent is in this state either before starting, or after being stopped.
+         */
+        STOPPED,
```
SylvainJuge: When agent is in `stopped` state after being stopped, can we transition to any other state ?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreakerConfiguration.java
```diff
@@ -0,0 +1,151 @@
...
+        .key("stress_monitor_cpu_num_measurements")
+        .configurationCategory(CIRCUIT_BREAKER_CATEGORY)
+        .tags("added[1.14.0]")
+        .tags("performance")
+        .description("The number of measurements required in order to determine whether the system is \n" +
```
SylvainJuge: Is that a number of consecutive measurements ? if yes (I guess it's the case) might be relevant to make it explicit here.

apm-agent-core/src/main/java/co/elastic/apm/agent/util/JmxUtils.java
```diff
@@ -0,0 +1,95 @@
...
+        "com.sun.management.OperatingSystemMXBean" // HotSpot
+    );
+
+    @Nullable
+    private static Class<?> operatingSystemBeanClass;
+
+    @Nullable
+    public synchronized static Method getOperatingSystemMBeanMethod(OperatingSystemMXBean operatingSystemBean, String methodName) {
+        if (!initialized) {
+            // lazy initialization - try loading the classes as late as possible
```
SylvainJuge: why do we need to do that ? can you elaborate a bit when this is required and how it helps ?
For example, do we need to wait another part of the application (like an application server startup) to execute before trying to use this code ?

apm-agent-core/src/test/java/co/elastic/apm/agent/AbstractServletTest.java
```diff
@@ -41,15 +49,30 @@
...
+
     @Nullable
     private static Server server;
     protected OkHttpClient httpClient;
 
+    @BeforeAll
+    static void setup() {
+        reporter = new MockReporter();
+        config = SpyConfiguration.createSpyConfig();
+        tracer = new ElasticApmTracerBuilder()
```
SylvainJuge: this will conflict with the changes I did in the JDBC PR, thus we need to not forget about this to be sure the right thing is merged afterwards.

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/ProfilingFactory.java
```diff
@@ -58,6 +58,16 @@ public void start(ElasticApmTracer tracer) {
         tracer.registerSpanListener(new ProfilingActivationListener(tracer, profiler));
     }
 
+    @Override
+    public void pause() throws Exception {
+        // todo - implement
```
SylvainJuge: why we have todos left here ? should we implement them or leave an empty impl. ?

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/SamplingProfiler.java
```diff
@@ -546,6 +546,16 @@ public void start(ElasticApmTracer tracer) {
...
     }
 
+    @Override
+    public void pause() throws Exception {
+        // todo - implement
+    }
+
+    @Override
+    public void resume() throws Exception {
+        // todo - implement
```
SylvainJuge: same as previous.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
eyalkoren: This is an (internal) tracer API, so I think package private is a good choice.
I use it widely in tests, which by itself qualifies to make it package private. 
Testing what happens when we pause the agent in the Quartz test through `stressDetected` is a much worse option in my opinion.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
eyalkoren: Currently no, there is no such option. This is actually something that bothered me when I introduced the state (the fact that there is no single definition to the `STOPPED` state).
I will add an `UNINITIALIZED` state to distinguish and will add proper commenting.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/circuitbreaker/CircuitBreakerConfiguration.java
eyalkoren: I am changing this one based on https://github.com/elastic/apm-agent-java/pull/1040#discussion_r386428151

apm-agent-core/src/main/java/co/elastic/apm/agent/util/JmxUtils.java
eyalkoren: > why do we need to do that ?

Because it is always a good idea to look for a class in the classpath as late as possible, as I elaborated. Otherwise this check may be premature.

> For example, do we need to wait another part of the application (like an application server startup) to execute before trying to use this code ?

Maybe 🙂 . I rather not wait for someone reporting on such scenario. Unless there is a good reason to avoid, safer code seems like the better option.

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/ProfilingFactory.java
eyalkoren: Yes, I left that to @felixbarny with a comment at the top, I hope it was not missed...

PR #1044, opened 2020-02-18
[docs] Add EOL table

bmorelli25:

This PR defines end of life and end of maintenance dates for the Java Agent. EOL dates were determined by going through old releases, finding the GA release date of the major Java Agent versions, and adding 18 months.

## Why is it important?

We love all our products, but sometimes we must say goodbye to a release so that we can continue moving forward on future development and innovation.

## What does EOL/Maintenance mean?

> When a version goes out of maintenance, it is no longer guaranteed to get bug fixes. Between the end of maintenance and EOL, bug fixes will be discretionary, e.g. critical security issues. Once it's EOL, it's no longer given any support attention; issues related to it will be closed with WONTFIX.

## Related issues

For elastic/apm#173.

## Screenshots

Screen Shot 2020-02-18 at 12 33 10 PM


bmorelli25:
Question. Does it make more sense to wait to define EOL for a major until after a new major has been released? The release cadence of majors for a lot of our agents is much slower than the stack... Which means EOL for 1.x could hit before 2.0 is released.

felixbarny:
> Question. Does it make more sense to wait to define EOL for a major until after a new major has been released?

Yes, I think so. Currently, we don't have plans for a v2.0 so I wouldn't know which date to pick. As much as possible we try to keep the current version backwards compatible so that we don't have to maintain two versions and have to start backporting.

bmorelli25:
My apologies for the confusion. I neglected to put the minor versions on the table which messed up the dates pretty bad. I've updated the EOL table, and the dates make a lot more sense now.

docs/upgrading.asciidoc
```diff
@@ -0,0 +1,38 @@
...
+forward on future development and innovation.
+Our https://www.elastic.co/support/eol[End of life policy] defines how long a given release is considered supported,
+as well as how long a release is considered still in active development or maintenance.
+The table below is a simplified description of this policy.
+
+[options="header"]
+|====
+|Agent version |EOL Date |Maintained until
+|1.13.x |2021-08-11 |1.14
+|1.12.x |2021-05-21 |1.13
```
felixbarny: I find this even more confusing now 😄 
EOL date is in the future (2021-05-21), yet we don't support it anymore because there's a new minor (1.13.0)

I think this table will only be really useful once we release a new Major version. In that case we can document how long the latest minor of the previous major is supported.

bmorelli25: I agree that this will only really be useful once a new major is released. As an example: https://github.com/elastic/apm-agent-nodejs/pull/1637/files#diff-bd9ef2ce28d1c05f980ae9ce34024b5bR38

 > EOL date is in the future (2021-05-21), yet we don't support it anymore because there's a new minor (1.13.0)

The [stack EOL table](https://www.elastic.co/support/eol) does the same thing:
7.3.x | 2021-01-31 | 7.4.0
7.4.x | 2021-04-01 | 7.5.0
7.5.x | 2021-06-02 | 7.6.0
7.6.x | 2021-08-11 | 7.7.0

Now I'm wondering if it just makes more sense to remove EOL dates altogether? Don't we really _only_ need a maintained until column?

felixbarny: > I agree that this will only really be useful once a new major is released.

As we're currently trying to make creating releases more seamless I'd really appreciate it if we could introduce that table lazily, once we create a new major version. It makes sense to add the description right now of course.

bmorelli25: Sorry Felix, I didn't realize I never followed up on this. Updated to remove the EOL table for now.

docs/upgrading.asciidoc
```diff
@@ -0,0 +1,18 @@
+[[upgrading]]
+== Upgrading
+Upgrades between minor versions of the agent, like from 1.1 to 1.2 are always backwards compatible.
+Upgrades that involve a major version bump often come with some backwards incompatible changes.
+
+Before upgrading the agent, be sure to review the:
+
+* <<release-notes,Agent release notes>>
+* {apm-overview-ref-v}/agent-server-compatibility.html[Agent and Server compatibility chart]
```
SylvainJuge: regarding updating the agent, since the latest agent version is the recommended one:
what about providing a bash one-liner to download the latest version from maven central ?

Also, it might be relevant to add that the recommended way to update (at least for now) is the following:
- shut-down the application
- update agent jar
- (optionally) change JVM settings if path to agent jar has changed due to a different filename for example
- re-start the application.

One extra thing that we could add maybe later is to have a shell one-liner script to download latest version of the agent for easier update/install.

felixbarny: Could be done like this: https://stackoverflow.com/a/26288401/1125055

bmorelli25: Thanks Sylvain. I've added in the upgrade steps. I'm having trouble adapting the one-liner that Felix provided to the APM jar

Issue #818, opened 2019-08-26
Meta  - Add Scala support


lreuven:
**Is your feature request related to a problem? Please describe.**

per customer request, I'm adding this as a FR.

**Describe the solution you'd like**
Scala Instrumentation as part of the java agent 

**Describe alternatives you've considered**

**Additional context**

felixbarny:
Which frameworks are the most important ones? I assume Akka and Play?

lreuven:
yap but I would like to have more customer feedback on that

PerWiklander:
I would add Lagom to that list, but since Lagom uses Akka and Play, and Play uses Akka, the answer might just be "Akka".

I guess it comes down which scope is feasible. Should the agent report on http requests only (Akka Http then) or general Akka actor interaction as well? And in case of the latter, is there a priority of what parts of Akka to add more details to? One that comes to mind is the PersistentActor, aka. Event Sourcing or the C in CQRS.

My personal wish list, working on a Lagom project would be:
- HTTP requests (Akka Http)
- DB (read) queries
  For me this should/could be covered by the existing JDBC support (I use Slick with Postgres), but the recommended/default db layer in Lagom projects is Cassandra.
- Commands -> Events -> DB (write) queries.
  This includes Akka Persistence and Persistence Query plus (for me) JDBC.

Existing solutions (?)
- https://github.com/levkhomich/akka-tracing
  Looks promising
- https://developer.lightbend.com/docs/telemetry/current/home.html
  The cheat sheet, Lightbend's own monitoring solution, but requires a Lightbend subscription ($$$).

cbrunnkvist:
Documenting some good practices and a showing proof-of-concept code for some "typical" bare bones monolith (maybe it's (Akka+)**Play** with **Slick**/**HikariCP**) would be good enough to allow people to get started in the right direction at least.

crocker:
Twitter's Finatra and Finagle frameworks would be good additions as well.

lssilva:
Finagle, I think with finagle a few others will be supported by default, Finatra, Finch

PR #1048, opened 2020-02-22
Add Scala concurrent plugin

milanvdm:
## What does this PR do?

This PR is a draft to add specific Scala support to the ElasticAPM agent.
The goal is to add basic support for Scala Futures.

## Known issues
* Test multiple Scala version (2.12.x and 2.13.x) -> cross-compiling multiple Scala versions doesn't seem to work/be possible with the Maven plugin
* LinkageError when the agent has a specific Scala dependency
* Not all edge-cases are caught in the tests, some of them were only found when running a small test-setup

## Checklist
- [ ] Added an instrumentation plugin? How did you make sure that old, non-supported versions are not instrumented by accident?

## Related issues
Part of https://github.com/elastic/apm-agent-java/issues/818


apm-agent-plugins/apm-scala-concurrent-plugin/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+    <name>${project.groupId}:${project.artifactId}</name>
+
+    <properties>
+        <apm-agent-parent.base.dir>${project.basedir}/../..</apm-agent-parent.base.dir>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>apm-agent-api</artifactId>
```
SylvainJuge: agent plugins do not need to rely on `apm-agent-api` and should use internal API instead, dependency is provided implicitly by `apm-agent-plugins` parent module.

milanvdm:
@SylvainJuge As I just started exploring the APM, I wanted to start with a test that just checks if it is matching one of the methods I wrote a `MethodMatcher` for.
Any tips/example on how to write a test for that?

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/scala/co/elastic/apm/agent/concurrent/FutureInstrumentation.scala
```diff
@@ -0,0 +1,41 @@
...
+class FutureInstrumentation extends ElasticApmInstrumentation {
+
+  override def getTypeMatcher: ElementMatcher[_ >: TypeDescription] =
+    hasSuperType[TypeDescription](named("scala.concurrent.Future"))
+      .or(hasSuperType(named("scala.concurrent.impl.Promise")))
+      .or(hasSuperType(named("scala.concurrent.impl.Promise$Transformation")))
+      .or(hasSuperType(named("scala.concurrent.Future$")))
+
+  override def getMethodMatcher: ElementMatcher[_ >: MethodDescription] =
+    named[MethodDescription]("onComplete").and(returns(classOf[Unit])).and(takesArguments(classOf[Try[_] => _]))
```
felixbarny, line 24: The type matchers should not refer to classes as that would cause loading these classes. Instead, use the `named` matcher, like `named("scala.util.Try")`

apm-agent-plugins/apm-scala-concurrent-plugin/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+                <groupId>net.alchim31.maven</groupId>
+                <artifactId>scala-maven-plugin</artifactId>
+                <version>4.3.0</version>
+                <configuration>
+                    <scalaVersion>2.12.10</scalaVersion>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
```
felixbarny, line 32: The executor service instrumentations can only be applied in the tests if you have a dependency on that module:
```suggestion
    <dependencies>
        <dependency>
            <groupId>${project.groupId}</groupId>
            <artifactId>apm-java-concurrent-plugin</artifactId>
            <version>${project.version}</version>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>${project.groupId}</groupId>
            <artifactId>apm-java-concurrent-plugin</artifactId>
            <version>${project.version}</version>
            <type>test-jar</type>
            <scope>test</scope>
        </dependency>
    </dependencies>
```

Add this snippet to the concurrent plugin in order to be able to access its test classes, in particular `WrappedExecutorService` (see also comment further below)>
```xml
    <build>
        <plugins>
            <plugin>
                <artifactId>maven-jar-plugin</artifactId>
                <executions>
                    <execution>
                        <goals>
                            <goal>test-jar</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>
```

apm-agent-plugins/apm-scala-concurrent-plugin/src/test/scala/co/elastic/apm/agent/concurrent/FutureInstrumentationSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+    private def blockingMethodOnMainThread(): Unit = {
+      try {
+        Await.result(nonBlockingMethodOnMainThread(), 10.seconds)
+      } catch {
+        case e: Exception => e.printStackTrace()
+      }
+    }
+
+    private def nonBlockingMethodOnMainThread(): Future[Unit] =
+      Future(methodOnWorkerThread())(ExecutionContext.fromExecutor(Executors.newFixedThreadPool(1)))
```
felixbarny, line 50: Maybe here's something going wrong with the instrumentation. One limitation of the unit tests is that you can't test instrumenting classes from the bootstrap class loader. `ThreadPoolExecutor`, which is returned from `Executors.newFixedThreadPool(1)` is such a class.

To work around this, we use `co.elastic.apm.agent.concurrent.ExecutorServiceWrapper` in our tests. This is a class that the agent can instrument in tests that just delegates to the wrapped `ExecutorService`.

For the built-in executor service instrumentation to work, you also have to declare a test dependency to the java-concurrent-plugin (see comment on pom.xml).

To access the `ExecutorServiceWrapper` class, you either need to copy/paste it to your module or build a test jar from the java-concurrent-plugin and add it as a test dependency as well (see comment on pom.xml).

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/scala/co/elastic/apm/agent/concurrent/FutureInstrumentation.scala
```diff
@@ -0,0 +1,41 @@
...
+//    .or(named[MethodDescription]("transform").and(returns(classOf[Future[_]])).and(takesArguments(classOf[Try[_] => Try[_]])))
+//    .or(named[MethodDescription]("transformWith").and(returns(classOf[Future[_]])).and(takesArguments(classOf[Try[_] => Future[_]])))
+    .and(not(isTypeInitializer[MethodDescription]()))
+
+  override def getInstrumentationGroupNames: util.Collection[String] = List("concurrent", "future").asJavaCollection
+
+  @Advice.OnMethodEnter(suppress = classOf[Throwable])
+  def onComplete(@Advice.Argument(value = 0, readOnly = true) callback: Try[_] => _): Unit = {
```
felixbarny, line 32: What are you trying to achieve by instrumenting the `onComplete` method? 

Is the `onComplete` executed in the context of the `Runnable` from the `Executor` that is used within the `ExecutionContext.fromExecutor`?

milanvdm: What should happen is when a `Future` is created, it should have the trace information. When a `Future` finishes its task, the method `onComplete` will be called on that same `Future`. Based on the trace information on the creation, it should then close the correct spans.

milanvdm: Another remark on using Java to write the Scala instrumentation. For example, this method would need to be replaced with some custom code to correctly close tracing contexts.
When writing all of this in Java, I don't immediately see a way to write the correct Scala code to replace the original method?

apm-agent-plugins/apm-scala-concurrent-plugin/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+<!--            <version>${project.version}</version>-->
+<!--            <scope>test</scope>-->
+<!--        </dependency>-->
+<!--        <dependency>-->
+<!--            <groupId>${project.groupId}</groupId>-->
+<!--            <artifactId>apm-java-concurrent-plugin</artifactId>-->
+<!--            <version>${project.version}</version>-->
+<!--            <type>test-jar</type>-->
+<!--            <scope>test</scope>-->
+<!--        </dependency>-->
```
milanvdm, line 33: Currently, disabling this dependency since adding it not only brings in the Wrapper classes, but for some reason also pulls in the JavaInstrumentation and method-matchers from the java-concurrent-plugin.
Comments welcome on how to prevent that :)

apm-agent-plugins/apm-scala-concurrent-plugin/src/test/scala/co/elastic/apm/agent/scala/concurrent/FutureInstrumentationSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+  private var coreConfiguration: CoreConfiguration = _
+  private var transaction: Transaction = _
+
+  override def beforeEach: Unit = {
+    reporter = new MockReporter
+    val config: ConfigurationRegistry = SpyConfiguration.createSpyConfig
+    coreConfiguration = config.getConfig(classOf[CoreConfiguration])
+    tracer = new ElasticApmTracerBuilder().configurationRegistry(config).reporter(reporter).build
+    ElasticApmAgent.initInstrumentation(tracer, ByteBuddyAgent.install)
+    transaction = tracer.startRootTransaction(null).withName("Transaction").activate()
```
milanvdm, line 35: I'm unsure on how to start the initial transaction for every test. Since the FutureInstrumentation doesn't create any transactions, I expect that I manually need to activate it here.
Although, I cannot find a similar thing in the `AsyncTraceMethodInstrumentationTest` of the java-concurrent plugin.

apm-agent-plugins/apm-scala-concurrent-plugin/src/test/scala/co/elastic/apm/agent/scala/concurrent/FutureInstrumentationSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  override def afterEach: Unit = {
+    transaction.deactivate().end()
+    ElasticApmAgent.reset()
+  }
+
+  "test" should "test" in {
+    new TestFutureTraceMethods().invokeAsync()
```
milanvdm, line 44: I see the FutureInstrumentation match correctly on this method-call which I find good progress :)

Im confused on why the test is not passing due to not having any transactions in the reporter although I manually create that root-transaction.
Any reason why it is not getting this root-transaction that gets created before every test?

felixbarny: That is because you are ending the transaction after the test has run (`afterEach` runs after `"test"`). Transactions are reported only after they are ended.

apm-agent-plugins/apm-scala-concurrent-plugin/pom.xml
milanvdm: Shouldn't the unit test only test the Scala instrumentation? If now the JavaConcurrent instrumentation also starts running, it makes it more difficult to for example check the debug logs of the tests since it gets 'polluted' with the JavaConcurrent instrumentation.

felixbarny: I was under the impression that you were trying to leverage the executor instrumentation from the java concurrent plugin?

milanvdm: Still unsure about it to be honest.
My question was mainly pointed towards you and the expectations around how plugins get tested and if there are any restrictions on how to test it properly.

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
```diff
@@ -0,0 +1,78 @@
...
+    @Nonnull
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Arrays.asList("concurrent", "future");
+    }
+
+    @VisibleForAdvice
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter() {
+        System.out.println("DEBUG2");
```
milanvdm, line 68: Im currently trying out to create a first real test to see how everything works.
Im currently stuck on why ByteBuddy is not executing this `onEnter` function even though I see method-matches on DEBUG level:
```
[ScalaTest-main-running-FutureInstrumentationSpec] DEBUG co.elastic.apm.agent.bci.ElasticApmAgent - Method match for instrumentation FutureInstrumentation: ((((name(equals(onComplete)) and returns(erasure(is(void)))) or (name(equals(transform)) and returns(erasure(name(equals(scala.concurrent.Future)))))) or (name(equals(transformWith)) and returns(erasure(name(equals(scala.concurrent.Future)))))) or name(equals(result))) matches public final void scala.concurrent.impl.Promise$DefaultPromise.onComplete(scala.Function1,scala.concurrent.ExecutionContext)
[ScalaTest-main-running-FutureInstrumentationSpec] DEBUG co.elastic.apm.agent.bci.ElasticApmAgent - Method match for instrumentation FutureInstrumentation: ((((name(equals(onComplete)) and returns(erasure(is(void)))) or (name(equals(transform)) and returns(erasure(name(equals(scala.concurrent.Future)))))) or (name(equals(transformWith)) and returns(erasure(name(equals(scala.concurrent.Future)))))) or name(equals(result))) matches public final java.lang.Object scala.concurrent.impl.Promise$DefaultPromise.result(scala.concurrent.duration.Duration,scala.concurrent.CanAwait) throws java.lang.Exception
```

I see that the log also mentions `throws java.lang.Exception` but it is unclear on how and when that Exception is thrown.
@felixbarny Any idea on what I'm missing on why `onEnter` is not being called even though there is a method match?

felixbarny: Have you tried setting a breakpoint at `Promise$DefaultPromise.onComplete` or a method breakpoint for `Promise.onComplete` to see if the methods get called at all?

> I see that the log also mentions throws java.lang.Exception but it is unclear on how and when that Exception is thrown.

This is just a part of the method signature to indicate this method may throw any `java.lang.Exception`.

milanvdm: Thanks for the tip, this helped me figure out the underlying process a lot better :)

Struggling with ByteBuddy now to figure out how I could add a transaction-id to a matched class as a field.
From my current understanding, a `Promise` will be created and at some point, the `run` method will be called on that `Promise`. Since the call on that `run` could happen on a different thread, I think it would make sense to somehow store the original context on that `Promise` and make sure to activate it again before the `run`.

I've tried working with `@Advice.FieldValue` and `@Advice.Local` but I'm not sure if those values are really kept on the `Promise` itself.

felixbarny: Our agent supports to be attached at runtime. One of the limitations when instrumenting classes that are already loaded is that you can't change the class file format. That means you can't add interfaces, methods or fields.

A common technique we use to overcome this is to put things like a `Promise` as the key of a `WeakConcurrentMap` and the corresponding `Span` as the value.

Example:
https://github.com/elastic/apm-agent-java/blob/842185664456651f1d6a99602d21a7c052ddf111/apm-agent-plugins/apm-redis-plugin/apm-lettuce-plugin/src/main/java/co/elastic/apm/agent/redis/lettuce/Lettuce5StartSpanInstrumentation.java#L57

milanvdm: What if it is a Scala `Promise`? I remember a comment that we should prevent loading Scala classes since they would not be available on all code-bases.
Especially, I'm not sure if Java has any Scala interop.

felixbarny: Defining a `WeakConcurrentMap<Promise, Span>` does not cause loading/linking the `Promise`.  That's because the javac compiler erases the generic arguments. We only put promises in at runtime as we encounter these promises so they are guaranteed to already have been loaded.

pom.xml
```diff
@@ -420,13 +420,8 @@
                 </plugin>
                 <plugin>
                     <artifactId>maven-surefire-plugin</artifactId>
-                    <version>2.19.1</version>
+                    <version>2.22.2</version>
```
milanvdm, line 423: Had to upgrade versions to be able to work with the Scala Munit test-framework.

pom.xml
```diff
@@ -438,6 +433,11 @@
                             <version>${version.junit-jupiter}</version>
                         </dependency>
                     </dependencies>
+                    <configuration>
+                        <includes>
+                            <include>**/*Spec.*</include>
```
milanvdm, line 438: Needed to match Scala `*Spec.scala` test files

milanvdm:
The tests I created are passing now. But it is hard for me to judge to what extent these tests are a good representation of reality and testing edge-cases.

The project I created to test a lot of tooling together would also be a good test run: https://github.com/milanvdm/scala-elastic-apm
For this Ill need to release the java-agent locally, not sure how to do this though.

felixbarny:
> For this Ill need to release the java-agent locally, not sure how to do this though.

Here are some instructions around that: https://github.com/elastic/apm-agent-java#build-form-source

milanvdm:
Thanks! I managed to release it locally and use it in my test project as well.

Without the specific scala-concurrent instrumentation, it was not able to trace all the different parts of the program. From some talks with the maintainer of Kamon, this seems to be related to how the scala execution-context handles an internal queue for scheduling.

With the newly added scala-concurrent instrumentation, it was able to trace all 20 http-calls and 20 database requests, all done in an async way: https://github.com/milanvdm/scala-elastic-apm/blob/master/src/main/scala/me/milan/MainF.scala

image

Not sure what the next steps would be to get this PR merged :)

felixbarny:
That's great progress!

> Not sure what the next steps would be to get this PR merged :)
I'll have a more in-depth review later but for starters, we'd have to revert the source and target level of the compiler plugin from 9 down to 7 again. That's because the agent still supports Java 7 and we don't want to change this with this PR. It's fine that the tests are compiled with Java 9, we have already configured it that way.

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
```diff
@@ -0,0 +1,101 @@
...
+        public static void onEnter(@Advice.This Promise<?> thiz) {
+            final TraceContextHolder<?> context = promisesToContext.getIfPresent(thiz);
+            if (tracer != null && context != null) {
+                tracer.activate(context);
+            }
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class)
+        public static void onExit(@Advice.This Promise<?> thiz) {
+            promisesToContext.remove(thiz);
```
felixbarny, line 96: Don't forget to deactivate the context.

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
```diff
@@ -0,0 +1,101 @@
...
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        public static void onEnter(@Advice.This Promise<?> thiz) {
+            final TraceContextHolder<?> context = promisesToContext.getIfPresent(thiz);
+            if (tracer != null && context != null) {
+                tracer.activate(context);
+            }
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class)
+        public static void onExit(@Advice.This Promise<?> thiz) {
```
felixbarny, line 95: Catch thrown exceptions `@Thrown @Nullable Throwable t` and `captureException(t)`

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return hasSuperType(named("scala.concurrent.impl.Promise$Transformation"));
+    }
+
+    public static class ConstructorInstrumentation extends FutureInstrumentation {
+
+        @Override
+        public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+            return isConstructor();
```
felixbarny, line 65: Promises could be created outside a transaction but used in the context of a transaction at the point a promise is scheduled. 
Is there a canonical place where promises are scheduled? Maybe something equivalent to `Executor#execute(Runnable)`?

apm-agent-core/pom.xml
```diff
@@ -165,6 +165,14 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>9</source>
```
felixbarny, line 172: Why is Java 9 required here? The agent must still support Java 7.  Loading the plugin (instantiating `FutureInstrumentation`) on an earlier JVM would result in an error.

milanvdm: I'll have to clean up the PR a bit later on, some parts were needed to get it running on my laptop. Otherwise, it was complaining about the java versions.

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
milanvdm: Mmmm, Ill have a look. 

Do you mean in the context of, someone creates a `Promise` outside of a transaction, without evaluating it. And then only start evaluating it later on, inside of a transaction?

felixbarny: Yes, exactly.

This would also allow us to do another nice thing. Currenly, all classes have to be evaluated whether they are a subclass of `Promise` (`hasSuperType(named("scala.concurrent.impl.Promise$Transformation"))`). This is a rather costly matcher as we can't exclude any class by name so all classes that are loaded have to be parsed.

See this example of the AsyncHttpClient instrumentation how we only instrument `AsyncHandler` classes that are passed to the `org.asynchttpclient.DefaultAsyncHttpClient#executeRequest(org.asynchttpclient.Request, org.asynchttpclient.AsyncHandler<T>)` method.

https://github.com/elastic/apm-agent-java/blob/75160feaf475a467ec347b32a98b24767d70cc52/apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java#L106

Using this strategy, we wouldn't have to instrument any `Promise`s to start with. Only those subclasses that pass the canonical place where they are scheduled would be instrumented at runtime.

milanvdm: But how would you handle the case that there is a thread-switch between the creation of a Promise and the scheduling? 
If the context is not propagated across that thread-switch, you would not know the transaction linked to this Promise and schedule it linked to no transaction.

felixbarny: Usually, the thread on which the promise is scheduled is the thread we want to propagate the context. 
Do you know any counter-examples?

milanvdm: 1. You start a transaction (thread 1)
2. You create a Promise (thread 1) - no context is saved on this Promise now
3. Some thread switching - as I understand, the transaction of thread 1 will not be propagated
4. Promise gets scheduled (thread 2) - it cannot find an active transaction

felixbarny: > Some thread switching 

This thread switching should also be covered by one of our instrumentation. For example, if the switch is done by another promise.

milanvdm: Oke, I'll dig a bit deeper into how Scala executes the Promises.

milanvdm: I've added some instrumentation on Scala's ExecutionContext instead of the Promise directly.
Some parts are working but Im having issues with `Future.sequence` which transforms a list of Future to one Future. 

For this, it uses the following logic:
```scala
/**
   * WARNING: Only ever execute logic which will quickly return control to the caller.
   *
   * This `ExecutionContext` steals execution time from other threads by having its
   * `Runnable`s run on the `Thread` which calls `execute` and then yielding back control
   * to the caller after *all* its `Runnable`s have been executed.
   * Nested invocations of `execute` will be trampolined to prevent uncontrolled stack space growth.
   *
   * When using `parasitic` with abstractions such as `Future` it will in many cases be non-deterministic
   * as to which `Thread` will be executing the logic, as it depends on when/if that `Future` is completed.
   *
   * Do *not* call any blocking code in the `Runnable`s submitted to this `ExecutionContext`
   * as it will prevent progress by other enqueued `Runnable`s and the calling `Thread`.
   * 
   * Symptoms of misuse of this `ExecutionContext` include, but are not limited to, deadlocks
   * and severe performance problems.
   *
   * Any `NonFatal` or `InterruptedException`s will be reported to the `defaultReporter`.
   */
  object parasitic extends ExecutionContextExecutor with BatchingExecutor {
    override final def submitForExecution(runnable: Runnable): Unit = runnable.run()
    override final def execute(runnable: Runnable): Unit = submitSyncBatched(runnable)
    override final def reportFailure(t: Throwable): Unit = defaultReporter(t)
  }
```

During debugging, I see the `SpanInScopeRunnableWrapper` be correctly submitted to this parasitic executor. But after the execution of the `Future.sequence`, the context seems to be lost.

milanvdm: Not sure if relevant, but in DEBUG mode, I see the following:
```
2020-04-24 20:14:59.428 [ForkJoinPool-1-worker-19] DEBUG co.elastic.apm.agent.mdc.MdcActivationListener - Class loader jdk.internal.loader.ClassLoaders$AppClassLoader@512ddf17 cannot load log4j API
java.lang.ClassNotFoundException: org.apache.log4j.MDC
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
	at co.elastic.apm.agent.mdc.MdcActivationListener$2.get(MdcActivationListener.java:76)
	at co.elastic.apm.agent.mdc.MdcActivationListener$2.get(MdcActivationListener.java:70)
	at co.elastic.apm.agent.cache.WeakKeySoftValueLoadingCache$CacheValue.<init>(WeakKeySoftValueLoadingCache.java:124)
	at co.elastic.apm.agent.cache.WeakKeySoftValueLoadingCache$CacheValue.<init>(WeakKeySoftValueLoadingCache.java:116)
	at co.elastic.apm.agent.cache.WeakKeySoftValueLoadingCache.get(WeakKeySoftValueLoadingCache.java:81)
	at co.elastic.apm.agent.mdc.MdcActivationListener.beforeActivate(MdcActivationListener.java:154)
	at co.elastic.apm.agent.impl.transaction.TraceContextHolder.activate(TraceContextHolder.java:101)
	at co.elastic.apm.agent.impl.transaction.AbstractSpan.activate(AbstractSpan.java:345)
	at co.elastic.apm.agent.impl.transaction.AbstractSpan.activate(AbstractSpan.java:42)
	at co.elastic.apm.api.AbstractSpanImpl.activate(AbstractSpanImpl.java:137)
	at me.milan.MainF$.$anonfun$program$3(MainF.scala:30)
```

milanvdm: @felixbarny Should I create an issue about the mismatch between traces in the UI and the actual logging of the transaction-id?

felixbarny: > One of the issues that I have is the logging though. With %X{transaction.id:-null} in my logback, I had all transaction ids of every operation when instrumenting the Promises directly.

Hm, that's strange.. Maybe there's another thread switch in-between? Could there be some work-stealing going on? Is it seemingly random when the MDC works and when it doesn't? In the cases where there's no MDC, does `ElasticApm.currentSpan()` (from the `apm-agent-api` module) return a noop span?


The debug message about `ClassNotFoundException: org.apache.log4j.MDC` can be ignored.

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        @Override
+        public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+            return named("run").and(returns(void.class));
+        }
+
+        @VisibleForAdvice
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        public static void onEnter(@Advice.This Promise<?> thiz) {
```
felixbarny, line 143: To properly deactivate on exit with considering wrapped promises, add `@Advice.Local("context") TraceContextHolder<?> context` to the parameters.

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
```diff
@@ -0,0 +1,157 @@
...
+            final TraceContextHolder<?> context = promisesToContext.getIfPresent(thiz);
+            if (tracer != null && context != null) {
+                tracer.activate(context);
+            }
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class)
+        public static void onExit(@Advice.This Promise<?> thiz) {
```
felixbarny, line 151:
```suggestion
        public static void onExit(@Advice.This Promise<?> thiz, @Nullable @Advice.Local("context") TraceContextHolder<?> context) {
```

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
```diff
@@ -0,0 +1,157 @@
...
+            promisesToContext.remove(thiz);
```
felixbarny, line 152:
```suggestion
            if (context != null) {
                promisesToContext.remove(thiz);
                tracer.deactivate(context);
            }
```

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
felixbarny: Maybe the non-deactivated context has something to do with the MDC issue.

milanvdm: Im not using the matchers on the Promise directly anymore. I'm only using matchers on the ExecutionContext directly using the wrapped Runnables:
```
co.elastic.apm.agent.scala.concurrent.FutureInstrumentation$ExecutionContextInstrumentation
co.elastic.apm.agent.scala.concurrent.FutureInstrumentation$BatchedExecutionContextInstrumentation
#co.elastic.apm.agent.scala.concurrent.FutureInstrumentation$ConstructorInstrumentation
#co.elastic.apm.agent.scala.concurrent.FutureInstrumentation$RunInstrumentation
```

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
milanvdm: At the moment it looks quite random. I'll into into it a bit more to see if I can reproduce it in a better way.
But as far as I understand it, the logging should be part of the submitted `Runnable` and therefore have exactly the same transaction as the action it is logging.

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
felixbarny: ah, got it

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
felixbarny: Yes, that's correct. Whenever a span is activated, it also sets the MDC.

milanvdm: > In the cases where there's no MDC, does ElasticApm.currentSpan() (from the apm-agent-api module) return a noop span?

Would you test this by adding doing something like:
```
println(ElasticApm.currentSpan())
log.info(testing MDC)
```

felixbarny: Yes, exactly

milanvdm: @felixbarny The `println` of `ElasticApm.currentSpan().getId()` returns an empty String so according to the docs it is indeed a noop.

I'll need to check how the logging is executed in a different thread for whatever reason.

milanvdm: It seems to be related to the usage of the Scala sttp library. It wraps the `AsyncHttpClient` from Java with https://github.com/reactive-streams/reactive-streams-jvm.

milanvdm: @felixbarny So basically what happens is that the Scala future will be completed (scheduled?) the moment `override def onCompleted(): Unit` is called on the `org.asynchttpclient.handler.StreamedAsyncHandler`.

If I add an `ElasticAPM.currentSpan()` in the `onCompleted`, it prints a Noop which shows that no trace-context is available at that moment. And therefore, also not on the Scala Future that gets completed by that method.

I assume because the `AsyncHttpClient` is instrumented on the `executeRequest`, it causes the span of the http request to still be correctly closed. 

But because I continue logging after the completion of the Scala future, the logging lost its context due to the `org.asynchttpclient.handler.StreamedAsyncHandler`.

felixbarny: > So basically what happens is that the Scala future will be completed (scheduled?) the moment override def onCompleted(): Unit is called on the org.asynchttpclient.handler.StreamedAsyncHandler

Thanks for the analysis!

> Unless you see it as a mistake that the trace-context is not passed into the org.asynchttpclient.handler.StreamedAsyncHandler#onCompleted?

Yes, that's a mistake.

This should be easy to fix in the AsyncHttpClient instrumentation:

https://github.com/elastic/apm-agent-java/blob/94c92159d99f57d67d6f2ca0aa16e373082f1684/apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java#L174-L187

We should activate the span in the `@OnMethodEnter` advice and deactivate it in the `@OnMethodExit` advice.

milanvdm: @felixbarny Cool :) Can I try to create a PR to fix the AsyncHttpClient or do you prefer to do it from your side?

Once that is merged, Ill continue with testing this PR and start cleaning it up.

felixbarny: > Can I try to create a PR to fix the AsyncHttpClient or do you prefer to do it from your side?

Sure, go ahead! I think we should do the activation on all currently instrumented `AsyncHandler` callback methods. This way, users can add custom stuff in their own callbacks to the spans like `ElasticApm.currentSpan().addCustom("foo", responseStatus.getStatusText());`

milanvdm: @felixbarny First attempt: https://github.com/elastic/apm-agent-java/pull/1172
Will need some help on how to add proper tests for this though.

milanvdm: @felixbarny I missed a method `onStream` on #1172 so will create a follow PR to add that one as well.
Locally, it seems the span is now correctly passed along the `AsyncHandler` but there is still a place in my test setup where the span is not correctly passed on through all the Scala Futures.

I will investigate further but I am maybe thinking there was a reason why other solutions instrument the `Promise` directly and not the `ExecutionContext`.

felixbarny: Both approaches seem to have their pros and cons. I currently don't have a clear opinion which is the best approach. I'm leaning towards instrument on `ExecutionContext#execute` as it's more efficient (no startup overhead) and supports propagating context for `Promise`s that are created outside of a transaction. But I also see that it requires more work and is potentially more brittle.

milanvdm: I went into the code again and found the behavior described in the accepted answer on https://stackoverflow.com/questions/59642314/scala-future-with-callback-and-context-switching.

What seems to happen is that if a `Future` is already completed, it will be executed immediately. Meaning that `final AbstractSpan<?> active = getActive();` in the `OnExecute` instrumentation, may return a null since the callback is executed on a thread that does not have an active span on it.

Since the `scala.concurrent.impl.Promise.DefaultPromise#dispatchOrAddCallback` is the method that is executing the `Future`, this method somehow needs to be able to track which span is linked to that `Future/Promise`.

I think this forces the choice to instrument on the `Future/Promise` directly instead of with the `ExecutionContext`.

One if the issues with instrumenting on the `Promise` directly, seems to be the extra dependency on `Scala` in the agent which cause some strange `LinkageErrors` sometimes. Is this related to the reason why the Elastic agent needs to shade libraries to avoid certain conflicts?

milanvdm: I added the instrumentation on the Promise back. It all works now for my test-setup.

I sometimes get a warning though:
`09:35:28.230 [ForkJoinPool-5-worker-19] WARN co.elastic.apm.agent.impl.ElasticApmTracer - Tried to start a span with a non-zero reference count 1 'GET postman-echo.com' 00-00e98d3ae2d1d3665c8286557ca405f1-05f18334e53a3ee6-01 (3c295061)`

As well as the LinkageError in some cases:
```
[ERROR] Exception in thread "main" java.lang.LinkageError: loader constraint violation: loader 'bootstrap' wants to load class scala.collection.immutable.LazyList$State$Empty$. A different class with the same name was previously loaded by 'app'. (scala.collection.immutable.LazyList$State$Empty$ is in unnamed module of loader 'app')
```

milanvdm: @felixbarny As discussed, I make the Map use an `Object` instead of a `Promise` but it still get the Linkage issue. Here is the full trace:

```
scala-elastic-apm[ERROR] Exception in thread "main" java.lang.LinkageError: loader constraint violation: loader 'bootstrap' wants to load class scala.collection.immutable.LazyList$State$Empty$. A different class with the same name was previously loaded by 'app'. (scala.collection.immutable.LazyList$State$Empty$ is in unnamed module of loader 'app')
scala-elastic-apm[ERROR] 	at scala.collection.immutable.LazyList$.<clinit>(LazyList.scala:949)
scala-elastic-apm[ERROR] 	at scala.package$.<clinit>(package.scala:89)
scala-elastic-apm 2020-05-15 13:09:22.792 [elastic-apm-init-instrumentation-shutdown-hook] INFO co.elastic.apm.agent.impl.ElasticApmTracer - Tracer switched to STOPPED state
scala-elastic-apm[ERROR] 	at scala.Predef$.<clinit>(Predef.scala:154)
scala-elastic-apm 2020-05-15 13:09:22.792 [elastic-apm-init-instrumentation-shutdown-hook] INFO co.elastic.apm.agent.report.ApmServerReporter - dropped events because of full queue: 0
scala-elastic-apm[ERROR] 	at scala.concurrent.Future$.<clinit>(Future.scala:520)
scala-elastic-apm[ERROR] 	at me.milan.main.future.Main$.delayedEndpoint$me$milan$main$future$Main$1(Main.scala:32)
scala-elastic-apm[ERROR] 	at me.milan.main.future.Main$delayedInit$body.apply(Main.scala:19)
scala-elastic-apm[ERROR] 	at scala.Function0.apply$mcV$sp(Function0.scala:39)
scala-elastic-apm[ERROR] 	at scala.Function0.apply$mcV$sp$(Function0.scala:39)
scala-elastic-apm[ERROR] 	at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:17)
scala-elastic-apm[ERROR] 	at scala.App.$anonfun$main$1(App.scala:73)
scala-elastic-apm[ERROR] 	at scala.App.$anonfun$main$1$adapted(App.scala:73)
scala-elastic-apm[ERROR] 	at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:553)
scala-elastic-apm[ERROR] 	at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:551)
scala-elastic-apm[ERROR] 	at scala.collection.AbstractIterable.foreach(Iterable.scala:920)
scala-elastic-apm[ERROR] 	at scala.App.main(App.scala:73)
scala-elastic-apm[ERROR] 	at scala.App.main$(App.scala:71)
scala-elastic-apm[ERROR] 	at me.milan.main.future.Main$.main(Main.scala:19)
```

felixbarny: Actually, the usage of `Promise` as a type parameter of the map shouldn't be a problem because of type erasure. 

What you could try is to use `Object` instead of `Promise<?>` in the advice method signatures and cast within the method body. Maybe that avoids some eager linking. For example:

```java
        public static void onExit(@Advice.This Object thiz) {
            final AbstractSpan<?> context = getActive();
            if (context != null) {
                promisesToContext.put(thiz, context);
            }
        }
```

milanvdm: Some things:
* When I use `Object` in the Map and method so that I can remove the Scala dependency, there is no LinkageError.
* Using Promise in the Map and in the method parameters Object does still have the LinkageError
* The LinkageError does only happen when forking the JVM (in case of running Debug from IntelliJ or using something like https://github.com/spray/sbt-revolver to start your application)

milanvdm: The LinkageError also happens when Cinnamon is enabled (and without running a forked JVM). So it seems specific to the Scala dependencies clashing somehow.

On the WARN that is happening:
`scala-elastic-apm 2020-05-15 14:08:53.201 [ForkJoinPool-1-worker-23] WARN co.elastic.apm.agent.impl.ElasticApmTracer - Tried to start a span with a non-zero reference count 1 'SELECT' 00-fbed3a699f6c244dffa8de3a3d95e48d-bd4655c8959cd4c8-01 (7b9e3d51)`

Im confused on how this warning can be generated. The method that logs this warning, first creates a new Span, and then counts the references. I don't see how the new Span can suddenly have any references right after being made and before the method actually returns the created Span.

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
```diff
@@ -0,0 +1,109 @@
...
+        @Override
+        public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+            return isConstructor();
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class)
+        public static void onExit(@Advice.This Promise<?> thiz) {
+            final AbstractSpan<?> context = getActive();
+            if (context != null) {
+                promisesToContext.put(thiz, context);
```
felixbarny, line 73:
```suggestion
                promisesToContext.put(thiz, context);
                // this span might be ended before the Promise$Transformation#run method starts
                // we have to avoid that this span gets recycled, even in the above mentioned case
                context.incrementReferences();
```

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
```diff
@@ -0,0 +1,109 @@
...
+        public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+            return named("run").and(returns(void.class));
+        }
+
+        @VisibleForAdvice
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        public static void onEnter(@Advice.This Promise<?> thiz, @Nullable @Advice.Local("context") AbstractSpan<?> context) {
+            context = promisesToContext.remove(thiz);
+            if (tracer != null && context != null) {
+                tracer.activate(context);
```
felixbarny, line 96:
```suggestion
                tracer.activate(context);
                // decrements the reference we incremented to avoid that the parent context gets recycled before the promise is run
                // because we have activated it, we can be sure it doesn't get recycled until we deactivate in the OnMethodExit advice
                context.decrementReferences();
```

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
felixbarny: I've added suggestions that are likely to fix the cause of this.

> I don't see how the new Span can suddenly have any references right after being made and before the method actually returns the created Span.

That's because the span it gets out of the pool is still being referenced.

milanvdm: The WARN is gone indeed :) 

So ElasticAPM uses a pool of Spans that it cycles through? And since the Span is creating in the Promise but only activated later on, it may be that the Span gets recycled before it is actually used?

Incrementing and decrementing makes sure the Span does not get recycled?

felixbarny: Exactly. The pool is basically a queue that is initially empty. When trying to get a span from an empty pool, a new span is created. If the reference count if a span is 0, it gets recycled. Activating a span increments the reference count so under normal circumstances, you don’t have to worry about the ref count. But in those async scenarios, it’s important to keep the span alive if it can be used after it’s ended.

felixbarny: > And since the Span is creating in the Promise but only activated later on, it may be that the Span gets recycled before it is actually used?

Actually, it's not creating any span for the Promise. It just propagates the span that has been active when the promise was created to the thread that executes the promise. But the span might be already ended once the promise is executed.

felixbarny:
Seems like just a small cleanup (for example the Java compiler settings) is needed and we're good to go 🙂

milanvdm:
@felixbarny I'm not so happy with the `Object` approach to not load Scala as a dependency and not cause the `LinkageError`. I'm afraid this will come back in the future when adding more features around Scala.

I'll start with cleaning the PR up but would like to fix this Linkage error before merging this.

felixbarny:
I'm not sure there's much we can do about the LinkageError right now.

> The LinkageError does only happen when forking the JVM (in case of running Debug from IntelliJ or using something like https://github.com/spray/sbt-revolver to start your application)

It seems these tools inspect classes and cause eager loading of classes. I'm not sure there's much we can do about it. You mentioned other tracing tools also have similar problems. So the best, for now, is probably to use `Object` vs `Promise`.
We have plans for a different class loading architecture in the agent. That would not append all classes to the bootstrap classloader but load them from an isolated class loader. This "hides" the agent from the regular class path so that class path scanning tools don't get confused. This may or may not help in this case.

milanvdm:
@felixbarny I'll add some docs and look into testing both Scala 2.12 and 2.13 as well.

Since Scala 2.11 is deprecated, I would also want to make sure that this version doesn't get instrumented.

As long as this gets marked as very experimental, I'm happy :) Will look into this LinkageError after this is merged then.

apm-agent-plugins/apm-scala-concurrent-plugin/src/main/java/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @VisibleForAdvice
+    @SuppressWarnings("WeakerAccess")
+    public static final WeakConcurrentMap<Object, AbstractSpan<?>> promisesToContext =
+        new WeakConcurrentMap.WithInlinedExpunction<>();
+
+    @Nonnull
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Arrays.asList("concurrent", "future");
```
felixbarny, line 53: Marking as experimental disables this instrumentation by default.

```suggestion
        return Arrays.asList("scala-future", "experimental");
```

milanvdm:
@felixbarny I tried to cross-compile against Scala 2.12 and 2.13 but it doesn't seem possible with Maven (unless you create a custom script that runs Maven 2 times with a different parameter).

So I added only 2.13 support in the docs as that is the tested version.

milanvdm:
@felixbarny Just a kind reminder :)

apm-agent-plugins/apm-scala-concurrent-plugin/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+        <groupId>co.elastic.apm</groupId>
+        <version>1.16.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>apm-scala-concurrent-plugin</artifactId>
+    <name>${project.groupId}:${project.artifactId}</name>
+
+    <properties>
+        <apm-agent-parent.base.dir>${project.basedir}/../..</apm-agent-parent.base.dir>
+        <maven.compiler.source>1.9</maven.compiler.source>
```
felixbarny, line 16: can this be removed?

milanvdm: Let me check since Scala does not support all Java version.
Ill also test Scala 2.12 in more detail.

milanvdm: I tested Scala 2.12 and it doesn't work with that version. Not sure how to prevent a plugin to run with a specific version though.

felixbarny:
I'm sorry if this PR feels like going in circles. But to go full circle, I'm not even sure whether the scala-specific instrumentation is needed at all. As `Promise` inherits from `Runnable`, we can just rely on the regular `Executor` instrumentation provided by the `apm-java-concurrent-plugin`. To demonstrate this, I have created https://github.com/milanvdm/apm-agent-java/pull/1

Maybe all that was needed was the additional context propagation for `AsyncHttpClient` (#1172 and #1175).

The big difference from the existing `Executor` instrumentation and this PR is that the context is captured when a `Promise`/`Future` is created (instrumenting the constructor). This circumvents issues of missing context propagation somewhat. But as `Promise` is a `Runnable`, why special-casing `Promise`s? Plus, we still have the problem that `Promise`s created outside of a transaction but used within one don't have the proper context. 

Another factor is the Scala version. Up until 2.12, there was a class called `scala.concurrent.forkjoin.ForkJoinTask` which has been removed in 2.13. I'm not too sure which Scala versions we cover with the `Executor` instrumentation. But I've successfully tested 2.11, 2.12 and 2.13 so far.

milanvdm:
@felixbarny From all the tests I ran, I remember the tests in the repo itself to work but fail in an example project when using the Java Executor.

I'll test your PR again on this project to see how it behaves to refresh my memory.

milanvdm:
@felixbarny I now realize that I forgot to enable experimental features on my test-project while testing Scala 2.12.

I'm rerunning my tests with the APM actually enabled on Scala.

milanvdm:
@felixbarny Ive tested your branch with my test-application and it is not able to propagate the trace-context correctly across all the thread switches when only using the Java executor.

milanvdm:
Scala 2.12 still doesn't work correctly for my example application.

As I mentioned in this PR, I feel like having a test case that is able to trigger the problem my test-application creates, would be quite valuable to add.

felixbarny:
> Scala 2.12 still doesn't work correctly for my example application.

Is this your example application: https://github.com/milanvdm/scala-elastic-apm?

Which test specifically?

> As I mentioned in this PR, I feel like having a test case that is able to trigger the problem my test-application creates, would be quite valuable to add.

++

Is that something you're currently trying to do?

At the moment, we don't instrument `ExecutorService#invokeAny`/`ExecutorService#invokeAll`. Maybe that could be a reason why context propagation doesn't work properly. You also mentioned that Scala executors do batching.

milanvdm:
> Which test specifically?

https://github.com/milanvdm/scala-elastic-apm/blob/master/src/main/scala/me/milan/main/future/Main.scala

> Is that something you're currently trying to do?

I will spend some time on it, but it will take some time to figure out how to trigger it :)

> Maybe that could be a reason why context propagation doesn't work properly. You also mentioned that Scala executors do batching.

https://stackoverflow.com/questions/59642314/scala-future-with-callback-and-context-switching.

felixbarny:
Here you're not closing the scope that's returned when activating the transaction:

https://github.com/milanvdm/scala-elastic-apm/blob/f9b6efe348be5c7c6841e5f8bb8e22d2eea4c77b/src/main/scala/me/milan/main/future/Main.scala#L33

I'm not sure what's the idiomatic way in Scala, but in Java, you'd execute within a [try-with-resources](https://www.elastic.co/guide/en/apm/agent/java/current/public-api.html#api-transaction-activate) block. Maybe you can do

```scala
    for {
      transaction <- Future(ElasticApm.startTransaction().setName("test-trace"))
      scope <- Future(transaction.activate())
      ...
      _ <- Future(transaction.end())
      _ <- Future(scope.close())
    } yield ()
```

But having a more isolated example would really help.

felixbarny:
I've tried the unit tests in this PR with different `ExecutionContext`s, including `parasitic` but I couldn't get it to fail.

milanvdm:
@felixbarny It's tricky indeed. I'll spend some time this weekend to see if I can come up with a minimal example where it fails.

milanvdm:
@felixbarny 
I've added tests to the branch that work on this branch but fail when only using the Java Executor.

apm-agent-plugins/apm-scala-concurrent-plugin/src/test/scala/co/elastic/apm/agent/scala/concurrent/FutureInstrumentationSpec.scala
```diff
@@ -0,0 +1,224 @@
...
+      .onComplete {
+        case Success(value) => promise.success(value)
+        case Failure(exception) => promise.failure(exception)
+      }
+
+    val future = promise
+      .future
+      .map(_ => tracer.currentTransaction().addCustomContext("future", true))
+
+    Await.ready(future, 10.seconds)
```
milanvdm, line 171: @felixbarny I saw in your branch that you were using this approach.

Technically, this is not needed as the test framework will correctly await the Future for you.

But I can imagine that this makes it more explicit when you don't know the test-framework or Scala Futures.

apm-agent-plugins/apm-scala-concurrent-plugin/src/test/scala/co/elastic/apm/agent/scala/concurrent/FutureInstrumentationSpec.scala
```diff
@@ -0,0 +1,237 @@
...
+  test("Handle a Future.sequence correctly") {
+
+    implicit val multiPoolEc: ExecutionContextExecutor =
+      ExecutionContext.fromExecutor(Executors.newFixedThreadPool(3))
+
+    val future = Future
+      .sequence(List(
+        Future(Thread.sleep(25))
+      ))
+      .map(_ => tracer.currentTransaction().addCustomContext("future", true))
```
milanvdm, line 147: From what I tested with a Java Executor only:
1. If the Thread.sleep(x) with x = 1, the test does not always fail
2. Adding a map with adding customContext on the Future inside of the Sequence, the context is correctly propagated
3. So only the map after the Sequence seems to lose context when only using the Java Executor

Sequence uses the parasitic Executor from Scala.

felixbarny:
Thanks for adding the tests! I could reproduce, isolate and fix the issue!


After these lines:

https://github.com/elastic/apm-agent-java/blob/842185664456651f1d6a99602d21a7c052ddf111/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/async/SpanInScopeRunnableWrapper.java#L60

https://github.com/elastic/apm-agent-java/blob/842185664456651f1d6a99602d21a7c052ddf111/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/async/SpanInScopeCallableWrapper.java#L61

Just add `tracer.allowWrappingOnThread();`

This fixes context propagation if a `Executor` that executes the runnable immediately on the same thread, such as parasitic, executes another runnable on another `Executor` that runs on a different thread.

I'll add the fix to the overhaul of our `Executor` instrumentation: https://github.com/elastic/apm-agent-java/pull/1206

milanvdm:
@felixbarny Ill test it out again to see if that fixes the issue in my test application 🙂 And hopefully my new test were representative to a more real life application.

milanvdm:
@felixbarny I've tested your fix with my test-application and it sadly doesn't fix the issue.

I'll try and figure out another failing test-scenario to represent this.

milanvdm:
@felixbarny I've tried to create a test case reproducing the issue on my test-application but until now, unsuccessfully.

I was able to narrow it down to the exact place where it loses context: https://github.com/milanvdm/scala-elastic-apm/blob/check-edge-case/src/main/scala/me/milan/main/future/DummyHttpBackend.scala#L93

I've tried to recreate that logic using only Future, but without the AsyncHttpClient it seems to be tricky.

I'll spend some time trying to make the setup as small as possible to make it easier to track.

felixbarny:
Could you try with the snapshot from this PR? https://github.com/elastic/apm-agent-java/pull/1206#issuecomment-635927423

Also, try to not run the tests in debug mode. I've seen IntelliJ's async stack trace capturing agent remove what our agent does.

milanvdm:
@felixbarny Same issue.

Also getting `2020-06-02 15:26:25,079 [ForkJoinPool-1-worker-17] WARN  co.elastic.apm.agent.impl.ElasticApmTracer - Deactivating a span ('test-trace' 00-2c6ad74b977875a96fcdfb3b93f5bedc-ae1b4e5cf89982aa-01 (4acfb22e)) which is not the currently active span (null). This can happen when not properly deactivating a previous span.`

milanvdm:
@felixbarny Anything I can do to make this move forward? :)

felixbarny:
To summarize my current thinking: Ideally, we'd capture context when it's scheduled for execution and restore the context on execution. It seems like with the current instrumentation that works well most of the time. Unless there's something that breaks context propagation like something related to `AsyncHttpClient`.

In an ideal world, we should fix the loss of context. Pragmatically, it's probably not a bad idea to just combine the two approaches. I expect nothing bad will happen if both the Promise instrumentation (which captures the context when the promise is created and restores when it executes) and the executor instrumentation capture and restore the same context.

I'm willing to give this a shot. In the `Future` (see what I did there? ;) we might be able to remove this instrumentation when we have more extensive coverage of reactive frameworks. Also, when we see that this instrumentation does interfere with the existing context propagation we can re-evaluate. But given that our Scala support is not really mature anyway (we don't have Akka HTTP support yet so I suspect we don't have many Scala users) I think it's well worth the risk of even having this on by default.

milanvdm:
@felixbarny Thanks for the write up :)

I fully agree with your arguments. Once this is merged, I'll continue with adding some extra libraries for Scala.

Some leftover comments (which may need to be solved in a different PR):
* I will test again if this works for Scala 2.12 or not. But besides that, I think we should not support deprecated 2.11. This will need to be checked somehow.
* I think using this Maven setup won't be Scalable (back at ya ;). So it may be something to consider to somehow combine an sbt project with Maven to be able to do cross-compilation and testing for the Scala plugins.

I can create some issues for that if you want to track it.

apm-agent-plugins/apm-scala-concurrent-plugin/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+        <apm-agent-parent.base.dir>${project.basedir}/../..</apm-agent-parent.base.dir>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.scalameta</groupId>
+            <artifactId>munit_2.13</artifactId>
+            <version>0.7.2</version>
+            <scope>test</scope>
+        </dependency>
```
felixbarny, line 24: Maybe one thing to add is a test-dependency on the java-concurrent plugin to test the effects of having both means of context propagation on at the same time.

milanvdm:
Shall I also create an issue pointing to my test-branch that shows an example on where the AsyncHttpClient seems to lose context?

felixbarny:
> Shall I also create an issue pointing to my test-branch that shows an example on where the AsyncHttpClient seems to lose context?

felixbarny:
As a last attempt, could you try out the artifact from https://github.com/elastic/apm-agent-java/pull/1230? I've added some fixes around `ForkJoinPool`s that might have an impact.

milanvdm:
@felixbarny Will have a look :) Do you build artifacts for branches or do I need to build it locally?

felixbarny:
Yes, there’s a link to the artifacts in the comment added by the CI

milanvdm:
@felixbarny Same issue when using the artifact from that branch

felixbarny:
Tests are failing with NPEs on CI. Could you have a look?

milanvdm:
@felixbarny For some reason, `tracer.startRootTransaction(null)` now returns a `null` which was not the case before.
Did something change regarding this?

felixbarny:
Ah, yes. You need to start the agent after building it and before starting transactions

PR #1049, opened 2020-02-24
added preventing the shade plugin from relocating org.slf4j.Logger to *.shaded.slf4j.Logger 

kananindzya:
… *.shaded.slf4j.Logger in error logging instrumentation

## Checklist

- [ ] I have rebased my changes on top of the latest master branch

- [ ] I have updated [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)
- [ ] Added an instrumentation plugin? How did you make sure that old, non-supported versions are not instrumented by accident?

## Screenshots
<!-- _(Optional)_


apm-agent-plugins/apm-error-logging-plugin/src/main/java/co/elastic/apm/agent/error/logging/AbstractLoggingInstrumentation.java
```diff
@@ -35,6 +35,7 @@
 import java.util.Collection;
 
 import static net.bytebuddy.matcher.ElementMatchers.nameContains;
+import static net.bytebuddy.matcher.ElementMatchers.nameContainsIgnoreCase;
```
SylvainJuge: unused import statement.

apm-agent-plugins/apm-error-logging-plugin/src/main/java/co/elastic/apm/agent/error/logging/Log4jLoggingInstrumentation.java
```diff
@@ -37,7 +37,7 @@
 
     @Override
     public ElementMatcher<? super TypeDescription> getTypeMatcher() {
-        return hasSuperType(named("org.apache.logging.log4j.Logger")).and(not(hasSuperType(named("org.slf4j.Logger"))));
+        return hasSuperType(named("org.apache.logging.log4j.Logger")).and(not(hasSuperType(named("org!slf4j!Logger".replaceAll("!", ".")))));
```
SylvainJuge: Probably nitpicking, but use `replaceAll('!','.')` and add a short comment to explain the 'why' and make it consistent with other place(s) it's done like this.

SylvainJuge:
On a side note, this issue also means that we lack a proper test for this feature, as relocation is done at packaging, we could probably add an integration test to prevent any further regression.

apm-agent-plugins/apm-error-logging-plugin/src/main/java/co/elastic/apm/agent/error/logging/AbstractLoggingInstrumentation.java
kananindzya: fixed

apm-agent-plugins/apm-error-logging-plugin/src/main/java/co/elastic/apm/agent/error/logging/Log4jLoggingInstrumentation.java
kananindzya: fixed

eyalkoren:
@kananindzya Your commit was not pushed yet

kananindzya:
> @kananindzya Your commit was not pushed yet

pushed

kananindzya:
> On a side note, this issue also means that we lack a proper test for this feature, as relocation is done at packaging, we could probably add an integration test to prevent any further regression.

added logging-integration-tests

eyalkoren:
@kananindzya this PR is very close to an end, only need to take another look at the test and add to the `CHANGELOG`. Do you want to finish it?

kananindzya:
> @kananindzya this PR is very close to an end, only need to take another look at the test and add to the `CHANGELOG`. Do you want to finish it?

I deleted submodule, changed the scope. But i don't know how to check that test fails without this fix.

eyalkoren:
> i don't know how to check that test fails without this fix

Your'e right, it's not a trivial thing, as in order to test shading effects you would need to test the built and packaged agent. One option is adding a test Servlet to the `simple-webapp` test app, that we use in the `application-server-integration-tests`. However, it is not required in order to merge this PR.
If you want to take a stab on it- go ahead.

Otherwise, since the added test doesn't test the fix and the basic functionality is already tested within the logging plugin, I suggest **removing this test and module altogether** and only verify in a manual test that it didn't work with an older agent and it works with an agent containing your fix.

Bottom line- it is an important fix we would like to merge, and manual test would be sufficient in this case. You would need to merge the latest master so that the `CHANGELOG` is up to date.
Thanks!

PR #1051, opened 2020-02-25
Set default test logging to INFO

cachedout:
This sets the default test logging to the `INFO` level instead of `DEBUG`. 

In doing so, this prevents us from logging environment variables during tests and just in general makes the test output easier to read and errors and warnings much easier to spot.

### DIFF
--- a/apm-agent-core/src/test/resources/elasticapm.properties
+++ b/apm-agent-core/src/test/resources/elasticapm.properties
@@ -1,4 +1,4 @@
-log_level=DEBUG
+log_level=INFO
 log_file=System.out
 # incubating instrumentations should be active in tests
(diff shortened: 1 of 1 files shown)


eyalkoren:
>  in general makes the test output easier to read and errors and warnings much easier to spot.

The output is easier to read, but we really need these logs in order to debug delicate failures. These include basic agent flows like starting/activating/deactivating/ending transactions and spans, which are really important if we want to know why stuff didn't happen as expected.

cachedout:
> The output is easier to read, but we really need these logs in order to debug delicate failures. 

Do you need it in the CI or just when debugging locally? Do you think there is any way to parametrize this so that a flag could be passed from the CLI?

Issue #989, opened 2020-01-13
Support for monitoring the apache dubbo rpc framework  


CoderLan0668:
**Is your feature request related to a problem? Please describe.**
i decide to use elastic apm to mornitor our apps, but our apps use the apache dubbo framework。
So far, it's not support in apm-agent-java

**Describe the solution you'd like**
i need the apche dubbo framework support in the apm-agent-java

**Describe alternatives you've considered**
i want to know that is someone  adding this feature now? or what's the elastic official plan for the apache dubbo? because i reads this page: https://discuss.elastic.co/t/java-agent-how-to-manually-instrument-an-rpc-framework-which-is-not-already-supported/172730

**Additional context**
None

felixbarny:
Hi and thanks for your feature request. Dubbo is currently not on our roadmap. Are you interested in providing a PR for it?

CoderLan0668:
ok , i will make it @felixbarny

CoderLan0668:
I had provided a PR for the dubbo @felixbarny  #1052

SylvainJuge:
Closing as related PR has already been merged

Issue #689, opened 2019-06-24
Follow Up on: Support scheduled tasks #418


wolframhaussig:
**Is your feature request related to a problem? Please describe.**
APM Server: 6.8.0
APM agent: 1.7.0
In elastic/apm-agent-java#418 support for the @Scheduled Annotations was added to the agent to create transactions when a scheduled task is being run. As we needed more flexibility we did not use the @Scheduled annotations and used org.springframework.scheduling.quartz.QuartzJobBean:
```
public class MyJob extends QuartzJobBean {
   @Override
   protected void executeInternal(JobExecutionContext context) throws JobExecutionException {
      doJobWork();//<--Job is implemented here
   }
}
```
The configuration is done in the Spring XML config:
```
	<bean id="detailJob1" class="org.springframework.scheduling.quartz.JobDetailFactoryBean">		
  		<property name="jobClass" value="com.mycompany.MyJob "/>
  		<property name="name" value="Job1" />
	</bean>
	<bean id="triggerJob1" class="org.springframework.scheduling.quartz.SimpleTriggerFactoryBean">
	    <property name="jobDetail" ref="detailJob1"/>
		<property name="repeatInterval" value="60000"/>
		<property name="group" value="my-group-name" />
		<property name="startDelay" value="2500"/>
	</bean>
	<bean id="scheduler" class="rg.springframework.scheduling.quartz.SchedulerFactoryBean" lazy-init="false">
		<property name="startupDelay" value="30" />
	    <property name="triggers" >   
	        <list>	  
				<ref bean="triggerJob1"/>     
            </list>
	    </property> 
	</bean>
```
This seems to be not yet(1.7.0) supported.

**Describe the solution you'd like**
It would be great if you could recognise calls to classes implementing org.quartz.Job (org.springframework.scheduling.quartz.QuartzJobBean implements that) and create a Transaction for each call. The transaction name should be the combination of group name and job name which can be read using `context.getJobDetail().getKey().toString()`.

**Describe alternatives you've considered**
Currently, we are creating transactions using the public API.

eyalkoren:
@wolframhaussig would you be interested in implementing such a plugin yourself?

wolframhaussig:
@eyalkoren I can try :-) I have never done it before, where should I start and where can I find information how to do it?

eyalkoren:
That's great ❤️  ! We are here to assist!

[...]

PR #1052, opened 2020-02-26
add support for the Apache Dubbo

CoderLan0668:
elastic/apm-agent-java#689  What does this PR do?

add instrumentation for the apache dubbo

## Checklist

## Related issues
- Relates elastic/apm-agent-java#989


felixbarny:
Thanks so much for this PR! We have recently got quite a few contributions so it may take a bit more time. In the meantime, please sign the [CLA](https://www.elastic.co/contributor-agreement) and make sure the email you submit matches with your GitHub email.

Are you currently using these changes in your application? Is it already running in production?
Are there any areas you specifically want to have feedback on or are there areas you're not sure?

felixbarny:
Seems like GitHub is a bit confused and can't correctly attribute the commits to your user. That's probably also the reason why our CLA checker can't validate your commits. I suggest adding the email you're using in your gitconfig to GitHub. Go to https://github.com/settings/emails and add the mail you see in https://github.com/elastic/apm-agent-java/pull/1052.patch. You'll then have to re-sign the CLA with the same email address.

Alternatively, you can set your GitHub email address in your gitconfig, squash, rebase and force-push your commits.

CoderLan0668:
@felixbarny  thanks for your reply
1、not running in production
2、CLA check is ok

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/ApacheDubboAttachmentHelperImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public <S> void forEach(String headerName, Invocation invocation, S state, HeaderConsumer<String, S> consumer) {
+        String headerValueStr = doGetHeader(headerName, invocation);
+        if (headerValueStr == null) {
+            return;
+        }
+        String[] headerValues = headerValueStr.split(SEPARATOR);
+        for (String headerValue : headerValues) {
+            consumer.accept(headerValue, state);
+        }
```
felixbarny: It's fine if the protocol doesn't support multiple headers. The agent will only write a single value per header but supports reading multiple values if the protocol supports it.

```suggestion        
        consumer.accept(invocation.getAttachment(headerName), state);
```

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/ApacheDubboAttachmentHelperImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public void setHeader(String headerName, String headerValue, Invocation invocation) {
+        String oldHeader = getFirstHeader(headerName, invocation);
+        String newHeader = headerValue;
+        if (oldHeader != null) {
+            newHeader = oldHeader + SEPARATOR + headerValue;
+        }
+        doSetHeader(headerName, newHeader, invocation);
```
felixbarny:
```suggestion
        invocation.setAttachment(headerName, headerValue)
```

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/advice/ApacheDubboFilterAdvice.java
```diff
@@ -0,0 +1,145 @@
...
+                                           @Advice.Local("apiClazz") Class<?> apiClazz,
+                                           @Advice.Local("transaction") Transaction transaction,
+                                           @Advice.Local("scope") Scope scope) {
+        RpcContext context = RpcContext.getContext();
+        String version = context.getUrl().getParameter("version");
+        Invoker<?> invoker = invocation.getInvoker();
+        apiClazz = invoker.getInterface();
+        DubboApiInfo dubboApiInfo = new DubboApiInfo(
+            apiClazz, invocation.getMethodName(),
+            invocation.getParameterTypes(), version);
```
felixbarny: Rather than allocating an object, just pass all the information as method parameters to the helper

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
```diff
@@ -0,0 +1,162 @@
...
+    private static String DUBBO_SUBTYPE = "dubbo";
+
+    public static final String PROVIDER_SERVICE_NAME_KEY = "elastic-apm-dubbo-provider";
+
+    public static void init(ElasticApmTracer tracer) {
+        DubboTraceHelper.tracer = tracer;
+    }
+
+    @VisibleForAdvice
+    public static String buildDubboRequestName(DubboApiInfo apiInfo) {
```
felixbarny: This allocates lots of objects on each invocation. In order to reduce allocations, could we go for a more simple span name to make it completely allocation free like this:

```java
span.appendToName(apiInfo.getApiClass().getName())
    .appendToName("#")
    .appendToName(apiInfo.getMethodName());
```

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @VisibleForAdvice
+    public static void fillTransaction(Transaction transaction, DubboApiInfo dubboApiInfo) {
+        transaction.withName(buildDubboRequestName(dubboApiInfo));
+        transaction.withType("dubbo");
+        transaction.activate();
+    }
+
+    public static boolean isBizException(Class<?> interfaceClass, Class<?> exp) {
```
felixbarny: Could you explain what this exactly does? I fear that this is going to have too much runtime overhead.

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
```diff
@@ -0,0 +1,162 @@
...
+        CoreConfiguration coreConfig = tracer.getConfig(CoreConfiguration.class);
+        CoreConfiguration.EventType captureBody = coreConfig.getCaptureBody();
+        if (CoreConfiguration.EventType.OFF.equals(captureBody) ||
+            (CoreConfiguration.EventType.ERRORS.equals(captureBody) && !hasError)) {
+            return;
+        }
+
+        captureArgs(transaction, args);
+        if (t != null) {
+            transaction.addCustomContext("throw", t.toString());
```
felixbarny: Would `transaction.captureEception(t);` work?

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
```diff
@@ -0,0 +1,162 @@
...
+        } else {
+            transaction.addCustomContext("return", returnValue.toString());
```
felixbarny: Custom context is reserved for user-defined properties. Core agent plugins should not use them.

apm-agent-plugins/apm-dubbo-plugin/src/test/java/co/elastic/apm/agent/dubbo/AbstractDubboInstrumentationTest.java
```diff
@@ -0,0 +1,313 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
+
+public abstract class AbstractDubboInstrumentationTest {
```
felixbarny: can this class extend `AbstractInstrumentationTest`?

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/ApacheDubboAttachmentHelperImpl.java
CoderLan0668: it doesn't support multiple headers, just simple KV pair.

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/advice/ApacheDubboFilterAdvice.java
CoderLan0668: good for performance, ok

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
CoderLan0668: ok i will reduce the allocations

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
CoderLan0668: the best pratice by dubbo, we will throw some exception that we expected, we call it as the business exception, so we don't want to let these  exceptions to be shown as error in kibana.
 
one way is ignoring these exception by setting the CoreConfigurations, but it's inconvenient to settings the configuration for each application， because each application has their own business exceptions, so i decide to handle it dynamically.

for performance, one exception class will only be judged whether it's business exception once, and the agent will cache the result.

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
CoderLan0668: some business exceptions are not expected to be capture as an exception, in these case,  i want to capture their as a result, and store it in custom context

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
CoderLan0668: sometimes i want to capture the arguments and return object, including exception thrown, and show them in kibana, where should i store them in?

CoderLan0668:
> Does Dubbo support any kind of async request processing or support for streaming or messaging? Or does it just have blocking request/response semantics?


it supports async request processing, but it never be used as far as i know, but it's better to support the async request processing in the agent plugin, i will make it later.

CoderLan0668:
@felixbarny
feel good to receive your sugguestions!
I will improve the dubbo plugin later , according to your great suggestions.

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/ApacheDubboAttachmentHelperImpl.java
felixbarny: And that's fine. There's no need to emulate multi-valued headers via string concatenation.

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
felixbarny: You are already trying to capture business exceptions with custom attributes. I would still just capture a business exception just like a regular one. Then you can both see business exceptions and actual errors in the UI.

Could you make an example of when you would throw a business exception? Would one example be someone trying to order an item when there's no stock for that item? Like `NoToiletPaperException`?

I would argue that it's still interesting to monitor how many of these exceptions are thrown over time.

CoderLan0668: ok, You are right.

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
```diff
@@ -0,0 +1,166 @@
...
+
+        Destination.Service service = destination.getService();
+        service.withType(EXTERNAL_TYPE).withResource(DUBBO_SUBTYPE).withName(DUBBO_SUBTYPE);
+
+        return span.activate();
+    }
+
+    private static void fillName(AbstractSpan<?> span, Class<?> apiClass, String methodName,
+                                 Class<?>[] paramClasses, String version) {
+        span.appendToName(apiClass.getName())
```
felixbarny, line 84: Can we make the name just `"${simpleClassName}#${methodName}"`.

This would make it more consistent how we name transactions for other technologies like Spring MVC and JAS-RS

felixbarny: Use `java.lang.StringBuilder#append(java.lang.CharSequence, int, int)` instead of `Class#getSimpleClassName` for that.

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
felixbarny: Are you planning to remove the biz exception concept?

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
CoderLan0668: what about "${simpleClassName}#${methodName}(${parameters})", because ${parameters} is a unique part of dubbo API

CoderLan0668:
@felixbarny 

1、the name of span or transaction : use '#' to concat the class name and method name instead of '.',  as other instrumentations do, for example SpringMVC, but i keep the  full class name  with package and keep the parameters part  as well.  

2、`Class#getName` or `Class#getSimpleName` they will cache the return value in `Class` Object, it  costs little,. There's no need to use `java.lang.StringBuilder#append(java.lang.CharSequence, int, int)`  instead.

3、as the support for Async Processing has been done, it will be another work to separate them into two PR, it will take more time to separate the code and run test case and validate in application. So,  just one PR, ok?

4、Yes， I‘m planning to use the Dubbo instrumentation for a production application， as far as it‘s merged into release branch.

felixbarny:
> 1、the name of span or transaction : use '#' to concat the class name and method name instead of '.', as other instrumentations do, for example SpringMVC, but i keep the full class name with package and keep the parameters part as well.

I'd prefer it to be consistent with other instrumentation to use the simple class name without parameters. What you could do is to register a listener and override the default name.

> 2、Class#getName or Class#getSimpleName they will cache the return value in Class Object, it costs little,. There's no need to use java.lang.StringBuilder#append(java.lang.CharSequence, int, int) instead.

`Class#getSimpleName` initializes `java.lang.Class.ReflectionData` which performs lots of reflective lookups which may be expensive. The result is also cached in a `SoftReference` which means that depending on available memory and how often it's called, it may have to be re-computed.

> 3、as the support for Async Processing has been done, it will be another work to separate them into two PR, it will take more time to separate the code and run test case and validate in application. So, just one PR, ok?

Ok 👌

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
felixbarny: Would it be possible to register an interceptor that is custom to your application that adds the custom context?

CoderLan0668:
The span names have been adjusted to "${simpleClassName}#${methodName}" and  the invocation of `Class.getSimpleName` has been removed.

CoderLan0668:
@felixbarny  

Is the PR in processing? what can i do to speed up the progress?

felixbarny:
Thanks for the ping. I went over the PR comments again and resolved the ones already addressed. Could you go over the unresolved comments again? We cannot merge this if `custom` is used. See also https://github.com/elastic/apm/blob/master/docs/agents/new-fields.md for more context.

apm-agent-plugins/apm-dubbo-plugin/src/main/java/co/elastic/apm/agent/dubbo/helper/DubboTraceHelper.java
CoderLan0668: Like `capture_body` for http request, it is quite common to capture something for dubbo request，for example, the params, the data was return or the exception thrown.

The difference is the dubbo agent store these data into the `custom context`. 

Is it proper to store these data into `SpanContext` or `TransactionContext`, like http?

felixbarny: There's currently no proper field for that. I agree that could be useful to store parameters and return values but it's not Dubbo specific. I also applies to gRPC, for example.
So I suggest leaving this out in this PR and open a follow-up discussion in https://github.com/elastic/apm instead, according to the guidelines in https://github.com/elastic/apm/blob/master/docs/agents/new-fields.md. This makes sure we align across agents and that the proposal is not Dubbo or Java specific.

> or the exception thrown.

We capture the transaction via `transaction.captureException` now, right?

CoderLan0668: ok， i will remove this out of this PR

CoderLan0668: It has been removed now.

felixbarny:
I've added some suggestions in https://github.com/CoderLan0668/apm-agent-java/pull/1

CoderLan0668:
@felixbarny  merged

docs/supported-technologies.asciidoc
```diff
@@ -262,6 +262,11 @@ The spans are named after the schema `<method> <host>`, for example `GET elastic
 |
 | 1.7.0
 
+|Apache Dubbo
+| 2.7.3 and upper,  2.6.x
```
felixbarny, line 266: Why exactly 2.7.3? Does it not work with 2.7.{0,1,2}? What happens when the agent is used with versions lower than 2.6.0? Does it break, does it work? Should we try to avoid instrumenting those versions?

felixbarny: > Why exactly 2.7.3? Does it not work with 2.7.{0,1,2}?

That's because the method `org.apache.dubbo.rpc.Result#whenCompleteWithContext` was not available previously. This means instrumenting a Dubbo 2.7.2 application would print lots of errors.

CoderLan0668: The agent will not work with dubbo 2.7.{0,1,2}, since 2.7.3, dubbo provides a better async callback like  `AsyncRpcResult.whenCompleteWithContext`, the most importance is the 2.7.{0,1,2}   are not recommended in production by dubbo official， so the agent do not compatible with 2.7.{0,1,2}。

the agent may be work with versions lower than 2.6.0, but they have not been well tested.

felixbarny: Thanks, I have tested with 2.5.x and it works. I've also ensured not to instrument unsupported versions. I think we are really close now! 🙂

PR #1058, opened 2020-02-28
Fully-automated release process

cachedout:
## Summary

This PR introduces a fully-automated release process.

## Dependencies

This PR cannot be merged until https://github.com/elastic/infra/pull/18234 is merged. Additionally, a new release of the APM Pipeline Library is needed which includes this PR: https://github.com/elastic/apm-pipeline-library/pull/413

## Design

The design of this PR is to be purely additive. 

The current release process can still be executed using the same means as prior to this change. Where previously, however, there were a number of manual steps, this new pipeline can be run which will attempt to automate the bulk of those.

## Use

After the pre-requisite PRs are merged and this PR is also merged, the internal CI server should have a new "post-release" job which contains the logic in this PR.

To run the post-build steps, the job will need to be run. It is important that when launching the job that a window is kept open because the job will present dialogue asking questions as the release process progresses. I would prefer if we perhaps set up a Zoom call during the first few invocations of this job so that I can be on-hand to help debug any errors. While this is tested wherever possible, the nature of this job makes testing certain elements somewhat challenging.

**For the time being, your release process should not change. We will take this as incrementally as you like.**

If there is a desire to split this into more discrete changes, or even to merge this in pieces, I am open to that option. Please provide feedback about this question if it's a concern.

## Maturity

This is being submitted as a draft in order to get feedback from the AQA team and from the apm-agent-java folks.

### For the AQA team:

 * Please make sure the Jenkins pipeline syntax looks correct. I'm sure there are a few places where things could be improved.

* In a few cases, this drops back to a shell. Please mark any cases where you think this would be better encapsulated in our pipeline library or where the same functionality can be achieved using existing Jenkins run-time variables or by composing pieces of our library. (I'm almost certain there are at least one or two of these.

* Bugs :)

### For the apm-agent-java team:

* I tried my best to leave plain-text documentation to make this easy to follow. In a few cases, it may not be obvious what is happening so please ask questions!

* If you don't want to automate any of these steps, it's very easy to put them behind gates and default them to being disabled. (Or just to remove them entirely.) Please mark any areas of concerns where you might like us to do this.

* I know that we discussed the automated CHANGELOG in another issue. I have included a step here which generates a CHANGELOG and just provides it as an artifact at the end of the build. I am happy to remove/disable it if you aren't interested in having this artifact available to inspect. It does not modify files in the repo or affect the build in any way.


.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+          steps {
+            dir("${BASE_DIR}") {
+              script {
+                def isMajor = sh(script: 'git tag|tail -1|cut -f2-3 -d "."|{ read ver; test $ver == "0.0"; }', returnStatus: true)
+                if (isMajor == 0) {
+                  // We need to create a new branch. First get the name of the branch
+                  def newBranchName = sh(script: 'git tag|tail -1|sed s/v//|cut -f1 -d "."|awk \'{print $1".x"}\'', returnStdout: true)
+                  echo "This appears to be a major version. We are creating a new branch for $newBranchName in the apm-agent-java repo on GitHub."
+                  // Now we need to branch
+                  sh(script: "git checkout -b $newBranchName")
```
v1v: What do you think to move this logic to a shell script?

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+          steps {
+            dir("${BASE_DIR}"){
+              script {
+                def curTag = sh(script: "git tag|tail -1", returnStdout: true)
+                // Construct the URL with anchor for the release notes
+                // Ex: https://www.elastic.co/guide/en/apm/agent/java/current/release-notes-1.x.html#release-notes-1.13.0
+                def baseUrl = "https://www.elastic.co/guide/en/apm/agent/java/current/release-notes-"
+                def dotX = sh(script: 'git tag|tail -1|sed s/v//|cut -f1 -d "."|awk \'{print $1".x"}\'', returnStdout: true)
+                def bareTag = sh(script: 'git tag|tail -1|sed s/v//', returnStdout: true)
+                def finalUrl = sh(script: "echo $baseUrl$dotX\.html#release-notes-$bareTag")
```
v1v: Shell script that returns the URL?

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+          }
+          steps {
+            sh(label: "Build Docker image", script: "scripts/jenkins/build_docker.sh")
+            // Get Docker registry credentials
+            dockerLogin(secret: "${ELASTIC_DOCKER_SECRET}", registry: 'docker.elastic.co')
+            sh(label: "Push Docker image", script: "scripts/jenkins/push_docker.sh")
+          }
+        }
+        // 14. Opbeans
+        stage('Opbeans') {
```
v1v: This should be tag-based, this pipeline is master based IIRC, therefore the tag conditional can be left as it was previously, what do you think?

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+pipeline {
+  agent { label 'linux && immutable' }
+  environment {
+    REPO = 'apm-agent-java'
+    BASE_DIR = "src/github.com/elastic/${env.REPO}"
+    NOTIFY_TO = credentials('notify-to')
+    DOCKERHUB_SECRET = 'secret/apm-team/ci/elastic-observability-dockerhub'
+    ELASTIC_DOCKER_SECRET = 'secret/apm-team/ci/docker-registry/prod'
+    NEXUS_SECRET = 'secret/apm-team/ci/nexus'
+    MAVEN_CONFIG = '-Dmaven.repo.local=.m2'
```
v1v: nit:
maybe DEFAULT_MAVEN_CONFIG ?

v1v: I said nothing, that's how it's defined in the other Jenkinsfile!

cachedout: Aye. I just moved it from there. :)

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+  options {
+    timeout(time: 3, unit: 'HOURS')
+    buildDiscarder(logRotator(numToKeepStr: '20', artifactNumToKeepStr: '20', daysToKeepStr: '30'))
+    timestamps()
+    ansiColor('xterm')
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+
+  }
+  parameters {
+    string(name: 'MAVEN_CONFIG', defaultValue: '-B -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn', description: 'Additional maven options.')
```
SylvainJuge: What are the use-cases of overriding maven config on a per build basis ? I'm pretty sure we haven't had to use this so far.

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+        }
+        // 2. Review project version.
+        stage('Review project version') {
+          steps {
+            script {
+              // Show the select input modal
+              def INPUT_PARAMS = input message: 'Do you wish to update the version?', ok: 'Next', parameters: [choice(name: 'UPDATE_CHOICE', choices:['Yes', 'No'], description: "This will run `mvn release:update-versions`")]
+              if (INPUT_PARAMS == 'Yes') {
+                  dir("${BASE_DIR}") {
+                    sh mvn release:update-versions
```
SylvainJuge: we switched to using `versions` plugin instead in latest version of the release instructions https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#releasing

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+      }
+    }
+    stage('Release') {
+      options { skipDefaultCheckout () }
+      stages{
+        // 1. Check to see if oss.sonatype.org is up. We throw an exception and fail the job if there is no response.
+        stage('Check oss.sonatype.org') {
+          steps {
+            // If this fails, an exception should be thrown and execution will halt
+            httpRequest(url: "https://oss.sonatype.org/#welcome")
```
SylvainJuge: does the `#welcome` part of the URL matters here ?

.ci/release/Jenkinsfile
SylvainJuge: Also, as an alternative, it might be easier to get project version from maven `pom.xml` rather than parsing it from git.

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+        }
+        // 14. Opbeans
+        stage('Opbeans') {
+          environment {
+            REPO_NAME = "${OPBEANS_REPO}"
+          }
+          steps {
+            deleteDir()
+            dir("${OPBEANS_REPO}"){
+              git credentialsId: 'f6c7695a-671e-4f4f-a331-acdce44ff9ba',
```
SylvainJuge: probably worth having a non-obscure credentials ID and/or a proper env variable for this.

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+        stage('Publish release on GitHub') {
+          steps {
+            githubReleasePublish(id: ${env.RELEASE_ID})
+          }
+        }
+        // 13. Publish Docker images
+        stage('Docker push') {
+          when {
+            beforeAgent true
+            expression { return params.push_docker }
```
SylvainJuge: What controls `params.push_docker` value ? Isn't it always `true` for release pipeline ?

SylvainJuge:
There have been a few recent changes in [release instructions](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#releasing).
- updating changelog early in the process
- use `versions` maven plugin to update versions (this one does not re-formats `pom.xml` as `release` plugin does, which is annoying.
- waiting for documentation to be published, otherwise users will get a 404 when opening link from changelog

.ci/release/Jenkinsfile
cachedout: Very good question. I used [the original Jenkinsfile](https://github.com/elastic/apm-agent-java/blob/master/Jenkinsfile#L34) as a skeleton for this one and that's where this came from. I don't see where it would come into play in this workflow, however, so I think it is safe to remove.

.ci/release/Jenkinsfile
cachedout: It doesn't matter. The root of the site redirects here so I just put it in to avoid the (slight) overhead of the redirect.

.ci/release/Jenkinsfile
cachedout: You are right. This came from the previous Jenkinsfile where it was gated, but that gate should not be necessary here. I will remove it. Thanks!

.ci/release/Jenkinsfile
cachedout: Using this method, we can use Jenkins secure credentials storage to ensure that we don't leak credentials to the log. Could you please elaborate on your reason for wanting this not to be obscured?

.ci/release/Jenkinsfile
cachedout: Ah! Glad you caught that. Thanks! I will make this change.

.ci/release/Jenkinsfile
felixbarny: For a future improvement: could we use this: https://status.maven.org/history.atom

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+        // 3. Execute the release Jenkins job on the internal ci server.
+            // For now we assume that the release has been pushed to staging already on the internal CI
+        // 4. Login to oss.sonatype.org, go to Staging Repositories, close and release the staged artifacts.
+        stage('Nexus release') {
+          steps {
+            script {
+              def spid = getVault('nexus')["staging-profile-id"]
+              dir("${BASE_DIR}"){
+                withEnvMask(vars: [[var: "SPID", password: spid]]){
+                  def foundStagingId = nexusFindStagingId(stagingProfileId: "${SPID}", groupId: "co.elastic.apm")
```
felixbarny: what happens if there are multiple staging repos (for example a leftover from a failed release)?

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+                  nexusCloseStagingRepository(stagingProfileId: "${SPID}", stagingId: foundStagingId)
```
felixbarny: Closing can take a minute or two. Does this return after triggering the close or after the close is finished?

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+                env.RELEASE_ID = ret['id']
+              }
+            }
+
+          }
+        }
+        // 10. Update cloudfoundry/index.yml
+        // Needs to append a line such as the following:
+        // 1.13.0: https://repo1.maven.org/maven2/co/elastic/apm/elastic-apm-agent/1.13.0/elastic-apm-agent-1.13.0.jar
+        stage('Update Cloudfoundry') {
```
felixbarny: let's to this after the `Wait for artifact to be available in Maven Central` step

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+            dir("${BASE_DIR}"){
+              script {
+                def curTag = sh(script: "git tag|tail -1", returnStdout: true)
+                // Construct the URL with anchor for the release notes
+                // Ex: https://www.elastic.co/guide/en/apm/agent/java/current/release-notes-1.x.html#release-notes-1.13.0
+                def baseUrl = "https://www.elastic.co/guide/en/apm/agent/java/current/release-notes-"
+                def dotX = sh(script: 'git tag|tail -1|sed s/v//|cut -f1 -d "."|awk \'{print $1".x"}\'', returnStdout: true)
+                def bareTag = sh(script: 'git tag|tail -1|sed s/v//', returnStdout: true)
+                def finalUrl = sh(script: "echo $baseUrl$dotX\.html#release-notes-$bareTag")
+                def ret = githubReleaseCreate(draft: true, body: "[Release Notes for $bareTag]($finalUrl)")
```
felixbarny: what's the title of the release? So far, we did `"Release $version"`.

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+        stage('Wait for artifact to be available in Maven Central') {
+          steps {
+            dir("${BASE_DIR}"){
+              script {
+                // We are looking for something like this:
+                // https://oss.sonatype.org/service/local/repositories/releases/content/co/elastic/apm/apm-agent-java/1.1.12/apm-agent-java-1.1.12.pom
+                def baseUrl = "https://oss.sonatype.org/service/local/repositories/releases/content/co/elastic/apm/apm-agent-java"
+                def bareTag = sh(script: 'git tag|tail -1|sed s/v//', returnStdout: true)
+                def fullUrl = sh(script: "$baseUrl/$bareTag/apm-agent-java-$bareTag\.pom")
+                waitUntil(initialRecurrencePeriod: 15000) {
```
felixbarny: Does that mean it first starts to try after 15 sec? Is there a timeout? It usually takes around 10 min.

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+                    sh mvn release:update-versions
+                  }
+              } else {
+                  echo 'Skipping version update'
+              }
+            }
+          }
+        }
+        // 3. Execute the release Jenkins job on the internal ci server.
+            // For now we assume that the release has been pushed to staging already on the internal CI
```
felixbarny: How does the release job get triggered?

.ci/release/Jenkinsfile
SylvainJuge: As mentioned in my comment, the change was mostly due to having the `release` plugin to change formatting in `pom.xml`, and not just update versions.

.ci/release/Jenkinsfile
SylvainJuge: There is no special reason, I was just surprised to not have something referred by a human-friendly name here.

.ci/release/Jenkinsfile
cachedout: By hand. I could probably improve this description but what I meant to indicate here is that this pipeline should be run after the existing release job is run.

.ci/release/Jenkinsfile
cachedout: cURL has its own time-out, and after it expires, it will trigger the retry in the `waitUntil`. which will retry every 15 seconds, starting immediately, with a back-off. 

I think that we can improve this with the following changes:

1) Sleep first for ~10 minutes before we enter the re-try loop.
2) Bump the initial re-try up to 1 minute from 15s.

.ci/release/Jenkinsfile
cachedout: Right now, we don't specify a name when we call the GitHub API but this is easy to change. I will make a change upstream in the APM Pipeline library to support this behavior.

.ci/release/Jenkinsfile
felixbarny: maybe open a dialog describing what to do which should be closed after the release job is done?

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+            // If this fails, an exception should be thrown and execution will halt
+            httpRequest(url: "https://oss.sonatype.org/#welcome")
+          }
+        }
+        // 2. Review project version.
+        stage('Review project version') {
+          steps {
+            script {
+              // Show the select input modal
+              def INPUT_PARAMS = input message: 'Do you wish to update the version?', ok: 'Next', parameters: [choice(name: 'UPDATE_CHOICE', choices:['Yes', 'No'], description: "This will run `mvn release:update-versions`")]
```
felixbarny: Instead of updating the version here, I wonder if we should instead require the release version to be set in the release job. See https://maven.apache.org/maven-release/maven-release-plugin/examples/non-interactive-release.html

We could then also require to set the next dev version. The release plugin offers system properties for the release and next dev version: `-DreleaseVersion=1.2.0  -DdevelopmentVersion=1.3.0-SNAPSHOT`.

@eyalkoren @SylvainJuge WDYT?

.ci/release/Jenkinsfile
felixbarny: BTW, I love that this just delegates to the existing and battle-tested release job instead of re-inventing this step which is definitely the most critical piece.

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+      options { skipDefaultCheckout () }
+      stages{
+        // 1. Check to see if oss.sonatype.org is up. We throw an exception and fail the job if there is no response.
+        stage('Check oss.sonatype.org') {
+          steps {
+            // If this fails, an exception should be thrown and execution will halt
+            httpRequest(url: "https://oss.sonatype.org/#welcome")
+          }
+        }
+        // 2. Review project version.
```
felixbarny: Could we add a step that checks whether the current build on master is green?
(Not required for the first version)

.ci/release/Jenkinsfile
SylvainJuge: I'm not really sure to understand how this would be different: do we make the release job just "release current version", which is read from `pom.xml` ?

What if we keep having this prompt for release version and the default value is read from `pom.xml`, if a different value is set just updates version before executing release ?

Having this version set/bump done at CI level allows to avoid running any local git command, which is one less opportunity for human errors.

I'm also ok for doing an increment on the minor version number, but would rather use `mvn versions:set` and not `mvn:release` plugin. for the reason I mentioned above (which I can definitely get by).

felixbarny: > but would rather use mvn versions:set and not mvn:release plugin

During a release:prepare and after a release:perform the versions get updated by the release plugin anyways.

So in the current state of this PR, we would update the version in step 2, then once again in on internal-ci via the `release:prepare` stage (it removes the `-SNAPSHOT`) and finally after the release is done, after the `release:perform` stage (increments the patch version and adds `-SNAPSHOT`, in most cases we actually want to increment the minor version).

What I'm suggesting is to modify the version 2 times instead of 3.

.ci/release/Jenkinsfile
cachedout: [This change is now merged](https://github.com/elastic/apm-pipeline-library/pull/427). I will update this call to use the new functionality.

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+                }
+              }
+            }
+
+          }
+        }
+        // 12. Publish release on Github. This will notify users watching repository.
+        stage('Publish release on GitHub') {
+          steps {
+            githubReleasePublish(id: ${env.RELEASE_ID})
```
eyalkoren: The GH release contains a link to a release-notes URL calculated in the `Create GitHub release draft` stage. This link will be available after the relevant branch gets built, after pushing the new tag in the `Branch creation` phase.
It takes time, so this step needs to block until the release-notes URL is available, otherwise we will publish a Release with a broken link.

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,235 @@
...
+        }
+      }
+    }
+    stage('Release') {
+      options { skipDefaultCheckout () }
+      stages{
+        // 1. Check to see if oss.sonatype.org is up. We throw an exception and fail the job if there is no response.
+        stage('Check oss.sonatype.org') {
+          steps {
+            // If this fails, an exception should be thrown and execution will halt
```
felixbarny: We have added another step since you have started working on this regarding documentation:

> Update [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc) to reflect the new version release:
Go over PRs or git log and add bug fixes and features.
Move release notes from the Unreleased sub-heading to the correct [[release-notes-{major}.x]] sub-heading ([Example PR](https://github.com/elastic/apm-agent-java/pull/1027/files) for 1.13.0 release).

This step seems a bit tricky to automate. But we could instead just show this instruction in an input. When done with manually reviewing and adjusting the release notes, the automated release continues.

.ci/release/Jenkinsfile
cachedout: A while ago I looked into using the Atom feed for this kind of thing but it's fraught with problems. Unfortunately, it's basically just free-text and there is no clear way to determine whether a given entry represents an ongoing outage, a recovery, an announcement, or free puppies. 🐶 I have some ideas about how to do this kind of thing in the future with Github since they have enabled webhooks. My current plan is to get that working and then see if Sonatype is open to the possibility of using webhooks on their page as well.

.ci/release/Jenkinsfile
cachedout: I have consolidated this into a single step [in this commit](d17f51e1add8472b5c93290be2a1b75677161993).

.ci/release/Jenkinsfile
cachedout: Great idea! I have [implemented this upstream](https://github.com/elastic/apm-pipeline-library/pull/432) and will update this pipeline to use the new functionality.

.ci/release/Jenkinsfile
cachedout: This has been removed. Thanks for catching this!

.ci/release/Jenkinsfile
cachedout: The above strategy is now implemented. Thanks for the feedback!

.ci/release/Jenkinsfile
cachedout: The call has been updated. Thanks for the suggestion!

.ci/release/Jenkinsfile
cachedout: At present, [we consider this an error requiring manual intervention](https://github.com/elastic/apm-pipeline-library/blob/master/vars/nexusFindStagingId.groovy#L68).

.ci/release/Jenkinsfile
cachedout: Good suggestion! I have [migrated this logic to a stand-alone shell script](b2488935d718ffc84965de047bdd474c2879db24).

.ci/release/Jenkinsfile
cachedout: We attempt to poll the repo for the actual close event, so this should block until the close actually happens.

.ci/release/Jenkinsfile
cachedout: @v1v Do you mean this this conditional specifically?

```
      when {
        tag pattern: 'v\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
      }
```

v1v: Let me clarify, I meant, this particular stage could be kept in the other CI Jenkinsfile. IIUC, this particular pipeline will be triggered for the master branch and as a consequence, it will create a tag/branch with the version.

Therefore, the other CI Jenkinsfile could be kept as it was, it was configured to listen for branches/tags/PRs and the `when condition` in place will care of running that particular stage. 

I think it might help to simplify this pipeline. what do you think?

.ci/release/Jenkinsfile
cachedout: Good idea! [I have introduced this capability upstream in this PR](https://github.com/elastic/apm-pipeline-library/pull/433).

.ci/release/Jenkinsfile
cachedout: Gotcha! I agree and have restored the original logic. Thanks!

.ci/release/Jenkinsfile
cachedout: I've moved this. Thanks!

.ci/release/Jenkinsfile
cachedout: I have updated this stage substantially. The new workflow is as follows:

1. Display the current version present in `pom.xml` and ask the user if they would like to do an update. If they select "no", then execution continues to the next stage.

2. If "yes" has been selected, an additional dialog appears again displaying the current version along with an input field for the user to enter a new version.

3. After submitting the new version, `mvn versions:set -DnewVersion=${new_version}` is run.

4. Execution continues from there.

.ci/release/Jenkinsfile
cachedout: I have introduced a step into this pipeline that warns if the master build is not passing and presents the user with an input asking if they would like to abort the build or continue.

.ci/release/Jenkinsfile
felixbarny: Or we could trigger the release job when a new tag is created. That way, we don't have a direct dependency to actively trigger the build in the internal CI form the public CI.

The public CI  would execute the [`release:prepare`](http://maven.apache.org/maven-release/maven-release-plugin/examples/prepare-release.html) phase of the maven release plugin which updates the versions and creates a new tag.

The creation of the tag triggers the release job on the internal CI which executes the [`release:perform`](http://maven.apache.org/maven-release/maven-release-plugin/examples/perform-release.html) phase of the release plugin.

This does not have to, and probably shouldn't, be in the first iteration of the automated release.

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,238 @@
...
+  environment {
+    REPO = 'apm-agent-java'
+    BASE_DIR = "src/github.com/elastic/${env.REPO}"
+    NOTIFY_TO = credentials('notify-to')
+    DOCKERHUB_SECRET = 'secret/apm-team/ci/elastic-observability-dockerhub'
+    ELASTIC_DOCKER_SECRET = 'secret/apm-team/ci/docker-registry/prod'
+    NEXUS_SECRET = 'secret/apm-team/ci/nexus'
+    MAVEN_CONFIG = '-B -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn -Dmaven.repo.local=.m2'
+    OPBEANS_REPO = 'opbeans-java'
+
```
v1v:
```suggestion
```

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,238 @@
...
+    string(name: 'branch_specifier', defaultValue: 'master')
+  }
+
+  stages {
+    stage('Initializing'){
+      options { skipDefaultCheckout() }
+      environment {
+        HOME = "${env.WORKSPACE}"
+        JAVA_HOME = "${env.HUDSON_HOME}/.java/java10"
+        PATH = "${env.JAVA_HOME}/bin:${env.PATH}"
```
v1v: These env variables are only defined in the top-level `Initializing` stage that it's only used in the nested checkout stage. Is this required for the other stages?

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,238 @@
...
+      }
+      stages {
+        /**
+         Checkout the code
+        */
+        stage('Checkout') {
+          steps {
+            dir("${BASE_DIR}") {
+              // This is required to ensure that Maven is working with a real branch
+              sh(script: "git checkout -f {$branch_specifier}")
```
v1v:
```suggestion
              sh(script: "git checkout -f ${branch_specifier}")
```

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,238 @@
...
+          }
+        }
+        // 2. Review project version.
+        stage('Review project version') {
+          steps {
+            // 2.1 Fetch the current version from pom.xml
+            def ver = mvnVersion(showQualifiers: true)
+            // 2.2 Raise a prompt asking if the version is correct or if they wish to update
+            script {
+              def should_continue = input(message: "Current version is ${ver}", parameters: [
```
v1v: Moved to the script closure

```suggestion
            script {
              // 2.1 Fetch the current version from pom.xml
              def ver = mvnVersion(showQualifiers: true)
              // 2.2 Raise a prompt asking if the version is correct or if they wish to update
              def should_continue = input(message: "Current version is ${ver}", parameters: [
```

.ci/release/Jenkinsfile
cachedout: I agree that this is a good long-term goal and I also agree that we shouldn't try to achieve this in the first iteration. Therefore, I'm going to resolve this comment because it's becoming a bit tricky to track all the feedback in this PR but we'll come back to it after we have this PR merged. Thanks, @felixbarny ! :)

.ci/release/Jenkinsfile
cachedout: Good catch. I moved them up to the top level where they belong.

.ci/release/Jenkinsfile
cachedout: Hi @felixbarny and @SylvainJuge We've let this discussion sit for a while. I don't have any strong opinions on the way we handle this, but I'd like to come to a consensus so we can get this merged. What's the best way to find a path forward here?

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,255 @@
...
+      options { skipDefaultCheckout() }
+      stages {
+        /**
+         Checkout the code
+        */
+        stage('Checkout') {
+          steps {
+            dir("${BASE_DIR}") {
+              // This is required to ensure that Maven is working with a real branch
+              sh(script: "git checkout -f ${branch_specifier}")
```
v1v, line 41: A few comments regarding the git commands
- Is it possible to use the `gitCheckout` step instead the `sh` step? If so, then we can remove the `githubEnv` references 
- Can we use the `withGitRelease` step before pushing with the command `gitPush`?
- Can we use the `setupAPMGitEmail` step to configure the user.email and user.name?

.ci/release/Jenkinsfile
felixbarny: don't see this as blocking the PR, that's something we can always improve later on

.ci/release/Jenkinsfile
cachedout: I moved this into a shell script. Could you take another look?

cachedout:
Hi friends. :)

This should now be ready for a final round of reviews. Hopefully I've addressed all the concerns but if there is any unresolved issue please let me know and we can either address it here or file an issue for a follow-up. Thanks!

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,229 @@
...
+                  nexusReleaseStagingRepository(stagingProfileId: "${SPID}", stagingId: foundStagingId)
+                }
+              }
+            }
+          }
+        }
+        stage('Branch creation') {
+          steps {
+            dir("${BASE_DIR}") {
+              setupAPMGitEmail()
```
v1v, line 138: See my previous comment

```suggestion
```

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,229 @@
...
+            )
+            stash(allowEmpty: true, name: 'source', useDefaultExcludes: false)
+            dir("${BASE_DIR}") {
+              setupAPMGitEmail()
+              script {
+                env.TAG_BARE = sh(script: "git tag | tail -1", returnStdout: true)
+                env.TAG_VER = sh(script: "git tag | tail -1 | sed s/v//", returnStdout: true)
+                env.TAG_DOT_X = sh(script: 'git tag|tail -1|sed s/v//|cut -f1 -d "."|awk \'{print $1".x"}\'', returnStdout: true)
+              }
+            }
```
v1v, line 51: What do you think to move the setupAPMGitEmail before the stash? Then all the other calls afteeward are not required:

```suggestion
            dir("${BASE_DIR}") {
              setupAPMGitEmail()
            }
            stash(allowEmpty: true, name: 'source', useDefaultExcludes: false)
            dir("${BASE_DIR}") {
              script {
                env.TAG_BARE = sh(script: "git tag | tail -1", returnStdout: true)
                env.TAG_VER = sh(script: "git tag | tail -1 | sed s/v//", returnStdout: true)
                env.TAG_DOT_X = sh(script: 'git tag|tail -1|sed s/v//|cut -f1 -d "."|awk \'{print $1".x"}\'', returnStdout: true)
              }
            }
```

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,229 @@
...
+        stage('Update Cloudfoundry') {
+          steps {
+            deleteDir()
+            unstash 'source'
+            dir("${BASE_DIR}"){
+              script {
+                env.BRANCH_NAME = "master"
+              }
+              withGitRelease() {
+                setupAPMGitEmail()
```
v1v, line 199: See my previous comments
```suggestion
              withGitRelease() {
```

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,229 @@
...
+                script {
+                  def curVer = mvnVersion()
+                  sh(script: "./scripts/jenkins/branch_creation.sh ${curVer}")
+                  def isMajor = sh(script: "./scripts/jenkins/is_major.sh ${curVer}", returnStatus: true)
+                  if (isMajor == 0) {
+                    input message: "This was a major version release. Please update the conf.yml in the docs repo before continuing", ok "Continue"
+                  }
+                }
+                gitPush()
+              }
```
v1v, line 152: IIRC, `BRANCH_NAME` is a read only env variable, `withEnv` could help to change the value for the release context, what do you think?

```suggestion
              withEnv(["BRANCH_NAME=tags/${env.TAG_BARE}"]){
                withGitRelease() {
                  script {
                    def curVer = mvnVersion()
                    sh(script: "./scripts/jenkins/branch_creation.sh ${curVer}")
                    def isMajor = sh(script: "./scripts/jenkins/is_major.sh ${curVer}", returnStatus: true)
                    if (isMajor == 0) {
                      input message: "This was a major version release. Please update the conf.yml in the docs repo before continuing", ok "Continue"
                    }
                  }
                  gitPush()
                }
              }
              
```

cachedout: Interesting. I had no problems in setting it directly when I was testing it locally but I agree that if we can use a context-based approach that it is much cleaner. Thanks for pointing this out. I wasn't familiar with `withEnv`.

.ci/release/Jenkinsfile
cachedout: I was wondering about that. I was just being extra-cautious here since there isn't much harm in setting it multiple times but if it will work this way than that's even better.

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,229 @@
...
+                def new_version = input(message: "Please enter version to change to:", parameters:
+                  [
+                    [
+                      $class: 'StringParameterDefinition',
+                      defaultValue: "${ver}",
+                      description: 'We will run mvn versions:set -DnewVersion=<NEW_VERSION>', name: 'New Version'
+                    ]
+                  ]
+                )
+                sh(name: "mavenVersionUpdate", script: "mvn versions:set -DnewVersion=${new_version}")
```
felixbarny, line 116: An issue with `versions:set` is that it creates a `pom.xml.versionsBackup` file for every `pom.xml` in the project.

I think `mvn --batch-mode release:update-versions -DdevelopmentVersion=${new_version}` would be the better option.

Also, I don't see the new changes being committed in this step, is that correct?

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,226 @@
...
+              } else {
+                echo "Skipping version update"
+              }
+            }
+          }
+        }
```
felixbarny, line 122: Add a step that just prints a message like.

`Start the release job on the internal CI. Click proceed once the job has succeeded. Click cancel if the release has failed and manually undo the release.`

felixbarny:
The build failure in the Javadoc step should be fixed once you merge the master in your branch.

.ci/release/Jenkinsfile
cachedout: Good call! Step added.

.ci/release/Jenkinsfile
cachedout: Thanks, @felixbarny. I have updated the `mvn` call per your suggestion. I have also added a `git commit`. Let me know if it looks good to you.

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,234 @@
...
+                  description: "Selecting 'Yes' will allow you to select the new version in the next step."
+                ]
+              ])
+              if (should_continue == 'Yes'){
+                def new_version = input(message: "Please enter version to change to:", parameters:
+                  [
+                    [
+                      $class: 'StringParameterDefinition',
+                      defaultValue: "${ver}",
+                      description: 'We will run mvn versions:set -DnewVersion=<NEW_VERSION>', name: 'New Version'
```
felixbarny, line 112:
```suggestion
                      description: 'We will update the project version in all pom.xml files. Set this to your desired <release-version>-SNAPSHOT (for example 1.2.3-SNAPSHOT if you want to release version 1.2.3).', name: 'New Version'
```

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,234 @@
...
+          }
+        }
+        stage('Review project version') {
+          steps {
+            script {
+              def ver = mvnVersion(showQualifiers: true)
+              def should_continue = input(message: "Current version is ${ver}", parameters: [
+                [
+                  $class: 'ChoiceParameterDefinition',
+                  name: "Do you wish to update the version?",
```
felixbarny, line 101:
```suggestion
                  name: "You are about to release version ${ver - '-SNAPSHOT'}. Do you wish to update the version?",
```

.ci/release/Jenkinsfile
```diff
@@ -0,0 +1,234 @@
...
+                    [
+                      $class: 'StringParameterDefinition',
+                      defaultValue: "${ver}",
+                      description: 'We will run mvn versions:set -DnewVersion=<NEW_VERSION>', name: 'New Version'
+                    ]
+                  ]
+                )
+                sh(name: "mavenVersionUpdate", script: "mvn --batch-mode release:update-versions -DdevelopmentVersion=${new_version}")
+                withGitRelease() {
+                  sh(script: "git commit -a -m 'Version bump'")
```
felixbarny, line 118: don't forget to push :)

cachedout: The reason I didn't push here is that it happens in the step shortly after. Do you think we should have a push here as well? I'm fine with either. :)

felixbarny: If we don't push here, the internal CI doesn't see the changes

cachedout: Ah! You're right of course. I'll make this change.

cachedout:
Just an update on this so people don't think it's abandoned. From my perspective, this is just awaiting a final review from @eyalkoren and @SylvainJuge . No hurry whatsoever, of course, but I wanted to make sure everybody is on the same page so there's no confusion. Thanks!

cachedout:
Tests are passing and so I am merging this [per the most recent comment by Felix](https://github.com/elastic/apm-agent-java/pull/1058#issuecomment-641994056). 

We can schedule a pair-release session whenever you like. Thanks all for the feedback and direction on this!

PR #1061, opened 2020-03-02
Add more loggers for diagnosing issues with inferred spans

felixbarny:
Also document known issues


felixbarny:
Also logs statistics about dropped and sent events when the agent shuts down. If log level is set to debug, it logs if events couldn't be added to the ring buffer.

PR #1065, opened 2020-03-04
Exclude directories from modification checks

eyalkoren:
@felixbarny @cachedout Not sure if this would work, but do you see a problem in trying this out?


cachedout:
I don't think this will do any harm but I would still like to understand why this is happening. It's quite unexpected. Thanks for merging this though!

PR #1070, opened 2020-03-04
Exclude Undertow classes from profiling

### LINKED ISSUE #1068: Exclude Undertow from profiling by default
Looking at the pre-release documentation for 1.14.0

`profiling_excluded_classes` currently includes all common Java servers except for Undertow. Adding `io.undertow.*` to the list would save its users from having to add it manually.

eyalkoren:
closes #1068


felixbarny:
Tests are failing because the docs need an update

PR #1073, opened 2020-03-04
Make JfrParserTest compatible with Java 9

eyalkoren:
The `java.net.URLDecoder#decode()` overload that accepts a Charset was only introduced in 10


felixbarny:
I think we can be a bit more aggressive when it comes to adopting new Java versions for the test target. The CI currently runs on Java 10 but I see no good reason not to update to Java 14, when it comes out later this month.

Having said that, we currently have `maven.compiler.testTarget` set to `9`. So we should be consistent and either raise the test target version or to merge this PR so the test sources match the `maven.compiler.testTarget`.

SylvainJuge:
So far the only reason I know that kept us on 10 was the unexpected
failures of javadoc with 11 and later that are reported as errors and not
warning during maven build.

eyalkoren:
One reason no to be too aggressive is that once we use a Java 14 feature, ALL current forks working on something will get test/build failures and be forced to install it once they fetch.
In any case, merging this PR seems the right thing as things stand now, regardless of my +1 on using a newer version.

PR #1075, opened 2020-03-04
Disable agent on early Java 8 versions

SylvainJuge:
# What does this PR do?

Early Java 8 versions are known to have bugs that are triggered when used with an agent doing instrumentation. Of course, those flaws might not be visible without the agent, thus agent becomes the obvious culprit whereas in practice we can't really work around those.

As a result, using our agent with one of those JVMs will likely result in:
- a crashed JVM
- likely some angry ops, because killing, even if it's a computer process is a bad thing.
- a lot of back & forth with user/customer and support trying to debug a core dump
- unmet expectations: Java 8 is officially supported.

## Checklist

## Author's Checklist
- [x]  run integration tests to make sure none of our docker images is using one of those JVMs
- [x]  test manually on working / not buggy versions

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/AgentMain.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/AgentMain.java
@@ -11,9 +11,9 @@
  * the Apache License, Version 2.0 (the "License"); you may
  * not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -72,8 +72,16 @@ public class AgentMain {
             // for example, Spring Boot restarts the application in dev mode
             return;
         }
+
+        String javaVersion = System.getProperty("java.version");
+        if (!isJavaVersionSupported(javaVersion)) {
+            // gracefully abort agent startup is better than unexpected failure down the road
+            System.err.println(String.format("Failed ot start agent - JVM version not supported: %s", javaVersion));
+            return;
+        }
+
         try {
-            FileSystems.getDefault();
+            FileSystems.getDefault(); // likely not required if it was only for java 7 compatibility
             final File agentJarFile = getAgentJarFile();
             try (JarFile jarFile = new JarFile(agentJarFile)) {
                 instrumentation.appendToBootstrapClassLoaderSearch(jarFile);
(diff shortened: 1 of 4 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/bci/AgentMain.java
```diff
@@ -90,6 +98,43 @@ public synchronized static void init(String agentArguments, Instrumentation inst
...
+            return false;
+        } else if (major == '7' || major > '8') {
+            return true;
+        } else {
+            // java 8
+            int updateIndex = javaVersion.lastIndexOf("_");
+            if (updateIndex <= 0) {
+                return false;
+            } else {
+                String updateVersion = javaVersion.substring(updateIndex + 1);
```
eyalkoren: One of our Tomcat tests fails due to the version called: `1.8.0_72-internal` 🙁

eyalkoren:
Indeed, one of the WebSphere tests fails with `[2020-03-05T07:06:27.540Z] [websphere] Failed ot start agent - JVM version not supported: 1.8.0`

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/AgentMain.java
```diff
@@ -72,8 +72,16 @@ public synchronized static void init(String agentArguments, Instrumentation inst
             // for example, Spring Boot restarts the application in dev mode
             return;
         }
+
+        String javaVersion = System.getProperty("java.version");
+        if (!isJavaVersionSupported(javaVersion)) {
+            // gracefully abort agent startup is better than unexpected failure down the road
+            System.err.println(String.format("Failed ot start agent - JVM version not supported: %s", javaVersion));
```
eyalkoren:
```suggestion
            System.err.println(String.format("Failed to start agent - JVM version not supported: %s", javaVersion));
```

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/AgentMain.java
SylvainJuge: fixed in the last commit.

SylvainJuge:
> We probably want to do that vendor-specific. Not sure about compatibility between Oracle and OpenJDK, but I assume IBM's J9 doesn't have to be under same restriction.

That's a very good question, I'm fine to disable this check for IBM JVMs, but only apply it to openjdk/oracle releases that are known to be problematic.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/AgentMain.java
```diff
@@ -99,43 +105,52 @@ public synchronized static void init(String agentArguments, Instrumentation inst
...
         if (major < '7') {
             // given code is compiled with java 7, this one is unlikely in practice
             return false;
         } else if (major == '7' || major > '8') {
             return true;
+        } else if (!vmName.contains("HotSpot(TM)")) {
+            // non-hotspot JVMs are not concerned (yet)
+            return true;
         } else {
             // java 8
```
eyalkoren:
```suggestion
            // HotSpot 8
```

PR #1076, opened 2020-03-05
JAX-RS transaction name should be based on first annotated method

### LINKED ISSUE #1062: Jax-RS Transaction name gets overriden when calling another Jax-RS
When the implementation of a JAX-RS service calls another Method that exposes a JAX-RS service, the name of the transaction is overriden to the name of the second method, obscuring the original call.

## Steps to reproduce
Consider the following class:

```
import javax.ws.rs.GET;
import javax.ws.rs.POST;
import javax.ws.rs.Path;

@Path("/test")
public class TestRestService {

    @GET
    public void methodA(){
        doSomeStuff()
        methodB();
        doMoreStuff()
    }

    @POST
    public void methodB(){
        doMethodBStuff()
    }

}


```
when methodA is called via HTTP, the Transaction will be named "methodB" by APM. The call to methodA is obscured. This happens because calls to JAX-RS endpoint rename the transaction even if the method is not the entry point to the transaction.

## Expected behavior
I would expec the APM-Transacton that gets traced when calling methodA via REST to be called "TestRestService#methodA" even if methodB is called inside the transaction.

## Sugestion
in #748 a priority system for overriding transaction names was introduced, but names with the same NamePriority as the current name can still override the current name. Maybe this could be changed to only higher priority? Just a thought.

eyalkoren:
## What does this PR do?

Introduce new internal API that allows setting a span name with the same priority.

## Checklist

## Related issues

Closes #1062 

## Use cases

JAX-RS-annotated method delegating to another JAX-RS-annotated method - since it is the same plugin (and instrumentation), both attempt to set the name with the same priority, but we prefer the first, rather than the second.

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
@@ -169,12 +169,15 @@ public abstract class AbstractSpan<T extends AbstractSpan> extends TraceContextH
 
     /**
      * Resets and returns the name {@link StringBuilder} if one of the following applies:
-     *  1.  the provided priority is {@code >} {@link #namePriority}
-     *  2.  the provided priority is {@code ==} {@link #namePriority} AND `overrideIfSamePriority` is `true`
+     * <ul>
+     *      <li>the provided priority is {@code >} {@link #namePriority}</li>
+     *      <li>the provided priority is {@code ==} {@link #namePriority} AND {@code overrideIfSamePriority} is {@code true}</li>
+     * </ul>
      * Otherwise, returns {@code null}
      *
-     * @param namePriority the priority for the name. See also the {@code AbstractSpan#PRIO_*} constants.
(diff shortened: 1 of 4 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -164,7 +164,22 @@ public StringBuilder getNameForSerialization() {
...
+    }
+
+    /**
+     * Resets and returns the name {@link StringBuilder} if one of the following applies:
+     *  1.  the provided priority is {@code >} {@link #namePriority}
+     *  2.  the provided priority is {@code ==} {@link #namePriority} AND `overrideIfSamePriority` is `true`
+     * Otherwise, returns {@code null}
+     *
+     * @param namePriority the priority for the name. See also the {@code AbstractSpan#PRIO_*} constants.
+     * @return the name {@link StringBuilder} if the provided priority is {@code >=} {@link #namePriority}, {@code null} otherwise.
```
eyalkoren: Note to self - adjust this comment and add description for the second parameter (copy + pasta)

PR #1082, opened 2020-03-12
Enhance JDBC connection string parsing for multiple hosts

### LINKED ISSUE #1031: enhance destination fields for Oracle
I was eager to see that the new 1.13.0 is finally out. I just tried it on our test system and would like a few enhancements:
We use a tomcat installation with multiple applications which are configured using property files. Our config for the JDBC connection typically looks like this:
`app.jdbc.url=jdbc:oracle:thin:@(DESCRIPTION=(ADDRESS=(PROTOCOL=TCP)(HOST=localhost)(PORT=6203))(CONNECT_DATA=(SERVER=DEDICATED)(SERVICE_NAME=DB.FQDN.ORG.DE)))`

The new agent creates the following destination fields:
```
 "destination": {
        "address": "(DESCRIPTION=(ADDRESS=(PROTOCOL=TCP)(HOST=localhost)(PORT=6203))(CONNECT_DATA=(SERVER=DEDICATED)(SERVICE_NAME=DB.FQDN.ORG.DE)))",
        "port": 1521,
        "service": {
          "resource": "oracle",
          "name": "oracle",
          "type": "db"
        }
      }
```
As the connection string can contain options which do not belong the destination server(e.g. SERVER=DEDICATED defines that the server will start a dedicated server process for our sessions) those strings might differ while still connecting to the same db.

We would like to see the port parsed correctly(6203 instead of 1521) and the address of the destination handled correctly(I am open for discussions about the _correct_ way :-) )
I also saw in the [testcases](https://github.com/eyalkoren/apm-agent-java/blob/e163e8562524de9c7ff3261ac0614bcc0c3a7cb5/apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaDataTest.java) that you do not currently extract the instance name - maybe this could be stored in the context.destination.service.resource?

## Additional context
Syntax of the connect strings are defined here: https://docs.oracle.com/en/database/oracle/oracle-database/18/netrf/local-naming-parameters-in-tnsnames-ora-file.html

I don't know how hard it is to implement a solution for the destination adress which fits all - or most - users:
- only use SERVICE_NAME or SID as destination (would work for us)
  - what happens if both fields are present?
  - does not fit if  SERVICE_NAME/SID is not globally unique(e.g. DB instead of DB.FQDN.DE)
- use the combination of HOST, PORT and SERVICE_NAME/SID
  - might create problems when applications running on the db server are connecting with HOST=localhost
  - might create problems when some applications are connecting with the hostname and some with the IP
  - what happens if multiple adresses are defined(failover)?
- make it configurable?

eyalkoren:
## What does this PR do?

Enhance the JDBC connection string parser to support dedicated syntaxes for multiple hosts in Oracle, MySQL and MariaDB.

## Checklist

## Related issues

Closes #1031

### DIFF
--- a/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
+++ b/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
@@ -28,8 +28,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Deque;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class ConnectionMetaData {
 
@@ -83,7 +90,7 @@ public class ConnectionMetaData {
             try {
                 ret = connectionUrlParser.parse(tmpUrl, user);
             } catch (Exception e) {
-                logger.error("Failed to parse connection URL", e);
+                logger.error("Failed to parse connection URL: " + tmpUrl, e);
             }
         } else {
             // Doesn't hurt to try...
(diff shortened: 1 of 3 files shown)


apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
```diff
@@ -165,36 +175,136 @@ ConnectionMetaData parse(String connectionUrl, String user) {
...
+                            if (stack.isEmpty()) {
+                                currentValueBuffer = null;
+                            } else {
+                                currentValueBuffer = stack.peek().value;
+                            }
+                            break;
+                        }
+                        default: {
+                            if (currentValueBuffer == null) {
+                                logger.warn("Failed to parse Oracle DB address list from: {}", connectionUrl);
```
SylvainJuge: do we really want to continue parsing in case of error ?

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
```diff
@@ -208,7 +318,16 @@ ConnectionMetaData parse(String connectionUrl, String user) {
         MYSQL("mysql") {
             @Override
             ConnectionMetaData parse(String connectionUrl, String user) {
-                return ConnectionUrlParser.defaultParse(connectionUrl, dbVendor, 3306, user);
+                String host = "localhost";
+                int port = 3306;
+                HostPort hostPort = parseMySqlFlavor(connectionUrl);
```
SylvainJuge: [minor] same here, having a `null` return value would avoid duplication of the `null` check and `port > 0` below.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
```diff
@@ -165,36 +175,136 @@ ConnectionMetaData parse(String connectionUrl, String user) {
...
                     }
+                }
 
-                    String[] parts = connectionUrl.split(":");
-                    if (parts.length > 0) {
-                        host = parts[0];
+                return new ConnectionMetaData(dbVendor, host, port, user);
+            }
+
+            private HostPort parseAddressList(String connectionUrl) {
```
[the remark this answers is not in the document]
eyalkoren: > maybe return `null` when `HostPort.host == null || HostPort.port < 0`

👍 for `HostPort.host == null`, however I am pretty sure the port is optional, so one may set a host and not the port (in which case the default will be used- there is a test for that)

> maybe naming this method with Oracle might be relevant as it's very Oracle-specific.

It's within the Oracle parser (enum)

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
eyalkoren: Could be an error or an unknown use of the connection string. However, I only care about very specific nodes in this tree, so if the parsing yields a valid tree in which I can find the tree nodes I am looking for, I don't care.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
eyalkoren: I don't think-  most (or all) make port configuration optional, each provider using a different default port if not configured.

PR #1087, opened 2020-03-18
Document and reorder classpath config

### LINKED ISSUE #959: Document and re-order classpath configuration priority
**Issue description**

Agent configuration can be provided through an `elasticapm.properties` file in the application classpath, which is required for applications that use [Attach API](https://www.elastic.co/guide/en/apm/agent/java/master/setup-attach-api.html).

There are a few issues:
- the feature is [not documented](https://www.elastic.co/guide/en/apm/agent/java/master/configuration.html#configuration), and should be at level 5 (see below), thus after the external `elasticapm.properties` in filesystem.
- due to an implementation detail, properties in classpath are always applied with the highest priority, see https://discuss.elastic.co/t/apm-config-overriding/207854 for details

**Proposal**

Document all the layers (including the implicit ones) of the configuration with their priority:
1. Remote configuration
1. `elastic.apm.*` JVM properties
1. `ELASTIC_APM_*` environment variables
1. external `elasticapm.properties` in filesystem (if any)
1. `elasticapm.properties` file in application classpath
1. Agent default configuration

Make agent configuration loading implement this ordering, which will be a breaking change for people who rely on current behavior where `elasticapm.properties` in classpath is always applied with highest priority which prevents overriding it remotely.

SylvainJuge:
## What does this PR do?

Change order of configuration sources and document all supported options.

As a bonus, there are a few minor tweaks to the documentation.

## Checklist

- ~~I have added tests that prove my fix is effective or that my feature works~~ >> testing manually for now
- ~~I have updated [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)~~

## Author's Checklist
- [x] manually test most configuration priorities

## Related issues
- Closes #959


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
```diff
@@ -101,10 +102,6 @@ public ElasticApmTracerBuilder reporter(Reporter reporter) {
         return this;
     }
 
-    public ElasticApmTracerBuilder withConfig(String key, String value) {
```
SylvainJuge: This method wasn't used outside tests.

docs/setup-javaagent.asciidoc
```diff
@@ -8,11 +8,12 @@
 The first step in getting started with the Elastic APM Java agent is to download the latest release of the agent jar file from
 link:https://search.maven.org/search?q=g:co.elastic.apm%20AND%20a:elastic-apm-agent[maven central].
 
-NOTE: Don't declare a dependency to the agent in your application.
+NOTE: Agents are not like regular application dependencies.
+Don't declare a dependency to the agent in your application.
```
SylvainJuge: just like with kids, it's often better to explain the `why` behind the `don't`

SylvainJuge:
Link to generated doc : http://apm-agent-java_1087.docs-preview.app.elstc.co/diff

docs/configuration.asciidoc
```diff
@@ -5,28 +5,35 @@ Please only make changes in configuration.asciidoc.ftl
...
+. runtime attach parameters, those are mutually exclusive +
   No prefix is required for the configuration keys.
-
-Central configuration overrides all other settings. Java system properties override Environment variables. Environment variables override `elasticapm.properties`
+.. `--config` parameter (<<setup-attach-cli>>)
+.. arguments of  `ElasticApmAttacher.attach(...)` (<<setup-attach-api>>)
+.. `elasticapm.properties` in classpath root with `ElasticApmAttacher.attach()` (<<setup-attach-api>>) +
+. `elasticapm.properties` file, by default in the same directory the agent jar. +
+  Location can be set with the <<config-config-file>> option. +
+  No prefix is required for the configuration keys.
```
felixbarny: Remove this? Seems to be a duplicate of 2.

docs/configuration.asciidoc
```diff
@@ -5,28 +5,35 @@ Please only make changes in configuration.asciidoc.ftl
...
+. default values
 
 Configuration options marked with Dynamic true can be changed at runtime via configuration sources which support dynamic reloading.
 {kibana-ref}/agent-configuration.html[Central configuration] and the `elasticapm.properties` file are such sources.
 Java system properties can be dynamic as well by being set from within the application.
 
+Please note that there are two distinct ways to use `elasticapm.properties`: as an external configuration file, and as a
+classpath resource. Only the external file can be used for dynamic configuration.
+
```
felixbarny: I think by using footnotes it's easier to understand and with less text.

docs/configuration.asciidoc
```diff
@@ -5,28 +5,35 @@ Please only make changes in configuration.asciidoc.ftl
...
-To adapt the Elastic APM agent to your needs,
-you can configure it using different configuration sources,
-which have different naming conventions for the property key.
-In descending order of precedence:
+The Elastic APM agent can be configured through different configuration sources,
+with the following descending order of precedence:
 
 [arabic]
 . {apm-app-ref}/agent-configuration.html[Central configuration]
+. `elasticapm.properties` file, by default in the same directory the agent jar. +
```
felixbarny: I think we could do a better job of flagging which configuration sources allow dynamic changes. Maybe we could use asciidoc footnotes for that.

```suggestion
. {apm-app-ref}/agent-configuration.html[Central configuration] footnote:dynamic[Configuration options marked with Dynamic true can be changed at runtime with this configuration source]
. `elasticapm.properties` file footnote:dynamic[], by default in the same directory the agent jar. +
```

disclaimer: I've never used footnotes before and have not tested this renders properly

SylvainJuge:
Using footnotes does not really work as expected here:
- there is no way to make a footnote close by where we define them (always at bottom of page), and our page is really long.
- we have no way to use small badges/labels that could be used instead

Thus since we can't have everything in the same list (priorities, dynamic support and few other details), I've moved paragraph about Dynamic config to it's own section (+references to it).

----------

Screenshot from 2020-03-24 17-08-32

----------

PR #1089, opened 2020-03-19
Make instrumentation tests more robust

eyalkoren:
Reusing the same tracer across different instrumentation tests.
It's far from being perfect, but it is a step in the right direction and most importantly, it lets us do what we want with the helpers initialization without breaking the tests.


eyalkoren:
For some reason, it seems it makes tests more flaky...

apm-agent-core/src/test/java/co/elastic/apm/agent/MockTracer.java
```diff
@@ -102,4 +146,18 @@ public static ElasticApmTracer create(ConfigurationRegistry configurationRegistr
         when(tracer.getConfig(any())).thenAnswer(invocation -> configurationRegistry.getConfig(invocation.getArgument(0)));
         return tracer;
     }
+
+    public static class MockInstrumentationSetup {
+        public final ElasticApmTracer tracer;
+        public final MockReporter reporter;
+        public final ConfigurationRegistry config;
+        public final TestObjectPoolFactory objectPoolFactory;
```
SylvainJuge: [minor] maybe provide getters for consistency, we don't have much direct field access in our codebase, and this part will be reused extensively into plugin tests.

apm-agent-core/src/test/java/co/elastic/apm/agent/MockTracer.java
```diff
@@ -81,6 +84,47 @@ public static ElasticApmTracer createRealTracer(Reporter reporter, Configuration
...
+    /**
+     * If an instrumentation has already been initialized by some other test, returns the static
+     * {@link co.elastic.apm.agent.bci.ElasticApmInstrumentation#tracer}.
+     * Otherwise, Creates a real tracer with a {@link MockReporter} and a mock configuration which returns default
+     * values that can be customized by mocking the configuration.
+     */
+    public static synchronized MockInstrumentationSetup getOrCreateInstrumentationTracer() {
+
+        ElasticApmTracer tracer = ElasticApmInstrumentation.tracer;
+        if (tracer == null || tracer.getState() == ElasticApmTracer.TracerState.STOPPED) {
```
SylvainJuge: why does a stopped tracer should trigger creating a new one ? Is that triggered for example if a test leaves the tracer in a stopped state ? Couldn't we just resume it when it's the case ?

What is a bit confusing here is that we have a call to `TracerInternalApiUtils.resumeTracer(tracer);` in the `else` block.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/AsyncInstrumentation.java
```diff
@@ -71,15 +71,12 @@
...
-            asyncHelperManager = HelperClassManager.ForSingleClassLoader.of(tracer,
-                "co.elastic.apm.agent.servlet.helper.AsyncContextAdviceHelperImpl",
-                "co.elastic.apm.agent.servlet.helper.AsyncContextAdviceHelperImpl$ApmAsyncListenerAllocator",
-                "co.elastic.apm.agent.servlet.helper.ApmAsyncListener");
+            if (asyncHelperManager == null) {
+                asyncHelperManager = HelperClassManager.ForSingleClassLoader.of(tracer,
+                    "co.elastic.apm.agent.servlet.helper.AsyncContextAdviceHelperImpl",
+                    "co.elastic.apm.agent.servlet.helper.AsyncContextAdviceHelperImpl$ApmAsyncListenerAllocator",
+                    "co.elastic.apm.agent.servlet.helper.ApmAsyncListener");
+            }
```
SylvainJuge: This null-check if statement is :heart_eyes:

apm-agent-core/src/test/java/co/elastic/apm/agent/MockTracer.java
```diff
@@ -81,6 +84,47 @@ public static ElasticApmTracer createRealTracer(Reporter reporter, Configuration
             .build();
     }
 
+    /**
+     * If an instrumentation has already been initialized by some other test, returns the static
```
felixbarny: Why do we allow tests to leak the tracer? It may be configured to only use certain instrumentations.

We could keep a stack trace of where the tracer initialized. That way we can find out which test has leaked the tracer and fix it instead of re-using a tracer that is in an unknown state.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -369,11 +369,15 @@ private static void logClassLoaderHierarchy(@Nullable ClassLoader classLoader, L
     /**
      * Reverts instrumentation of classes and re-transforms them to their state without the agent.
      * <p>
-     * This is only to be used for unit tests
+     * NOTE: THIS IS ONLY TO BE USED FOR UNIT TESTS
```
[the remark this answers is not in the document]
eyalkoren: > Do you see a problem with using this method to detach the agent in the future?

No, just thought of emphasizing this note 🙂 . However, before using that we should probably disable the tracer and make sure everything stops using it, because this may happen in bulks, so some code may be instrumented while other is not (a strange undefined state). 

> What about using 'are assertions enabled' heuristic and throw an exception if this code is used in production ?

Based on the above, we may DO start using it.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -369,11 +369,15 @@ private static void logClassLoaderHierarchy(@Nullable ClassLoader classLoader, L
...
      * </p>
      */
     public static synchronized void reset() {
-        if (resettableClassFileTransformer == null || instrumentation == null) {
+        if (instrumentation == null) {
+            return;
+        }
+
+        if (resettableClassFileTransformer == null) {
```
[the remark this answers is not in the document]
eyalkoren: I will add

apm-agent-core/src/test/java/co/elastic/apm/agent/MockTracer.java
eyalkoren: This is according to the way we defined the `STOPPED` state [as irreversible](https://github.com/elastic/apm-agent-java/blob/96b7d60e4edcb39242ab5d7922111d52fd9c1023/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java#L784).

> Is that triggered for example if a test leaves the tracer in a stopped state ?

Yes, for example - [this one](https://github.com/elastic/apm-agent-java/blob/842185664456651f1d6a99602d21a7c052ddf111/apm-opentracing/src/test/java/co/elastic/apm/opentracing/OpenTracingBridgeCloseTest.java#L48).

apm-agent-core/src/test/java/co/elastic/apm/agent/MockTracer.java
eyalkoren: The basic problem is the way we spread around tracer's references, so it is held as a field in many places (all Instrumentations, numerous helpers etc.), which means it leaks by design whenever multiple instrumentation tests are executed on the same JVM. Currently this is only a problem in tests, but it may become a real issue when we implement detach/re-attach.

I think the proper solution is to refactor the tracer to be a singleton, then we can easily control all accesses centrally.

Issue #1056, opened 2020-02-27
publish snapshot as part of the regular build pipeline


SylvainJuge:
# What we currently have

- one scheduled job on `internal-ci` that rebuilds the whole agent from `master` with the `@daily` schedule and publishes the snapshot to sonatype maven repository takes (40 to 60 minutes).
- one build pipeline that rebuilds the agent on every update of `master` (and also other branches).

# Improvement proposal

- add a step to the pipeline that will publish last built (and tested) snapshot to sonatype repository
- make this step only executed on the `master` branch.
- remove the scheduled job to publish snapshot

This is independant (but somehow related) to the ongoing effort to automate release process https://github.com/elastic/apm-agent-java/issues/991

# Benefits

- one less build job to take care of
- everything is described in pipeline
- avoid testing and rebuilding

cachedout:
> add a step to the pipeline that will publish last built (and tested) snapshot to sonatype repository

I assume here that you mean [this pipeline ](https://github.com/elastic/apm-agent-java/blob/master/Jenkinsfile) which runs on the [public-facing Jenkins instance](https://apm-ci.elastic.co).

We can do this, but in the past there have been some reservations expressed by @felixbarny and perhaps others about interacting with Nexus via the public Jenkins, and the risk of potential leaks of credentials, etc.

Before I start this work, I want to make sure we're all comfortable with the public-facing CI doing this work. Otherwise, we can take another look at doing this on the internal CI or find another strategy.

cc: @felixbarny and @eyalkoren

felixbarny:
IMO, the snapshot build should run on the same Jenkins with a very similar configuration than the release job as it's a good proxy for whether the release job might fail.

But as we have removed the testing step on the release job (we instead require the master to be green) we could do something similar for the snapshot job. Or even trigger the snapshot job after a successful build on master.

cachedout:
> IMO, the snapshot build should run on the same Jenkins with a very similar configuration than the release job as it's a good proxy for whether the release job might fail.

If we keep this on the internal CI, I see the following possible changes available to us:

[...]

PR #1096, opened 2020-03-24
Script to detect changes in remote build

cachedout:
## What does this PR do?
This allows our internal CI system to poll a remote CI to determine if the most recent successful build has a new commit hash.

## Checklist

- [ ] ~~My code follows the [style guidelines of this project]~~(CONTRIBUTING.md#java-language-formatting-guidelines)

- [ ]~~ ~~New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#testing) pass locally with my changes~~

## Related issues

- Relates https://github.com/elastic/apm-agent-java/issues/1056
## Screenshots
<!-- _(Optional)_

### DIFF
--- a/scripts/jenkins/get_remote_build.py
+++ b/scripts/jenkins/get_remote_build.py
@@ -28,7 +28,7 @@ if __name__ == '__main__':
     tmp_file = '/tmp/apm-ci-jenkins-last-hash'
 
     last_hash = read_hash(tmp_file)
-    cur_hash = getSCMInfroFromLatestGoodBuild("http://apm-ci.elastic.co", "apm-agent-go/opbeans-go-mbp/master")
+    cur_hash = getSCMInfoFromLatestGoodBuild("http://apm-ci.elastic.co", "apm-agent-go/opbeans-go-mbp/master")
     write_last(tmp_file, cur_hash)
     
     print(last_hash == cur_hash)
--- a/scripts/jenkins/get_remote_build.py
+++ b/scripts/jenkins/get_remote_build.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+
+"""
+This script retreives the latest successful build from
+apm-ci.elastic.co and returns the commit hash
+"""
+import os
+from jenkinsapi.jenkins import Jenkins
+
+def getSCMInfroFromLatestGoodBuild(url, jobName, username=None, password=None):
+    J = Jenkins(url, username, password)
+    job = J[jobName]
+    lgb = job.get_last_good_build()
+    return lgb.get_revision()
+
+def write_last(tmp_file, hash):
+    with open(tmp_file, 'w+') as f_:
+        f_.write(hash)
+
+def read_hash(tmp_file):
+    if not os.path.exists(tmp_file):
+        return ''
+    with open(tmp_file, 'r') as f_:
+        git_hash = f_.readline()
+    return git_hash
+
+if __name__ == '__main__':
+    tmp_file = '/tmp/apm-ci-jenkins-last-hash'
+
+    last_hash = read_hash(tmp_file)
+    cur_hash = getSCMInfroFromLatestGoodBuild("http://apm-ci.elastic.co", "apm-agent-go/opbeans-go-mbp/master")
+    write_last(tmp_file, cur_hash)
+    
+    print(last_hash == cur_hash)
--- a/scripts/jenkins/get_remote_build.py
+++ b/scripts/jenkins/get_remote_build.py
@@ -7,7 +7,7 @@ apm-ci.elastic.co and returns the commit hash
 import os
 from jenkinsapi.jenkins import Jenkins
 
-def getSCMInfroFromLatestGoodBuild(url, jobName, username=None, password=None):
+def getSCMInfoFromLatestGoodBuild(url, jobName, username=None, password=None):
     J = Jenkins(url, username, password)
     job = J[jobName]
     lgb = job.get_last_good_build()
--- a/scripts/jenkins/get_remote_build.py
+++ b/scripts/jenkins/get_remote_build.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 """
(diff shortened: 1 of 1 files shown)


scripts/jenkins/get_remote_build.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
+
+"""
+This script retreives the latest successful build from
+apm-ci.elastic.co and returns the commit hash
+"""
+import os
+from jenkinsapi.jenkins import Jenkins
+
+def getSCMInfroFromLatestGoodBuild(url, jobName, username=None, password=None):
```
mdelapenya: I guess it's a typo: "getSCM`Infro`FromLatestGoodBuild"

scripts/jenkins/get_remote_build.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
+
+"""
+This script retreives the latest successful build from
```
v1v:
```suggestion
This script retrieves the latest successful build from
```

scripts/jenkins/get_remote_build.py
```diff
@@ -0,0 +1,34 @@
...
+        return ''
+    with open(tmp_file, 'r') as f_:
+        git_hash = f_.readline()
+    return git_hash
+
+if __name__ == '__main__':
+    tmp_file = '/tmp/apm-ci-jenkins-last-hash'
+
+    last_hash = read_hash(tmp_file)
+    cur_hash = getSCMInfroFromLatestGoodBuild("http://apm-ci.elastic.co", "apm-agent-go/opbeans-go-mbp/master")
```
v1v:
```suggestion
    cur_hash = getSCMInfoFromLatestGoodBuild("http://apm-ci.elastic.co", "apm-agent-go/opbeans-go-mbp/master")
```

scripts/jenkins/get_remote_build.py
cachedout: You're right! Thanks for catching this.

SylvainJuge:
I'm not really familiar with python scripts, but sounds a good approach, even if it requires to use polling.

I have a few questions regarding the behavior in those case
- what is returned if the last build is not successful ?
- what is returned if there is no git commit id in the remote repository ?

How will this be used in the snapshot deploy job ? should we have something in shared library that will allow us to use this directly from the Jenkins pipeline ?

cachedout:
Thanks for the questions @SylvainJuge . I will answer them below

> what is returned if the last build is not successful 

This always polls for the most recent _successful_ build. A build that fails will not be evaluated by this script. I believe this is the behavior that we want but please correct me if I'm wrong:

1. The script polls the most recent successful build and finds a commit hash of X.

2. A new build runs and fails.

3. The script polls again, but still just for the most recent successful build (from step 1). The git hash is still X, so nothing happens and polling continues.

> what is returned if there is no git commit id in the remote repository ?

Under what conditions would this happen? We know that the job on the public CI always checks out from the HEAD of master. There should always be a commit id at the HEAD of the branch.

That said, if for some reason this does occur (or perhaps Jenkins somehow fails to return it), I do think we should handle that case. I will modify this script so that if we don't see a hash that we do nothing, which seems like the safest option.

>How will this be used in the snapshot deploy job ? 

I am going to experiment with a few options today. I do believe that it should be possible to run the script in the pipeline, get the return value and then simply exit the pipeline without continuing if the script tells us that nothing has changed. It's a little tricky to know what status to set the build to, though. I am thinking that we always set it to 'Passing' even if some invocations upload a snapshot and others do not. However, I am interested to hear other opinions and I'm very flexible on what we do here.

> should we have something in shared library that will allow us to use this directly from the Jenkins pipeline ?

We could go this route but I wanted to use the Python library to interact with Jenkins in this case just because it was easy. I don't have any real argument against moving it to the shared library other than it would take a little more time and Python was the shortest route for me to get the job done in this case. :)

Let me know what you think. Thanks, @SylvainJuge !

cachedout:
> I've got some questions regarding if the infra/nginx is a bit flaky how can we ensure there is a retry? In other words, if 500 or another http error is raised, could we have some logic in place to sleep/retry? Or this responsibility should be in the script that uses this one?

Originally I thought that implementing retries should be unnecessary because the entire job will already be executed on a relatively short polling interval.

Now that I think more about this, though, a timeout would fail the job and we can avoid that noise by implementing the retry logic directly into the script. So, I will do this today. Thanks for the suggestion, @v1v !

SylvainJuge:
I agree that 'not having git commit' is more a corner case, but it's always better to have a known behavior when unexpected happens (and it often does). Same if the remote server is not available for example.

For the snapshot deploy job, if it was a Jenkins pipeline I would have suggested to do a 2 step pipeline
- checkout
- publish snapshot: this step is executed only if there is something to do

When the job is not able to query for the status of the remote job, it should fail.

If you go to the "simple python script" route, then this should also be fine:
- make build pass if there is nothing to do
- make build fail when there is any error or unable to poll the remote job

cachedout:
I believe I have uncovered an upstream bug in the `jenkinsapi` Python library wherein it returns the incorrect hash if there are two jobs with similar names. In our case, we're getting the hash for the most recent successful job in the `apm-pipeline-library` job. 

I am investigating alternatives for moving forward on this issue in light of that discovery.

cachedout:
OK, I have found the issue and submitted a PR upstream to work around it: https://github.com/pycontribs/jenkinsapi/pull/772

I think that we should see if we can get that PR merged and into a release. If that doesn't happen relatively quickly, we can always point to a fork in order to get this going.

cachedout:
Just by way of status here, I am still waiting on an upstream maintainer to respond to my request for a merge in pycontribs/jenkinsapi#772.

cachedout:
I am going to close this in favor of the approach outlined in https://github.com/elastic/apm-agent-java/pull/1167 . We can always re-open this if it's needed in the future.

PR #1108, opened 2020-03-31
Exclude Quartz 1 from instrumentation

felixbarny:
To avoid IncompatibleClassChangeError: Found class org.quartz.JobExecutionContext, but interface was expected

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -32,6 +32,9 @@ endif::[]
 ===== Bug fixes
 * When Servlet-related Exceptions are handled through exception handlers that return a 200 status code, agent
 shouldn't override with 500 - {pull}1103[#1103]
+* Exclude Quartz 1 from instrumentation to avoid
+  `IncompatibleClassChangeError: Found class org.quartz.JobExecutionContext, but interface was expected` - {pull}1108[#1108]
+
 
 [[release-notes-1.x]]
 === Java Agent version 1.x
--- a/apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentation.java
+++ b/apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentation.java
@@ -11,9 +11,9 @@
  * the Apache License, Version 2.0 (the "License"); you may
  * not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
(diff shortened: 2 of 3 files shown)


SylvainJuge:
Quartz 1.8.x is about 7 to 8 years old, what would it cost to provide a compatible instrumentation for it ? Is there any issue with binary compatibility where we expect it to be an interface in 2.x versions and it was a class in older version ?

felixbarny:
> Quartz 1.8.x is about 7 to 8 years old, what would it cost to provide a compatible instrumentation for it ?

We'd have to have another module compiled against Quartz 1.8.
I'd wait for user requests to do that. It shouldn't be hard but if no one uses it anyway it's probably not a good use of time to implement that.

> Is there any issue with binary compatibility where we expect it to be an interface in 2.x versions and it was a class in older version ?

Correct. As we compile against the interface and at runtime, it's a class the JVM bails out.

eyalkoren:
> Correct. As we compile against the interface and at runtime, it's a class the JVM bails out.

Ahh, I wondered why it failed, so BB adds an `invokeinterface` instruction while it needs to be `invokevirtual`/`invokespecial`?

felixbarny:
That is not Byte Buddy's fault. Byte Buddy takes the advice method as a template and inlines it into the instrumented method. The `javac` compiler does the `INVOKEINTERFACE` call. See also the byte code for the relevant method:

https://github.com/elastic/apm-agent-java/blob/96b7d60e4edcb39242ab5d7922111d52fd9c1023/apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameAdvice.java#L59-L61

```
L14
    LINENUMBER 59 L14
    ALOAD 3
    ALOAD 0
    INVOKEINTERFACE org/quartz/JobExecutionContext.getJobDetail ()Lorg/quartz/JobDetail; (itf)
    INVOKEINTERFACE org/quartz/JobDetail.getKey ()Lorg/quartz/JobKey; (itf)
    INVOKEVIRTUAL org/quartz/JobKey.toString ()Ljava/lang/String;
    INVOKEVIRTUAL co/elastic/apm/agent/impl/transaction/Transaction.withName (Ljava/lang/String;)Lco/elastic/apm/agent/impl/transaction/AbstractSpan;
    CHECKCAST co/elastic/apm/agent/impl/transaction/Transaction
    LDC "scheduled"
```

eyalkoren:
Of course it's not Byte Buddy's fault, it's ours by compiling against an interface, there is no blaming here 🙂 
However, it is the one adding the instructions in practice...

PR #1117, opened 2020-04-02
Fix parent/child issues with inferred spans

### LINKED ISSUE #1112: Regular spans can't be a child of profiler-inferred spans
## Describe the bug
APM Timeline starts with DB spans and shows them at top, but as per the application execution Java spans should be shown first and then DB spans.

## Steps to reproduce
Attached screenshots of our application.

In the below screenshot the transaction starts with DB Span:

APM-SPAN Order Mismatch

Below screenshot shows Java Spans after DB spans:

Java Span Misorder2

## Expected behavior
Various spans hierarchy  (like JAVA, DB etc) should be as per the application execution flow in the APM timeline.

## Debug logs

<details>
<summary>Click to expand</summary>

```
replace this line with your debug logs
```
</details>

felixbarny:
Inferred spans can track a list of child or successor ids
With transitive reduction, the UI can then make regular spans a child of inferred spans

## What does this PR do?

closes #1112 

## Checklist

- ~[ ] I have updated [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)~
- ~[ ] Added an instrumentation plugin? How did you make sure that old, non-supported versions are not instrumented by accident?~


apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/CallTree.java
```diff
@@ -391,6 +414,25 @@ public void resetState() {
         }
     }
 
+    public void addChildId(long id) {
+        if (childIds == null) {
+            childIds = new LongList();
```
eyalkoren, line 419: Doesn't worth pooling?

apm-agent-plugins/apm-profiling-plugin/src/test/java/co/elastic/apm/agent/profiler/CallTreeTest.java
```diff
@@ -262,6 +265,30 @@ void testDectivationBeforeEnd() throws Exception {
...
+     *  [a          ]      [a          ]
+     *   [2   ]  [3]        [b    ] [3]   <- b is supposed to stealChildIdsFom(a)
+     *    [b   ]            [2   ]           however, it should only steal 2, not 3
+     */
+    @Test
+    void testDectivationBeforeEnd2() throws Exception {
+        assertCallTree(new String[]{
+            "   bbbb b      ",
+            " a aaaa aa a a ",
+            "1 2    2  3 3 1"
```
eyalkoren, line 279:
```suggestion
            "   bbbb b      ",
            " a aaaa a a a ",
            "1 2    2 3 3 1"
```
Just to make sure that one measurement is enough

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/CallTree.java
felixbarny: Not really. Lists are quite tricky to pool as they can have vastly different sizes. The danger is that over time, all lists in the pool will be quite large even though there are only few that actually contain a lot of elements.
Judging from the benchmarks, the allocations caused by the inferred spans processing increase by 10%. That seems reasonable and it's definitely not the major source of allocations.

apm-agent-plugins/apm-profiling-plugin/src/test/java/co/elastic/apm/agent/profiler/CallTreeTest.java
felixbarny: Hm, that doesn't work because `b` is still considered the top of the stack at the time `3` gets activated.

eyalkoren: Why isn't that enough, if you know that activation of `3` happens after deactivation of `2` and you know that `b` cannot be a child of `3`?

felixbarny: Because of situations like these:

```
 aa  aa 
1  22  1
```

We have to add each activation to the `childIds` of the current top of the stack.

However, it least to errors in situations like these:
```
 bbb      
 aaa aa a 
1   2  2 1
```

As 2 is falsely declared a child of b when it's actually a child of a. Currently not sure how to fix it without introducing regressions.

felixbarny: I think I've managed to fix it

apm-agent-plugins/apm-profiling-plugin/src/test/java/co/elastic/apm/agent/profiler/CallTreeTest.java
```diff
@@ -266,29 +272,181 @@ void testDeactivationBeforeEnd() throws Exception {
...
+     *    [b   ]           [2   ]          however, it should only steal 2, not 3
      */
     @Test
     void testDectivationBeforeEnd2() throws Exception {
         assertCallTree(new String[]{
-            "   bbbb b      ",
-            " a aaaa aa a a ",
-            "1 2    2  3 3 1"
+            "   bbbb b     ",
+            " a aaaa a a a ",
```
eyalkoren, line 284: 💪

apm-agent-plugins/apm-profiling-plugin/src/test/java/co/elastic/apm/agent/profiler/CallTreeTest.java
```diff
@@ -266,29 +272,181 @@ void testDeactivationBeforeEnd() throws Exception {
...
+            "1    23   32 1"
+        }, new Object[][]{
+            {"a", 8},
+            {"  b", 3},
+            {"  c", 3},
+        }, new Object[][]{
+            {"1", 13},
+            {"  a", 11},
+            {"    b", 2},
+            {"  2", 6},
```
eyalkoren, line 394: How come `2` is not a child of `a`?

felixbarny: Wow, good catch! From now on, I call you the eagle eye. Or maybe Eyeal 🤔 

Submitting fix shortly that searches for common ancestors, instead of always adding the child_ids on the top of the stack

apm-agent-plugins/apm-profiling-plugin/src/test/java/co/elastic/apm/agent/profiler/CallTreeTest.java
```diff
@@ -262,6 +271,210 @@ void testDectivationBeforeEnd() throws Exception {
...
+            {"    b",     2},
+            {"    2",     6},
+            {"      3",   4},
+            {"        c", 2}
+        });
+
+        // this is empty but not null because a has stolen the child ids from b
+        assertThat(spans.get("b").getChildIds().getSize()).isEqualTo(0);
+        assertThat(spans.get("c").getChildIds()).isNull();
+        // has both 2 and 3 as child_ids, even though only 2 is a direct child
```
felixbarny, line 402: This might be problematic for the UI: https://github.com/elastic/kibana/issues/62952#issuecomment-624472852

felixbarny: Fixed it

PR #1123, opened 2020-04-06
Remove incubating, introduce experimental

### LINKED ISSUE #876: Improving documentation - stop using "incubating" and explain the expectations
**Is your feature request related to a problem? Please describe.**
JMS instrumentation of ElasticAPM Java agent is marked as "in incubating" since version 1.7.0,
leaving developers with a lot of uncertainty in regards to production readiness. 
Additionally, - even when uncertainty is put aside - this instrumentation is activated by "not deactivating any other instrumentation" leading to situations where one could not use the "incubating" JMS-instrumentation, e.g. when another instrumentation needs to be deactivated.  

**Describe the solution you'd like**

- Clarification in the documentation about what incubating means.
- Ideally an ETA when feature leaves this state.
- Separate configuration key to explicitly activate incubating features (e.g. enable_incubating_features=jms). A sensible default behavior in regards to incubating state change would be to ignore this particular entry when features leave incubating and print out a warn-message in the logs.

**Describe alternatives you've considered**
The alternative considered and currently in-place is to implement some sort of decorator-class which resembles the behavior of the JMS-instrumentation in question (e.g. decorator for MessageListener).
As this decorator only resembles the behavior of the ElasticAPM-instrumentation, both the incubating instrumentation and the homegrown solution might diverge from a UI-users perspective.

eyalkoren:
Removing the `incubating` term from our documentation if favour of `experimental` and add some expectation setting regarding features labeled as `experimental`.
As part of that, the `incubating` instrumentation group is deprecated, but this PR is still maintaining backwards compatibility for `disable_instrumentation` configurations that use it.

## Checklist

## Related issues
Closes #876

### DIFF
--- a/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
+++ b/apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
@@ -11,9 +11,9 @@
  * the Apache License, Version 2.0 (the "License"); you may
  * not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -79,7 +79,7 @@ public class HibernateSearch6Instrumentation extends ElasticApmInstrumentation {
 
     @Override
     public Collection<String> getInstrumentationGroupNames() {
-        return Arrays.asList(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE, "incubating");
+        return Arrays.asList(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE, "experimental");
     }
 
     public static class Hibernate6ExecuteAdvice {
--- a/docs/intro.asciidoc
+++ b/docs/intro.asciidoc
@@ -29,20 +29,6 @@ You can then use the APM app in Kibana to gain insight into latency issues and e
 More detailed information on how the Agent works can be found in the <<faq-how-does-it-work,FAQ>>.
 
 [float]
-[[experimental]]
-=== Experimental features
-
-Some agent features or <<supported-technologies-details,plugins>> are marked as `experimental`, which means:
-
-* They are opt-in (disabled by default).
-* There are no guarantees regarding changes in such features in future releases.
-* In some cases they have special caveats or unknowns that come with them.
-In such cases, the specific feature/plugin documentation should indicate that.
-
-There are cases where we decide to release a feature as `experimental` to be on the safe side and only make it an official feature after we get some feedback about it.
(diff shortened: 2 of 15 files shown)


apm-agent-plugins/apm-hibernate-search-plugin/apm-hibernate-search-plugin-6_x/src/main/java/co/elastic/apm/agent/hibernate/search/v6_x/HibernateSearch6Instrumentation.java
```diff
@@ -79,7 +79,7 @@
 
     @Override
     public Collection<String> getInstrumentationGroupNames() {
-        return Arrays.asList(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE, "incubating");
+        return Arrays.asList(HibernateSearchConstants.HIBERNATE_SEARCH_ORM_TYPE, "experimental");
```
felixbarny, line 82: I'd add both `incubating` and `experimental`. Otherwise, it's a breaking change for people that have set `disable_instrumentation=incubating,foo`.

eyalkoren: Good point! 
Not really related to this location- I will internally treat `incubating` as if it is an `experimental` tag

docs/intro.asciidoc
```diff
@@ -28,6 +28,20 @@ You can then use the APM app in Kibana to gain insight into latency issues and e
...
+Some agent features or <<supported-technologies-details,plugins>> are marked as `experimental`, which means:
+
+* They are opt-in (disabled by default).
+* There are no guarantees regarding changes in such features in future releases.
+* In some cases they have special caveats or unknowns that come with them.
+In such cases, the specific feature/plugin documentation should indicate that.
+
+There are cases where we decide to release a feature as `experimental` to be on the safe side and only make it an official feature after we get some feedback about it.
+Any such feedback will be highly appreciated, through a related GitHub issue or through the https://discuss.elastic.co/c/apm[APM discuss forum].
+
```
bmorelli25, line 44: I would remove this entire `Experimental features` section. I think it's repetitive. Here's why:

```
* They are opt-in (disabled by default).
* There are no guarantees regarding changes in such features in future releases.
```

The note below, included with all experimental features, already explains both of these points. 

```
NOTE: This feature is currently experimental, which means it is disabled by default and it is not guaranteed to be backwards compatible in future releases.
```

The third bullet point explains that documentation will indicate if there are caveats or unknowns.

The only thing missing is the feedback point. If that's important, I say we just update the experimental tag to something like this: 

```
NOTE: This feature is currently experimental, which means it is disabled by default and it is not guaranteed to be backwards compatible in future releases. Have feedback? Visit the https://discuss.elastic.co/c/apm[APM discuss forum].
```

eyalkoren: I agree. I added this because there are places that are not part of the configuration docs (like [the tuning page](http://apm-agent-java_1123.docs-preview.app.elstc.co/guide/en/apm/agent/java/master/tuning-and-overhead.html#circuit-breaker) or [supported technologies page](http://apm-agent-java_1123.docs-preview.app.elstc.co/guide/en/apm/agent/java/master/supported-technologies-details.html#supported-java-methods) where we just say `(experimental)`.
However, since in such places we DO refer to the enablement configuration, we can probably remove this section.

> The only thing missing is the feedback point

I don't think I want to add that into these configuration docs multiple times. It's ok to leave out.

bmorelli25: Sounds good to me. Thanks!

Issue #1104, opened 2020-03-30
docs: Improve configuration precedence


bmorelli25:
As you know, we're extremely limited in how we can visually customize the documentation. However, I think there are still things we can do to improve the [configuration precedence list](https://www.elastic.co/guide/en/apm/agent/java/master/configuration.html#configuration-dynamic). As-is, it feels a bit squished to me, and it seems difficult to scan.

I played around with it this morning, and moved the content to a table with badges for dynamic configuration. The padding is pretty bad, and I wish we could customize the column width.... but does this seem like an improvement to you?

<img width="684" alt="Screen Shot 2020-03-30 at 10 46 33 AM" src="https://user-images.githubusercontent.com/5618806/77944723-4d7c8380-7274-11ea-81d3-15c3a9bd2dec.png">

cc @felixbarny / @SylvainJuge

felixbarny:
Looks great! Let's get it merged in.

SylvainJuge:
Nice! Can we just have `dynamic` instead of `dynamic configuration` in the badge ?

PR #1124, opened 2020-04-06
docs: update config precedence and add dynamic badge

bmorelli25:
## What does this PR do?
This PR moves the configuration precedence information to a table and adds badges for dynamic configuration. The dynamic badges are links to the `=== Dynamic configuration` section heading.

## Related issues
Closes https://github.com/elastic/apm-agent-java/issues/1104.
For https://github.com/elastic/observability-dev/issues/639.

## Screenshots
Screen Shot 2020-04-06 at 2 58 56 PM


bmorelli25:
Badges for dynamic config were added after the `${option.description}`. Do we want to rename the badge from `dynamic` to `reloadable`?

bmorelli25:
We could also swap configs to horizontal tables. They do take up a bit more room, but maybe with a collapsible option, that's OK? How do you feel about this:

java-settins-tables

bmorelli25:
Please note that I'm not sure why the preview and my screenshots above are not showing a border around the collapsible sections. See this comment for a realistic visual: https://github.com/elastic/docs/pull/1794#issuecomment-609864528

bmorelli25:
One other option. Keep Default / Type / Dynamic in a table. Move Java System Properties, Property file, and Environment into a collapsible section that defaults to closed.

SylvainJuge:
I really like the `Dynamic` badge, it's really clear and understandable.

Regarding configuration options, I feel that
- tables (especially without borders) don't really fit long strings like our option names
- using a collapsible option might help to save some space here, especially with the vertical layout
- the `> Properties of xxxx` could be replaced with `> xxxxxx (default_value)`, which allows to have something useful even when collapsed.
- what about adding the whole option description to the collapsible part ? that means that our reference configuration page will be mostly a list of all available options ?

bmorelli25:
Thanks for the feedback @SylvainJuge!

> the > Properties of xxxx could be replaced with > xxxxxx (default_value), which allows to have something useful even when collapsed.

I like the thought, but I don't think that adding the default to the drop-down will work. Some of the configurations have very long defaults, e.g., https://www.elastic.co/guide/en/apm/agent/java/master/config-core.html#config-sanitize-field-names. I tried adding the type to the drop-down, but it looks funky to me:

Screen Shot 2020-04-13 at 10 21 43 AM

> using a collapsible option might help to save some space here, especially with the vertical layout

> what about adding the whole option description to the collapsible part ? that means that our reference configuration page will be mostly a list of all available options ?

Yeah, the goal with the collapsible block was to save space. But the more I think about it, the configuration options should be expanded by default. If we collapse everything, I worry we'll make it harder for users who typically `ctrl + f` the docs.

I'm going to remove the collapsible sections from this PR. I don't think we have the right tooling to make these changes yet.

PR #1137, opened 2020-04-14
Don't skip Javadoc when building agent

felixbarny:
Currently, the Javadoc task gets skipped when building the agent. If there's an error in the Javadoc we only learn that during the release phase which is not ideal.

I'm not sure if there was a reason behind skipping the Javadoc or if it was a leftover from setting up the pipeline.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -81,7 +81,7 @@ pipeline {
               dir("${BASE_DIR}"){
                 sh """#!/bin/bash
                 set -euxo pipefail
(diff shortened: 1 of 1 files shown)


SylvainJuge:
Actually there is a dedicated stage in the pipeline for javadoc generation, the part you changed just avoids generating javadoc in the build (see line 253 of the pipeline).

Screenshot from 2020-04-14 16-22-08

felixbarny:
Good call, updated to enable the Javadoc pipeline step instead

PR #1148, opened 2020-04-20
Fix system metrics on J9

felixbarny:
fixes https://discuss.elastic.co/t/java-apm-agent-not-pushing-cpu-memory-metric-ibm-java-on-websphere-liberty/228064

If `/proc/meminfo` is available, prefer reading `MemTotal` for `system.memory.total`  rather than `OperatingSystemMXBean#getSystemCpuLoad` for better consistency.

This is in line with how the go agent/metricbeat collects the total memory:
https://github.com/elastic/go-sysinfo/blob/2e4263c5c548ff5352fd86b0a29330fb08b5c98f/providers/linux/memory_linux.go#L40-L41

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/SystemMetrics.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/SystemMetrics.java
@@ -99,24 +99,18 @@ public class SystemMetrics extends AbstractLifecycleListener {
     }
 
     void bindTo(MetricRegistry metricRegistry) {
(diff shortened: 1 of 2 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/SystemMetrics.java
```diff
@@ -99,27 +99,21 @@ public void start(ElasticApmTracer tracer) {
     }
 
     void bindTo(MetricRegistry metricRegistry) {
-        metricRegistry.addUnlessNegative("system.cpu.total.norm.pct", Labels.EMPTY, new DoubleSupplier() {
+        // J9 always returns -1 on the first call
```
eyalkoren, line 102: 🤦

PR #1149, opened 2020-04-21
TLS v1.3 Fallback for JDK bug

### LINKED ISSUE #1080: Force TLS 1.2 for apm-server transport with older JDKs
With the 7.6.0, TLS v1.3 support was added as a acceptable default protocol   https://github.com/elastic/apm-server/issues/3452 revealed an issue with the JDK that has since been fixed but leaves some users with a manual workaround needed during upgrade.

Consider forcing TLS v1.2 for the connection to APM Server under affected java versions.

SylvainJuge:
## What does this PR do?

This PR implements a fallback for #1080 

TLS 1.3 has been added in JDK 11, but due to a JDK bug ([JDK-8236039](https://bugs.openjdk.java.net/browse/JDK-8236039)), it can't be used in affected versions.

By default, there is no fallback on other versions of the protocol, even if those are both available and supported by the remote server.

This PR implements the following fallback:
- applied only when SSL socket handshake fails with a known error type & message pattern and `TLS v1.3` protocol is enabled.
- creates a new socket without `TLS v1.3` and restarts handshake.

When applied, the following warning is present in agent log:
```
WARN co.elastic.apm.agent.report.ssl.TLSFallbackSSLSocket - Workaround for JDK Bug JDK-8236039 applied, will connect without TLS v1.3
```

## Checklist
- ~~I have made corresponding changes to the documentation~~
- ~~I have updated [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)~~

## Author's Checklist
- [x] manual testing with affected JDK (11.0.5)
- [x] manual testing with unaffected JDK (8.0.x)

## Related issues

- Closes #1080


apm-agent-core/src/main/java/co/elastic/apm/agent/report/ssl/TLSFallbackSSLSocket.java
```diff
@@ -0,0 +1,408 @@
...
+        // known JDK bug: JDK-8236039
+        // automatically fallback to another protocol when triggered
+        try {
+            socket.startHandshake();
+        } catch (SSLHandshakeException e) {
+
+            Set<String> enabledProtocols = new HashSet<>(Arrays.asList(socket.getEnabledProtocols()));
+            if (enabledProtocols.contains(TLS_v_1_3) && e.getMessage().contains("should not be presented in")) {
+
+                logger.warn("Workaround for JDK Bug JDK-8236039 applied, will connect without TLS v1.3");
```
felixbarny, line 75: Seems like we're re-triggering this exception for every request to APM Server, right? Could we exclude TLSv1.3 if we know it triggers the bug on the current JVM?

felixbarny: Maybe also suggest the user what to do when the exception occurs, like updating the JDK, or removing 1.3 for the server's setting (https://www.elastic.co/guide/en/apm/server/current/agent-server-ssl.html#_supported_protocols_2).

SylvainJuge: I agree with both comments, having  a single warning is better, and an actionable message is definitely better than having to deal with that in an FAQ somewhere :-).

SylvainJuge: Also, there would be the option to disable TLS 1.3 client side, but that would impact the whole JVM, or provide a configuration for agent, which implies some extra complexity for documentation/support.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ssl/TLSFallbackSSLSocket.java
```diff
@@ -0,0 +1,417 @@
...
+        }
+        try {
+            socket.startHandshake();
+        } catch (SSLHandshakeException e) {
+
+            if (hasTLS13 && e.getMessage().contains("should not be presented in")) {
+
+                boolean hasBeenWarned = skipTLS13.getAndSet(true);
+                if (!hasBeenWarned) {
+                    logger.warn("Workaround for JDK Bug JDK-8236039 applied, will connect without TLS v1.3. Update JRE/JDK to fix this, or disable TLS v1.3 on apm-server as a workaround");
```
felixbarny, line 86:
```suggestion
                    logger.warn("Workaround for JDK Bug JDK-8236039 applied, will connect without TLS v1.3. Update JRE/JDK to fix this, or disable TLS v1.3 on APM Server as a workaround (apm-server.ssl.supported_protocols)");
```

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
```diff
@@ -121,6 +122,11 @@ private void setServerUrls(List<URL> serverUrls) {
         return copy;
     }
 
+    private static void tlsFallback(HttpsURLConnection connection) {
+        SSLSocketFactory newFactory = SslUtils.getTLSFallbackSocketFactory(connection.getSSLSocketFactory());
```
felixbarny, line 126: Why don't we create our own SSLSocketFactory from scratch?

SylvainJuge: You mean, why do we wrap the default `SSLSocketFactory` where we could just have created a new one ?

For me that's just a way to make sure that our implementation does not introduce obvious security bugs or unexpected behaviors here. Also, the default implementation is not public thus we can't inherit from it.

felixbarny: I mean similar to this: https://gist.github.com/fkrauthan/ac8624466a4dee4fd02f

PR #1150, opened 2020-04-21
Add span_min_duration flag

### LINKED ISSUE #1094: Need DB spans duration threshold, similar to trace_methods_duration_threshold
Need DB spans duration threshold, similar to trace_methods_duration_threshold. If the DB spans does not have threshold even the minor spans (1 or 2 micro seconds spans)are reported in APM timeline and messing up the timeline for large transactions.

Please provide configuration parameter for DB spans so that spans above threshold only would be captured and showed in the timeline.

felixbarny:
closes #1094


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
```diff
@@ -79,11 +79,6 @@
     @Nullable
     private String result;
 
-    /**
-     * Noop transactions won't be reported at all, in contrast to non-sampled transactions.
-     */
-    private boolean noop;
```
eyalkoren, line 85: It will be great to get rid of this state, just make sure that the combination of `discardRequested`, `discardable`  and `sampled` provide the same meaning as `noop`.

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/DiscardSpanTest.java
```diff
@@ -0,0 +1,94 @@
...
+import co.elastic.apm.agent.impl.transaction.Transaction;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class DiscardSpanTest {
```
eyalkoren, line 39: Add tests for Errors and for non-sampled transactions

felixbarny:
> However, a risky one- may yield broken traces.

Yeah, I have to agree. For example, if someone uses the API to propagate the context to another thread, we might discard the span even though it's a parent of a longer async operation that starts after the initial span has already finished.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContextHolder.java
```diff
@@ -147,14 +146,43 @@ public boolean isExit() {
...
     }
 
+    /**
+     * Sets Trace context text headers, using this context as parent, on the provided carrier using the provided setter
+     *
+     * @param carrier      the text headers carrier
+     * @param headerSetter a setter implementing the actual addition of headers to the headers carrier
+     * @param <C>          the header carrier type, for example - an HTTP request
+     */
+    public abstract <C> void setOutgoingTraceContextHeaders(C carrier, TextHeaderSetter<C> headerSetter);
```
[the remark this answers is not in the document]
felixbarny: I've tried that but it doesn't seem feasible as we allow to activate a `TraceContextHolder` and we want to be able to do `tracer.getActive().setOutgoingTraceContextHeaders`. This is needed for the public API and the OpenTracing API bridge, currently.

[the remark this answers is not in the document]
eyalkoren: Do you think it is problematic to allow that it OP and API only if `trace.getActive` is `instanceof AbstractSpan`? 
If you don't want to introduce this restriction, I would at least add a very clear javadoc warning saying that implementations are not equal and the `AbstractSpan` overload should be preferred whenever possible.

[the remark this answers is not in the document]
felixbarny: It required some refactoring but I've managed to remove `TraceContextHolder`. Now `TraceContext#setOutgoingTraceContextHeaders` is package private.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
felixbarny: It wasn't equivalent. I've reverted the removal of the noop flag. It's also something only needed for the OT API... So maybe remove in a dedicated PR but this one changes too much already.

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ElasticApmTracerTest.java
```diff
@@ -286,8 +286,10 @@ void testEnableDropSpans() {
             transaction.end();
         }
         assertThat(reporter.getFirstTransaction().isSampled()).isTrue();
-        assertThat(reporter.getFirstTransaction().getSpanCount().getDropped().get()).isEqualTo(1);
-        assertThat(reporter.getFirstTransaction().getSpanCount().getStarted().get()).isEqualTo(1);
+        assertThat(reporter.getFirstTransaction().getSpanCount().getDropped()).hasValue(0);
```
eyalkoren, line 289: This leftover is what fails the test now..

apm-agent-plugins/apm-opentracing-plugin/src/main/java/co/elastic/apm/agent/opentracing/impl/ApmSpanBuilderInstrumentation.java
```diff
@@ -82,22 +82,34 @@ public static void createSpan(@Advice.Argument(value = 0, typing = Assigner.Typi
...
-            return null;
+            if (result != null) {
+                // This reference count never gets decremented, which means it will be handled by GC rather than being recycled.
+                // The OpenTracing API allows interactions with the span, such as span.getTraceContext even after the span has finished
+                // This makes it hard to recycle the span as the life cycle is unclear.
+                // See also https://github.com/opentracing/opentracing-java/issues/312
+                // Previously, we kept a permanent copy of the trace context around and recycled the span on finish.
+                // But that meant lots of complexity in the internal API,
+                // as it had to deal with the fact that a TraceContext might be returned by ElasticApmTracer.getActive.
+                // The complexity doesn't seem worth the OT specific optimization that a bit less memory gets allocated.
```
eyalkoren, line 109: 👍

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -543,28 +531,6 @@ public void recycle(SpanInScopeCallableWrapper<?> wrapper) {
         callableSpanWrapperObjectPool.recycle(wrapper);
     }
 
-    public Runnable wrapRunnable(Runnable delegate, TraceContext traceContext) {
```
eyalkoren, line 546: 👍

felixbarny:
@v1v I got a message regarding a build failure even though the tests are still running 🤔

felixbarny:
@eyalkoren raised some valid concerns:

> I can see why a user may want to see ALL important events (like HTTP or DB), regardless of duration, but still cast a wide net on some package through `trace_methods` and remove fast spans. Similarly with inferred spans.
What do you say we use `span_min_duration` for all thresholds if set, but if the specific ones are set- give them precedence? If we are worried about too-many-config-options, we can make them internal . It’s just that we added the `trace_methos` threshold to address a pain, and now we will force people to accept losing meaningful events in order to use that.

In order to move forward with this, let's postpone whether or not to deprecate or replace or align the naming for the other configuration options. 
Thus, I'll remove the `deprecated` tag and restore the `profiling_inferred_spans_min_duration` flag.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/methodmatching/TraceMethodInstrumentation.java
```diff
@@ -97,8 +93,8 @@ public static void onMethodExit(@Advice.Local("span") @Nullable AbstractSpan<?>
             final long endTime = span.getTraceContext().getClock().getEpochMicros();
             if (span instanceof Span) {
                 long durationMicros = endTime - span.getTimestamp();
-                if (traceMethodThresholdMicros <= 0 || durationMicros >= traceMethodThresholdMicros || t != null) {
-                    span.setDiscard(false);
+                if (traceMethodThresholdMicros > 0 && durationMicros < traceMethodThresholdMicros && t == null) {
```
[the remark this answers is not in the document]
felixbarny: Removed the deprecated flag (see https://github.com/elastic/apm-agent-java/pull/1150#issuecomment-621825177). The `span_min_duration` flag is checked in `ElasticApmTracer#endSpan`

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ActivationListener.java
```diff
@@ -24,34 +24,50 @@
...
+    void beforeActivate(AbstractSpan<?> span) throws Throwable;
 
     /**
-     * A callback for {@link TraceContextHolder#deactivate()}
+     * A callback for {@link ErrorCapture#activate()}
+     *
+     * @param error the {@link ErrorCapture} that is being activated
+     * @throws Throwable if there was an error while calling this method
+     */
+    void beforeActivate(ErrorCapture error) throws Throwable;
```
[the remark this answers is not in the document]
felixbarny: Yes, that's true. I chose this name to be consistent with spans, but I don't have a strong option about it.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -334,26 +406,56 @@ public final void end(long epochMicros) {
...
-    @Override
-    public boolean isChildOf(TraceContextHolder other) {
-        return getTraceContext().isChildOf(other);
-    }
-
-    @Override
     public T activate() {
         incrementReferences();
-        return super.activate();
+        List<ActivationListener> activationListeners = tracer.getActivationListeners();
```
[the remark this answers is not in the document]
felixbarny: Yes, this way there's no chance to skip activation listeners by directly calling `ElasticApmTracer#endSpan`

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/DiscardSpanTest.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        assertThat(transaction.getSpanCount().getTotal()).hasValue(1);
+        assertThat(transaction.getSpanCount().getDropped()).hasValue(0);
+        assertThat(transaction.getSpanCount().getReported()).hasValue(1);
+    }
+
+    @Test
+    void testErrorCapturingMakesSpansNonDiscardable() {
+        when(tracer.getConfigurationRegistry().getConfig(CoreConfiguration.class).getTransactionMaxSpans()).thenReturn(2);
```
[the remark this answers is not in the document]
felixbarny: !

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/SamplingProfiler.java
```diff
@@ -733,7 +735,7 @@ private void stopProfiling(SamplingProfiler samplingProfiler) {
                     logger.debug("End call tree ({}) for thread {}", deserialize(samplingProfiler, traceContextBuffer), threadId);
                 }
                 samplingProfiler.profiledThreads.remove(threadId);
-                callTree.end(samplingProfiler.callTreePool, samplingProfiler.config.getInferredSpansMinDuration().getMillis() * 1_000_000);
+                callTree.end(samplingProfiler.callTreePool, samplingProfiler.coreConfig.getSpanMinDuration().getMillis() * 1_000_000);
```
eyalkoren, line 738: The greater between them?

CHANGELOG.asciidoc
```diff
@@ -37,7 +36,8 @@ This is not a breaking change, so former
 * Add workaround for JDK bug JDK-8236039 with TLS 1.3 {pull}1149[#1149]
 * Add log level `OFF` to silence agent logging
 * Adds <<config-span-min-duration,`span_min_duration`>> option to exclude fast executing spans.
-  This deprecates `trace_methods_duration_threshold` and removes `profiling_inferred_spans_min_duration`.
+  When set together with one of the more specific thresholds - `trace_methods_duration_threshold` or `profiling_inferred_spans_min_duration`,
+  the higher threshold will determine spans will be discarded.
```
eyalkoren, line 40:
```suggestion
  the higher threshold will determine which spans will be discarded.
```

apm-agent-core/src/test/java/co/elastic/apm/agent/bci/methodmatching/TraceMethodInstrumentationTest.java
```diff
@@ -184,24 +189,50 @@ void testAgentPaused() {
...
     void testDiscardMethods_DiscardAll() {
         new TestDiscardableMethods(tracer).root(false);
         assertThat(reporter.getTransactions()).hasSize(1);
         assertThat(reporter.getSpans()).hasSize(0);
     }
 
     @Test
-    @Tag("200ms")
+    @Tag("span_min_duration=50ms")
+    @Tag("trace_methods_duration_threshold=200ms")
```
eyalkoren, line 201: 😍

apm-agent-plugins/apm-java-concurrent-plugin/src/test/java/co/elastic/apm/agent/concurrent/AsyncTraceMethodInstrumentationTest.java
```diff
@@ -85,15 +90,24 @@ void testWithDefaultConfig() {
...
+    @Tag("span_min_duration=50ms")
+    @Tag("trace_methods_duration_threshold=200ms")
+    void testWithHigherSpecificThreshold() {
+        new TestAsyncTraceMethodsClass().invokeAsync();
+        assertThat(reporter.getTransactions()).hasSize(1);
+        assertThat(reporter.getSpans()).hasSize(2);
+    }
+
+    @Test
+    @Tag("span_min_duration=50ms")
```
eyalkoren, line 110: I'd add another test with only `@Tag("trace_methods_duration_threshold=50ms")`

PR #1153, opened 2020-04-23
ci(dependabot): configuration file

v1v:
If we would like to enable the dependabot automation process to bump the version of the packages automatically, then it's required this particular file.

### How does it work?

There is a weekly pipeline that will care for all the projects with this particular file `ci/dependabot.yml` and if so it will run the validation and PR creations.

### For instance

I ran the below command locally

```bash
docker run -ti \
    --rm -e GITHUB_ACCESS_TOKEN=${GITHUB_ACCESS_TOKEN} \
    -e PROJECT_PATH=v1v/apm-agent-java \
    -e PACKAGE_MANAGER=maven \
    docker.elastic.co/observability-ci/dependabot
```

and caused the below PRs in my forked repo:

https://github.com/v1v/apm-agent-java/pulls?q=is%3Apr+is%3Aclosed+label%3Adependencies

### Actions

- [ ] Verify if this particular dependabot is a good candidate to be enabled in this project.
- [ ] Assign?

### DIFF
--- a/.ci/dependabot.yml
+++ b/.ci/dependabot.yml
@@ -1,0 +1,5 @@
+### The list of GitHub accounts that will be used to assign the PRs to (no space and comma separated)
+assign: ''
+
+### The package manager to be used (no space and comma separated if more than one package)
+manager: maven


felixbarny:
Most of our dependencies are test dependencies. For some of them, we already specify a version range in the `pom.xml` to always test the latest version.

That strategy doesn't work well for production dependencies as we want to have reproducible build artifacts. This is where dependabot could help us.

Is there a way to exclude or include certain dependencies that dependabot should keep up-to-date?  While we'd like to keep core dependencies, like Byte Buddy up-to-date, we don't want to update the dependencies for tests that are supposed to verify that our agent works fine when using with an old spring version, for example.

v1v:
> Is there a way to exclude or include certain dependencies that dependabot should keep up-to-date? While we'd like to keep core dependencies, like Byte Buddy up-to-date, we don't want to update the dependencies for tests that are supposed to verify that our agent works fine when using with an old spring version, for example.

Thanks Felix, I'll add this particular requirement to the list of use cases that are required/wished. I'll close this PR now for the time being, as we are still gathering details about what's the tool to be used.

Issue #550, opened 2019-03-28
Logging implementation is fixed


OrangeDog:
This project uses slf4j for logging, but appears to fix the implementation to `SimpleLogger` instead of what the host application has on its classpath.

This largely defeats the point of using slf4j, and severely reduces the ability to configure logging.

felixbarny:
The problem is that the Java agent is not like any other library. It contains all the code, including dependencies and a logging implementation. The reason for that is that it runs inside the system/bootstrap class loader and we can't assume the instrumented application actually uses slf4j. Even if slf4j is used, it could be a different version, causing version conflicts. We also shade (i.e. relocate) slf4j from `org.slf4j` to `co.elastic.apm.shaded.slf4j` to avoid that these classes conflict with the classes in your application and also to avoid them being present on the classpath twice.

I can understand that it's a bit annoying to not be able to configure logging in the usual way but I'm afraid there's no real way around that.

However, we do have plans to offer more options to configure the logging. See also https://github.com/elastic/apm-agent-java/issues/442. What options are you currently missing and how would you like to configure the logging?

OrangeDog:
I want to send the logs via https://github.com/logstash/logstash-logback-encoder, customise the log format, and manage loggers over JMX.

Since it's a classloader agent anyway, can you not detect the presence of SLF4J bindings and only add `SimpleLogger` if there isn't one?
Alternatively, can you provide an alternate artefact that relies on the classpath entirely for logging?

I'm using the `ElasticApmAttacher` in a Spring Boot application, tied to the `ApplicationEnvironmentPreparedEvent`, so I can load the agent properties via the Spring environment.

felixbarny:
That's currently not easily possible as the `ElasticApmAttacher` just attaches the regular java agent which contains all of the shaded dependencies, including the shaded slf4j. But in theory, it's something that is possible. I'll give it some thought.

OrangeDog:
Note you do already have a dependency on the application's own slf4j, for `enable_log_correlation`.

felixbarny:
Yes, but that does not have a direct dependency but rather uses `MethodHandle`s for decoupling. But maybe we could do integration with an application own slf4j in a similar way 🤔.

felixbarny:
@MrThreepwood would you expect the agent logs to be in a separate file or in the same log file as your main application?

OrangeDog:
Not who you asked, but I would expect to be able to choose, via my application’s logging configuration.

[...]

Issue #442, opened 2019-01-24
Implement log rotation 


felixbarny:

gitck:
I would really appreciate this feature, since currently there is no way to disable logging of the agent (there is no support for log_level=OFF). If the apm server is not available, the log file is flooded since every connection timeout problem is logged as an ERROR (see example below). 
Is there any progress on this issue?

--------------------------------------------------------------------
2020-03-20 08:52:47.078 [elastic-apm-remote-config-poller] ERROR co.elastic.apm.agent.configuration.ApmServerConfigurationSource - connect timed out
2020-03-20 08:53:11.643 [elastic-apm-server-reporter] ERROR co.elastic.apm.agent.report.IntakeV2ReportingEventHandler - Failed to handle event of type METRICS with this error: connect timed out
--------------------------------------------------------------------

felixbarny:
Not yet, but we definitely need an overhaul of the logging system soon.

PR #1154, opened 2020-04-23
Allow logs to be redirected to /dev/null

msievers:
## What does this PR do?

The documentation of the APM agent states that [the logging output can be redirected to a file](https://www.elastic.co/guide/en/apm/agent/java/master/config-logging.html#config-log-file). Currently, this does not work when specifying `/dev/null` to silence the APM logs completely. When trying to do so, the agent complains with

```
Log file /dev/null is not writable. Falling back to System.out.
```

## Checklist

## Author's Checklist

dir exists | dir writeable | file exists | file writeable | action
---------- | ---           | ---         | ---            | ---
false      | -             | -           | -              | Create directory and file
true       | false         | false       | -              | Print error, fallback to STDOUT
true       | false         | true        | false          | Print error, fallback to STDOUT
true       | false         | true        | true           | Use given file (e.g. `/dev/null`)
true       | true          | false       | -              | Create new file
true       | true          | true        | false          | Print error, fallback to STDOUT
true       | true          | true        | true           | Use given file

## Related issues

* https://github.com/elastic/apm-agent-java/issues/550
* https://github.com/elastic/apm-agent-java/issues/442#issuecomment-602042731

## Use cases

There are companies that have very strict requirements regarding application logs/log formats which may prevent the usage of Elastic APM agent, because there is currently no way, to configure/alter the log format, for example to output JSON.

As a workaround, it should be possible to silence the APM agent by redirecting its logs to `/dev/null`.

## Screenshots
<!-- _(Optional)_

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
@@ -34,6 +34,9 @@ import org.stagemonitor.configuration.source.ConfigurationSource;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.List;
 
 /**
@@ -169,11 +172,17 @@ public class LoggingConfiguration extends ConfigurationOptionProvider {
         if (!logDir.exists()) {
             logDir.mkdir();
         }
-        if (!logDir.canWrite()) {
+
+        final Path logFilePath = Paths.get(logFile);
+        if (Files.exists(logFilePath) && !Files.isWritable(logFilePath) || !Files.exists(logFilePath) && !logDir.canWrite()) {
             System.err.println("Log file " + logFile + " is not writable. Falling back to System.out.");
             return SYSTEM_OUT;
         }
-        System.out.println("Writing Elastic APM logs to " + logFile);
+
+        if(!logFile.equalsIgnoreCase("/dev/null")) {
+            System.out.println("Writing Elastic APM logs to " + logFile);
+        }
+
         return logFile;
     }
 
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -45,6 +45,7 @@ This is not a breaking change, so former
 * Fix flaky gRPC server instrumentation {pull}1122[#1122]
 * Fix side effect of calling `Statement.getUpdateCount` more than once {pull}1139[#1139]
 * Stop capturing JDBC affected rows count using `Statement.getUpdateCount` to prevent unreliable side-effects {pull}1147[#1147]
+* Allow logs to be redirected to `/dev/null` {pull}1154[#1154]
 
 
 [[release-notes-1.x]]


apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
```diff
@@ -169,11 +171,15 @@ static String getActualLogFile(@Nullable String agentHome, String logFile) {
...
+        if (!Files.isWritable(Paths.get(logFile)) && !logDir.canWrite()) {
             System.err.println("Log file " + logFile + " is not writable. Falling back to System.out.");
             return SYSTEM_OUT;
         }
-        System.out.println("Writing Elastic APM logs to " + logFile);
+
+        if(!logFile.equalsIgnoreCase("/dev/null")) {
+            System.out.println("Writing Elastic APM logs to " + logFile);
+        }
+
```
msievers, line 182: Although this isn't very elegant, it seems to be the only way to retain the actual behavior while making the agent completely silent in case of `/dev/null`.

felixbarny:
> There are companies that have very strict requirements regarding application logs/log formats which may prevent the usage of Elastic APM agent, because there is currently no way, to configure/alter the log format, for example to output JSON.

You might be happy to hear that we're experimenting with some things regarding logging here.
Firstly, we make it possible to set the log format to `JSON` which uses https://github.com/elastic/ecs-logging-java to format the logs.
Additionally, the JSON logs can be automatically shipped to APM Server via the agent. It's of course also possible to do this manually by shipping the log with Filebeat, for example, or to log JSON to std out which can be useful in a container-based setup.

Out of curiosity, why can't just log to a file and ignore it? That way you at least have the option to manually look at the file if you need to.

msievers:
We tried with logging to a file first, but for instance, in case of errors, the agent logs a huge amount of messages. So the fear is, that this may silently flood our storages.

Actually, the problem why it refuses to log to /dev/null is simply the writeable check on the parent directory, which for instance would also prevent you to log into a file you are allowed to write to, but if the directory is for instance not writeable.

We are currently facing the problem, that our OPs department want's us to shut down all agents if we don't get the logging right or silence to agent at all. That's why I would really appreciate it if we could allow logging to /dev/null as a workaround until some more sophisticated solution is available.

msievers:
Btw, I just found a combination of directory/file permission where the MR fails, so I will push a fix together with a list of tested combinations (dir exists/writeable/file exists/writeable within the next hours).

felixbarny:
> We tried with logging to a file first, but for instance, in case of errors, the agent logs a huge amount of messages. So the fear is, that this may silently flood our storages.

This is also something we're tackling. You'll be able to configure a max log file size.

In the meantime, I think your fix can't do harm.

msievers:
Nice, thank you :tada:  I'll ping you when I finished testing all possible combinations and updated the pull requests description.

msievers:
@felixbarny I just double-checked all possible combinations of directory/file existence/writeability and wrote them down in the pull requests description under `Author's Checklist`.

eyalkoren:
@msievers Thanks! Your permissions-result matrix in the Author's checklist is 👍 
I did a [tests PR](https://github.com/msievers/apm-agent-java/pull/1), but wait for the conclusion of the following discussion before merging.

@felixbarny I think this wouldn't work with the new logging we are about to add - since we are going to use rolling logs and the state file, the directory probably needs to be writeable. 
It may make it a breaking change to @msievers when we migrate to it.
Nevertheless, we need to offer a way to make the agent silent, maybe through configuration?

eyalkoren:
@msievers we discussed that - while your solution works great now, we are soon to introduce a very useful enhancement to the agent logging that will rely on the directory being writeable.

However, what you really need is a silent agent, so let's do that 🙂. We thought of adding an `OFF` option to the [`log_level`](https://www.elastic.co/guide/en/apm/agent/java/current/config-logging.html#config-log-level) config to facilitate that. Would you like to do that?

msievers:
@eyalkoren Sure, I can give it a try. I think I can report back by the end of the week :+1:

msievers:
@eyalkoren I couldn't keep my fingers from this and created a new PR to add `OFF` as log level. I will close this PR in favour of https://github.com/elastic/apm-agent-java/pull/1160

PR #1156, opened 2020-04-24
Update gRPC test version to 1.29.0

SylvainJuge:
## What does this PR do?

Updates gRPC version used for testing to the latest available version 1.29.0.

There is strictly no change introduced in generated code when compared to the previous version (1.27.1).

## Checklist
- [x] My code follows the [style guidelines of this project](CONTRIBUTING.md#java-language-formatting-guidelines)
- [x] I have rebased my changes on top of the latest master branch
- ~~I have made corresponding changes to the documentation~~
- ~~I have added tests that prove my fix is effective or that my feature works~~
- [x] New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#testing) pass locally with my changes
- ~~I have updated [CHANGELOG.asciidoc](CHANGELOG.asciidoc)~~
- [x] I have updated [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)
- ~~Added an API method or config option? Document in which version this will be introduced~~
- ~~Added an instrumentation plugin? How did you make sure that old, non-supported versions are not instrumented by accident?~~

## Related issues

- Closes #1152

PR #1157, opened 2020-04-24
test gRPC with latest released version

SylvainJuge:
## What does this PR do?

Whenever there is a new gRPC version released, we have to update our agent build to ensure it's properly working as expected, this has quite a few drawbacks
- does not test latest version
- intermediate versions might not be properly tested
- requires to update project: PR/review/build while it's a simple and minimal change.

This change will make build modify `pom.xml` to use the latest grpc dependency on every build.
This implies that doing a release (that requires no local modifications) will fail unless the change has been commited.

## Checklist
- ~~I have added tests that prove my fix is effective or that my feature works~~
- ~~I have updated [CHANGELOG.asciidoc](CHANGELOG.asciidoc)~~

## Author's Checklist

- [x] testing locally by simulating a `1.27` -> `1.29` update

## Related issues

#1156 : updating to latest version (1.29.0 as of today).


apm-agent-plugins/apm-grpc/apm-grpc-test-latest/pom.xml
```diff
@@ -21,6 +21,30 @@
...
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>versions-maven-plugin</artifactId>
+                <version>2.7</version>
+                <executions>
+                    <execution>
+                        <phase>initialize</phase>
+                        <goals>
+                            <!-- because grpc version is set using a shared property -->
+                            <goal>update-properties</goal>
```
felixbarny, line 35: When setting `grpc.version` to `[1.29.0,)`, do we still need the `versions-maven-plugin`?

SylvainJuge: Yes, it's even simpler without the plugin.

The only difference is that we don't have to update the pom.xml as it automatically (and silently) picks the latest version.

PR #1163, opened 2020-04-29
Fix NoClassDefFoundError in gRPC instrumentation

felixbarny:
Reported here: https://discuss.elastic.co/t/elastic-apm-cannot-collect-grpc-requests/230086/8


felixbarny:
User reported this fixed the problem

PR #1167, opened 2020-04-30
Snapshots

cachedout:
## What does this PR do?
During a test run on the master branch, if the tests pass, then we continue to this stage which pushes a snapshot to the artifact repository. 

## Checklist

- [ ] ~~New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#testing) pass locally with my changes~~

## Related issues

Closes https://github.com/elastic/apm-agent-java/issues/1056

## Points to consider 

I modified the Maven command so that it includes a `-DskipTests` because we've already run them in this case but I just want to make sure that's OK with the team since it's a change from how snapshots are currently deployed.

I am submitting this as a draft because I think there's still some discussion surrounding this. I will fill in more details over at https://github.com/elastic/apm-agent-java/issues/1056

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -298,6 +298,20 @@ pipeline {
         githubNotify(context: "${env.GITHUB_CHECK_ITS_NAME}", description: "${env.GITHUB_CHECK_ITS_NAME} ...", status: 'PENDING', targetUrl: "${env.JENKINS_URL}search/?q=${env.ITS_PIPELINE.replaceAll('/','+')}")
       }
     }
+    stage('Snapshots') {
+      when {
+        allOf {
+          expression { return env.ONLY_DOCS == "false" }
+          anyOf {
+            branch 'master'
+            expression { return params.Run_As_Master_Branch }
+          }
+        }
+      }
+      steps {
+        doSnapshot()
+      }
+    }
     stage('AfterRelease') {
       options { skipDefaultCheckout() }
       when {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -298,6 +298,20 @@ pipeline {
...
+        allOf {
+          expression { return env.ONLY_DOCS == "false" }
+          anyOf {
+            branch 'master'
+            expression { return params.Run_As_Master_Branch }
+          }
+        }
+      }
+      steps {
+        doSnapshot()
```
v1v, line 312: Run in the `BASE_DIR` context

```suggestion
        dir("${BASE_DIR}") {
          doSnapshot()
        }
```

scripts/jenkins/snap.sh
```diff
@@ -0,0 +1,33 @@
...
+set +x
+
+# Import the key into the keyring
+echo $KEYPASS | gpg --batch --import keyfile > /dev/null 2>&1
+
+set -x
+
+# Deploy the snapshot
+# Sensitive data may be logged to stderr. Remove for debugging.
+$BASE_DIR/mvnw -s .ci/settings.xml -Pgpg clean deploy -DskipTests --batch-mode 2>/dev/null
```
v1v, line 33: `BASE_DIR` context at the pipeline level, otherwise `-c .ci/settings.xml` might not work as it's relative to the `BASE_DIR` context:

```suggestion
mvnw -s .ci/settings.xml -Pgpg clean deploy -DskipTests --batch-mode 2>/dev/null
```

cachedout: Ah, I didn't realize that the context doesn't propagate.. Thanks for finding this!

scripts/jenkins/snap.sh
```diff
@@ -0,0 +1,33 @@
...
+    rm -rf $WORKSPACE@tmp/keyring
+    echo "$WORKSPACE@tmp/keyring deleted"
+    exit $ARG
+}
+trap clean_up EXIT
+
+set +x
+
+# Import the key into the keyring
+echo $KEYPASS | gpg --batch --import keyfile > /dev/null 2>&1
```
mdelapenya, line 27: Where does this `keyfile` value come?

cachedout: Guh. You're right. I updated this and forgot to change this variable. Thanks for catching this. Will fix.

cachedout:
Hi @SylvainJuge . In case you missed it, we're just waiting for a review on this change. Not urgent at all though. :) Thanks!

felixbarny:
Why are the nightly snapshots plus the snapshot builds for PR artifacts not good enough?

I'm not sure it's worth the additional risk of exposing credentials. I think it's also important that the snapshot build is performed in a very similar way than the actual release (see also https://github.com/elastic/apm-agent-java/issues/1056#issuecomment-600636664).

SylvainJuge:
After discussing this a bit with @felixbarny and @eyalkoren , we don't think this is a good idea to move this forward as-is.

The initial idea (https://github.com/elastic/apm-agent-java/issues/1056) was to cut down build time and avoid duplicate build executions, which made sense if everything was executed in a single CI system.

For now, we think that keeping things as-is (with snapshots executed daily on internal CI), is a better option for the following reasons:
1. using maven repository credentials on the public CI is dangerous and could lead to a credentials leak
1. while close, both CIs aren't 100% identical and it often happens that something fails on one, but not on the other. Because we want to prevent discovering issues at release time, having a snapshot build daily is a good proxy to detecting potential issues early.
1. When an ad-hoc snapshot is needed, for example to help diagnose a customer issue, using PR artifacts links to the public CI can be used.

cachedout:
Hi @SylvainJuge . Thanks for taking the time to write this summary. I'm totally fine with this and agree with all of your conclusions. I'll go ahead and close this unmerged.

PR #1168, opened 2020-04-30
Adding support for new Atlassian OSGi config

### LINKED ISSUE #1164: Confluence 7.4.0 fails start with elastic-agent-java 1.15.0
## Describe the bug
Confluence 7.4.0 fails start with elastic-agent-java 1.15.0
agent start with this options:

-javaagent:C:\elastic\elastic-apm-agent-1.15.0.jar
-Delastic.apm.service_name=confl_local_test_v7
-Delastic.apm.application_packages=org.example,org.another.example 
-Delastic.apm.server_urls=http://elastic:8200

Error in browser when try open Confluence:
[confl_error_in_browser.txt](https://github.com/elastic/apm-agent-java/files/4552257/confl_error_in_browser.txt)

## Steps to reproduce
Start confluence 7.4 with apm agent

## Debug logs
Confluence startup log in file:
[atlassian-confluence.log](https://github.com/elastic/apm-agent-java/files/4552300/atlassian-confluence.log)

eyalkoren:
## What does this PR do?

Using the new and superior Atlassian OSGi config. I added an internal config to determine which to use and set its default to use the new config, however this means we are not setting the old one (which requires appending to all default packages), which means it may be a breaking change for existing users, so we will decide about the default later. Currently this is only for analysis.
## Checklist

## Related issues

Closes #1164


eyalkoren:
@felixbarny the current state of this PR solves the reported issue (#1164), however it is a breaking change for users of the older Attlasian versions, because the current default setting of the new config option I added sets the `.extra` property and does not set the default one.
I think we should merge with the default to behave as older agent versions, and then I see two options:
1.  Just tell this user (and future users reporting the same) to actively set this config (probably also document something about it)
2. Try adding the missing packages that are required in newer Atlassian versions (at least `jdk.` according to the issue). The risk with that is that we can't be certain it wouldn't affect older versions.
WDYT?

felixbarny:
Is there any way we can detect at runtime whether the `.extra` property should be used? In which version was the `.extra` property introduced?

eyalkoren:
There probably is, but I wouldn't invest too much on it (needs to be tested on multiple versions), especially since we are going to address the OSGi issue in general.
The user that reported the related issue now [reported](https://github.com/elastic/apm-agent-java/issues/1164#issuecomment-624579762) that adding the `jdk.internal` package to the legacy atlassian bootdelegation config did the trick, so I am thinking on merging it as is, and have the `extra` option available for us through the internal config if we need it in future versions.

felixbarny:
Makes sense 👍

PR #1170, opened 2020-05-01
Automatic tracing for Jobs defined in quartz-jobs dependency

theobisproject:
## What does this PR do?

While going through our traced Quartz schedules I noticed that the directory scanning schedules are missing which we implement via [quartz-jobs](https://github.com/quartz-scheduler/quartz/tree/master/quartz-jobs) artifact of the quartz project.  
In this use case we do not implement a quartz job but only a callback interface the job will call. As described in the documentation I would have to add the `org.quartz.jobs` package currently to the `application_packages` to make this work.  

In my opinion it would be more user friendly when the jobs provided by the quartz project would work without extra configuration.

This PR add's the instrumentation of the job from the `quartz-jobs` dependency. I decided to not add the package name to the `applicationPackages` collection because the collection we get is immutable.

## Checklist

## Use cases

The `quartz-jobs` dependency provides ready to use jobs for the following use-cases

- Scanning directories for changed files via lastModified date
- Scanning files for changes via lastModified date
- Executing native executables
- Sending mails
- Invoke JMX
- Sending messages to JMS
- Invoke EJB

## Screenshots
<!-- _(Optional)_

### DIFF
--- a/apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentation.java
+++ b/apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentation.java
@@ -40,6 +40,7 @@ import static co.elastic.apm.agent.bci.bytebuddy.CustomElementMatchers.isInAnyPa
 import static net.bytebuddy.matcher.ElementMatchers.declaresMethod;
 import static net.bytebuddy.matcher.ElementMatchers.hasSuperType;
 import static net.bytebuddy.matcher.ElementMatchers.isInterface;
+import static net.bytebuddy.matcher.ElementMatchers.nameStartsWith;
 import static net.bytebuddy.matcher.ElementMatchers.named;
 import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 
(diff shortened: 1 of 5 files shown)


eyalkoren:
Looks good! Thanks @theobisproject!
Please update the CHANGELOG and the supported technologies with a minimal sentence (maybe updating the existing `NOTE: only classes from packages configured in application_packages will be instrumented.`)

theobisproject:
I will update both. Should the sentence for the CHANGELOG be in the `feature` section?

apm-agent-plugins/apm-quartz-job-plugin/src/main/java/co/elastic/apm/agent/quartz/job/JobTransactionNameInstrumentation.java
```diff
@@ -56,6 +57,7 @@ public JobTransactionNameInstrumentation(ElasticApmTracer tracer) {
     @Override
     public ElementMatcher<? super TypeDescription> getTypeMatcher() {
         return isInAnyPackage(applicationPackages, ElementMatchers.<NamedElement>none())
+            .or(nameStartsWith("org.quartz.job"))
```
eyalkoren, line 60: This works, but I prefer the boolean expression more explicit and clearer for reading:
```suggestion
        (isInAnyPackage(applicationPackages, ElementMatchers.<NamedElement>none()).or(nameStartsWith("org.quartz.job")))
```
Not insisting on it though.

Issue #1003, opened 2020-01-22
Create a plugin for Grails framework


SylvainJuge:
## Is your feature request related to a problem?

Currently [Grails framework](https://grails.org/) is not natively supported by our agent.

Grails relies on Servlets, and all transactions are merged together, whatever the actual Grails controler actually handles the request since they have the same entry point.

While providing a better naming for transactions by using an `Interceptor` class as [explained here](https://github.com/elastic/apm-contrib/tree/master/apm-agent-java/frameworks/grails3) is easy, it requires to add an extra dependency and to modify the application.

## Describe the solution you'd like

A native plugin that would provide the best experience out of the box.
This plugin would provide a simple transaction naming strategy, probably the same as the one  [described here](https://github.com/elastic/apm-contrib/tree/master/apm-agent-java/frameworks/grails3)

## Expected challenges

- multiple versions of the framework needs to be supported: 2.x, 3.x and 4.x
- extensive testing with an integration test might be tedious and require to create multiple test webapps, manual testing might be acceptable given the simplicity at what this plugin will do (just providing transaction naming).

kananindzya:
Hi @SylvainJuge , currently I am trying to writie instrumentation for grails framework. I wrote working instrumentation for grail 4.0.1 version and tested with grails application. Transaction name is set correctly.

But I have problem with writing integration test: grails 4 version uses gradle. 

![image](https://user-images.githubusercontent.com/48118512/80863663-b1dd9900-8c9f-11ea-8bf8-9a525762fdef.png)

felixbarny:
> But I have problem with writing integration test: grails 4 version uses gradle.

Instead of writing a full integration test, could you just manually call `GrailsDispatcherServlet` in a unit test, using mocked requests and responses?

PR #1171, opened 2020-05-02
Name transactions based on Grails controllers

kananindzya:
## What does this PR do?
closes elastic/apm-agent-java#1003 

## Checklist

- [ ] Added an instrumentation plugin? How did you make sure that old, non-supported versions are not instrumented by accident? By adding a matcher on a class documented as added in version 3 (`classLoaderCanLoadClass("org.grails.web.mapping.mvc.GrailsControllerUrlMappingInfo")`)

## Screenshots
<!-- _(Optional)_


apm-agent-plugins/apm-grails-plugin/src/main/java/co/elastic/apm/agent/grails/GrailsTransactionNameInstrumentation.java
```diff
@@ -0,0 +1,137 @@
...
+    public Collection<String> getInstrumentationGroupNames() {
+        return Collections.singletonList("grails");
+    }
+
+    @VisibleForAdvice
+    public static class HandlerAdapterAdvice {
+
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        static void setTransactionName(@Advice.Argument(2) Object handler) {
+            if (tracer != null) {
```
SylvainJuge, line 96: We try to keep advice code minimal, thus moving most of the logic to `setName` method would be better here.

Two minor code style changes could be done here:
- Accessing logger directly from the advice could be removed as it does not provide much value (`null` transaction or tracer will not help to debug, but changing the transaction name will).
- using shortcut `return` statements helps to avoid numerous nested if statements, which helps readability (while it's not a very general "good code" practice, we tend to use it in advice code to make it easier to read).
- generating an exception to get the stack trace will likely not be required in production, thus we could remove it also.

apm-agent-plugins/apm-grails-plugin/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+
+    <properties>
+        <apm-agent-parent.base.dir>${project.basedir}/../..</apm-agent-parent.base.dir>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.grails</groupId>
+            <artifactId>grails-web-mvc</artifactId>
+            <version>[4.0.1,)</version>
```
SylvainJuge, line 24: in supported versions in doc, you've written it was working for versions 3+, and here we test only with 4.0+, thus I guess you've tested with both 3.x and 4.0.x. What happens with 2.x ? Is it supported or is it just ignored by this instrumentation (for example it does not have the same classes/interfaces instrumented by this plugin) ?

CHANGELOG.asciidoc
```diff
@@ -33,6 +33,7 @@ There will always be one history file `${log_file}.1`.
 * Add <<config-log-format-sout>> and <<config-log-format-file>> with the options `PLAIN_TEXT` and `JSON`.
 The latter uses https://github.com/elastic/ecs-logging-java[ecs-logging-java] to format the logs.
 * Exposing <<config-classes-excluded-from-instrumentation>> config - {pull}1187[#1187]
+* Add support for naming transactions based on Grails controllers
```
SylvainJuge, line 36: adding a version range in changelog would be great just to set user expectations.

apm-agent-plugins/apm-grails-plugin/pom.xml
felixbarny: > thus I guess you've tested with both 3.x and 4.0.x

Yes, I've tested that manually. The reason I didn't set version `[3.0.0,)` is that maven would download every single minor and bugfix version between 3.0.0 and now which would take forever.

> What happens with 2.x ?

Grails 2.x is excluded via the `classLoaderCanLoadClass("org.grails.web.mapping.mvc.GrailsControllerUrlMappingInfo")` matcher

apm-agent-plugins/apm-grails-plugin/src/main/java/co/elastic/apm/agent/grails/GrailsTransactionNameInstrumentation.java
felixbarny: > We try to keep advice code minimal, thus moving most of the logic to setName method would be better here.

Not sure if that's possible in this case. We can only refer to Grails and Spring types within the advice.

I agree that the layout of the code and the loggers are not ideal but it's adapted from `co.elastic.apm.agent.spring.webmvc.SpringTransactionNameInstrumentation.HandlerAdapterAdvice#setTransactionName`. So I suggest to either fix both here or just do a followup with some cleanup, for consistency.

felixbarny: I have removed the loggers and decreased nesting by adding early returns

PR #1172, opened 2020-05-04
Keep spans active during AsyncHandler methods in AsyncHttpClient

milanvdm:
## What does this PR do?
Currently, when using a custom `AsyncHandler` with the `AsyncHttpClient`, there is no active span in the callback methods.
This PR makes sure that span are activated when entering those methods making it possible for users to use tracing-context in those methods.

## Checklist


apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
```diff
@@ -49,11 +46,32 @@ public AsyncHttpClientInstrumentationTest(RequestExecutor requestExecutor) {
         this.requestExecutor = requestExecutor;
     }
 
+    public static AsyncHandler<Response> customAsyncHandler = new AsyncCompletionHandler<Response>() {
+        @Override
+        public State onStatusReceived(HttpResponseStatus responseStatus) {
+            assert(tracer.currentTransaction() != null);
```
milanvdm, line 52: Not sure how to add a proper test making sure the context in the methods are active or not.

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java
```diff
@@ -179,6 +179,14 @@ public AsyncHandlerOnCompletedInstrumentation() {
 
         @Advice.OnMethodEnter(suppress = Throwable.class)
         private static void onMethodEnter(@Advice.This AsyncHandler<?> asyncHandler) {
+            final Span span = handlerSpanMap.getIfPresent(asyncHandler);
```
felixbarny, line 182: remove from the map and assign to the span paramenter

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java
```diff
@@ -179,6 +179,14 @@ public AsyncHandlerOnCompletedInstrumentation() {
 
         @Advice.OnMethodEnter(suppress = Throwable.class)
         private static void onMethodEnter(@Advice.This AsyncHandler<?> asyncHandler) {
```
felixbarny, line 181: Making sure that wrapped `AsyncHandler`s work as expected (activate and deactivate only in outer-most `AsyncHandler`
```suggestion
        private static void onMethodEnter(@Advice.This AsyncHandler<?> asyncHandler, @Advice.Local("span") Span span) {
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java
```diff
@@ -179,6 +179,14 @@ public AsyncHandlerOnCompletedInstrumentation() {
...
+            final Span span = handlerSpanMap.getIfPresent(asyncHandler);
+            if (span != null) {
+                span.activate();
+            }
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class)
+        private static void onMethodExit(@Advice.This AsyncHandler<?> asyncHandler) {
             final Span span = handlerSpanMap.remove(asyncHandler);
```
felixbarny, line 190:
```suggestion
        private static void onMethodExit(@Advice.This AsyncHandler<?> asyncHandler, @Nullable @Advice.Local("span") Span span) {
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
felixbarny, line 52: Use AssertJ, for consistency

Add static import for `org.assertj.core.api.Assertions.assertThat`

```suggestion
            assertThat(tracer.activeSpan()).isNotNull();
            assertThat(tracer.isExit()).isTrue();
```

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java
milanvdm: Bit unsure on what you mean with activate and deactive in the outer-most AsyncHandler?

felixbarny: Following the decorator pattern, calling `AsyncHandler#onComplete` may result in the handler calling another, wrapped/nested `AsyncHandler#onComplete`. We want to make sure that only the first `AsyncHandler` activates and deactivates the span.

In your first draft, the first `AsyncHandler` would activate the span but the last `AsyncHandler` in the chain would deactivate it. This would result in the span not being active in the first `AsyncHandler` after it called `onComplete` on its delegate handler.


```
    private final AsyncHandler<Response> delegate;

    @Override
    public Response onCompleted(Response response) {
        try {
            return delegate.onCompleted(response);
        } finally {
             tracer.getActive() // returns null
        }
    }
```

By propagating the active span via a `@Advice.Local`, we have fixed that problem.

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
```diff
@@ -49,11 +47,35 @@ public AsyncHttpClientInstrumentationTest(RequestExecutor requestExecutor) {
...
+
+        @Override
+        public void onThrowable(Throwable t) {
+            assertThat(tracer.currentTransaction()).isNotNull();
+            assertThat(tracer.currentTransaction().isExit()).isTrue();
+        }
+
+        @Override
+        public Response onCompleted(Response response) {
+            assertThat(tracer.currentTransaction()).isNotNull();
```
felixbarny, line 66: To double-check, do the tests fail when doing `isNull()`?

milanvdm: Ill double-check all PR requirements now and make sure the tests make sense :)

milanvdm: Mmm, the changes make the existing tests fail (even when I remove the newly added `customAsyncHandler`. 

```
testHttpCall[3]  Time elapsed: 0.519 sec  <<< ERROR!
org.awaitility.core.ConditionTimeoutException:
Assertion condition defined as a lambda expression in co.elastic.apm.agent.MockReporter
Expecting actual not to be empty within 500 milliseconds.
```

Seems to me that is a local problem linked to my laptop. Will try to figure out why it is doing that. Any chance you have a docker-image able to run the tests?

felixbarny: The tests are failing on CI for the same reason. Not sure what causes that 🤔

milanvdm: Oh wow, interesting. Very strange that the new instrumentations make the current tests time-out. Will have a look later today when I have a bit more time :)

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java
```diff
@@ -223,7 +223,7 @@ public AsyncHandlerOnStatusReceivedInstrumentation() {
 
         @Advice.OnMethodEnter(suppress = Throwable.class)
         private static void onMethodEnter(@Advice.This AsyncHandler<?> asyncHandler, @Advice.Local("span") Span span, @Advice.Argument(0) HttpResponseStatus status) {
-            span = handlerSpanMap.remove(asyncHandler);
+            span = handlerSpanMap.get(asyncHandler);
```
felixbarny, line 226: 👍

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
milanvdm: Fixed it :)
Trying to get the new tests to work now as they currently only pass on the assertion of `isNull()`.

Could it be that activating the span is not enough and that the tracer needs to set the transaction to active instead?

apm-agent-plugins/apm-asynchttpclient-plugin/src/main/java/co/elastic/apm/agent/asynchttpclient/AbstractAsyncHttpClientInstrumentation.java
```diff
@@ -178,8 +178,15 @@ public AsyncHandlerOnCompletedInstrumentation() {
...
+        private static void onMethodEnter(@Advice.This AsyncHandler<?> asyncHandler, @Advice.Local("span") Span span) {
+            span = handlerSpanMap.remove(asyncHandler);
+            if (span != null) {
+                span.activate();
+            }
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class)
+        private static void onMethodExit(@Advice.This AsyncHandler<?> asyncHandler, @Nullable @Advice.Local("span") Span span) {
             if (span != null) {
```
felixbarny, line 190: Don't forget to deactivate it in all exit advices.

milanvdm: What is the difference between `end` and `deactivate`?

felixbarny: `end` sets the end timestamp and schedules the span to be reported. `deactivate` removes the span from the thread local which avoids the span from being leaked into different executions.

apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
felixbarny: hm, strange.. No, activating the span is enough. You don't have to activate the transaction on this thread.

milanvdm: I tried to figure it out but tbh, I have no idea.
Only thing I could think of was that Im using a `new AsyncCompletionHandler<Response>` in the test and somehow the methods of this subclass don't get matched properly.

milanvdm: @felixbarny I've played around with it a bit more:
```
        private static void onMethodEnter(@Advice.This AsyncHandler<?> asyncHandler, @Advice.Local("span") Span span) {
            span = handlerSpanMap.remove(asyncHandler);
            if (span != null) {
                span.activate();
                System.out.println(tracer.currentTransaction());
                System.out.println("=======================");
            }
        }
```

This prints out:
```
null
=======================
```

So it seems that activating the span does not mean I can do `assertThat(tracer.currentTransaction()).isNull();` in the test.

felixbarny: Yes, that's true. You should assert that `tracer.getActive()` is not null instead. I agree that if `tracer.getActive()` returns non-null, `tracer.getTranaction()` should do, too. I have an idea how to accomplish that. In the meantime, please assert on `tracer.getActive()`.

felixbarny: See also #1174

milanvdm: Cool, thanks :)
Ive applied the changes and the tests now pass correctly!

felixbarny:
Great work, thanks! ❤️

PR #1173, opened 2020-05-05
Upgrading async profiler version to 1.7.1 and adding safemode config

eyalkoren:
Adding support for the `safemode` config by async profiler through an internal agent config.


apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/ProfilingConfiguration.java
```diff
@@ -60,6 +60,13 @@
         .tags("added[1.15.0]", "experimental")
         .buildWithDefault(false);
 
+    private final ConfigurationOption<Integer> recoverMode = ConfigurationOption.<Integer>integerOption()
+        .key("recover_mode")
+        .configurationCategory(PROFILING_CATEGORY)
+        .dynamic(false)
+        .tags("internal")
+        .buildWithDefault(0);
```
felixbarny, line 68: This disables all stack trace recoveries by default. Is that intentional?

apm-agent-plugins/apm-profiling-plugin/src/test/java/co/elastic/apm/agent/profiler/RecoverSamplingProfilerTest.java
```diff
@@ -0,0 +1,182 @@
...
+            .until(() -> profiler.getProfilingSessions() > 1);
+    }
+
+    @AfterEach
+    void tearDown() throws Exception {
+        tracer.stop();
+    }
+
+    @Test
+    void testAll() throws Exception {
```
felixbarny, line 85: What's the value of testing the different options?

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/ProfilingConfiguration.java
eyalkoren: This is not what I plan to push to master, this is only an analysis agent, where we want to start by testing that the recovery code is not the cause. Only if required, we will ask the user to set this to other values.
I did it as a PR just so I have the artefacts publicly available.

apm-agent-plugins/apm-profiling-plugin/src/test/java/co/elastic/apm/agent/profiler/RecoverSamplingProfilerTest.java
eyalkoren: Only sanity check that enabling discrete parts of the recovery logic doesn't break anything. Again - not to be eventually merged.

eyalkoren:
@felixbarny any idea how to test the new state?

felixbarny:
I don’t think testing is required here. We just bass this through to async profiler. We shouldn’t test async profiler itself in this project.

PR #1187, opened 2020-05-17
Adding class exclusion from instrumentation config

eyalkoren:
Adding (and exposing) the ability to exclude classes from instrumentation without having to override the default excluded classes. 
For example - https://discuss.elastic.co/t/how-to-ignore-certain-spans/232625

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -334,11 +334,22 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
         .builder(new ListValueConverter<>(new WildcardMatcherValueConverter()), List.class)
         .key("classes_excluded_from_instrumentation")
         .configurationCategory(CORE_CATEGORY)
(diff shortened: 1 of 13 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -334,11 +334,22 @@
...
+        .description("Use to exclude specific classes from being instrumented. In order to exclude entire packages, \n" +
+            "use wildcards, as in: `com.project.exclude.*`" +
+            "\n" +
+            WildcardMatcher.DOCUMENTATION)
+        .dynamic(false)
+        .buildWithDefault(Collections.<WildcardMatcher>emptyList());
+
+    private final ConfigurationOption<List<WildcardMatcher>> defaultClassesExcludedFromInstrumentation = ConfigurationOption
+        .builder(new ListValueConverter<>(new WildcardMatcherValueConverter()), List.class)
+        .key("default_classes_excluded_from_instrumentation")
```
felixbarny, line 346: Maybe `classes_excluded_from_instrumentation_default` so that the prefix is the same? Feel free to ignore though

PR #1189, opened 2020-05-18
Verifying Ubuntu cgroup line parsing for container ID

eyalkoren:
Verifying that no false positive is created as reported in https://github.com/elastic/apm-agent-go/issues/776


eyalkoren:
This only validates this one reported cgroup path pattern, we could still get false positives with other similar patterns, but as long as we don't get such reported, no need to change.

PR #1196, opened 2020-05-21
gRPC memory leak fix + bench test app

### LINKED ISSUE #1191: [GRPC] memory leak
## Describe the bug
memory leak when using grpc instrumentation
Screenshot 2020-05-18 at 11 16 30
Screenshot 2020-05-18 at 11 19 58

It took about 48h to OOM a 1GB application with 100 requests per minute

## Steps to reproduce
I did not try a minimal example, but our application is really really simple already. 
NOTE: This link is Elastic internal. 
https://github.com/elastic/demos/blob/master/observe/docker/adservice/src/main/java/hipstershop/AdService.java

### Disable Instrumentation 
leak disappears when setting `ELASTIC_APM_DISABLE_INSTRUMENTATIONS=grpc` 

## Expected behavior
No memory leak :) 

## Debug logs
heap dump
NOTE: This link is Elastic internal. 
https://drive.google.com/file/d/1NP-NQXnJMbXUgEpPOEuzwgQyGk_XK51S/view?usp=sharing

SylvainJuge:
Fixes #1191 .

This PR contains:
- a fix for the memory leak instrumentation implementation
- a "benchmark mode" for test applications that we can use from CLI, which helps to reproduce similar errors (and could be used later to measure performance impact).
- a simplification of existing implementation thanks to @felixbarny that makes memory leaks harder to create.


SylvainJuge:
I've added a "benchmark" mode for the sample applications that should allow to reproduce this locally.

```
java -jar apm-grpc-test-1.6.1/target/testapp.jar --benchmark --count 100 --wait
```

Currently when limiting the number of server-side thread pool, execution becomes really slow.
If we don't, then the number of threads grows and execution time is getting slower during benchmark than it was during warmup. I think that our test application is probably missing some flow-control (does not stop sending messages when server is busy).

SylvainJuge:
Quick update, The issue is now easily reproducible with latest update.

Screenshot from 2020-05-29 20-38-04

This was triggered in a few minutes with the following command line, without agent there is no leak, with agent there is a very nice staircase to OutOfMemoryError.

```
java -jar apm-grpc-test-1.6.1/target/testapp.jar --benchmark --count 100000 --wait
```

SylvainJuge:
current status: There are still some test failures in CI, but I can't reproduce them locally, either in IDE or command-line.

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/ServerCallListenerInstrumentation.java
```diff
@@ -151,7 +148,14 @@ public FinalMethodCall(ElasticApmTracer tracer) {
...
                 // --> end of unary call (success)
-                .or(named("onComplete"));
+                .or(named("onComplete"))
+                //
+                // client completed all message sending, but can still cancel the call
+                // --> for unary calls, actual method invocation is done within 'onHalfClose' method, and there is no
+                // call to 'onComplete', thus consider it as 'final' allows to properly perform cleanup as it's the last
+                // method that will be invoked on the listener.
+                //
+                .or(named("onHalfClose"));
```
SylvainJuge, line 158: this change was the one that fixed the memory leak, as it seems that some calls to `onComplete` are not always executed at all, or are executed in an unexpected order.

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/test/java/co/elastic/apm/agent/grpc/GrpcTest.java
```diff
@@ -26,10 +26,11 @@
 
 import co.elastic.apm.agent.grpc.testapp.GrpcApp;
 import co.elastic.apm.agent.grpc.testapp.GrpcAppProvider;
+import co.elastic.apm.agent.testutils.TestPort;
 
 public class GrpcTest {
 
-    private static final int PORT = 50051;
+    private static final int PORT = TestPort.getAvailableRandomPort();
```
SylvainJuge, line 33: this allows to have more than one test running on the same host without conflict.

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/test/java/co/elastic/apm/agent/grpc/testapp/AbstractGrpcAppTest.java
```diff
@@ -86,9 +87,15 @@ void simpleErrorCall() {
...
-            strategy.sendAndCheckMessage(app, "rob", 2, "nested(2)->nested(1)->hello(rob)");
+            // because nested call is done with a blocking call, we have to make sure that we don't use all sever
+            // threads otherwise we get exceeded deadlines
+            for (int depth = 0; depth < HelloServer.POOL_SIZE; depth++) {
+                StringBuilder prefix = new StringBuilder();
+                for (int i = 1; i <= depth; i++) {
+                    prefix.insert(0, String.format("nested(%d)->", i));
+                }
+                strategy.sendAndCheckMessage(app, "joe", depth, prefix + "hello(joe)");
+            }
```
SylvainJuge, line 98: this one deserves a facepalm and was the cause of the failure on CI, as CI server has less cores than a decent laptop (likely 4), thus server was running with 2 threads in thread pool, which makes the 3rd nested call fail with a missed deadline.

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/test/java/co/elastic/apm/agent/grpc/testapp/HelloClient.java
```diff
@@ -37,17 +38,23 @@
...
-
     private final ManagedChannel channel;
+    private final AtomicLong errorCount;
 
     protected HelloClient(ManagedChannel channel) {
         this.channel = channel;
+        this.errorCount = new AtomicLong(0);
+    }
+
+    protected static Deadline getDeadline() {
```
SylvainJuge, line 56: adding a deadline makes remote calls fail with timeout instead of waiting forever.

SylvainJuge:
This PR is now ready for review, feel free to merge it @felixbarny if approved.

apm-agent-core/src/test/java/co/elastic/apm/agent/testutils/TestPort.java
```diff
@@ -0,0 +1,59 @@
...
+        int port;
+        do {
+            port = MIN_PORT + rand.nextInt(MAX_PORT - MIN_PORT + 1);
+        } while (!isAvailablePort(port));
+        return port;
+    }
+
+    private static boolean isAvailablePort(int port) {
+        try {
+            ServerSocket serverSocket = ServerSocketFactory.getDefault().createServerSocket(port, 1, InetAddress.getByName("localhost"));
```
felixbarny, line 52: You can also use port 0 which makes it open a random free port, without you having to manually try port after port.

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/ServerCallListenerInstrumentation.java
felixbarny: I don't think that's the issue. The problem is that we're only instrumenting server call listeners, that contain `Unary` in the class name. The `UnaryServerCallListener` doesn't override the noop `onComplete` callback.

Another issue is that the transaction is ended in `co.elastic.apm.agent.grpc.ServerCallInstrumentation#onExit` (on `ServerCall#close`). The effect is that even when instrumenting `io.grpc.ServerCall.Listener#onComplete`, the helper is not called, because the transaction has already ended.

Furthermore, the fact that `inFlightServerListeners` is of type `WeakConcurrentMap<ServerCall.Listener<?>, ServerCall<?, ?>>` imposes a risk of a memory leak as `ServerCall` is not a weak reference. Have you considered mapping directly to the transaction (`WeakConcurrentMap<ServerCall.Listener<?>, Transaction>`)?

Similar story for client instrumentation. Changing the signature to `WeakConcurrentMap<ClientCall.Listener<?>, Span> inFlightClientListeners` would eliminate the chance of leaks and allows direct lookup of a span given a listener.

Another simplification of the client instrumentation could be done by removing `co.elastic.apm.agent.grpc.ClientCallImplInstrumentation.Constructor` and doing that inside `co.elastic.apm.agent.grpc.ChannelInstrumentation` (which instruments `io.grpc.Channel#newCall`). That would be more in line with the server instrumentation in `co.elastic.apm.agent.grpc.ServerCallHandlerInstrumentation` (instrumenting `io.grpc.ServerCallHandler#startCall`). And allows you to start the span before the constructor is invoked. In the exit advice you can then use the return value.

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/ServerCallHandlerInstrumentation.java
```diff
@@ -69,35 +67,25 @@ public ServerCallHandlerInstrumentation(ElasticApmTracer tracer) {
...
-    }
-
-    @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
-    private static void onExit(@Advice.Thrown Throwable thrown,
-                               @Advice.Argument(0) ServerCall<?, ?> serverCall,
-                               @Advice.Return ServerCall.Listener<?> listener,
-                               @Advice.Local("transaction") @Nullable Transaction transaction) {
-
-        if (tracer == null || grpcHelperManager == null || transaction == null) {
+        if (serverCall.getMethodDescriptor().getType() != MethodDescriptor.MethodType.UNARY) {
```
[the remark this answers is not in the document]
SylvainJuge: Good question, first, that's on par with what other elastic APM agents provide, avoid extra complexity when adding this feature, and then there was the case of bi-directional streaming which might not fit the concept of "transaction" if it's used as a long-lived communication channel, but that might really be a marginal use. Thus we could probably lift this limitation in the near future in a separate PR.

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/ServerCallHandlerInstrumentation.java
```diff
@@ -69,35 +67,25 @@ public ServerCallHandlerInstrumentation(ElasticApmTracer tracer) {
...
+                               @Advice.Return ServerCall.Listener<?> listener) {
 
-        if (tracer == null || grpcHelperManager == null) {
+        if (tracer == null || grpcHelperManager == null || thrown != null) {
             return;
         }
 
-        GrpcHelper helper = grpcHelperManager.getForClassLoaderOfClass(ServerCall.class);
-        if (helper != null) {
-            transaction = helper.startTransaction(tracer, clazz.getClassLoader(), serverCall, headers);
```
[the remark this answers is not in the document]
SylvainJuge: fair point, it was a way to make instrumentation a bit simpler with only one single call to helper class. I've changed that in last commit.

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/helper/GrpcHelperImpl.java
```diff
@@ -185,22 +158,21 @@ public void exitServerListenerMethod(@Nullable Throwable thrown, ServerCall.List
         if (null != thrown) {
             // when there is a runtime exception thrown in one of the listener methods the calling code will catch it
             // and set 'unknown' status, we just replicate this behavior as we don't instrument the part that does this
-            endTransaction(Status.UNKNOWN, thrown, transaction);
-
-            // listener won't be called anymore
-            inFlightServerListeners.remove(listener);
+            setTransactionStatus(Status.UNKNOWN, thrown, transaction);
```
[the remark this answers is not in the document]
SylvainJuge: server-side tests have been added in latest version of PR.

SylvainJuge:
After adding extra tests with exceptions thrown within server and client listeners, I think we are quite well covered with tests now.

However, there is still one minor thing bothering me: trying to trigger exceptions from `onClose` on client side completely breaks the client part, thus it's not covered in tests. That means if such an exception occurs, we might leave an active span, but that should not create a leak this time.

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/ServerCallHandlerInstrumentation.java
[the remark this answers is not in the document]
felixbarny: Please create a follow-up issue for that.

felixbarny:
The changelog is missing I think

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/ServerCallHandlerInstrumentation.java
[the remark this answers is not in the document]
SylvainJuge: done: https://github.com/elastic/apm-agent-java/issues/1233

PR #1203, opened 2020-05-26
Jul deadlock workaround

### LINKED ISSUE #1202: Agent initializes Java truststore
### Environment
Wildfly 17.0.1
Java 11

We made a big version jump of the Elastic Apm Agent version in one of our environments (1.4.0 -> 1.16.0) and immediately saw our ldap connection fail. This connection uses internal certificates which are not stored in the default java truststore. The truststore is not set in the jvm options but will be set from the application at deployment like this example
```java
System.setProperty("javax.net.ssl.trustStore", truststoreLocation);
System.setProperty("javax.net.ssl.trustStorePassword", truststorePassword);
```

Debugging via `-Djavax.net.debug=all` showed us that the apm agent since version 1.13.0 initializes the default truststore before Wildfly even starts which breaks our truststore initialization.

## Steps to reproduce

1. Download Wildfly from https://wildfly.org/downloads/ and extract it
2. Add the following JAVA_OPTS in the wildfly config at `<wildfly-dir>/bin/standalone.conf` (Linux) or `<wildfly-dir>/bin/standalone.conf.bat` (Windows)
`-javaagent:<path-to-agent>/elastic-apm-agent.jar -Delastic.apm.service_name=App -Delastic.apm.server_urls=http://localhost:8200 -Delastic.apm.application_packages=com.example -Delastic.apm.environment="development" -Djavax.net.debug=all`
3. Start wildfly with the start script found in `<wildfly-dir>/bin/standalone.sh` (Linux) or `<wildfly-dir>/bin/standalone.bat` (Windows)
4. With apm versions >= 1.13.0 the whole default truststore will be printed to the console/log before this log message appears
`
INFO co.elastic.apm.agent.report.ApmServerHealthChecker - Elastic APM server is available: {  "build_date": "2020-05-12T00:04:54Z",  "build_sha": "64e91c95329991c36b16ff94fd34ea75230c06c2",  "version": "7.7.0"}
`

## Expected behavior
Expected would be that the agent like before does not initialize the Java truststore.

## Debug logs
If needed I can add the debug logs.

eyalkoren:
An alternative workaround for the JULI deadlock reported at #954.

The deadlock fix was verified using the following test (where uncommenting the first line eliminates it):
```
    public static void main(String[] args) throws InterruptedException {
        // java.util.logging.Logger.getLogger("noop");
        ExecutorService executorService = Executors.newFixedThreadPool(2);
        System.setProperty("java.util.logging.config.class", "Logging enabled");
        executorService.submit(() -> new URL("http://www.elastic.co").openConnection());
        executorService.submit(ManagementFactory::getPlatformMBeanServer);
        executorService.shutdown();
        executorService.awaitTermination(5, TimeUnit.SECONDS);
    }
```

No side effects are expected with this, as this only initializes the Logger manager a bit before it would be initialized anyway.

Closes #1202

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
@@ -117,6 +117,8 @@ public class ElasticApmTracerBuilder {
     }
 
     public ElasticApmTracer build() {
+        initializeJavaUtilLogging();
+
         boolean addApmServerConfigSource = false;
         List<LifecycleListener> lifecycleListeners = new ArrayList<>();
 
@@ -151,6 +153,21 @@ public class ElasticApmTracerBuilder {
         lifecycleListeners.addAll(extraLifecycleListeners);
         tracer.start(lifecycleListeners);
         return tracer;
+    }
+
+    /**
+     * A noop method aimed for working around the JULI deadlock reported at https://github.com/elastic/apm-agent-java/issues/954.
+     * The main thread creates an {@link ApmServerClient} and then starts the Tracer, which start the {@link LifecycleListener}s.
+     * One of the listeners is the JMX metric tracker, that initializes the JUL mechanism. Concurrently, another thread is
+     * doing healthcheck with the APM server (kicked from the {@link ApmServerClient}), and this may cause a deadlock.
+     * Calling this method before both, should prevent this deadlock.
+     */
+    private void initializeJavaUtilLogging() {
+        try {
+            java.util.logging.Logger.getLogger("noop");
+        } catch (Throwable th) {
+            logger.info("Failed to initialized the Java Util Logging mechanism", th);
+        }
     }
 
     private LifecycleListener scheduleReloadAtRate(final ConfigurationRegistry configurationRegistry, final int rate, TimeUnit seconds) {
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
@@ -117,8 +117,6 @@ public class ElasticApmTracerBuilder {
     }
 
     public ElasticApmTracer build() {
-        initializeJavaUtilLogging();
-
         boolean addApmServerConfigSource = false;
         List<LifecycleListener> lifecycleListeners = new ArrayList<>();
 
@@ -153,21 +151,6 @@ public class ElasticApmTracerBuilder {
         lifecycleListeners.addAll(extraLifecycleListeners);
         tracer.start(lifecycleListeners);
         return tracer;
-    }
-
-    /**
-     * A noop method aimed for working around the JULI deadlock reported at https://github.com/elastic/apm-agent-java/issues/954.
-     * The main thread creates an {@link ApmServerClient} and then starts the Tracer, which start the {@link LifecycleListener}s.
-     * One of the listeners is the JMX metric tracker, that initializes the JUL mechanism. Concurrently, another thread is
-     * doing healthcheck with the APM server (kicked from the {@link ApmServerClient}), and this may cause a deadlock.
-     * Calling this method before both, should prevent this deadlock.
-     */
-    private void initializeJavaUtilLogging() {
-        try {
-            java.util.logging.Logger.getLogger("noop");
-        } catch (Throwable th) {
-            logger.info("Failed to initialized the Java Util Logging mechanism", th);
-        }
     }
 
     private LifecycleListener scheduleReloadAtRate(final ConfigurationRegistry configurationRegistry, final int rate, TimeUnit seconds) {
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -41,6 +41,7 @@ The latter uses https://github.com/elastic/ecs-logging-java[ecs-logging-java] to
   This also makes the profiler more resilient by just removing the call tree related to the exception (which might be in an invalid state)
   as opposed to stopping the profiler when an exception occurs.
 * Fix `NumberFormatException` when parsing Ingres/Actian JDBC connection strings - {pull}1198[#1198]
+* Prevent agent from overriding JVM configured truststore when not using HTTPS for communication with APM server - {pull}1203[#1203]
 
 [[release-notes-1.x]]
 === Java Agent version 1.x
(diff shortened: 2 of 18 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
```diff
@@ -153,6 +155,21 @@ public ElasticApmTracer build() {
...
+    /**
+     * A noop method aimed for working around the JULI deadlock reported at https://github.com/elastic/apm-agent-java/issues/954.
+     * The main thread creates an {@link ApmServerClient} and then starts the Tracer, which start the {@link LifecycleListener}s.
+     * One of the listeners is the JMX metric tracker, that initializes the JUL mechanism. Concurrently, another thread is
+     * doing healthcheck with the APM server (kicked from the {@link ApmServerClient}), and this may cause a deadlock.
+     * Calling this method before both, should prevent this deadlock.
+     */
+    private void initializeJavaUtilLogging() {
+        try {
+            java.util.logging.Logger.getLogger("noop");
```
felixbarny, line 167: I'm pretty sure that this will fail the JBoss integration tests. JBoss needs to be able to set the `java.util.logging.manager` system property to override the default `LogManager`.

See also

https://github.com/elastic/apm-agent-java/blob/96b7d60e4edcb39242ab5d7922111d52fd9c1023/apm-agent-plugins/apm-jmx-plugin/src/main/java/co/elastic/apm/agent/jmx/JmxMetricTracker.java#L96-L144

I assume because the health check runs in a background thread, JBoss has enough time to init the log manager. Maybe we can make the above more code more generic and fire the start lifecycle event only after `java.util.logging.manager` has been set if running on JBoss and fire immediately otherwise.

eyalkoren: That's odd, the fact that they it does get into this deadlock means that the server healthcheck is done quickly enough. But we can
>  make the above more code more generic and fire the start lifecycle event only after java.util.logging.manager has been set if running on JBoss and fire immediately otherwise

to be on the safe side.

eyalkoren: OK, there are a lot of hidden subtleties that produce misleading behaviours here. 

1. Why didn't we see this before with the healthcheck? This is the relevant thread stack taken from the deadlock thread dump:
```
"apm-server-healthcheck":
        at java.util.logging.LogManager.ensureLogManagerInitialized(LogManager.java:317)
        - waiting to lock <0x0000000755bb5ef0> (a org.apache.juli.ClassLoaderLogManager)
        at java.util.logging.LogManager.getLogManager(LogManager.java:378)
        at java.util.logging.Logger.getPlatformLogger(Logger.java:572)
        at java.util.logging.LoggingProxyImpl.getLogger(LoggingProxyImpl.java:41)
        at sun.util.logging.LoggingSupport.getLogger(LoggingSupport.java:100)
        at sun.util.logging.PlatformLogger$JavaLoggerProxy.<init>(PlatformLogger.java:602)
        at sun.util.logging.PlatformLogger$JavaLoggerProxy.<init>(PlatformLogger.java:597)
        at sun.util.logging.PlatformLogger.<init>(PlatformLogger.java:239)
        at sun.util.logging.PlatformLogger.getLogger(PlatformLogger.java:198)
        - locked <0x0000000756becbe0> (a java.lang.Class for sun.util.logging.PlatformLogger)
        at sun.net.www.protocol.http.HttpURLConnection.<clinit>(HttpURLConnection.java:431)
        at sun.net.www.protocol.http.Handler.openConnection(Handler.java:62)
        at sun.net.www.protocol.http.Handler.openConnection(Handler.java:57)
        at java.net.URL.openConnection(URL.java:979)
        at co.elastic.apm.agent.report.ApmServerClient.startRequestToUrl(ApmServerClient.java:125)
        at co.elastic.apm.agent.report.ApmServerClient.executeForAllUrls(ApmServerClient.java:236)
        at co.elastic.apm.agent.report.ApmServerHealthChecker.call(ApmServerHealthChecker.java:69)
        at co.elastic.apm.agent.report.ApmServerHealthChecker.call(ApmServerHealthChecker.java:47)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
```
The `sun.net.www.protocol.http.HttpURLConnection` static initializer calls `sun.util.logging.PlatformLogger.getLogger()` explicitly, which in turn creates a new `PlatformLogger` instance. This constructor uses the `JavaLoggerProxy` (resulting in the rest of the stack above) only if some `java.util.logging.config` system properties were previously set (see what I needed to do in my [code snippet](https://github.com/elastic/apm-agent-java/pull/1203#issue-423224248) above in order to be able to reproduce the deadlock). If such are not set, it will instantiate a `PlatformLogger.DefaultLoggerProxy` instead, which will not continue with the JUL initialization chain, thus not interfering with WildFly's Log Manager obsession.
2. So why don't the WildFly/JBoss integration tests fail when calling `Logger.getLogger` explicitly? Because of shading! The call I added is replaced with `co.elastic.apm.agent.logging.JulBridgeLogger.getLogger` 🤦 

Going back to the drawing board...

eyalkoren:
I ended up synching the minimal required context - `URL.openConnection()` and `ManagementFactory.getPlatformMBeanServer()`. I started with a more fancy approach of only locking until either one is first invoked, but since MBean server lookup occurs only once and opening a connection is heavy and infrequent enough to make locking overhead negligible, I decided it doesn't worth the complexity.

eyalkoren:
BTW, we cannot verify it with a unit test because JUnit apparently initilizes JUL, so verified with this test:
```
    public static void main(String[] args) throws InterruptedException, IOException {
        ExecutorService executorService = Executors.newFixedThreadPool(2);
        System.setProperty("java.util.logging.config.class", "Logging enabled");
        final Lock lock = new ReentrantLock();
        executorService.submit(() -> {
            lockAndSubmit(lock, () -> {
                try {
                    new URL("http://www.elastic.co").openConnection();
                } catch (IOException e) {
                    e.printStackTrace();
                }
            });
        });
        executorService.submit(() -> lockAndSubmit(lock, ManagementFactory::getPlatformMBeanServer));
        executorService.shutdown();
        executorService.awaitTermination(5, TimeUnit.SECONDS);
    }

    private static void lockAndSubmit(Lock lock, Runnable runnable) {
//        lock.lock();
        try {
            runnable.run();
        } finally {
//            lock.unlock();
        }
    }
```

PR #1205, opened 2020-05-27
ignore jps JVM for runtime attach

SylvainJuge:
When using runtime attach, especially the `--continuous` option, the JVM that is running the `jps` process is always listed, and makes agent output an error message with stack trace because this short-lived JVM is already terminated.

This PR makes JVMs running `jps` ignored when listing, which makes it behave slightly differently than the `jps -lv` command (which always includes itself). Also, it resolves path to `jps` binary using `JAVA_HOME` environment variable in case it's not available in `PATH`.

### DIFF
--- a/apm-agent-attach/src/main/java/co/elastic/apm/attach/JvmDiscoverer.java
+++ b/apm-agent-attach/src/main/java/co/elastic/apm/attach/JvmDiscoverer.java
@@ -27,7 +27,6 @@ package co.elastic.apm.attach;
 import com.sun.jna.Platform;
 
 import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 import java.io.File;
 import java.io.FileFilter;
 import java.io.FilenameFilter;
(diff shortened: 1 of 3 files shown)


apm-agent-attach/src/main/java/co/elastic/apm/attach/JvmDiscoverer.java
```diff
@@ -96,19 +99,45 @@ private static String getJpsOutput() throws IOException, InterruptedException {
             Set<JvmInfo> set = new HashSet<>();
             for (String s : jpsOutput.split("\n")) {
                 JvmInfo parse = JvmInfo.parse(s);
-                set.add(parse);
+                // ignore jps command that we just started as it's already terminated and not relevant for attachment
+                if(!parse.packageOrPathOrJvmProperties.contains("sun.tools.jps.Jps")){
```
felixbarny, line 103: [minor] might make it more resilient when packages are renamed.
```suggestion
                if(!parse.packageOrPathOrJvmProperties.contains(".Jps")){
```

PR #1206, opened 2020-05-27
Instrument Runnables/Callables instead of wrapping them

### LINKED ISSUE #987: Enhance Executors instrumentation
**Problem description**
Currently, we wrap `Callable`s and `Runnable`s when they are submitted for execution. The problem with wrapping is the risk of type-mismatch in cases where the framework expects (and casts to) a specific subtype that we do not extend. In some cases we can explicitly exclude such frameworks, but in other cases this is not an option, as it means we would not properly trace. See elastic/apm-agent-java#977 for example.

**Solution description**
Instead of wrapping, we can instrument the actual `Callable#call` and `Runnable#run` APIs. Following are the related problems and suggested solutions:
1. _Startup overhead:_ since any anonymous class can implement these interfaces, we must match the entire classpath, which is expensive. The dynamic matching mechanism added in elastic/apm-agent-java#985 can be applied similarly in this case, thus eliminating this problem (using the `execute`/`submit` methods to make sure the submitted callable/runnable class is instrumented). 
2. _Runtime overhead:_ we don't want to instrument ALL callables and runnables, as many will be irrelevant for this purpose. Since normally each runnable/callable type is used for a specific task, the solution of bullet 1 will be effective for this purpose as well.
3. _Lambdas:_ we cannot instrument lambdas, but we can wrap them because they do not suffer from the same problem of the wrapped types- a lambda only implements one known method. In other words, we keep the current mechanism only for lambdas, which can be identified like we do for [JMS listeners](https://github.com/elastic/apm-agent-java/blob/935c73367b596d9b41ca86817e1a5fdd21f45d6f/apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelperImpl.java#L121).

felixbarny:
## What does this PR do?
closes elastic/apm-agent-java#987 

Bonus content:
* Support for `ForJoinPool` and `ScheduledExecutorService`
* Support for `ExecutorService#invokeAny` and `ExecutorService#invokeAll`

## Checklist


apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/JavaConcurrent.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ForkJoinTask;
+
+public class JavaConcurrent {
+
+    private static final WeakConcurrentMap<Object, AbstractSpan<?>> contextMap = new WeakConcurrentMap<Object, AbstractSpan<?>>(false);
```
eyalkoren, line 44: We need some cleanup of stale entries, maybe use `WeakConcurrentMap.WithInlinedExpunction`?
On the other hand, given that this map may be accessed VERY frequently, using a cleaner thread makes more sense

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/JavaConcurrent.java
```diff
@@ -0,0 +1,229 @@
...
+    public static <T> Collection<? extends Callable<T>> withContext(@Nullable Collection<? extends Callable<T>> callables, @Nullable ElasticApmTracer tracer) {
+        if (callables == null || tracer == null) {
+            return null;
+        }
+        if (callables.isEmpty()) {
+            return callables;
+        }
+        final Collection<Callable<T>> wrapped;
+        if (needsWrapping(callables)) {
+            wrapped = new ArrayList<>(callables.size());
```
eyalkoren, line 179: This introduces a risk of causing `ClassCastException` due to the collection type. I think it's better to try/catch using `Collection.remove` and `Collection.add` and avoid wrapping (with logging) if not supported by the collection.

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/JavaConcurrent.java
```diff
@@ -0,0 +1,229 @@
...
+        if (runnable instanceof RunnableLambdaWrapper || runnable == null || tracer == null || needsContext.get() == Boolean.FALSE) {
+            return runnable;
+        }
+        needsContext.set(Boolean.FALSE);
+        AbstractSpan<?> active = tracer.getActive();
+        if (active == null) {
+            return runnable;
+        }
+        if (isLambda(runnable)) {
+            runnable = new RunnableLambdaWrapper(runnable);
```
eyalkoren, line 96: What about pooling wrappers? Could be used very frequently.

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/RunnableCallableForkJoinTaskInstrumentation.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Arrays.asList("concurrent", "executor");
+    }
+
+    @Nullable
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    private static AbstractSpan<?> onEnter(@Advice.This Object thiz) {
+        return JavaConcurrent.restoreContext(thiz, tracer);
```
eyalkoren, line 77: [minor] - proposal: make the activation here, rather than implicit, making it symmetric with `onExit`.

apm-agent-core/src/main/java/co/elastic/apm/agent/util/CallDepth.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.util.Map;
+import java.util.WeakHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A utility that makes it easy to detect nested method calls.
+ */
+public class CallDepth {
+    private static final ThreadLocal<Map<Class<?>, AtomicInteger>> callDepthPerThread = new ThreadLocal<Map<Class<?>, AtomicInteger>>();
```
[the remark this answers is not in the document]
felixbarny: > Also, if it's about identification of nesting only (without knowing the level of nesting), no need for a map - a stack of classes is sufficient.

Not quite sure what you mean by that. Note that often you also need to know in the exit advice whether it's a nested invocation.

> If it's thread local, why using an AtomicInteger?

Mainly out of convenience. We could create something like a `MultableInt` to get rid of the volatile reads/writes if that's what you're hinting at?

> Lastly, using class name may be better as it won't require weak referencing.

The new version is using a string which is important so the call depth is a global vs a per classloader counter.

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/JavaConcurrent.java
felixbarny: Good catch, it's not using any extinction at all 🙈 

We should create a global registry for `WeakConcurrentMap`s so that we only need one cleaner thread for all vs one per map. But for now, `WithInlinedExpunction` should do.

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/JavaConcurrent.java
felixbarny: That would impact the order of the callables in the collection. But we could clear and re-fill it.

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/JavaConcurrent.java
felixbarny: By putting the runnable in a `WeakConcurrentMap`, we're already allocating a `WeakReference` and a `Map.Entry` so pooling the Runnable wrapper won't make it allocation free, unfortunately. Given that, I don't think it's worth it.

apm-agent-core/src/main/java/co/elastic/apm/agent/util/CallDepth.java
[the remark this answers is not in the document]
eyalkoren: > Note that often you also need to know in the exit advice whether it's a nested invocation.

Why isn't a stack enough for that? In any case, if you know in the enter advice, it should be easy to tell the exit advice...

> Mainly out of convenience. We could create something like a MultableInt to get rid of the volatile reads/writes if that's what you're hinting at?

The volatile and the additional overhead of atomically fetching and incrementing/decrementing. Why does it have to be mutable and not an `Integer` you override when incrementing/decrementing? You would really need a deep nesting to be out of the Integer cache

> The new version is using a string which is important so the call depth is a global vs a per classloader counter.

Do you think the usage of same class names loaded by different class loaders on the same stack is a concern?

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/JavaConcurrent.java
eyalkoren: I also had that in mind, but didn't see an API in `WeakConcurrentMap` to control which thread is the cleaner thread.

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/JavaConcurrent.java
```diff
@@ -0,0 +1,229 @@
...
+        }
+        // When an Executor executes directly on the current thread we need to enable this thread for context propagation again
+        needsContext.set(Boolean.TRUE);
+        AbstractSpan<?> context = contextMap.remove(o);
+        if (context == null) {
+            return null;
+        }
+        if (tracer.getActive() != context) {
+            context.activate();
+            context.decrementReferences();
```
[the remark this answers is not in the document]
felixbarny: The decrement should happen after the activate. Otherwise, the span might be recycled if the reference count goes down to 0.

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/JavaConcurrent.java
felixbarny: I think calling `add` and `remove` is more likely to fail than creating an entirely new list because that's a good chance the collection is immutable.
Maybe leave as-is and remove if we see it causes problems?

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/JavaConcurrent.java
felixbarny: I've added a central registry to get and clean weak maps

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/RunnableCallableForkJoinTaskInstrumentation.java
felixbarny: I did that on purpose to minimize the null checks you need to do.

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/JavaConcurrent.java
eyalkoren: My vote is actually to go the other way around - fail (meaning - don't wrap lambdas in collections if the collection is immutable) or not try to wrap in collections and see if we get requests to support that. I prefer questions about why something is not working than issues with breaking applications.
If you are not convinced - leave as is.

felixbarny: That may cause subtle context losses that are very hard to debug. I'd like to resort to these defensive measures only if there are actual problems in the wild which will surface in a more obvious way by throwing an exception. I'll add a new instrumentation group name `executor-collection` to disable just this instrumentation more easily.

I think in practice it's unlikely that an executor implementation depends on the type of the collection or even overrides the default implementation in `AbstractExecutorService`.

eyalkoren: This entire PR is about executors exposing APIs that allow interfaces but expect specific types 🙂

PR #1207, opened 2020-05-27
lock jar file when writing to avoid multiple jvm override him

twillouer:


twillouer:
> x Author of the following commits did not sign a [Contributor Agreement](https://www.elastic.co/contributor-agreement):
> [be5230f](https://github.com/elastic/apm-agent-java/commit/be5230f5491744ffd536794b33b067250888c014)
> 
> Please, read and sign the above mentioned agreement if you want to contribute to this project

done

apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
```diff
@@ -190,10 +192,17 @@ private static File getAgentJarFile() {
...
                 if (!tempAgentJar.exists()) {
-                    try (OutputStream out = new FileOutputStream(tempAgentJar)) {
-                        byte[] buffer = new byte[1024];
-                        for (int length; (length = agentJar.read(buffer)) != -1;) {
-                            out.write(buffer, 0, length);
+                    try (FileOutputStream out = new FileOutputStream(tempAgentJar)) {
+                        FileChannel channel = out.getChannel();
+                        // If multiple JVM start on same compute, they can write in same file
+                        // and this file will be corrupted.
+                        try(FileLock ignored = channel.lock()) {
```
felixbarny, line 199:
```suggestion
                        try (FileLock ignored = channel.lock()) {
```

twillouer:
> Nice catch!
> Could you apply the same changes to `co.elastic.apm.agent.util.IOUtils#exportResourceToTemp`?
> 
> This methods contains duplicated code because there's no good way to share code across the `apm-agent-attach` and other modules.

Your right !
Done !

felixbarny:
I gave this some thought and I think there's an even better and more idiomatic way to guard against this. The problem currently is that checking `file.exists()` and writing to the file is a classic check-and-act race condition.

A better way is to use the `FileChannel#open` API to atomically create a new file if it does not exist already:
```java
try (FileChannel fileChannel = FileChannel.open(path, CREATE_NEW, WRITE)) {
    // write to file
} catch (FileAlreadyExistsException ignore) {
}
```

Let me know if you need help with the details.

twillouer:
@felixbarny I only do the job for IOUtils, tell me if I'm wrong ?
I will report the solution in ElasticApmAttacher if your agree with this fix.

I'm not sure this is the best way to handle this

felixbarny:
I think I have missed a key point in your implementation. The `channel.lock()` call has the big benefit that other JVMs that are concurrently trying to write to the file will wait until the JVM that acquired the lock first has fully written the file before proceeding. This is of course much better than having to manually sleep. Sorry for the confusion, your initial implementation was better. Just don't forget to also check for `if (tempFile.length() == 0)` in `IOUtils`

PR #1216, opened 2020-06-04
ci: run benchmark on a PR basis if comment

v1v:

Enable the benchmark stage on a PR basis if using the GH comment `jenkins run the benchmark tests please` or a more minimalistic comment with `jenkins run benchmark tests` or even with `run benchmark tests`

## Why is it important?

Help to run the benchmark on a PR basis using a GitHub comment can facilitate the CI interactions.

## Questions

- [ ] Can we run this stage safely for the PRs? Or will it mess the benchmark data?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -217,7 +217,7 @@ pipeline {
                 branch 'master'
                 branch "\\d+\\.\\d+"
                 branch "v\\d?"
-                tag "v\\d+\\.\\d+\\.\\d+*"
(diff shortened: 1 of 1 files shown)


v1v:
@felixbarny 

> What we'll also need then is the ability to filter the dashboards based on the branch.

Do you want me to add this within this PR or could it be done in a follow up?

felixbarny:
Let's do it in a follow-up, thanks! 🙂

PR #1218, opened 2020-06-05
Fix failing tests due to url-encoded file paths

SylvainJuge:
Of course, this only fails on internal CI otherwise life is too simple to debug.

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/util/IOUtils.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/util/IOUtils.java
@@ -32,7 +32,6 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.math.BigInteger;
-import java.net.URLDecoder;
 import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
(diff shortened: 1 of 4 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/util/IOUtils.java
```diff
@@ -221,11 +220,7 @@ public static synchronized File exportResourceToTemp(String resource, String tem
                 throw new IllegalStateException(resource + " not found");
             }
             String hash = md5Hash(IOUtils.class.getResourceAsStream("/" + resource));
-
-            // temp folder might be url-encoded
-            String tempFolder = URLDecoder.decode(System.getProperty("java.io.tmpdir"), StandardCharsets.UTF_8.name());
-
-            File tempFile = new File(tempFolder, tempFileNamePrefix + "-" + hash + tempFileNameExtension);
+            File tempFile = new File(System.getProperty("java.io.tmpdir"), tempFileNamePrefix + "-" + hash + tempFileNameExtension);
```
SylvainJuge, line 223: this just reverts to the previous changes applied directly to master branch.

SylvainJuge:
I've managed to reproduce it locally simply by having a space in the path of the project folder (because it's what happens when you have a space in your jenkins job name).

With that I've found another place where it was failing in a similar way, and we can now re-enable some tests that were skipped on CI due to this :tada:

PR #1223, opened 2020-06-09
wip: added deactivate for Span in public api

### LINKED ISSUE #780: Add Span#deactivate to the public API
Currently, the public API does not allow for `deactivation` of active Transactions and Spans arbitrarily. The only way to deactivate is by keeping the `Scope` reference returned from the `activate` API. 

An example where this would be required - [this discuss topic](https://discuss.elastic.co/t/spring-kafka-instrumentation-kafkalistenerendpointcontainer-mono-thread/193810).

kananindzya:
closes #780 
[ ] tests
[ ] docs

### DIFF
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/AbstractSpanImpl.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/AbstractSpanImpl.java
@@ -138,6 +138,11 @@ public abstract class AbstractSpanImpl implements Span {
     }
 
     @Override
+    public void deactivate() {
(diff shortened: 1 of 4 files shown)


felixbarny:
I'm not convinced we should have that API. Wrong Scope management is one of the most common sources of error when doing manual instrumentation. The safest thing is to always use activations in combination with try-with-resources.

kananindzya:
> I'm not convinced we should have that API. Wrong Scope management is one of the most common sources of error when doing manual instrumentation. The safest thing is to always use activations in combination with try-with-resources.

Yes, indeed, then I close pr?

felixbarny:
Yes, let's to that for now. I'll raise my concerns in the issue (should have done that earlier).

Thanks anyways 🙂

PR #1236, opened 2020-06-16
Move CI to Java 11

cachedout:
Move the pipeline to Java 11.

PR #1237, opened 2020-06-16
Fix javadoc error on Java 11

felixbarny:
Prerequisite for #1236


SylvainJuge:
For the record, switching to java11 from java 10 for build also requires to add `<detectJavaApiLink>false</detectJavaApiLink>` (https://github.com/elastic/apm-agent-java/pull/1307/commits/e988686691ee6477b18235afe52b85ed2dfa1fde) to make it work, otherwise we still get the following error:

```
Exit code: 1 - javadoc: error - The code being documented uses packages in the unnamed module, but the packages defined in https://docs.oracle.com/en/java/javase/11/docs/api/ are in named modules.
```

PR #1238, opened 2020-06-16
fix: check that docker is installed

kuisathaverat:
## What does this PR do?

It checks that Docker is installed before run it.

## Why is it important?

Some worker types do not have Docker installed this breaks the packer build.

### DIFF
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -5,7 +5,7 @@ JAVA_HOME=$HOME/.java/java10 ./mvnw clean verify \
   --fail-never -q -B \
   -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
 
-if [ -f "$(command -v docker)" ]; then
+if [ -x "$(command -v docker)" ]; then
   (retry 2 docker pull docker.elastic.co/observability-ci/weblogic:12.2.1.3-dev) \
     && docker tag docker.elastic.co/observability-ci/weblogic:12.2.1.3-dev store/oracle/weblogic:12.2.1.3-dev
 fi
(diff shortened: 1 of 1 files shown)


.ci/packer_cache.sh
```diff
@@ -5,5 +5,7 @@ JAVA_HOME=$HOME/.java/java10 ./mvnw clean verify \
   --fail-never -q -B \
   -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
 
-(retry 2 docker pull docker.elastic.co/observability-ci/weblogic:12.2.1.3-dev) \
-  && docker tag docker.elastic.co/observability-ci/weblogic:12.2.1.3-dev store/oracle/weblogic:12.2.1.3-dev
+if [ -f "$(command -v docker)" ]; then
```
kuisathaverat, line 8:
```suggestion
if [ -x "$(command -v docker)" ]; then
```

felixbarny:
> Some worker types do not have Docker installed this breaks the packer build.

Isn't that what's supposed to happen? We need the worker to have Docker installed. We wouldn't want a packer to be built without the cached docker images.

kuisathaverat:
>Isn't that what's supposed to happen? We need the worker to have Docker installed. We wouldn't want a packer to be built without the cached docker images.

This affects to workers are not used to build the Java agent, but because the job to build all packer images for the apm-ci is the same we have to make the check on every packer script. So nothing changes for the Java agent.

felixbarny:
As part of the test which are run in the packer cache script, the agent also pulls docker containers.

Or are you saying the packager for the Java agent is guaranteed to have Docker installed? But why do we need to guard for it then?

kuisathaverat:
> As part of the test which are run in the packer cache script, the agent also pulls docker containers.
Or are you saying the packager for the Java agent is guaranteed to have Docker installed? But why do we need to guard for it then?

The Java agent issues Ubuntu 18.04 workers those workers have Docker, when we build the image that has Docker so they add the packer cache related to the Java Agent. But we have more worker types more centOS (6,7,8), OracleLinux, Ubuntu (12.04, 14.04) some of them do not have Docker but they run the packer script so we have to make the script to work in all our workers.

kuisathaverat:
@felixbarny Are you ok to merge this?

PR #1243, opened 2020-06-17
Missing version tag

OrangeDog:


eyalkoren:
This config actually existed before, only it wasn't documented, so if you try to use it with older versions you can.
Thanks for keeping an eye on us and reporting though! 🙂

OrangeDog:
Ah, so that's what "exposing" meant.

PR #1244, opened 2020-06-18
Relaxing k8s pod ID discovery regex

eyalkoren:


felixbarny:
This is something all agents need to do, right?

Could you create a PR to update https://github.com/elastic/apm/blob/master/docs/agents/agent-development.md#containerkubernetes-metadata?

Maybe we should extract the regexes to a Gherkin test?

eyalkoren:
Definitely needs to be looked at by all agents, I am still waiting for the customer's input about where it is expected and the additional input about the other Minikube pattern (I ended up "relaxing" the regex, but if it's only the first, we could have kept it stricter, allowing one or more `kubepods/`, so I want to know).

felixbarny:
relates to https://github.com/elastic/apm/pull/284

PR #1245, opened 2020-06-18
Add enabled config

### LINKED ISSUE #1086: Add ENABLED config
See elastic/apm#92

felixbarny:
closes #1086


apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -172,6 +181,7 @@ public void onChange(ConfigurationOption configurationOption, Object oldValue, O
                 }
             });
         }
+        tracer.start();
```
eyalkoren, line 184: As discussed offline, I think it's better if there's only one path to start the tracer from where it's built, rather than here

PR #1258, opened 2020-06-29
Avoid missing jboss.as MBeans on JBoss

felixbarny:
let JBoss init the platform MBean server

Fixes an issue reported in https://discuss.elastic.co/t/no-jmx-metrics-in-jboss-eap-7-3-0/238438

## Checklist

- [x] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else


apm-agent-plugins/apm-jmx-plugin/src/main/java/co/elastic/apm/agent/jmx/JmxMetricTracker.java
```diff
@@ -186,20 +192,28 @@ private void registerMBeanNotificationListener(final MBeanServer server) {
...
+                        logger.debug("MBean added at runtime: {}", jmxMetric.getObjectName());
+                        register(Collections.singletonList(jmxMetric), server);
                     }
                 }
             }
         };
         try {
             server.addNotificationListener(MBeanServerDelegate.DELEGATE_NAME, listener, filter, null);
-        } catch (InstanceNotFoundException e) {
+        } catch (Exception e) {
```
SylvainJuge, line 216: why do we need to catch such a broad exception ? Are there common NPE or IllegalStateException thrown ?

felixbarny: No, didn't see any of those in the wild but if there are any I think we should still continue and try monitor already registered MBeans than to give up jmx monitoring entirely.

In theory, this could happen if there's a customer MBean server that does not allow to add notification listeners.

SylvainJuge:
For the record, here are some references to this "statistics=pool" JMX attributes in JBoss/Wildfly codebase (in tests)

[DataSourcePoolStatisticsTestCase](https://github.com/wildfly/wildfly/blob/14f01528af29a9d162ecd452f8c24f8044c5ccda/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/statistics/DataSourcePoolStatisticsTestCase.java#L185
)
[JCAMetrictsTestBase](https://github.com/wildfly/wildfly/blob/1bf993be72b570a0c780a58ba05cabafe1ce7935/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/metrics/JCAMetrictsTestBase.java#L49)

Here are 3 possible ways to solve this issue:

1) use polling for JMX beans
:-1: adds another agent thread

2) JBoss specific hack (what this PR currently implements)
:+1: no extra thread
:+1: already implemented
:-1: only covers the exact known reported case, there might be other occurrences somewhere else in the JMX tree

3) Instrument MBeanServer implementations and trigger our own notifications
:-1: extra complexity
:-1: likely to hit implementation differences between application servers (JBoss is fun and OSS, Websphere & Weblogic aren't).
:-1: Does not cover current case as impacted MBeans do not appear to be registered (hence no notification).

Thus for now I would advocate for option (2), if that's OK I'll try to add integration tests so we are sure it's properly covered.
WDYT @felixbarny ?

PR #1260, opened 2020-06-30
fix NPE for Dubbo plugin

CoderLan0668:
## What does this PR do?

To fix NPE for dubbo plugin.
it might be a problem in dubbo, when the application is both provider and consumer,  `context.isConsumerSide()` is invoking on MonitorFilter exit, NPE will be thrown.

But actually, there is no need to call `context.isConsumerSide()`, so we can use `span != null` instead.


CoderLan0668:
@felixbarny sorry, A NPE problem was found in some conditions, i push a commit to fix it

felixbarny:
Could you add this to the release notes?

PR #1261, opened 2020-06-30
Dynamic logger ecs reconfig

eyalkoren:
## Checklist

- [ ] This is a bugfix
- [x] This is a new plugin
  - [x] I have added information to the [Log correlation documentation](https://www.elastic.co/guide/en/apm/agent/java/current/log-correlation.html)
- [ ] This is something else


apm-agent-plugins/apm-log-shader-plugin/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
```diff
@@ -0,0 +1,102 @@
...
+    private final LoggingConfiguration loggingConfiguration;
+    private final String serviceName;
+
+    public AbstractLogShadingHelper(ElasticApmTracer tracer) {
+        this.tracer = tracer;
+        loggingConfiguration = tracer.getConfig(LoggingConfiguration.class);
+        serviceName = tracer.getConfig(CoreConfiguration.class).getServiceName();
+    }
+
+    private static final WeakConcurrentMap<Object, Object> appenderToShadeAppender = new WeakConcurrentMap.WithInlinedExpunction<Object, Object>();
```
felixbarny, line 53: Use the new `WeakMapSupplier.createMap()`. Also, maybe type to `<A, Object>`?

apm-agent-plugins/apm-log-shader-plugin/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Checks whether the given appender is a shading appender, so to avoid recursive shading
+     * @return true if the provide appender is a shading appender; false otherwise
+     */
+    protected abstract boolean isShadingAppender(A appender);
+
+    @Nullable
+    protected abstract A createAndConfigureAppender(A originalAppender);
+
+    // todo: find more accurate service name
```
felixbarny, line 92: you could do `tracer.getMetaData().getService().getName()`

apm-agent-plugins/apm-log-shader-plugin/src/main/java/co/elastic/apm/agent/log/shader/log4j2/Log4j2LogShadingInstrumentation.java
```diff
@@ -0,0 +1,155 @@
...
+import static net.bytebuddy.matcher.ElementMatchers.not;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+
+public abstract class Log4j2LogShadingInstrumentation extends AbstractLogShadingInstrumentation {
+
+    // Logback class referencing is allowed thanks to type erasure
+    @VisibleForAdvice
+    @Nullable
+    public static HelperClassManager<AbstractLogShadingHelper<AbstractOutputStreamAppender<?>>> helperClassManager;
```
felixbarny, line 56: have you considered migrating to indy plugins?

apm-agent-plugins/apm-log-shader-plugin/src/main/java/co/elastic/apm/agent/log/shader/log4j2/helper/Log4j2LogShadingHelper.java
```diff
@@ -0,0 +1,122 @@
...
+            return null;
+        }
+
+        String shadeFile = Utils.computeShadeLogFilePath(logFile);
+
+        EcsLayout ecsLayout = EcsLayout.newBuilder()
+            .setServiceName(getServiceName())
+            .setIncludeMarkers(false)
+            .setIncludeOrigin(false)
+            .setStackTraceAsArray(true)
```
felixbarny, line 92: Although this makes it more readable it probably has to be set to false so it works with the logs intake API which is currently spec'd to assume ND-JSON

apm-agent-plugins/apm-log-shader-plugin/src/main/java/co/elastic/apm/agent/log/shader/logback/helper/LogbackLogShadingHelper.java
```diff
@@ -0,0 +1,93 @@
...
+        RollingFileAppender<ILoggingEvent> shadeAppender = new RollingFileAppender<>();
+        String shadeFile = Utils.computeShadeLogFilePath(originalAppender.getFile());
+        shadeAppender.setFile(shadeFile);
+
+        EcsEncoder ecsEncoder = new EcsEncoder();
+        ecsEncoder.setServiceName(getServiceName());
+        // todo read from configuration??
+        ecsEncoder.setIncludeMarkers(false);
+        ecsEncoder.setIncludeOrigin(false);
+        ecsEncoder.setStackTraceAsArray(true);
```
felixbarny, line 63: probably better to set to false, see above

apm-agent-plugins/apm-log-shader-plugin/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
eyalkoren: Changed to the new sdk API.
As for typing - this is a static map, so I cannot type to an instance generic type

apm-agent-plugins/apm-log-shader-plugin/src/main/java/co/elastic/apm/agent/log/shader/log4j2/Log4j2LogShadingInstrumentation.java
eyalkoren: Yes, I assume I will migrate to that before I un-draft it

apm-agent-plugins/apm-log-shader-plugin/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
eyalkoren: I used that, but what I was really looking for is a way to capture the actual service name of the relevant context. The problem is that ECS-lib layouts store the service name statically.
What do you think of adding a way to register a `ServiceNameProvider#getServiceName` callback to the ECS layout that will be called when the logging event is serialized?

felixbarny: Actually, I think we don't need to set the service name on the ECS logger implementation. When sending the logs to APM Server, we're already sending it as part of the metadata: https://github.com/elastic/apm/blob/f5a3dbfd9bc731bbe6b920f31b4f9cd611cfb70d/docs/agents/log-shipper.md#metadata

eyalkoren: When log correlation is on, the log events should have the right context - they will contain a `trace.id` and `transcation.id` that we may report separately with a different service name

eyalkoren:
I ended up changing a bit the performance-related config - instead of **overriding** the original logs with the ECS-formatted JSONs, I added a config to **replace** the original logs, meaning- those will not be written to and we will write to the `.ecs.json` logs instead. The problem with overriding is that we would need to create an appender which is of the same type as the original appender and copy over the same configurations, so that we provide the same behaviour.
In log4j2 it is actually not possible to create an appender that is incompatible with the original appender writing to the same file, so this is a rigid requirement in this framework.
If we want to override, we can change that later.

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
```diff
@@ -159,8 +159,38 @@ public void assertValid(Boolean value) {
...
+
+    private final ConfigurationOption<Boolean> logShadingReplace = ConfigurationOption.booleanOption()
+        .key("log_shading_replace")
+        .configurationCategory(LOGGING_CATEGORY)
+        .tags("performance")
+        .description("By default, when Log Shading is enabled, application logs will be duplicated so that the \n" +
+            "ECS-formatted logs are written to new files having the `.ecs.json` extension. In order to reduce the \n" +
+            "related overhead, set this option to true to replace the original log files with the ECS-compatible ones.")
+        .dynamic(false)
+        .buildWithDefault(false);
```
felixbarny, line 179: This overlaps/conflicts with https://github.com/elastic/apm/issues/373 as the replacement should not only happen for file appenders but for any appenders that use a plain-text formatter.
We might have to do that at the time the formatter is set on the appender as we need the appender name to set the `event.dataset` (`${service.name}.${appender.name}`). This implies that it would not work for runtime attachment. But I think that's fair as it ensures that there's no mix of plain-text and JSON logs.

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
```diff
@@ -159,8 +159,38 @@ public void assertValid(Boolean value) {
         })
         .buildWithDefault(false);
 
+    private final ConfigurationOption<Boolean> logShadingEnabled = ConfigurationOption.booleanOption()
```
felixbarny, line 162: Let's make it clear in the docs that log shading is experimental and may be removed in future releases and that we recommend replacing the plain-text formatters with ECS formatters (elastic/apm#373).
Should we make the log shading options internal?

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/test/java/co/elastic/apm/agent/log/shader/LogShadingInstrumentationTest.java
```diff
@@ -0,0 +1,241 @@
...
+     * Disabled by default, as this is a very slow test. Can be used for manual testing of shade file rolling.
+     * Note: logback and log4j2 rollover before appending an event, which means the two log files will contain messages.
+     * As opposed to those, log4j1 rolls over after appending an event, which means that the active log file (log4j1.log)
+     * will be empty when the test ends.
+     */
+     //@Test
+     public void testShadeLogRolling() throws IOException {
+         when(config.getConfig(LoggingConfiguration.class).getLogFileSize()).thenReturn(100L);
+         logger.trace("First line");
+         sleep();
```
felixbarny, line 217: An idea to make the test faster
```suggestion
         await().untilAsserted(() -> new File(getShadeLogFilePath()).length() == 0);
```

felixbarny (commented):
I'm fine with merging this after the comments in LoggingConfiguration have been addressed. As we've discussed, I see the future in elastic/apm#373 (replacing formatters) rather than forking/shadowing the log stream to another file. But I'm fine to have that option implemented in just the Java agent so that we can get feedback from users and so that we can use this PR as a basis for elastic/apm#373 as the general setup of injecting ecs-logging-java is pretty much the same.
However, if the type of feedback we get doesn't justify adding the shadowing/forking feature in all other agents, we should eventually remove the feature rather than leaving it experimental forever or promoting it to beta or GA.

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/test/java/co/elastic/apm/agent/log/shader/LogShadingInstrumentationTest.java
eyalkoren: Unfortunately, this is only possible for log4j1, which rolls the file at the end of the event logging. 
It is also possible in log4j2 if I set `immediateFlush=false`, but this fails all other tests. I used two different loggers, one for the rolling test (with `immediateFlush==false`) and one for the rest, but it ended up being very close in duration to using a sleep of 50ms, which seems safe in log4j2. 
Logback required a very long sleep, but it works well and fast with `immediateFlush==false`.

Eventually I got to a state where these tests take 200-600ms, which is good enough to enable by default. If they will prove to be flaky we can extend the sleep a bit or disable.

eyalkoren: Already proved flaky in the first test 🙁 
I raised sleep time to 4X. Let's see where it takes us.

eyalkoren: This test was very useful for me to test our log file rolling configurations. It will be as useful in the future if we would like to change something like our rolling-decision strategy. However, it is not as important for continuous-integration testing, which is why I disabled it and still left it. 
It now adds about 1.5 seconds to the build for the three logging frameworks combined, so as long as it is not flaky, we can leave it on and disable later if required.

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
eyalkoren: We decided on using a new config option called `log_ecs_reformatting` that will have multiple options:
- OFF
- SHADE - writing to an additional file in ECS format
- REPLACE - reformat to ECS like SHADE, but block writing the original log file
- OVERRIDE - use the same logger output stream, but override with ECS formatting (the option you are referring to, which will be relevant to non-file-appenders as well)

This way they will not overlap or conflict.

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
eyalkoren: The new `log_ecs_reformatting` config will be marked experimental, so it will give us time to gather feedback on how useful these options are without worrying about removing them in the future.
My next task in this area will be the addition of the `OVERRIDE` option, where I can add a note that it is preferred if we decide so.

eyalkoren:
OK, at last ready for a review. Supported technologies page is updated with supported versions (and the addition of Logback). Legacy tests use the earliest-supported version and there is a test for a recent version. For log4j1, the earliest supported version is the latest version, as it is the first compiled with Java 1.4.

Also worth noticing: the ECS `service.name` field is set with what should be the most relevant service name, meaning with preference to the the service name of the current active span, if there is such, otherwise the configured service name. Assuming that logs are not ingested not through APM, this should allow the best correlation between log events and APM event.

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
```diff
@@ -159,8 +159,53 @@ public void assertValid(Boolean value) {
         })
         .buildWithDefault(false);
 
+    private final ConfigurationOption<LogEcsReformatting> logEcsReformatting = ConfigurationOption.enumOption(LogEcsReformatting.class)
+        .key("log_ecs_reformatting")
+        .configurationCategory(LOGGING_CATEGORY)
+        .tags("added[1.22.0]", "experimental")
+        .description("Specifying whether and how the agent should automatically reformat application logs \n" +
+            "into ECS-compatible JSON, suitable for ingestion into Elasticsearch for further Log analysis. \n" +
```
felixbarny, line 167: I have not tested the ref works but I think it makes sense to link to the ecs docs so people can learn more about it.
```suggestion
            "into {ecs-logging-ref}/index.html[ECS-compatible JSON], suitable for ingestion into Elasticsearch for further Log analysis. \n" +
```

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/Utils.java
```diff
@@ -0,0 +1,80 @@
...
+import co.elastic.apm.agent.logging.LoggingConfiguration;
+
+import javax.annotation.Nullable;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class Utils {
+
+    private static final String SHADE_FILE_EXTENSION = ".ecs.json";
+    private static final LoggingConfiguration config = GlobalTracer.requireTracerImpl().getConfig(LoggingConfiguration.class);
```
felixbarny, line 37: As this class has a dependency (`LoggingConfiguration`), I'd prefer it not to be a static utility class but to have a constructor argument for `LoggingConfiguration`.
It'd simplify `UtilsTest` (or at least make it more lightweight) in that it doesn't need to extend from `AbstractInstrumentationTest` and you could just inject a mock instance for  `LoggingConfiguration`.

Also, consider creating a constructor argument for `LoggingConfiguration` in `AbstractLogShadingHelper`. Instead of making the implementations singletons, have a look at `co.elastic.apm.agent.jdbc.JdbcInstrumentation#getJdbcHelper`.

By using that pattern, the fact that advices are static does not leak into the rest of the codebase which can use regular constructor-based "dependency-injection". I think it makes sense to follow a similar pattern throughout the codebase so that it's more cohesive and can serve as a reference for new plugins.

felixbarny: Actually, I think `co.elastic.apm.agent.log4j2.Log4j2LogShadingHelper#instance` can probably stay as-is and just change `private static final Log4j2LogShadingHelper INSTANCE = new Log4j2LogShadingHelper();` to `private static final Log4j2LogShadingHelper INSTANCE = new Log4j2LogShadingHelper(GlobalTracer.requireTracerImpl());`

I like the fact that it's a final field rather than a volatile field in `JdbcInstrumentation`

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
```diff
@@ -0,0 +1,146 @@
...
+
+    private final ElasticApmTracer tracer;
+    private final LoggingConfiguration loggingConfiguration;
+    @Nullable
+    private final String configuredServiceName;
+
+    public AbstractLogShadingHelper() {
+        this.tracer = GlobalTracer.requireTracerImpl();
+        loggingConfiguration = tracer.getConfig(LoggingConfiguration.class);
+        configuredServiceName = new ServiceFactory().createService(tracer.getConfig(CoreConfiguration.class), "").getName();
```
felixbarny, line 62: Why not like this?
```suggestion
        configuredServiceName = tracer.getConfig(CoreConfiguration.class).getServiceName();
```

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Nullable
+    protected abstract A createAndConfigureAppender(A originalAppender, String appenderName);
+
+    @Nullable
+    protected String getServiceName() {
+        String serviceName = configuredServiceName;
+        AbstractSpan<?> active = tracer.getActive();
+        if (active != null) {
+            String runtimeServiceName = active.getTraceContext().getServiceName();
```
felixbarny, line 133: This may yield different service names for the same webapp depending on whether the log has been created inside or outside of an active transaction.
Not sure what the alternative is, though.

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/Utils.java
eyalkoren: `Utils` is a static utility by its nature. The argument of having a "dependency" shouldn't interfere with that. Depending on the logging framework as in `LoggerFactory.getLogger()`, for example, would be the same but would not exclude a class from being a static utility. 
Doing what you suggest can create the exact opposite problem of enforcing the utility user to introduce this dependency, which it might don't even need for anything else. And given that it is a utility, this will affect multiple users.

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
eyalkoren: Because it is a bit more future-proof. I want to get here the exact same service name that is reported in the metadata, so if in the future we need to make it all-capital, or append a prefix, or apply any logic instead of only using the configured name as is, it will still remain in sync.

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
eyalkoren: Yes, I am aware and agree this is not ideal. However, don't forget that when logging correlation is enabled, we also add `trace.id` and `transaction.id`. What I tried to do is keep these Log events in sync with the APM events that we report in that context. So I guess the question is: what's more important - the sync with APM events or the ability to aggregate for a single service name in the logging app?

eyalkoren: (I honestly don't have an answer, I think you are in a better position to judge, so we'll go with what you think is better)

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/Utils.java
felixbarny: IMHO, a static utility should not depend on another class that is stateful. But what do you think about just adding a `String shadeFileDestinationDir` parameter that the caller can read from `config.getLogEcsFormattingDestinationDir()`?

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
felixbarny: One thing I missed is that `getServiceName` only gets called in the context of `co.elastic.apm.agent.log.shader.AbstractLogShadingHelper#createAndConfigureAppender` which is only called once, when the first log after instrumenting the appender is logged.

This may or may not be within a transaction. Likely, it happens before the service name can be determined for a `.war` web application or before being able to detect the `spring.application.name`.

It's tricky. But at least, the service name should stay consistent throughout the lifetime of the application.

apm-agent-plugins/apm-log-shader-plugin/apm-log4j1-plugin/src/main/java/co/elastic/apm/agent/log4j1/Log4j1LogShadingInstrumentation.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return named("org.apache.log4j.WriterAppender");
+    }
+
+    public static class ShadingInstrumentation extends Log4j1LogShadingInstrumentation {
+
+        @Override
+        public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+            return named("subAppend").and(takesArgument(0, named("org.apache.log4j.spi.LoggingEvent")));
```
felixbarny, line 65: Javadoc links to the instrumented method would be helpful on all instrumentations 🙂

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
eyalkoren: That is correct, and now I recall we already had this discussion 🙂 
If we think it's important enough to maintain consistency (for future purposes as well), we can add the service name to the MDC and in the ECS logger first look for such key and only if it is not there, use the one configured during the appender configuration.
Really, it's your call (sorry, I didn't understand from the last comment if made this decision)

apm-agent-plugins/apm-log-shader-plugin/apm-logback-plugin/apm-logback-plugin-impl/src/main/java/co/elastic/apm/agent/logback/LogbackLogShadingHelper.java
```diff
@@ -0,0 +1,107 @@
...
+        rollingPolicy.setContext(defaultLoggerContext);
+        rollingPolicy.start();
+        shadeAppender.setRollingPolicy(rollingPolicy);
+
+        SizeBasedTriggeringPolicy<ILoggingEvent> triggeringPolicy = new SizeBasedTriggeringPolicy<>();
+        try {
+            VersionUtils.setMaxFileSize(triggeringPolicy, getMaxLogFileSize());
+        } catch (Throwable throwable) {
+            // We cannot log here because this plugin escapes slf4j package reallocation.
+            System.out.println("Failed to set max file size for log shader file-rolling strategy. Using the default " +
```
felixbarny, line 81: why do we have to do the shading trick for slf4j on this module? could we just remove this portion of `apm-logback-plugin-impl/pom.xml`? It does not use any slf4j classes in a context where we need to reference the target cl's slf4j.

```xml
<relocation>
    <pattern>org.slf4j</pattern>
    <shadedPattern>co.elastic.apm.agent.logshading.shaded.slf4j</shadedPattern>
</relocation>
```

docs/configuration.asciidoc
```diff
@@ -702,7 +704,7 @@ you should add an additional entry to this list (make sure to also include the d
...
+Valid options are `annotations`, `apache-commons-exec`, `apache-httpclient`, `asynchttpclient`, `bootdelegation`, `concurrent`, `dubbo`, `elasticsearch-restclient`, `exception-handler`, `executor`, `executor-collection`, `experimental`, `fork-join`, `grails`, `grpc`, `hibernate-search`, `http-client`, `jax-rs`, `jax-ws`, `jdbc`, `jdk-httpclient`, `jedis`, `jms`, `jsf`, `kafka`, `lettuce`, `log4j`, `log4j1`, `log4j2`, `logback`, `logging`, `micrometer`, `mongodb-client`, `okhttp`, `opentracing`, `process`, `public-api`, `quartz`, `rabbitmq`, `redis`, `redisson`, `render`, `scala-future`, `scheduled`, `servlet-api`, `servlet-api-async`, `servlet-api-dispatch`, `servlet-input-stream`, `slf4j`, `spring-mvc`, `spring-resttemplate`, `spring-service-name`, `spring-view-render`, `ssl-context`, `timer-task`, `urlconnection`.
```
felixbarny, line 707: There's already a `log4j` instrumentation group name for `Log4j2LoggerErrorCapturingInstrumentation`. It's hard to know from just looking at this list what the difference between `log4j` and `log4j1` is.

apm-agent-plugins/apm-log-shader-plugin/apm-log4j1-plugin/src/main/java/co/elastic/apm/agent/log4j1/Log4j1LogShadingHelper.java
```diff
@@ -0,0 +1,91 @@
...
+    protected WriterAppender createAndConfigureAppender(WriterAppender originalAppender, String appenderName) {
+
+        RollingFileAppender shadeAppender = null;
+        if (originalAppender instanceof FileAppender) {
+            try {
+                FileAppender fileAppender = (FileAppender) originalAppender;
+                String shadeFile = Utils.computeShadeLogFilePath(fileAppender.getFile());
+
+                EcsLayout ecsLayout = new EcsLayout();
+                ecsLayout.setServiceName(getServiceName());
```
felixbarny, line 68: It's more of an idea or a thing to discuss but I think it makes sense to add the `appenderName` to the event dataset so that it's possible to filter by different log streams of an application.
```suggestion
                ecsLayout.setServiceName(getServiceName());
                ecsLayout.setEventDataset(getServiceName() + "." + appenderName);
```

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
```diff
@@ -0,0 +1,146 @@
...
+        this.tracer = GlobalTracer.requireTracerImpl();
+        loggingConfiguration = tracer.getConfig(LoggingConfiguration.class);
+        configuredServiceName = new ServiceFactory().createService(tracer.getConfig(CoreConfiguration.class), "").getName();
+    }
+
+    private static final WeakConcurrentMap<Object, Object> appenderToShadeAppender = WeakMapSupplier.createMap();
+
+    @Nullable
+    public A getOrCreateShadeAppenderFor(A originalAppender) {
+        if (isShadingAppender(originalAppender)) {
```
felixbarny, line 69: Also return `null` if the appender already uses an ECS encoder/layout.

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
felixbarny: Got it. I think to make it even more future-proof and to solve the issue about which service name to use, we should expose a `ElasticApmTracer#getServiceName(ClassLoader)` method. In contrast to what the existing private method does, it should get the service name from the `MetaData` if there's no override for the provided class loader. We can probably get the class loader of the application by calling `Thread.currentThread().getContextClassLoader()`.

One complication is that the metadata is inside a future, but that should be resolvable as the service name doesn't depend on the cloud metadata.

Another issue is that the service name might not be known that early in the application startup. For example, the detection of the `spring.application.name` happens after the first event has been logged. But at least for spring boot, there seems to be a fix. We can instrument either `org.springframework.boot.SpringApplication#prepareEnvironment` or `org.springframework.boot.context.config.ConfigFileApplicationListener#onApplicationEvent`, or similar which allows us to read `environment.getProperty("spring.application.name")` before the first event is logged.

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
felixbarny: I think it's important that the service name stays consistent regardless of whether or not a log is emitted during a transaction. So setting it once when creating the ecs formatter/encoder does make sense. But I think we can refine the way we gather the service name. See also https://github.com/elastic/apm-agent-java/pull/1261/files#r587523067

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/Utils.java
eyalkoren: There is such method already: https://github.com/elastic/apm-agent-java/pull/1261/files#diff-2f5c514f954a475ff220cebc3a8551dce7915534f14d15bcffa86021ad108403R322
but it only returns the configured value or `null`.

The documentation for the `log_ecs_reformatting_dir` config option says: 
```
If `log_ecs_reformatting` is set to `SHADE` or `REPLACE`, 
the shade log files will be written alongside 
the original logs in the same directory by default. 
Use this configuration in order to write the shade logs 
into an alternative destination. Omitting this config or 
setting it to an empty string will restore the default 
behavior. If relative path is used, this path will be used 
relative to the original logs directory.
```
In order to support that, there is some logic required, as well as an input argument for the original log file, so I felt this is not very appropriate to be located within the config class.

apm-agent-plugins/apm-log-shader-plugin/apm-logback-plugin/apm-logback-plugin-impl/src/main/java/co/elastic/apm/agent/logback/LogbackLogShadingHelper.java
eyalkoren: Not sure now, maybe just a copy+paste+easier-to-just-leave 🙂 
I'll remove and see if anything breaks

docs/configuration.asciidoc
eyalkoren: OK, I'll change both

apm-agent-plugins/apm-log-shader-plugin/apm-log4j1-plugin/src/main/java/co/elastic/apm/agent/log4j1/Log4j1LogShadingHelper.java
eyalkoren: Yeah, good idea. I'll apply this suggestion, unless you want to think of an alternative. As long as this is experimental, I think it's OK to start with something that may change later.

apm-agent-plugins/apm-log-shader-plugin/apm-logback-plugin/apm-logback-plugin-impl/src/main/java/co/elastic/apm/agent/logback/LogbackLogShadingHelper.java
eyalkoren: We cannot add an slf4j dependency in the logback module, but we will add a "facade" logging method in the abstract helper that the helper subclasses can use. The common module should be able to have this dependency

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
eyalkoren: We decided to start with leaving it as it is now and see how we enhance it later on. In any case, when the user is setting the `service_name` explicitly, then it will work as expected.

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LogEcsReformatting.java
```diff
@@ -0,0 +1,33 @@
...
+ * #L%
+ */
+package co.elastic.apm.agent.logging;
+
+public enum LogEcsReformatting {
+    OFF,
+    SHADE,
+    REPLACE,
+    // NOT IMPLEMENTED YET
+    //OVERRIDE
```
felixbarny, line 32: As I find it hard to recall what the difference between `REPLACE ` and `OVERRIDE ` is, just some ideas how we could maybe make it clearer.

```suggestion
    SHADE_FILE,
    REPLACE_FILE,
    // NOT IMPLEMENTED YET
    // REFORMAT
```

The `_FILE` options only work for file appenders while `REFORMAT` works with potentially any appender.

felixbarny (approved):
We might have to refine how the `service.name` is computed but let's not block this PR on that. It's an experimental feature anyway.

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LogEcsReformatting.java
eyalkoren: I actually prefer the simplicity in enum names, especially ones you use for configuration, so I think these single-verb-word names make sense. I see what you say about the meaning of each, but I don't think this will make a big change - in any case you would read the documentation and this is where I put efforts to very clearly explain what each does.
In addition, `shade` is currently only applied for files because this is how I implemented it, but it may do the same for console/stdout appenders as well if required for the same reason (keep existing log stream but also ingest ECS format of it to ES).

PR #1262, opened 2020-06-30
Delay init by default in Tomcat and JBoss, only once

### LINKED ISSUE #1246: Bug: incompatibility with OpenJDK 1.8.0_242 (AdoptOpenJDK)
## Describe the bug
java-agent 1.17.0 is incompatible with OpenJDK 1.8.0_242 (AdoptOpenJDK).
if I launched a tomcat server (8.5.54) with java-agent 1.17.0, and OpenJDK 1.8.0_242 (AdoptOpenJDK), web applications are not started.
If I run the same tomcat server with same apps, but with OpenJDK 1.8.0_212 (AdoptOpenJDK) or Oracle JDK 1.8.0_202, it works fine.

## Steps to reproduce
Just start tomcat server with above configuration.

## Expected behavior
tomcat starts cleanly and especially web apps are loaded.

## Debug logs
I see no tomcat log at all at startup (when used with OpenJDK 1.8.0_212, I get tomcat startup logs).

APM logs:
```
2020-06-19 13:45:50,557 [main] INFO  co.elastic.apm.agent.util.JmxUtils - Found JVM-specific OperatingSystemMXBean interface: com.sun.management.OperatingSystemMXBean
2020-06-19 13:45:52,384 [main] INFO  co.elastic.apm.agent.configuration.StartupInfo - Starting Elastic APM 1.17.0 as my-project on Java 1.8.0_242 (AdoptOpenJDK) Linux 2.6.32-754.29.2.el6.x86_64
2020-06-19 13:45:52,475 [main] INFO  co.elastic.apm.agent.impl.ElasticApmTracer - Tracer switched to RUNNING state
```

eyalkoren:
## What does this PR do?
Fixes #1246 (main issue is the infamous JUL deadlock, in addition - a bug of multiple delays).

## Checklist

- [x] This is a bugfix
  - ~[ ] I have added tests that would fail without this fix~ This one is tested manually
- [ ] This is a new plugin
- [ ] This is something else


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -538,28 +539,35 @@ void init(List<LifecycleListener> lifecycleListeners) {
...
+    }
+
+    private synchronized void startWithDelay(final long delayInitMs) {
         ThreadPoolExecutor pool = ExecutorUtils.createSingleThreadDeamonPool("tracer-initializer", 1);
         pool.submit(new Runnable() {
             @Override
             public void run() {
                 try {
-                    long delayInitMs = coreConfiguration.getDelayInitMs();
                     logger.info("Delaying initialization of tracer for " + delayInitMs + "ms");
```
felixbarny, line 564: Can we end the delay once the platform MBean server is initialized (`!MBeanServerFactory.findMBeanServer(null).isEmpty()`)? In that case, it should be guaranteed that JUL is initialized.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -538,28 +539,35 @@ void init(List<LifecycleListener> lifecycleListeners) {
...
+        if (delayInitMs > 0) {
+            startWithDelay(delayInitMs);
         } else {
             startSync();
         }
     }
 
-    private synchronized void startWithDelay() {
+    private boolean shouldDelay() {
+        return ClassLoader.getSystemClassLoader().getResource("org/apache/catalina/startup/Bootstrap.class") != null;
```
felixbarny, line 555: Can we do the delay for specific versions of Java, like only 1.8?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
eyalkoren: You mean poll every 500ms or so? It still doesn't guarantee anything.
I think we can live with the 5 seconds delay in Tomcat on 1.8 installed via premain

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -538,28 +539,35 @@ void init(List<LifecycleListener> lifecycleListeners) {
         }
     }
 
-    public synchronized void start() {
-        if (getConfig(CoreConfiguration.class).getDelayInitMs() > 0) {
-            startWithDelay();
+    public synchronized void start(boolean premain) {
+        long delayInitMs = getConfig(CoreConfiguration.class).getDelayInitMs();
+        if (shouldDelay()) {
+            delayInitMs = Math.max(delayInitMs, 5000L);
```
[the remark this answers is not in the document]
eyalkoren: Logging this info in the delay itself. We will be able to see there is no `delay_instrumentation` configured, so we should be able to know that

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
felixbarny: You'd only wait up to 500ms more than necessary vs waiting up to 5s more than necessary and possibly not waiting long enough. We could use the same mechanism for `delay_init` and the MBean server polling. Both conditions need to be fulfilled to end the delay.

> I think we can live with the 5 seconds delay in Tomcat on 1.8 installed via premain

I think so too, just mentioning for completeness

PR #1263, opened 2020-06-30
Re-enable non-Scala tests

felixbarny:
https://github.com/elastic/apm-agent-java/pull/1048 accidentally disabled all non-Scala tests


milanvdm:
@felixbarny I think this will also disable the Scala tests again

felixbarny:
You're right! Should be fixed by the latest commit.

PR #1269, opened 2020-07-02
Instrumenting LogManager to overcome JUL deadlock

### LINKED ISSUE #1246: Bug: incompatibility with OpenJDK 1.8.0_242 (AdoptOpenJDK)

eyalkoren:
## What does this PR do?
Supersedes #1262 
Fixes #1246 (main issue is the infamous JUL deadlock, in addition - a bug of multiple delays).

## Checklist

- [x] This is a bugfix
  - ~[ ] I have added tests that would fail without this fix~ Tested manually
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-plugins/apm-jul-deadlock-plugin/src/main/java/co/elastic/apm/agent/jul/LogManagerInstrumentation.java
+++ b/apm-agent-plugins/apm-jul-deadlock-plugin/src/main/java/co/elastic/apm/agent/jul/LogManagerInstrumentation.java
@@ -1,0 +1,89 @@
+/*-
+ * #%L
+ * Elastic APM Java agent
+ * %%
+ * Copyright (C) 2018 - 2020 Elastic and contributors
+ * %%
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * #L%
+ */
+package co.elastic.apm.agent.jul;
+
+import co.elastic.apm.agent.bci.ElasticApmInstrumentation;
+import co.elastic.apm.agent.bci.VisibleForAdvice;
+import co.elastic.apm.agent.util.GlobalLocks;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+import java.util.Collection;
+import java.util.Collections;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public class LogManagerInstrumentation extends ElasticApmInstrumentation {
+
+    /**
+     * A state making sure we lock when JUL is used only at startup. No need to be volatile, better lock few too many
+     * times then making volatile accesses for the entire JVM lifetime
+     */
+    @VisibleForAdvice
+    public static boolean initialized = false;
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return named("java.util.logging.LogManager");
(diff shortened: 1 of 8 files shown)


apm-agent-plugins/apm-jul-deadlock-plugin/src/main/java/co/elastic/apm/agent/jul/LogManagerInstrumentation.java
```diff
@@ -0,0 +1,89 @@
...
+        return LogManagerAdvice.class;
+    }
+
+    public static class LogManagerAdvice {
+
+        @Advice.OnMethodEnter(suppress = Throwable.class)
+        public static boolean lock() {
+            boolean locked = false;
+            if (!initialized) {
+                GlobalLocks.JUL_INIT_LOCK.lock();
```
SylvainJuge, line 75: Why use global lock here ? If it also covers both agent usage through http client & jmx and other third-party libraries we could make this lock local to the plugin.

apm-agent-plugins/apm-jul-deadlock-plugin/src/main/java/co/elastic/apm/agent/jul/LogManagerInstrumentation.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Collections.singletonList("jul");
+    }
+
+    @Override
+    public Class<?> getAdviceClass() {
+        return LogManagerAdvice.class;
```
SylvainJuge, line 66: [style] we could even inline the advice into instrumentation class

apm-agent-plugins/apm-jul-deadlock-plugin/src/main/java/co/elastic/apm/agent/jul/LogManagerInstrumentation.java
```diff
@@ -0,0 +1,89 @@
...
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+import java.util.Collection;
+import java.util.Collections;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public class LogManagerInstrumentation extends ElasticApmInstrumentation {
```
SylvainJuge, line 40: [doc] Maybe add some javadoc with link to instrumented method and a short description of why we do this.

apm-agent-plugins/apm-jul-deadlock-plugin/src/main/java/co/elastic/apm/agent/jul/LogManagerInstrumentation.java
eyalkoren: Right, a matter of habit. I usually like to compile advices in separate classes because they typically have dependencies I don't want in the instrumentation, but not in this case.

apm-agent-plugins/apm-jul-deadlock-plugin/src/main/java/co/elastic/apm/agent/jul/LogManagerInstrumentation.java
eyalkoren: It covers them only if they can lock on it as well

eyalkoren:
Even though the instrumentation of JUL's `LogManager` added in this PR provides full thread-safety between agent and other threads, we still need to also verify we don't execute our JUL-initializing tasks too early, otherwise we don't allow setting a custom `LogManager`.
So we need to delay initialization in this case.
Therefore, closing this PR.

Issue #937, opened 2019-11-21
Add support for community supported plugins


SylvainJuge:
**Issue description**

Community contributions are split in two broad categories
1. agent "core" and common infrastructure code
2. agent "plugins", which provide support for specific frameworks/libraries.

Plugins (2) are the easiest way to add features to the agent and are also the most common contribution type.
We (Elastic Java Agent team) currently take ownership of all those added features and merge them to the plugins, which means asking for changes in pull-requests to meet quality/style standards.

While doing so seems fair, it means that merging contributions takes a considerable effort and time, and that our team is often the bottleneck here, often with back and forth with contributors.
Also, keeping pull-requests open for a while makes them harder to merge.

**Proposed solution**

https://docs.google.com/document/d/1LFTBDLM3-Sj7cU1pHFxfgI3donyTMp8tvjFpyJM7dF8/edit

- Create `apm-agent-plugin-sdk`: used both by internal and external/community plugins
  - agnostic of the tracer API
- External plugins can use the public API or the OpenTracing API bridge
- External plugins are loadable from `$AGENT_HOME/plugins`
- Testing: relies on internal API and internal test helpers such as `AbstractInstrumentationTest` and `MockReporter`
  - Creating and maintaining abstractions that are backwards compatible doesn't seem feasible and will always lag behind
  - The important objective of the plugin being compatible for the users of the plugin is still maintained
  - When the plugin maintainer bumps the agent version, they have to resolve breaking changes in test code
  - Makes it easier to eventually pull external plugins into the core agent as the same testing framework is used.
- Plugin registry
  - Would be just the supported technologies page linking to external GH repos
  - The plugins can be downloaded from the releases page

Tasks
- [x] #1337 
- [x] #1338 
- [x] #1339 
- [x] #1340 
- [x] #2198
- [x] #2199

SylvainJuge:
Also, we currently have "incubating" features (currently JMS is one of them), those are disabled by default because `disable_instrumentations=incubating` in the default configuration.

There are two possible approaches:
1. reuse the same feature and set `disable_instrumentations=incubating,community` by default
2. create a dedicated opt-in option for enabled community instrumentations `enabled_community_instrumentations` that would be empty by default.

Pros/cons
1. will automatically enable community-supported plugins if default value is changed, even if only a few users do, that would be a breaking change
2. adds another configuration option

felixbarny:
I would prefer the explicit opt-in via `enabled_community_instrumentations`

[...]

PR #1271, opened 2020-07-06
First step towards enabling external plugins

felixbarny:
## What does this PR do?
Implements https://github.com/elastic/apm-agent-java/issues/937#issuecomment-651768676

Makes it possible to load external plugins in an isolated class loader. The plugins depend on the `apm-agent-plugin-sdk` and the `apm-agent-api` artifacts to create a jar that can be place it into the `plugins_dir`. An example for such a plugin is in `integration-tests/external-plugin-test`.

As some parts are still missing, this is not mentioned in the changelog yet and the corresponding configuration is marked internal (thus not included in the generated docs).

- Creates `apm-agent-plugin-sdk`: used both by internal and external/community plugins
- I decided against fractoring out a `apm-agent-tracer-impl` module from `apm-agent-core` for now. It was more tricky than expected and not needed for external plugins.

Missing bits before we can open this to the public:
- Remove `ElasticApmPlugin#indyPlugin`, making indy plugins the only option
- stop shading Byte Buddy so that users can depend on `elastic-apm-agent` in their tests to activate `apm-agent-api`
   - agent has to be loaded in its dedicated class loader first to avoid conflicts with the application
- publish test module for apm-agent-core


felixbarny:
In a follow-up, we may want to include some of the `CustomElementMatchers` in the plugin SDK such as `classLoaderCanLoadClass`.

felixbarny:
@elastic/apm-agent-java any changes you want to do in the `ElasticApmInstrumentation` class? After we publish the SDK, we can only add but never remove or rename methods.

One thing we might want to consider, although being a very minor thing, is to remove the `get` prefixes, for example from `getTypeMatcher`. It feels a bit awkward when writing an instrumentation because a type matcher is created, not retrieved when the method is called.

I have also changed the method name `getImplementationVersionPostFilter` to `getProtectionDomainPostFilter` to give it a more generic name.

integration-tests/external-plugin-test/src/main/java/co/elastic/apm/plugin/PluginInstrumentation.java
```diff
@@ -0,0 +1,68 @@
...
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+import java.util.Collection;
+import java.util.Collections;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public class PluginInstrumentation extends ElasticApmInstrumentation {
```
SylvainJuge, line 40: [minor] might also be interesting to have an example that extends `TracerAwareInstrumentation` as relying on API is not the only option.

integration-tests/external-plugin-test/src/main/java/co/elastic/apm/plugin/PluginInstrumentation.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return named(System.getProperty("elastic.apm.plugin.instrumented_class", "co.elastic.apm.plugin.PluginInstrumentationTest"));
```
SylvainJuge, line 44: [minor] if we use this plugin as an example, that's not the canonical way to deal with configuration as it does not provides proper documentation. A comment to explain it's a shortcut might be relevant.

integration-tests/external-plugin-test/src/main/java/co/elastic/apm/plugin/PluginInstrumentation.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
+    public static Object onEnter(@Advice.Origin(value = "#m") String methodName) {
+        return ElasticApm.startTransaction().setName(methodName);
+    }
+
+    @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class, inline = false)
+    public static void onExit(@Advice.Thrown Throwable thrown, @Advice.Enter Object transactionObject) {
+        Transaction transaction = (Transaction) transactionObject;
```
SylvainJuge, line 64: [question] why can't we have `Transaction` directly in the advice signature ? is that to avoid leaking an extra type in the instrumented class stack frames ?

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -325,8 +371,10 @@ public boolean matches(TypeDescription typeDescription, ClassLoader classLoader,
         if (offsetMapping != null) {
             withCustomMapping = withCustomMapping.bind(offsetMapping);
         }
-        if (instrumentation.indyPlugin()) {
-            validateAdvice(instrumentation.getAdviceClass().getName());
+        // external plugins are always indy plugins
+        if (!(instrumentation instanceof TracerAwareInstrumentation)
```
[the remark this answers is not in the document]
felixbarny: The instrumentation might be an instance of `ElasticApmInstrumentation`, indicating that it's an external plugin and thus and indy plugin by default.

PR #1274, opened 2020-07-06
Add scala future test case using multiple threads

henrikno:
## What does this PR do?
I wanted to give the [scala futures implementation](https://github.com/elastic/apm-agent-java/pull/1048) a spin but, hit a snag when using it with multiple threads.
It seems like the transaction is not activated/deactivated at the right time in some cases.
The existing test cases only use a single transaction so it wouldn't catch it.

Sometimes it fails because current tracer != the tracer passed as parameter. 
Also sometimes `num` is different some the transaction name.
Example output:
`thread=35, trace='Transaction7' 00-852e1e7ca03813413465fc7c21b4ab5e-42a231f21779d2aa-01 (4cd1e659) start transaction num=20`
`num` should be the same as the number in the transaction name. It looks like the transaction from one future is still active when starting a different unrelated future.

The test sometimes passes. The Thread.sleep is not necessary for it to fail but makes it happen more often.

Not intended to be merged as-is, just to show the issue.

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-plugins/apm-scala-concurrent-plugin/src/test/scala/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation2Spec.scala
+++ b/apm-agent-plugins/apm-scala-concurrent-plugin/src/test/scala/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation2Spec.scala
@@ -1,0 +1,66 @@
+package co.elastic.apm.agent.scala.concurrent
+
+import java.util.concurrent.Executors
+
+import co.elastic.apm.agent.MockReporter
+import co.elastic.apm.agent.bci.ElasticApmAgent
+import co.elastic.apm.agent.configuration.{CoreConfiguration, SpyConfiguration}
+import co.elastic.apm.agent.impl.transaction.Transaction
+import co.elastic.apm.agent.impl.{ElasticApmTracer, ElasticApmTracerBuilder}
+import munit.FunSuite
+import net.bytebuddy.agent.ByteBuddyAgent
+import org.stagemonitor.configuration.ConfigurationRegistry
+
+import scala.concurrent.duration._
+import scala.concurrent.{Await, ExecutionContext, ExecutionContextExecutor, Future}
+
+class FutureInstrumentation2Spec extends FunSuite {
+
+  private var reporter: MockReporter = _
+  private var tracer: ElasticApmTracer = _
+  private var coreConfiguration: CoreConfiguration = _
+
+  override def beforeEach(context: BeforeEach): Unit = {
+    reporter = new MockReporter
+    val config: ConfigurationRegistry = SpyConfiguration.createSpyConfig
+    coreConfiguration = config.getConfig(classOf[CoreConfiguration])
+    tracer = new ElasticApmTracerBuilder().configurationRegistry(config).reporter(reporter).build
+    ElasticApmAgent.initInstrumentation(tracer, ByteBuddyAgent.install)
+    tracer.start()
+  }
+
+  override def afterEach(context: AfterEach): Unit = ElasticApmAgent.reset()
+
+  test("Scala Future should not propagate the tracing-context to unrelated threads") {
+    implicit val executionContext: ExecutionContextExecutor =
+      ExecutionContext.fromExecutor(Executors.newFixedThreadPool(10))
+
+    def startFuture(num: Int): Future[(Transaction, Int)] = {
+      Future {
+        println(s"thread=${Thread.currentThread().getId}, trace=${tracer.currentTransaction()} before num=$num")
+        val transaction = tracer.startRootTransaction(getClass.getClassLoader).withName("Transaction" + num).activate()
+        println(s"thread=${Thread.currentThread().getId}, trace=${tracer.currentTransaction()} start transaction num=$num")
+        Thread.sleep(10)
+        (transaction, num)
+      }
+    }
+
+    val futures = (1 to 10).map(x =>
+      startFuture(x)
+        .map { case (transaction: Transaction, num: Int) =>
+          Thread.sleep(10)
+          println(s"thread=${Thread.currentThread().getId}, trace=${tracer.currentTransaction()} map1 $num")
+          val x = transaction == tracer.currentTransaction()
+          assertEquals(transaction, tracer.currentTransaction())
+          (transaction, x)
+        })
+
+    val future = Future.sequence(futures)
+
+    val result = Await.result(future, 10.seconds).toList
+    assertEquals(result.forall(x => x._2), true)
+    //    result.foreach(x => x._1.deactivate().end())
+    result.foreach(x => x._1.end())
+  }
+
+}


milanvdm:
I will have a.look but the current implementation is already tested on multithreaded scenarios so I expect you hitting an edge-case rather than multithreads being the issue.

apm-agent-plugins/apm-scala-concurrent-plugin/src/test/scala/co/elastic/apm/agent/scala/concurrent/FutureInstrumentation2Spec.scala
```diff
@@ -0,0 +1,66 @@
...
+  override def afterEach(context: AfterEach): Unit = ElasticApmAgent.reset()
+
+  test("Scala Future should not propagate the tracing-context to unrelated threads") {
+    implicit val executionContext: ExecutionContextExecutor =
+      ExecutionContext.fromExecutor(Executors.newFixedThreadPool(10))
+
+    def startFuture(num: Int): Future[(Transaction, Int)] = {
+      Future {
+        println(s"thread=${Thread.currentThread().getId}, trace=${tracer.currentTransaction()} before num=$num")
+        val transaction = tracer.startRootTransaction(getClass.getClassLoader).withName("Transaction" + num).activate()
```
felixbarny, line 41: This transaction is never deactivated on the thread and can therefore leak into other operations.

A more sensible test scenario would be to start a transaction before doing the multi-threaded operations and check if the transaction successfully propagates to the futures executed in different threads. I've tried this out and it even works when commenting out the instrumentations in `apm-agent-plugins/apm-scala-concurrent-plugin/src/main/resources/META-INF/services/co.elastic.apm.agent.bci.ElasticApmInstrumentation`. That's because Scala relies on the Java concurrent framework that we already instrument.

The instrumentations in this module are only needed when the context gets lost along the way which seems to be the case for some AsyncHttpClient callbacks.

```scala
val transaction = tracer.startRootTransaction(getClass.getClassLoader).withName("Transaction").activate()
val futures = (1 to 10).map(x => Future {
  Thread.sleep(10)
  println(s"thread=${Thread.currentThread().getId}, trace=${tracer.currentTransaction()}")
  val x = transaction == tracer.currentTransaction()
  assertEquals(transaction, tracer.currentTransaction())
  (transaction, x)
})
val future = Future.sequence(futures)
val result = Await.result(future, 10.seconds).toList
transaction.deactivate().end()
assertEquals(result.forall(x => x._2), true)
```

henrikno: > This transaction is never deactivated on the thread and can therefore leak into other operations.
But if I deactivate it, it won't propagate... right?

How do I "activate" the transaction inside the future, so that it's passed on to the next .map() operation, but while deactivating it so that it's not passed on to unrelated operation that happens on the same thread?

Looks like your example is using a single transaction. What I'm trying to simulate is a "http server" scenario, which has multiple transactions corresponding to http requests concurrently. If I take your example and make it run that 10 times in different threads I get the same issue.
https://gist.github.com/henrikno/71e54b7c2d43e63936fd33fdd45411bb
Interestingly it happens less often with ForkJoinPool than with FixedThreadPool.

If the test is wrong, could you help explain how it _should_ be used with multiple "requests"/threads handing different transactions concurrently (each of which will typically spawn nested futures to do the actual work)?

felixbarny: The test in the linked gist looks good. I had to increase the number of threads though. Seems like it does expose some edge cases where the context isn't properly transferred.

To make investigation easier, we'd ideally want a test that's a bit simpler but that still fails.

milanvdm: Ill spend some time on this tomorrow 🙂

milanvdm: @felixbarny I am not sure how the elastic-agent handles the following case for JavaFutures:

1. Create multiple transactions (could be on different or the same threads) - https://gist.github.com/henrikno/71e54b7c2d43e63936fd33fdd45411bb#file-test-scala-L41
2. After you have multiple transactions active, create a new `Future` - https://gist.github.com/henrikno/71e54b7c2d43e63936fd33fdd45411bb#file-test-scala-L43

In this case, the problem seems to be, how does the `Future` from (2) know which transaction to use? 

That's how I currently understand the issue show-cased on the Gist.

felixbarny: There can only be one transaction active on a thread at a time. I think that's maintained as each Future is either executed in its own thread concurrently or steals time from the main thread and executes synchronously.

> how does the Future from (2) know which transaction to use?

The [`map`](https://gist.github.com/henrikno/71e54b7c2d43e63936fd33fdd45411bb#file-test-scala-L43) schedules the `Future` for execution (by calling eventually `Executor#execute`). Whichever transaction is active on the current thread when the `Future` is scheduled is going to be restored when the `Future` is executed.

milanvdm: @felixbarny That `map` is not actually being called on any `Future`. Therefore, it will actually create a new `Future`  and use whatever `Transaction` is active at that point. 

I think that is the exact problem here. Transactions are started in parallel but are not followed up on correctly to pass the correct context.

So somehow, we need to have Transactions created inside of a Future, actually be linked to that Future and continue the logic with `map`, `flatMap` on that Future.

I'm not sure if the above not already happens though, say: `Future(startTransaction()).flatMap(logic)` maybe always runs on the same thread and use the same transaction. But Im not sure about that.

felixbarny: In my tests, I have commented out the scala-specific `Future` instrumentation. The tests do fail more often when doing that but when using a `ForkJoinPool`, it works a bit more reliably.

Calling `map` ends up calling `Executor#execute` and executing a `Future` ends up calling `Runnable#run`. These methods are instrumented by the java-concurrent plugin which captures and restores the context. Something must be going wrong in some cases there.

> So somehow, we need to have Transactions created inside of a Future, actually be linked to that Future and continue the logic with map, flatMap on that Future.

When transactions are activated, they are attached to the current thread. If on that thread, a future gets scheduled for execution via `map`, the transaction gets associated with that future and will be restored when the future is executed.

milanvdm: @felixbarny So since the Future that creates the transaction will be the same Runnable that executes the underlying Futures using that transaction, the context should be passed along correctly?

Thanks for the information as always :)

milanvdm:
@felixbarny  I've spent some time investigating this issue again :)

[Kamon's implementation](https://github.com/kamon-io/Kamon/blob/master/instrumentation/kamon-scala-future/src/main/scala-2.13/kamon/instrumentation/futures/scala/FutureChainingInstrumentation.scala) has some nice descriptions on how and why they instrument certain methods to make the Future context propagation work correctly.

Seems like currently the part of https://github.com/kamon-io/Kamon/blob/master/instrumentation/kamon-scala-future/src/main/scala-2.13/kamon/instrumentation/futures/scala/FutureChainingInstrumentation.scala#L39-L49 is implemented in the `elastic-agent` but the other cases are missing. Might be the cause of certain cases not working atm.

I tried implementing these parts but Im stuck on https://github.com/kamon-io/Kamon/blob/master/instrumentation/kamon-scala-future/src/main/scala-2.13/kamon/instrumentation/futures/scala/FutureChainingInstrumentation.scala#L56 where it cleans a context on a certain match. I cannot figure out how that should be done in the `elastic-agent`.

felixbarny:
That's going to be tricky as the Elastic agent allows to be attached at runtime. At that point, the type initializer of Future has most likely already been executed. Also, the Kamon agent adds interfaces/mixins such as `HasContext` which also can't be done using runtime attachment.

Maybe an alternative could be to avoid context propagation by checking if the future is (as in reference equality (`==`)) `scala.concurrent.Future.unit`.

milanvdm:
@felixbarny Been debugging this a bit more again :)

I have some strange behavior that I have a question around:

The following code is instrumenting a specific method:
```
@Nullable
        public static Object onEnter(@Advice.This Object thiz) {
            logger.warn(promisesToContext.toString());
            AbstractSpan<?> context = promisesToContext.remove(thiz);
            if (context != null) {
                logger.warn("==============");
                logger.warn("ACTIVATE Run on " + Thread.currentThread().getId() + " - " + context);
                context.activate();
                logger.warn(tracer.currentTransaction().toString());
                logger.warn("==============");
                // decrements the reference we incremented to avoid that the parent context gets recycled before the promise is run
                // because we have activated it, we can be sure it doesn't get recycled until we deactivate in the OnMethodExit advice
                context.decrementReferences();
            }
            return context;
 ```
 
 But in the logs I get the following:
 ```
 2021-06-29 22:38:13,721 [ForkJoinPool-1-worker-13] WARN  co.elastic.apm.agent.cache.WeakKeySoftValueLoadingCache - ==============
2021-06-29 22:38:13,721 [ForkJoinPool-1-worker-13] WARN  co.elastic.apm.agent.cache.WeakKeySoftValueLoadingCache - ACTIVATE Run on 30 - '10' 00-97d38655e32835994ec9d4277289de50-0e7cd552f87e9631-01 (7c2f827d)
2021-06-29 22:38:13,721 [ForkJoinPool-1-worker-13] WARN  co.elastic.apm.agent.cache.WeakKeySoftValueLoadingCache - '6' 00-ae27a70994d59099ff6d696c1b290a03-6bb34461a653557e-01 (fef99bd)
2021-06-29 22:38:13,721 [ForkJoinPool-1-worker-13] WARN  co.elastic.apm.agent.cache.WeakKeySoftValueLoadingCache - '6' 00-ae27a70994d59099ff6d696c1b290a03-6bb34461a653557e-01 (fef99bd)
2021-06-29 22:38:13,721 [ForkJoinPool-1-worker-13] WARN  co.elastic.apm.agent.cache.WeakKeySoftValueLoadingCache - ==============
thread=30 transactionNumber=10, trace='6' 00-ae27a70994d59099ff6d696c1b290a03-6bb34461a653557e-01 (fef99bd) before futureNumber=1, 6
```

So even after calling `context.activate();`, the same thread will return a different context when calling `tracer.currentTransaction().toString()`.  I am probably missing some understanding on how the `context.activate()` works?

felixbarny:
You'll need to call `tracer.getActive()`, which returns a `AbstractSpan<?>` (either a `Span` or a `Transaction`). The method `tracer.currentTransaction()` returns the transaction. Transactions are the entry point into the service, whereas spans are operations within that.

milanvdm:
@felixbarny Im not sure how the `getActive` and the `currentTransaction` can give a different result though.

In the following test:
```
val fs = (1 to 10).map(transactionNumber => Future {
      Thread.sleep(10)

      val transaction = tracer.startRootTransaction(getClass.getClassLoader).withName(transactionNumber.toString).activate()

      println(s"thread=${Thread.currentThread().getId} transaction=$transactionNumber, span=${tracer.getActive}, trace=${tracer.currentTransaction()} starting transaction")

      val futures = (1 to 1)
        .map(futureNumber => Future {
          Thread.sleep(10)

          val currentTransactionNumber = tracer.currentTransaction().getNameAsString

          println(s"thread=${Thread.currentThread().getId} transactionNumber=$transactionNumber, span=${tracer.getActive}. trace=${tracer.currentTransaction()} before futureNumber=$futureNumber, $currentTransactionNumber")

          assertEquals(transaction, tracer.currentTransaction())
          assertEquals(currentTransactionNumber.toInt, transactionNumber)

          (transaction, futureNumber)
        }
```
        
Since there are only rootTransactions started, shouldnt both methods always return the same transaction?

The test is failing due to this. It is activating the correct context at the right times. So `getActive` will correctly propagate the contexts between the Futures. But the `currentTransaction` method doesn't return the correct context for some reason.

felixbarny:
Seems you are activating multiple transactions that belong to different traces on the same thread. That's an illegal state but it seems we currently don't guard against that. What happens when calling `currentTransaction()` is that the bottom-most span's transaction is returned. 

https://github.com/elastic/apm-agent-java/blob/48d697eb4504b50b9dcb844ed2d9d4ec10a94a10/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java#L252-L255

That means if you activate multiple transactions, the first activated one will be returned.
I believe the specific issue in your example is that you don't deactivate the transaction after the inner `map` operation.

milanvdm:
@felixbarny It indeed looks like the spans are correctly passed across threads and being linked to the correct `Future`. Meaning that `tracer.getActive()` will give the expected results.

As you mention, the difference is with the `tracer.currentTransaction`. In the current implementation, `tracer.currentTransaction` will always return the transaction that was created on the thread that the `currentTransaction` is run on.

So here I am a bit lost. I've got a solution that correctly activates and deactivates spans across different threads correctly for `getActive`. So I dont see how (and why) `tracer.getActive` and `tracer.currentTransaction` return different results.
How can I deactivate and active transactions for `currentTransaction` in the same way I do for `getActive`?

eyalkoren:
I didn't read through everything, but I hope I can help:
We maintain a stack of transactions/spans for each thread. `getActive` returns the topmost span/transaction. `currentTransaction` returns the bottommost transaction or the parent transaction if the bottom is a span.
A span/transaction is pushed to the stack when you `activate` it and popped when you `deactivate` it. This has nothing to do with where the transaction/span was created.

If you run multiple futures on a thread pool, you must make sure any activated transaction is also deactivated before the thread is returned to the pool to run the next task. Otherwise, the next time you `activate` a transaction on that thread (either explicitly or through the context propagation mechanism), you would have two activated transactions, which is an illegal state and in which case `getActive() != currentTransaction()`.

eyalkoren:
@henrikno a lot has changed since this was opened. Is this still relevant? Would you want to follow up on this, or should we close it for now?

PR #1275, opened 2020-07-08
Validate that tests don't leaks tracers, scopes and pooled objects

felixbarny:
## What does this PR do?

Depends on #1271 

With #1270 we have a good foundation to make sure tracers don't leak to other tests classes. This makes `GlobalTracer` more restrictive so that a tracer can only be set via `init` that throws if the current tracer is not a noop.

I have also added the recycling tests on the `@AfterEach` method of `AbstractInstrumentationTest` which required some adjustments to tests to make sure all transactions and spans are ended and thus recycled after being reported. It also checks the bookkeeper object pool to make sure all objects have returned. This also implicitly tests for scope leaks as those would prevent objects from being recycled.

As a bonus, I've found and fixed a scope leak in the currently unreleased instrumentation of `Executor#executeAll`.

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [x] This is something else: Test improvements
    - [ ] ~I have updated [CHANGELOG.asciidoc](CHANGELOG.asciidoc)~

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
@@ -143,7 +143,7 @@ public class Destination implements Recyclable {
     @Override
     public void resetState() {
         address.setLength(0);
-        port = -1;
+        port = 0;
         service.resetState();
     }
 
--- a/apm-agent-plugins/apm-scheduled-annotation-plugin/src/test/java/co/elastic/apm/agent/spring/scheduled/TimerTaskInstrumentationTest.java
+++ b/apm-agent-plugins/apm-scheduled-annotation-plugin/src/test/java/co/elastic/apm/agent/spring/scheduled/TimerTaskInstrumentationTest.java
@@ -25,6 +25,7 @@
 package co.elastic.apm.agent.spring.scheduled;
 
 import co.elastic.apm.agent.AbstractInstrumentationTest;
+import co.elastic.apm.agent.impl.transaction.AbstractSpan;
 import co.elastic.apm.agent.impl.transaction.Transaction;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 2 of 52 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
```diff
@@ -143,7 +143,7 @@ public boolean hasContent() {
     @Override
     public void resetState() {
         address.setLength(0);
-        port = -1;
+        port = 0;
```
eyalkoren, line 146: why?

apm-agent-plugins/apm-scheduled-annotation-plugin/src/test/java/co/elastic/apm/agent/spring/scheduled/TimerTaskInstrumentationTest.java
```diff
@@ -56,14 +55,13 @@ void testTimerTask_scheduleWithFixedRate() throws InterruptedException {
...
     void testTimerTask_scheduleWithFixedDelay() throws InterruptedException {
-        reporter.reset();
         TestTimerTask timerTask = new TestTimerTask();
         Timer timer = new Timer("Timer");
         timer.schedule(timerTask, 1L, 10L);
 
         reporter.awaitUntilAsserted(1000L, () -> {
-            timer.cancel();
             assertThat(reporter.getTransactions()).isNotEmpty();
+            timer.cancel();
```
eyalkoren, line 64: Test is failing. Maybe revert to happen before the assertion and wait 20ms before asserting.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Destination.java
felixbarny: It made a test fail. In general, when calling `resetState`, the instance should be in the same state as if it had just been instantiated. The field is initialized with the default value of `0`.

apm-agent-plugins/apm-scheduled-annotation-plugin/src/test/java/co/elastic/apm/agent/spring/scheduled/TimerTaskInstrumentationTest.java
felixbarny: I rewrote the test a little. It should be much more deterministic and resilient now.

PR #1289, opened 2020-07-13
Minor changes to release process pipeline

cachedout:
## What does this PR do?
1. Introduces a new parameter for the release build which allows the check for failures in the branch build to be skipped.
1. Allows the CHANGELOG to be updated during the build process instead of requiring a restart.

Once this PR is merged, it should be safe to re-try to release process. The previous problems with the job failing to start have [been resolved separately](https://github.com/elastic/apm-agent-java/pull/1279).

### DIFF
--- a/.ci/release/Jenkinsfile
+++ b/.ci/release/Jenkinsfile
@@ -150,6 +150,7 @@ pipeline {
         stage('Branch creation') {
           steps {
             dir("${BASE_DIR}") {
+              git credentialsId: 'f6c7695a-671e-4f4f-a331-acdce44ff9ba', url: 'git@github.com:elastic/apm-agent-java.git'
               withEnv(["BRANCH_NAME=tags/${env.TAG_BARE}"]){
                 withGitRelease() {
                   script {
--- a/.ci/release/Jenkinsfile
+++ b/.ci/release/Jenkinsfile
@@ -87,10 +87,12 @@ pipeline {
             Go over PRs or git log and add bug fixes and features.
             Move release notes from the Unreleased sub-heading to the correct [[release-notes-{major}.x]] sub-heading (Example PR for 1.13.0 release).
 
-            Click 'Proceed' to confirm that this step has been completed or Abort to stop the build in order to complete this step.
+            Click 'Proceed' to confirm that this step has been completed and changes have been pushed or Abort to stop the build.
             """
             )
-            sh(script: "git pull")
+            dir("${BASE_DIR}") {
+              git credentialsId: 'f6c7695a-671e-4f4f-a331-acdce44ff9ba', url: 'git@github.com:elastic/apm-agent-java.git'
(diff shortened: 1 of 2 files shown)


.ci/release/Jenkinsfile
```diff
@@ -68,27 +69,30 @@ pipeline {
...
             Go over PRs or git log and add bug fixes and features.
             Move release notes from the Unreleased sub-heading to the correct [[release-notes-{major}.x]] sub-heading (Example PR for 1.13.0 release).
 
-            Click 'Proceed' to confirm that this step has been completed or Abort to stop the build in order to complete this step.
+            Click 'Proceed' to confirm that this step has been completed and changes have been pushed or Abort to stop the build.
             """
             )
-            sh(script: "git pull")
+            dir("${BASE_DIR}") {
+              git credentialsId: 'f6c7695a-671e-4f4f-a331-acdce44ff9ba', url: 'git@github.com:elastic/apm-agent-java.git'
```
v1v, line 94: I don't see any `deleteDir` step and the `BASE_DIR` should already contain the source code, What's the reason for cloning the repo again?

cachedout: Because we want to pull in any additional changes to the changelog that might be made as a result of the reminder and prompt just above this.

.ci/release/Jenkinsfile
```diff
@@ -68,27 +69,30 @@ pipeline {
...
-        //}
+        stage('Check master build status') {
+        when { expression { params.ignore_branch_ci_status } }
+         steps {
+           script {
+             // If this build is not green: https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/job/master/
+             if(!buildStatus(host: 'apm-ci.elastic.co', job: ['apm-agent-java', 'apm-agent-java-mbp', 'master'], return_boolean: true)) {
+               input(message: "WARNING! The master build is not passing. Do you wish to continue?")
+             }
+           }
```
kuisathaverat, line 80:
```suggestion
             // If this build is not green: https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/job/master/
             whenTrue(!buildStatus(host: 'apm-ci.elastic.co', job: ['apm-agent-java', 'apm-agent-java-mbp', 'master'], return_boolean: true)) {
               input(message: "WARNING! The master build is not passing. Do you wish to continue?")
             }
```

.ci/release/Jenkinsfile
felixbarny: In `branch_creation.sh`, we just do a plain git pull instead. Is that intentional?

cachedout: Good catch. I agree that we should be consistent here and I have replaced that shell call with the approach we are using in this step.

PR #1295, opened 2020-07-19
Allow empty server_urls

### LINKED ISSUE #1281: Ability to disable reporting to APM node but keep recording on?

No

Right now spans and transactions always generate a noop transaction if `recording=false`. I'd love to be able to turn this on but disable the reporting to the APM server (i.e. `server_urls` is empty). This could allow things like span and transaction ids for distributed tracing to work in other contexts such as developers running local development stacks where the logging functionality is still super useful but running a full ES/APM/Kibana stack is far too heavy.

I've played around with several of the configuration options w/o any luck. It's either on and trying to report/communicate with the APM server and logging a bunch of failures, or off and we get no tracing which is super useful for adding tracing http request headers, log correlation, etc.

Thoughts?

eyalkoren:
## What does this PR do?

Fixes #1281 
- Does not create arithmetic error of dividing by zero when server url list is empty
- Allows setting `server_urls` with an empty string, making the agent trace as usual with any sampling rate, but without attempting to communicate with the APM server and:
    - Metrics are collected as usual with buffers being replaced in the specified interval
    - Events are dropped as long as there is no server URL configured
    - Log shipper can resume working when server urls are set

## Checklist for reviewer
Verify that when `server_urls` are set to an empty string:
- [ ] Events do not leak
- [ ] No communication with the APM server is attempted, including health check and remote config polling
- [ ] File shipper works as expected
- [ ] Threads communicating with APM server are backing off properly

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
@@ -172,12 +172,14 @@ public class MetricRegistry {
         }
     }
 
+    /**
+     * Replace the active metrics container with a new one in a thread-safe manner.
+     */
     public void resetBuffers() {
         try {
             phaser.readerLock();
-            ConcurrentMap<Labels.Immutable, MetricSet> temp = inactiveMetricSets;
             inactiveMetricSets = activeMetricSets;
-            activeMetricSets = temp;
+            activeMetricSets = new ConcurrentHashMap<>();
             phaser.flipPhase();
         } finally {
             phaser.readerUnlock();
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
@@ -162,15 +162,15 @@ public class MetricRegistry {
     /**
      * Executes the following steps within a single read-operation critical section:
      * <ul>
-     *     <li>Switch between active and inactive MetricSet buffers</li>
-     *     <li>Report the inactivated buffer (optional)</li>
-     *     <li>Reset the inactivated buffer</li>
+     *     <li>Switch between active and inactive MetricSet containers</li>
+     *     <li>Report the inactivated MetricSets (optional)</li>
+     *     <li>Reset the inactivated MetricSets</li>
      * </ul>
      *
-     * @param metricsReporter a reporter to be used for reporting the inactivated metrics buffer. May be {@code null}
+     * @param metricsReporter a reporter to be used for reporting the inactivated MetricSets. May be {@code null}
      *                        if reporting is not required.
      */
-    public void switchBuffersAndReport(@Nullable MetricsReporter metricsReporter) {
+    public void flipPhaseAndReport(@Nullable MetricsReporter metricsReporter) {
         try {
             phaser.readerLock();
             ConcurrentMap<Labels.Immutable, MetricSet> temp = inactiveMetricSets;
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
@@ -159,28 +159,30 @@ public class MetricRegistry {
         return null;
     }
 
-    public void report(MetricsReporter metricsReporter) {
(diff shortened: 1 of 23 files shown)


felixbarny:
@elastic/apm-agent-python You have an option called [`disable_send`](https://www.elastic.co/guide/en/apm/agent/ruby/1.x/configuration.html#config-disable-send). Have you considered allowing to set the `server_url` to an empty string (overriding the default value `http://localhost:8200`) to get the same behavior?

eyalkoren:
@felixbarny I chatted with @basepi yesterday - `disable_send` is not entirely the same as what we do when setting `server_urls=`. The `disable_send` option exclusively deals with dropping events instead of sending them. The Python agent would still periodically poll for remote configuration for example.

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
```diff
@@ -172,6 +172,18 @@ public void report(MetricsReporter metricsReporter) {
         }
     }
 
+    public void resetBuffers() {
```
felixbarny, line 175: This is called each time any event is processed from the queue when `server_urls` is empty, correct?
What is the purpose of flipping active and inactive metric sets in this case? Note it doesn't reset the counters and timers as that's only done during serialization. Conflating serialization and metrics resetting is probably a code smell and that's on me. But do we have to reset the metrics at all? If so, I think this also resets even if a span is reported, for example, which doesn't feel right. Or is this serving an entirely different purpose?

Documenting the purpose here would help.

eyalkoren: > This is called each time any event is processed from the queue when server_urls is empty, correct?

No, every time a metrics event is reported - https://github.com/elastic/apm-agent-java/pull/1295/files#diff-517f2e7d31c3425c5ec7719efd32a06dR151.
The purpose is to continue collecting metrics in the right intervals, because the `server_urls` is a dynamic config.

> Note it doesn't reset the counters and timers as that's only done during serialization.

Right. I was going to look into that next. What do you say simply replacing the buffers with new ones, as in:
```
    public void resetBuffers() {
        try {
            phaser.readerLock();
            inactiveMetricSets = activeMetricSets;
            activeMetricSets = new ConcurrentHashMap<>();
            phaser.flipPhase();
        } finally {
            phaser.readerUnlock();
        }
    }
```

> Documenting the purpose here would help.

👍

felixbarny: > simply replacing the buffers with new ones

That won't work for gauges that still need to remain in the registry. Having to re-add transaction timers would also increase the allocation rate.

eyalkoren: @felixbarny check out the last commit. I separated reset from reporting. It costs a bit more because of the additional iteration, but since it is done as a secluded read-lock-operation, I think it's lightweight enough to worth the code simplification (especially now that we need to do that when not reporting as well).
I would even go further with providing a separate `resetState` operation for the `MetricRegistry`, but since it needs to be done on the same read-operation critical section with buffer switch and reporting, I didn't.

eyalkoren: Added tests as well

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
```diff
@@ -159,14 +159,30 @@ public DoubleSupplier getGauge(String name, Labels labels) {
...
+     * <ul>
+     *     <li>Switch between active and inactive MetricSet buffers</li>
+     *     <li>Report the inactivated buffer (optional)</li>
+     *     <li>Reset the inactivated buffer</li>
+     * </ul>
+     *
+     * @param metricsReporter a reporter to be used for reporting the inactivated metrics buffer. May be {@code null}
+     *                        if reporting is not required.
+     */
+    public void switchBuffersAndReport(@Nullable MetricsReporter metricsReporter) {
```
felixbarny, line 173: [minor] maybe call it flip phase and report to be consistent with the reader-writer-phaser terminology? the word `buffer` is not used elsewhere in this class.

Issue #60, opened 2018-04-16
Spring WebFlux support


prafsoni:
I don't know if this is a correct place for a feature request. Please add support for Spring WebFlux Framework too along with Web MVC. The new frameworks provide native support to build reactive web services. We are currently considering and moving towards Webflux, support in a future release will be awesome.

felixbarny:
Hi and thanks for that feature request. We definitely do consider adding support for this, however it is not targeted for version 1 which we see as a production ready MVP which supports auto instrumentation via `-javaagent`. Once we are at that stage, we will add support for more frameworks.

felixbarny:
Make sure to fill out the survey if you did not already: https://docs.google.com/forms/d/e/1FAIpQLScd0RYiwZGrEuxykYkv9z8Hl3exx_LKCtjsqEo1OWx8BkLrOQ/viewform

We use the results to prioritize framework support.

prafsoni:
Thanks, filled the survey. Maybe 1.1 will have support for this?

felixbarny:
I can't promise anything at this stage, but spring web reactive is currently the most requested reactive web framework in terms of agent support, so we will probably work on it soon after 1.0.

prafsoni:
Cool, is there a roadmap for 1.0 or it will be ready when it’s ready?

felixbarny:
The roadmap is basically the issues + milestones. However, there is no fixed date for 1.0/GA.

prafsoni:
Gotcha!

felixbarny:
The roadmap is just GitHub issues + milestones. Note that there is not much for post GA yet. Anything, in particular, you are interested in?

digz6666:
@felixbarny Yes I'm interested to know about what features are currently ready and what is coming on what version. And also want to know about the general architecture, if someone wants to contribute by sending pull request, where he or she might look for things that could be done.

b0c1:
It will contain both annotation and functional API?

felixbarny:
Ideally, yes. But we have not started working on this feature and I don't have a lot of insight into the framework's internals yet.

felixbarny:
@digz6666 FYI, we have added a [CONTRIBUTING.md](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md) which also links to an architecture overview.

zhongdj:
any progress on this?

felixbarny:
Unfortunately not. Make sure to 👍 the OP

[...]

PR #1221, opened 2020-06-08
spring webflux & reactor netty plugins

PGoski:
I've improved #827 and made apm-spring-weblux to work with real life app, added netty support also:

java agent jar for testing:
https://github.com/PGoski/apm-agent-java/releases/tag/1.16.1-SNAPSHOT

Tested with Spring Boot 2.2.2.RELEASE

With Netty:
```
<dependency>
  <groupId>org.springframework.boot</groupId>
  <artifactId>spring-boot-starter-webflux</artifactId>
</dependency>
```

Tomcat:
```
                   <dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-webflux</artifactId>
			<exclusions>
				<exclusion>
					<groupId>org.springframework.boot</groupId>
					<artifactId>spring-boot-starter-reactor-netty</artifactId>
				</exclusion>
			</exclusions>
		</dependency>
		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-tomcat</artifactId>
		</dependency>
```
       
## Checklist          
- [ ] My code follows the [style guidelines of this project](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#java-language-formatting-guidelines)
- ~[ ] I have made corresponding changes to the documentation~
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#testing) pass locally with my changes
- [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
- [ ] I have updated [supported-technologies.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/docs/supported-technologies.asciidoc)
- ~[ ] Added an API method or config option? Document in which version this will be introduced~
- [ ] Added an instrumentation plugin? Describe wow you made sure that old, non-supported versions are not instrumented by accident.

My goal was to implement for Spring Boot 2.2.2.RELEASE, tried to keep injection logic as lightweight as possible and I don't guarantee that some older versions are working. However tried to set Spring Boot version to a lower number in pom and tests passed, which is a good sign.

PR #1305, opened 2020-07-22
add Spring Webflux server support

SylvainJuge:
## What does this PR do?

This pull-request is an improvement on the #1221 community pull-request.
Given the large number of changes, it was simpler to open another PR than recycle #1221.

Fixes https://github.com/elastic/apm-agent-java/issues/60

## Features

Test application:
  - available as an executable jar and also used for unit-testing instrumentation
  - provides both functional and annotated controllers endpoints
  - executable with `netty` or `tomcat` as server, which allows to cover Webflux with or without Servlets
  - provides a client to call itself (based on Webflux client), thus using any third-party HTTP client is not necessary

Instrumentation:
- Webflux server: covers both functional and annotated endpoints variants
- Webflux client: not covered yet, see #1306 
- using new plugin API and indy dispatcher.
- adds reactor instrumentation for context-propagation

## Implementation checklist

- [x] provide context-propagation, maybe through reactor-core instrumentation
- [x] test supported webflux/reactor versions
- [x] performance benchmark to ensure minimal overhead
- [ ] in-depth allocation benchmark to estimate overhead (current implementation is expected to allocate a bit more).
- [x] capture upstream transaction HTTP headers (if any) for context propagation

## Checklist

- [x] This is a new plugin


elastic-apm-agent/pom.xml
```diff
@@ -14,6 +14,9 @@
     <properties>
         <elastic.apm.agent>co.elastic.apm.agent.premain.AgentMain</elastic.apm.agent>
         <apm-agent-parent.base.dir>${project.basedir}/..</apm-agent-parent.base.dir>
+
+        <!-- default value that will be overloaded through commit-id plugin -->
+        <git.commit.id.abbrev>UNKNOWN</git.commit.id.abbrev>
```
SylvainJuge, line 19: allows to skip git commit plugin execution and still get a valid property

pom.xml
```diff
@@ -107,6 +107,7 @@
         <version.log4j>2.12.1</version.log4j>
         <version.log4j2-ecs-layout>1.0.1</version.log4j2-ecs-layout>
         <version.spring>5.0.15.RELEASE</version.spring>
+        <version.spring-boot>2.2.2.RELEASE</version.spring-boot>
```
SylvainJuge, line 110: not sure if it's worth having such a global version as most plugins use their own. We could align it in another PR if it makes sense.

eyalkoren (commented):
Unfortunately only a partial review at this point 😞

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Url.java
```diff
@@ -108,17 +138,17 @@ public Url withHostname(@Nullable String hostname) {
     }
 
     /**
-     * The port of the request, e.g. '443'
+     * The port of the request, e.g. 443
      */
-    public StringBuilder getPort() {
+    public int getPort() {
```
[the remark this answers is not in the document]
SylvainJuge: As of `7.0.0` we can have `url.port` as a numeric field, making this an `int` allows to better fit the current schema. I've changed the behavior to only allocate when using a pre `7.0.0` server.

apm-agent-core/src/main/java/co/elastic/apm/agent/util/SpanConcurrentHashMap.java
```diff
@@ -0,0 +1,104 @@
...
+ * @param <V> context type
+ */
+public class SpanConcurrentHashMap<K, V extends AbstractSpan<?>> extends ConcurrentHashMap<K, V> {
+
+    public static <K, V extends AbstractSpan<?>> WeakConcurrentMap<K, V> createWeakMap() {
+        SpanConcurrentHashMap<AbstractWeakConcurrentMap.WeakKey<K>, V> map = new SpanConcurrentHashMap<>();
+        WeakConcurrentMap<K, V> result = new NullSafeWeakConcurrentMap<>(false, map);
+        WeakMapSupplier.registerMap(result);
+        return result;
+    }
```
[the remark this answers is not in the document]
SylvainJuge: We can't because `WeakMapSupplier` is part of the `plugin-sdk` module, which has no dependency on `core` module for `AbstractSpan` definition. I agree with you that this is definitely not super elegant to have it this way.

apm-agent-plugins/apm-reactor-plugin/src/main/java/co/elastic/apm/agent/reactor/ReactorInstrumentation.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return isStatic().and(named("onAssembly"));
+    }
+
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Collections.singletonList("reactor");
```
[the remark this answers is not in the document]
SylvainJuge: That's definitely a very good idea, that would prevent missing expectations if things do not work 100% as expected on the 1st release.

apm-agent-plugins/apm-kafka-plugin/apm-kafka-headers-plugin/pom.xml
```diff
@@ -27,5 +27,10 @@
             <artifactId>apm-kafka-base-plugin</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.ivy</groupId>
+            <artifactId>ivy</artifactId>
+            <scope>test</scope>
+        </dependency>
```
SylvainJuge, line 34: We might find a better solution to avoid duplication of common plugin test dependencies, but it would be better handled in a separate cleanup PR.

apm-agent-plugins/apm-reactor-plugin/src/main/java/co/elastic/apm/agent/reactor/TracedSubscriber.java
```diff
@@ -0,0 +1,267 @@
...
+                    // no active context, we have nothing to wrap
+                    return subscriber;
+                }
+
+                log.trace("wrapping subscriber {} publisher {} with active span/transaction {}", subscriber.toString(), publisher, active);
+
+                return new TracedSubscriber<>(subscriber, tracer, active);
+            }
+        });
+    }
```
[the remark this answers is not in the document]
SylvainJuge: I don't think it's worth the effort now, while I don't have exact measurement of allocation, it's currently on the same level as what other agents do. That would however be easy to further optimize if that proves to be an issue later on.

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-plugin/src/main/java/co/elastic/apm/agent/springwebflux/TransactionAwareSubscriber.java
```diff
@@ -0,0 +1,385 @@
...
+            .updateFull();
+
+        copyHeaders(serverRequest.getHeaders(), request.getHeaders());
+
+        for (Map.Entry<String, List<HttpCookie>> cookie : serverRequest.getCookies().entrySet()) {
+            for (HttpCookie value : cookie.getValue()) {
+                request.getCookies().add(cookie.getKey(), value.getValue());
+            }
+        }
+
```
eyalkoren, line 357: How complicated would it be to use `serverRequest.getBody()` to capture the request body?

apm-agent-premain/src/main/java/co/elastic/apm/agent/premain/AgentMain.java
```diff
@@ -174,6 +177,43 @@ private static File getAgentJarFile() throws URISyntaxException {
         if (!agentJar.getName().endsWith(".jar")) {
             throw new IllegalStateException("Agent is not a jar file: " + agentJar);
         }
-        return agentJar.getAbsoluteFile();
+        // verify agent shading at runtime, because tests might not have been always executed (when creating plugin)
+        // and failure to properly shade will create unexpected and complex behavior that is hard to diagnose and
+        // debug, thus the cost of doing this at runtime is worth it.
```
eyalkoren, line 182: The problem is that when we add something intentionally to the jar, we need to maintain both here and `co.elastic.apm.servlet.AgentFileIT#testEverythingIsShaded`. I don't see how something can be merged without making sure tests pass. So the latest we should stumble into this, is when running the integration tests, or running manual tests (which is normally done after integration tests pass).
If you feel this is important, at least add a reference from each of the location to the other, so that when one is updated, the other is updated as well.

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-plugin/src/main/java/co/elastic/apm/agent/springwebflux/TransactionAwareSubscriber.java
SylvainJuge: I don't think it would be easy to capture body using this, as we get a `Flux` and not a direct access to the request body byte buffer. Would better off keeping that for a follow-up PR.

apm-agent-premain/src/main/java/co/elastic/apm/agent/premain/AgentMain.java
SylvainJuge: I agree, that completely duplicates that IT test, the issue I had with that was more related to the long feedback loop that it implies:
- integration tests tend to take a while, thus I don't run them all the time
- they are finally run in CI when PR/branch is pushed, thus we are sure that we won't merge with such test failing
- the unexpected side effects of improper packaging (things not working as expected) can thus be silently ignored until IT are run

When trying to have a short feedback loop and thus skipping long-running tests (which tends to be common, at least in my current plugin dev workflow), we might spend quite some time until figuring out that there is a packaging issue.

Given this issue might definitely go away once we have proper external plugin and do not require shading anymore, it's not worth keeping that here, I have removed it and kept only the `testEverythingIsShaded` variant.

apm-agent-premain/src/main/java/co/elastic/apm/agent/premain/AgentMain.java
```diff
@@ -176,4 +176,5 @@ private static File getAgentJarFile() throws URISyntaxException {
         }
         return agentJar.getAbsoluteFile();
     }
+
```
eyalkoren, line 179:
```suggestion
```

eyalkoren (approved):

kihyun-leee:
Hi all,
I have a customer who wants to use JAVA APM agent with webflux backend.
It seems this PR is related... Is this PR applied to production? if not is there any ETA?

eyalkoren:
The best place to find out what is supported or not is in our [documentation](https://www.elastic.co/guide/en/apm/agent/java/current/supported-technologies-details.html#supported-web-frameworks).
Since this PR is merged long ago, it is of course released already (included since 1.24.0 and marked as experimental, as you can see in docs).

PR #1314, opened 2020-07-28
Fixes problem with Vault in release automation

cachedout:
## What does this PR do?
This fixes a problem with Vault which is described in an issue in a private repository which I will link back to here from.

This uses a script to pull down the proper configuration from the Vault instead of relying on specific tooling from the APM Pipeline Library to do so.

## Blocked

This PR will not be functional until a corresponding PR on the infra side is merged. I will link back to this PR from that PR. To ensure proper ordering, I will submit this PR as a draft so that it is not merged right away.

### DIFF
--- a/.ci/release/Jenkinsfile
+++ b/.ci/release/Jenkinsfile
@@ -140,7 +140,7 @@ pipeline {
         stage('Nexus release') {
           steps {
             script {
-              def spid = getVaultSecret('nexus')["staging-profile-id"]
+              def spid = sh(script: "./scripts/jenkins/fetch_nexus_id.sh", returnStdout: true).trim()
               dir("${BASE_DIR}"){
                 withEnvMask(vars: [[var: "SPID", password: spid]]){
                   def foundStagingId = nexusFindStagingId(stagingProfileId: "${SPID}", groupId: "co.elastic.apm")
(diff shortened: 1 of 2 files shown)


felixbarny:
For fixes on the release job, feel free to merge those in without waiting on my review. You and the robots are the better judges of the implementation details.

.ci/release/Jenkinsfile
```diff
@@ -140,7 +140,7 @@ pipeline {
         stage('Nexus release') {
           steps {
             script {
-              def spid = getVaultSecret('nexus')["staging-profile-id"]
+              def spid = sh(script: "./scripts/jenkins/fetch_nexus_id.sh", returnStdout: true).trim()
```
kuisathaverat, line 143: I am not sure why this command will not work

```suggestion
              def spid = getVaultSecret('secret/apm-team/ci/nexus')["staging-profile-id"]
```

cachedout: It will  not work because we are not running on `apm-ci`. The implementation of `getVaultSecret` depends specifically on credentials for Vault which are not deployed on the internal CI. This implementation more closely mirrors approaches which are typically employed on the internal CI.

felixbarny: Btw, I'm not objecting having the post-release job on apm-ci as long as the job that pushes to maven central remains in the internal ci. This is especially true if there's already a precedent of using secrets from vault in a public CI and if you consider that safe.

cachedout:
jenkins, test this please

PR #1315, opened 2020-07-28
Update to transaction_ignore_urls config name

### LINKED ISSUE #1313: Add TRANSACTION_IGNORE_URLS option
See https://github.com/elastic/apm/issues/144

eyalkoren:
Closes #1313 

@felixbarny I believe all we are required to do is renaming, while still allowing the old name so it is not a breaking change, right?
cc @gregkalapos 🙂

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/web/WebConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/web/WebConfiguration.java
@@ -11,9 +11,9 @@
  * the Apache License, Version 2.0 (the "License"); you may
  * not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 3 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/web/WebConfiguration.java
```diff
@@ -58,7 +58,8 @@
 
     private final ConfigurationOption<List<WildcardMatcher>> ignoreUrls = ConfigurationOption
         .builder(new ListValueConverter<>(new WildcardMatcherValueConverter()), List.class)
-        .key("ignore_urls")
+        .key("transaction_ignore_urls")
+        .aliasKeys("ignore_urls")
```
gregkalapos, line 62: Wow, you can even do an alias in your config-system; very fancy! 👏 👏

eyalkoren: Yeah, we use a fancy third party for that 😉

felixbarny: Let's call it a 2.5 party 😛

felixbarny:
@eyalkoren could you create a changelog for this?

eyalkoren:
> could you create a changelog for this?

@felixbarny it was added before merging. You probably looked at the RC release notes and not the [unreleased](https://www.elastic.co/guide/en/apm/agent/java/master/_unreleased.html#_features)

felixbarny:
ah, right 🤦‍♂️

Issue #1316, opened 2020-07-29
log4j using by the agent will attempt to loading the application's log4j2 class, leads to  ClassNotFoundException


CoderLan0668:
the log here 
```
ERROR StatusLogger Unable to create custom ContextSelector. Falling back to default.
 java.lang.ClassNotFoundException: org/apache/logging/log4j/core/async/AsyncLoggerContextSelector
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:264)
	at co.elastic.apm.agent.shaded.apache.logging.log4j.util.LoaderUtil.loadClass(LoaderUtil.java:168)
	at co.elastic.apm.agent.shaded.apache.logging.log4j.util.LoaderUtil.newInstanceOf(LoaderUtil.java:207)
	at co.elastic.apm.agent.shaded.apache.logging.log4j.util.LoaderUtil.newCheckedInstanceOf(LoaderUtil.java:228)
	at co.elastic.apm.agent.shaded.apache.logging.log4j.util.LoaderUtil.newCheckedInstanceOfProperty(LoaderUtil.java:253)
	at co.elastic.apm.agent.shaded.apache.logging.log4j.core.util.Loader.newCheckedInstanceOfProperty(Loader.java:338)
	at co.elastic.apm.agent.shaded.apache.logging.log4j.core.impl.Log4jContextFactory.createContextSelector(Log4jContextFactory.java:98)
	at co.elastic.apm.agent.shaded.apache.logging.log4j.core.impl.Log4jContextFactory.<init>(Log4jContextFactory.java:59)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at co.elastic.apm.agent.shaded.apache.logging.log4j.LogManager.<clinit>(LogManager.java:94)
	at co.elastic.apm.agent.shaded.apache.logging.log4j.spi.ThreadContextMapFactory.createThreadContextMap(ThreadContextMapFactory.java:99)
[...]

PR #1320, opened 2020-07-30
Fix log4j2 indy plugin context issues

felixbarny:
By calling classLoader.getParent() log4j tries to determine the logging context of a class loader
The IndyPluginClassLoader, having a MultipleParentClassLoader as it's parent, always returned null,
making it impossible to reuse context from the agent classloader

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/classloading/IndyPluginClassLoader.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/classloading/IndyPluginClassLoader.java
@@ -11,9 +11,9 @@
  * the Apache License, Version 2.0 (the "License"); you may
  * not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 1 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/bci/classloading/IndyPluginClassLoader.java
```diff
@@ -40,7 +40,7 @@
  */
 public class IndyPluginClassLoader extends ByteArrayClassLoader.ChildFirst {
     public IndyPluginClassLoader(@Nullable ClassLoader targetClassLoader, ClassLoader agentClassLoader, Map<String, byte[]> typeDefinitions) {
-        super(new MultipleParentClassLoader(Arrays.asList(agentClassLoader, targetClassLoader)), true, typeDefinitions, PersistenceHandler.MANIFEST);
+        super(new MultipleParentClassLoader(agentClassLoader, Arrays.asList(agentClassLoader, targetClassLoader)), true, typeDefinitions, PersistenceHandler.MANIFEST);
```
eyalkoren, line 43: The javadoc for this constructor says about the single-parent argument: "This explicit parent should only be set if the current platform does not allow creating a class loader that extends the bootstrap loader". Do you know why?

felixbarny: Not sure. Haven't noticed any issues with providing an explicit parent so far.

felixbarny:
> Making sure- not related to #1316 , but fixes the Indy plugin problem you encountered, right?

This is a fix for the issue I encountered specifically. I'm not sure if it helps that issue, too. Maybe.

PR #1324, opened 2020-08-04
Fixing default service name configuration

### LINKED ISSUE #1323: the service_name option of the core configuration not works
I run my application as the cmd `java -Delastic.apm.service_name=foo -jar foo.jar`, but the apllication.properties has an item `spring.application.name=bar`.

expected:  the apm agent will report the data with service.name = foo
But actually, it's bar, the service name has been replaced!

From ElasticApmTracer : 

`

    @Override
    public void overrideServiceNameForClassLoader(@Nullable ClassLoader classLoader, @Nullable String serviceName) {
        // overriding the service name for the bootstrap class loader is not an actual use-case
        // null may also mean we don't know about the initiating class loader
        if (classLoader == null
            || serviceName == null || serviceName.isEmpty()
            // if the service name is set explicitly, don't override it
            || !coreConfiguration.getServiceNameConfig().isDefault()) {
            return;
        }
        if (!serviceNameByClassLoader.containsKey(classLoader)) {
            serviceNameByClassLoader.putIfAbsent(classLoader, ServiceNameUtil.replaceDisallowedChars(serviceName));
        }
    }

`

and from ServiceNameUtil

`

    public static String getDefaultServiceName() {
        return getDefaultServiceName(System.getProperty("sun.java.command"));
    }

    static String getDefaultServiceName(@Nullable String sunJavaCommand) {
        String serviceName = parseSunJavaCommand(sunJavaCommand);
        if (serviceName != null) {
            serviceName = replaceDisallowedChars(serviceName);
            serviceName = serviceName.trim();
        }
        if (serviceName == null || serviceName.isEmpty()) {
            serviceName = "my-service";
        }
        return serviceName;
    }

`

Although the `elastic.apm.service_name` has been set by command line, but it do not work, because it is equals with default value, which from the system property `sun.java.command`.

eyalkoren:
## What does this PR do?
Fixes #1323 

## Checklist

- [x] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else


felixbarny:
What's the error condition here? Is it that the `spring.application.name` form `application.properties` is used instead of the configured `service_name` in case it is the same as the one derived from the name of the jar (`sun.java.command`)?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -696,7 +696,7 @@ public void overrideServiceNameForClassLoader(@Nullable ClassLoader classLoader,
         if (classLoader == null
             || serviceName == null || serviceName.isEmpty()
             // if the service name is set explicitly, don't override it
-            || !coreConfiguration.getServiceNameConfig().isDefault()) {
+            || coreConfiguration.getServiceNameConfig().getUsedKey() != null) {
```
felixbarny, line 699: neat!

eyalkoren:
> What's the error condition here? Is it that the spring.application.name form application.properties is used instead of the configured service_name in case it is the same as the one derived from the name of the jar (sun.java.command)?

Yes, the test I added fails without the fix

PR #1326, opened 2020-08-05
Fix service name caching based on context path

eyalkoren:
## What does this PR do?
Fixes service name caching. 

### Context 
A bug reported in https://discuss.elastic.co/t/apm-agent-not-working-after-updating-war-in-tomcat/239750 - after un-deploying and re-deploying a web app on a Servlet container, the web-app-specific service name (based on auto-discovery) is not used anymore. Instead, events are reported with the generic process service name.
The problem is with two inconsistent caches of the service name - it is cached in `ServletTransactionHelper` based on the context path and then in `ElasticApmTracer` based on the instrumented class's initiating class loader. When redeploying an app, the same context app is used with a different class loader and this causes the problem.

NOTE: I am not sure the first caching even worth it, only a few String comparisons. Maybe it's enough to leave only the second (class-loader-based) caching in `ElasticApmTracer`).

## Checklist

- [x] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletGlobalState.java
+++ b/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletGlobalState.java
@@ -25,6 +25,8 @@
 package co.elastic.apm.agent.servlet;
 
 import co.elastic.apm.agent.sdk.state.GlobalState;
+import co.elastic.apm.agent.sdk.weakmap.WeakMapSupplier;
+import com.blogspot.mydailyjava.weaklockfree.WeakConcurrentMap;
 
 import java.util.Collections;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletGlobalState.java
```diff
@@ -33,7 +35,7 @@
 @GlobalState
 public class ServletGlobalState {
 
-    public static final Set<String> nameInitialized = Collections.newSetFromMap(new ConcurrentHashMap<String, Boolean>());
+    public static final WeakConcurrentMap<ClassLoader, Boolean> nameInitialized = WeakMapSupplier.createMap();
```
felixbarny, line 38: Just FYT, in #1303, I've create a `WeakMapSupplier.createSet()` method.

eyalkoren: If I remember that after merging #1303, I will switch to using it instead.

PR #1057, opened 2020-02-28
add  RocketMQ support

gdutChenxj:
## What does this PR do?
add instrumentation for the Apache RcoektMQ

## Checklist

- [x] My code follows the [style guidelines of this project](CONTRIBUTING.md#java-language-formatting-guidelines)
- [x] I have rebased my changes on top of the latest master branch

- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#testing) pass locally with my changes

- [x] I have updated [CHANGELOG.asciidoc](CHANGELOG.asciidoc)
- [x] I have updated [supported-technologies.asciidoc](docs/supported-technologies.asciidoc)
- [ ] Added an API method or config option? Document in which version this will be introduced
- [x] Added an instrumentation plugin? How did you make sure that old, non-supported versions are not instrumented by accident?

## Screenshots
<!-- _(Optional)_
Add here screenshots about how the project will be changed after the PR is applied. They could be related to web pages, terminal, etc, or any other image you consider important to be shared with the team.

PR #1328, opened 2020-08-09
Add RabbitMQ support

hectorespert:
## What does this PR do?

RabbitMQ support based on the [RocketMQ support](https://github.com/elastic/apm-agent-java/pull/1057)

## Checklist

- [x] This is a new plugin
  - n/a ~Added an API method or config option? Document in which version this will be introduced~
 -> Testing the instrumentation with old versions of the RabbitMQ client


felixbarny:
Thanks for your PR ❤️ 

We're currently migrating plugins from `HelperClassManager` to indy plugins. See also https://github.com/elastic/apm-agent-java/pull/1329 for more context.

The new indy plugins should make it much more pleasant to write plugins as you don't need those helper interfaces and you can just reference both agent classes and RabbitMQ classes from anywhere in your code.

Would you mind migrating your plugin to be an indy plugin?

I noticed you marked this as a draft PR. What's still missing in your view? Which parts do you specifically want to get feedback on?

hectorespert:
> Would you mind migrating your plugin to be an indy plugin?
I will try it.

hectorespert:
> I noticed you marked this as a draft PR. What's still missing in your view? Which parts do you specifically want to get feedback on?

Consumer instrumentation seems that are working, but producer instrumentation isn't work and I can't do to it pass the tests.
I don't kwon if I do something wrong or I need to add something in the exit span.

apm-agent-plugins/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/RabbitMQProducerInstrumentation.java
```diff
@@ -0,0 +1,134 @@
...
+
+            // TODO: Propagate trace context
+            //final TextHeaderSetter<HashMap<String, Object>> textHeaderSetter = new RabbitMQTextHeaderSetter();
+            //basicProperties = propagateTraceContext(exitSpan, basicProperties, textHeaderSetter);
+
+            /*
+            TODO: Transaction context
+            exitSpan.getContext().getMessage().withQueue(exchange);
+            exitSpan.getContext().getDestination().getService().withType("messaging").withName("rabbitmq")
+                .getResource().append("rabbitmq/").append(exchange);*/
```
hectorespert, line 100: Is context required in exit spans? I can't get the producer's instrumentation to work.

eyalkoren: Without looking too much into the implementation, what do you mean you can't get it to work? What's not working? 
- If you use `log_level=DEBUG`, can you see whether the instrumentation is applied? You should see a `Method match for instrumentation RabbitMQProducerInstrumentation` indicating that the matchers found the method it was looking for
- Can you make sure the producer method is invoked when you expect it to be invoked in the test? If so, you can add a `System.out` and make sure the advice code runs when it is called as well.
- Can you see whether the trace context headers are added to the message? If not, it may be read-only at this point
- Try to print out the message headers on the receiver advice

hectorespert:
I migrated the plugin to be an indy plugin, but I have a problem in the Producer instrumentation.

```java
java.lang.IllegalStateException: Cannot define writable field access for com.rabbitmq.client.AMQP$BasicProperties arg1 when using delegation
```
How can I modify a method field in the indy plugins?

felixbarny:
You can use `@AssignTo.Field`

https://github.com/elastic/apm-agent-java/blob/cd1608b0205ad77824057f8ae17d39d233cd17f4/apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/advice/AssignTo.java#L138

hectorespert:
> You can use `@AssignTo.Field`
> 
> https://github.com/elastic/apm-agent-java/blob/cd1608b0205ad77824057f8ae17d39d233cd17f4/apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/advice/AssignTo.java#L138

Thanks. It was fixed using your advice.

apm-agent-plugins/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/RabbitMQProducerInstrumentation.java
hectorespert: I've updated the `RabbitMQProducerInstrumentationTest` test to check if the trace context headers are added to the message. They are added but the test is failling in the recycling validation.

```
java.lang.RuntimeException: Condition not fulfilled within 1000 ms
Expected :false
Actual   :true
<Click to see difference>


	at co.elastic.apm.agent.MockReporter.awaitUntilAsserted(MockReporter.java:423)
	at co.elastic.apm.agent.MockReporter.awaitUntilAsserted(MockReporter.java:401)
	at co.elastic.apm.agent.MockReporter.assertRecycledAfterDecrementingReferences(MockReporter.java:372)
	at co.elastic.apm.agent.AbstractInstrumentationTest.cleanUp(AbstractInstrumentationTest.java:91)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.RunAfters.invokeMethod(RunAfters.java:46)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:33)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:220)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:53)
Caused by: org.opentest4j.AssertionFailedError: [should not have any reference left, but has 2 : 'Rabbit-Test Transaction' 00-d7055f0e479e136460efa24fbccdfd1f-23255d7121bbf626-01 (23ee2ccf)] 
Expecting:
 <true>
to be equal to:
 <false>
but was not.
	at jdk.internal.reflect.GeneratedConstructorAccessor18.newInstance(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at co.elastic.apm.agent.MockReporter.lambda$assertRecycledAfterDecrementingReferences$8(MockReporter.java:384)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at java.base/java.util.Collections$SynchronizedCollection.forEach(Collections.java:2067)
	at java.base/java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1085)
	at co.elastic.apm.agent.MockReporter.lambda$assertRecycledAfterDecrementingReferences$9(MockReporter.java:381)
	at co.elastic.apm.agent.MockReporter.awaitUntilAsserted(MockReporter.java:415)
	... 31 more
```

felixbarny:
> using your advice

Pun intended?

apm-agent-plugins/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/RabbitMQProducerInstrumentation.java
felixbarny: That could be because you did not call `span.end()` in `RabbitProducerAdvice#afterBasicPublish()`

hectorespert:
> > using your advice
> 
> Pun intended?

No pun intended

apm-agent-plugins/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/RabbitMQProducerInstrumentation.java
hectorespert: Yes, seems that it is the error :facepalm: 

Thank you for your patience

SylvainJuge:
I've opened a PR against your fork as it was simpler to do the changes directly to avoid back & forth: https://github.com/hectorespert/apm-agent-java/pull/1

You really did an awesome job here, the changes are mostly details and the instrumentation worked really well !!

hectorespert:
@SylvainJuge I merged your changes. Thanks for your help. What are your plans for this PR?

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
```diff
@@ -92,7 +92,7 @@
         when(apmServerClient.isAtLeast(any())).thenReturn(true);
         dslJsonSerializer = new DslJsonSerializer(mock(StacktraceConfiguration.class), apmServerClient);
         SPAN_TYPES_WITHOUT_ADDRESS = Set.of("jms");
-        SPAN_ACTIONS_WITHOUT_ADDRESS = Map.of("kafka", Set.of("poll"));
+        SPAN_ACTIONS_WITHOUT_ADDRESS = Map.of("kafka", Set.of("poll"), "rabbitmq", Set.of("send"));
```
felixbarny, line 95: What functionality is impacted by that? Will RabbitMQ be visible as part of the service map?

What's preventing the capturing of the address? Maybe summarize the problems in a follow-up issue.

hectorespert: > What functionality is impacted by that? Will RabbitMQ be visible as part of the service map?

For my side, I don't know if any feature is affected. I used the same solution that use the kafka plugin.

> What's preventing the capturing of the address? Maybe summarize the problems in a follow-up issue.

There isn't any attribute in the instrumented methods that allows obtain the address directly. I don't know if it can be obtained using other methods.

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/ChannelInstrumentation.java
```diff
@@ -0,0 +1,180 @@
...
+import static net.bytebuddy.matcher.ElementMatchers.not;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+
+
+public abstract class ChannelInstrumentation extends BaseInstrumentation {
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        // Instrumentation applied at runtime, thus no need to check type
```
eyalkoren, line 60: What does that means? How come we don't look for specific types?

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
eyalkoren: > What functionality is impacted by that? Will RabbitMQ be visible as part of the service map?

The address is not required for the service map, only for the SIEM integration.

> There isn't any attribute in the instrumented methods that allows obtain the address directly. I don't know if it can be obtained using other methods.

If you know the address when the channel is created, you can store it in a map (channel object ---> address) and when the method is called, get it through the `@Advice.This` reference in the instrumented method.

felixbarny: I see. Maybe one day I'll remember. Thanks for explaining again 😉 
So I would say it's not too critical at the moment.

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/ChannelInstrumentation.java
felixbarny: It's because only known types explicitly get instrumented via `DynamicTransformer`. See `ConnectionInstrumentation`.

Also note that the service descriptor `META-INF/services/co.elastic.apm.agent.sdk.ElasticApmInstrumentation` does not contain `ChannelInstrumentation`.

SylvainJuge: We could probably create a dedicated subclass for those "instrumentations applied at runtime", that would probably make things a bit more obvious instead of repeating this comment.

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
SylvainJuge: There isn't even a need to use a map in this case as we can use `Channel.getConnection()`, I've implemented that in the t commit.

hectorespert:
Could be activated your CI environment for this PR?

SylvainJuge (approved):
self-approve to allow running in CI.

SylvainJuge:
jenkins run the tests please

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -683,10 +683,10 @@ public static void ensureInstrumented(Class<?> classToInstrument, Collection<Cla
                         .with(FailSafeDeclaredMethodsCompiler.INSTANCE);
                     AgentBuilder agentBuilder = getAgentBuilder(byteBuddy, config, logger, AgentBuilder.DescriptionStrategy.Default.HYBRID, false);
                     for (Class<? extends ElasticApmInstrumentation> instrumentationClass : instrumentationClasses) {
+                        ElasticApmInstrumentation apmInstrumentation = instantiate(instrumentationClass);
                         pluginClassLoaderByAdviceClass.put(
-                            instrumentationClass.getName(),
+                            apmInstrumentation.getAdviceClass().getName(),
```
SylvainJuge, line 688: without this, it works only with advice that is directly within the instrumentation class, and not delegated to it's own class.

eyalkoren: Can we do this caching based on advice class names only, instead of instrumentation classes and their **instantiation** here? I mean that `ensureInstrumented` will be done directly on advice class names

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/ConnectionInstrumentation.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class, inline = false)
+    public static void onExit(@Nullable @Advice.Thrown Throwable thrown,
+                              @Advice.Return @Nullable Channel channel) {
+
+        if (thrown != null || channel == null) {
+            return;
+        }
+        DynamicTransformer.Accessor.get().ensureInstrumented(channel.getClass(), CHANNEL_INSTRUMENTATIONS);
```
eyalkoren, line 85: I am not sure I understand why this is required instead of simple type matching for `Channel` implementations. 
However, since we already have this instrumentation, let's use it to store channel destination metadata (address and port) and use it to fill the send span destination details in the `basicPublish` instrumentation. Looks like very little is left in order to collect this info.

docs/supported-technologies.asciidoc
```diff
@@ -348,6 +348,12 @@ If both producer and consumer are traced, then distributed tracing is supported,
 side will be a part of the same trace containing the producer side as well. Kafka streams are not traced.
 |1.13.0
 
+|RabbitMQ
+|3.x - 5.x
+|The agent captures RabbitMQ Message sends and consumes.  
+When the producer send message within a traced transaction, a span will be created and a transaction is created when consumes a message.
```
eyalkoren, line 354:
```suggestion
When the producer sends a message within a traced transaction, a send span will be created and the transaction resulting from the message consumption will be correlated to the same trace. When a message is consumed, a transaction is created, regardless if the send action is traced.
```

eyalkoren (changes requested):
Very nice! 👏 

Some comments in addition to the ones embedded in code:

- Transaction and span naming - please align with Kafka and JMS, where we use the queue names. This adds valuable information, plus it supports the ability to disable tracing of specific queues.
In RabbitMQ, the queue name seems to be inaccessible in the publisher side, however it looks like the combination of exchange and routing key is meaningful and is available on both sides (through the `Envelope` in the consumer side). Therefore, I suggest using `<exchange-name>/<routing-key>` as the queue identifier. Otherwise, only exchange may be good enough.
I imagine that most users use custom exchange names, which we can test through the `com.rabbitmq.client.Channel#exchangeDeclare()` API
- Missing functionalities:
   - Header capturing on consumer side (based on configuration, with sanitation, see example in Kafka plugin)
   - Body capturing on consumer side (eg if UTF8-encoded, based on related configuration - see example in Kafka plugin)
   - [`ignore_message_queues`](https://www.elastic.co/guide/en/apm/agent/java/current/config-messaging.html#config-ignore-message-queues) functionality.

SylvainJuge:
@hectorespert If you don't mind, I will start working on doing the post-review changes right now.
Hopefully we will be able to merge this very soon.

hectorespert:
> @hectorespert If you don't mind, I will start working on doing the post-review changes right now.
> Hopefully we will be able to merge this very soon.

Thanks @SylvainJuge, go ahead.

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/ChannelInstrumentation.java
SylvainJuge: has been covered in the last commit.

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/test/java/co/elastic/apm/agent/rabbitmq/RabbitMQTest.java
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    @AfterAll
+    static void after() {
+        container.close();
+    }
+
+
+    @Test
+    void contextPropagationWithProperties() throws IOException, InterruptedException {
```
eyalkoren, line 117:
```suggestion
    void contextPropagationWithoutProperties() throws IOException, InterruptedException {
```

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/test/java/co/elastic/apm/agent/rabbitmq/RabbitMQTest.java
```diff
@@ -0,0 +1,405 @@
...
+        performTest(null);
+    }
+
+    @Test
+    void contextPropagationWithoutProperties() throws IOException, InterruptedException {
```
eyalkoren, line 122:
```suggestion
    void contextPropagationWithProperties() throws IOException, InterruptedException {
```

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/ConnectionInstrumentation.java
eyalkoren: Or what you did - use `com.rabbitmq.client.Channel#getConnection`, much simpler 😄

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/BaseInstrumentation.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            headers = properties.getHeaders();
+        }
+
+        Message message = context.getContext().getMessage()
+            .withQueue(exchange)
+            .withAge(age);
+
+        if (isCaptureHeaders() && headers != null && headers.size() > 0) {
```
eyalkoren, line 88: We should only capture headers on message read when starting a transaction.
Capturing outgoing headers in not aligned with other messaging plugins nor HTTP client plugins.
(Similarly if we decide to capture body in the future).

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-test-5/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+        <artifactId>apm-rabbitmq</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.19.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>apm-rabbitmq-test-5</artifactId>
+    <name>${project.groupId}:${project.artifactId}</name>
+
+    <properties>
+        <version.rabbitmq>5.9.0</version.rabbitmq>
```
eyalkoren, line 16: Same version as the common module. This module seems redundant.

eyalkoren (changes requested):
Looking good!
Some minor changes requested.

fgabolde:
I'm interested in RMQ support (the one last big part of our stack that apm-agent-java didn't support! thanks to all involved) but I'm a little wary of using the routing key in transaction names, as I understand this PR is doing (if I misunderstood, sorry about that, and you can safely discard the rest of this comment). IMHO, this should be part of metadata, like HTTP transaction URLs are.

The reason is that routing keys can be extremely high cardinality, e.g. user IDs. The RMQ exchanges can bind `user.#` to deliver all user IDs however they want; this is how the various sharding plugins work, for instance.

SylvainJuge:
Thanks for that feedback @fgabolde ! We often miss some real-world usage feedback for all the frameworks & libraries that we instrument, thus it's more than welcome !

In the current version we currently include exchange & routing key as part of span and transactions naming as you can find in the tests code for example:

```java
        assertThat(span.getNameAsString())
            .isEqualTo("RabbitMQ SEND to %s/%s", exchange, routingKey);
```

There are multiple ways to make it fit high-cardinality routing keys, which might get added in a follow-up/improvement PR, for example by providing an option to not include those in transaction names (only keeping queue/exchange).

Also, while this is a _possible usage_ of routing keys, would it be something that would prevent you from using it as-is ?

Anyone is invited to add a :+1: (if it's fine) or :-1: (if it's not) on this comment so we have feedback if that's a blocking thing that we should address right now or we can do it in a follow-up PR.

fgabolde:
@SylvainJuge, I think you're right that the feature as implemented in this PR is already useful for some of my use cases. E.g. I have several services consuming task queues from RMQ where the routing key is not really interesting, so we mostly do direct exchanges and the routing key never varies (all the interesting info is in the payload). These services also tend to be ones that would most benefit from this feature; since the message is the point of entry and generates a single tree of spans, it's often the most interesting span.

I'm thinking of some other services that would not be suitable though. They consume large amounts of messages from sharded queues, so the routing key varies wildly. Then again, they tend to do things that don't work well with distributed tracing, like batching or reordering messages, so a given message is not the only one responsible for any other span we might generate. E.g. we might gather a batch of 100 messages, filter out some of them, process them somewhat, and bulk-write everything to a database in a single insert query. I'm not sure how Elastic APM (or any other similar tool) can help me here, to be honest. If you have any tips I'm very interested.

All in all, I do think that having the routing key in names can be useful, but it feels odd that it's the opposite of the HTTP use case for instance, where the transaction/span name is usually a method name (so cardinality is lower). E.g. for an HTTP service, the root span will be something like `FooControllerName#getFoo`; you wouldn't put foo IDs there. That's what the `elastic.apm.use_path_as_transaction_name` option is for, but it's disabled by default, which makes sense to me. It would be more consistent if RMQ support was similar (low cardinality by default, with an option to bring back the details in the transaction names if necessary).

eyalkoren:
@fgabolde just to add a bit of context: in other messaging systems, we use topic or queue names and so far it seems to resonate with users. In RabbitMQ, we are not so sure the queue name is meaningful always and I am not sure it is accessible in the receiver side (although I have to admit I did not put a lot of effort investigating that). On the other hand, we are worried that the exchange names would provide too-low cardinality. Therefore I suggested using the routing key in combination with the exchange, but I was not aware of using it as you mention, e.g. for sharding.
Ideally, what would you define as the best cardinality-factor for RabbitMQ - the exchange or the queue?

@SylvainJuge the problem with very high cardinality of transaction names is that it makes a lot of our UI much less useable, as all transaction summaries are based on that, so we should avoid this default experience OOTB as much as possible. I think it is much better to err on the opposite side, meaning - providing too-low cardinality. And it may be the case that even allowing opt-in on routing key doesn't make sense in most cases.

fgabolde:
> @fgabolde just to add a bit of context: in other messaging systems, we use topic or queue names and so far it seems to resonate with users. In RabbitMQ, we are not so sure the queue name is meaningful always and I am not sure it is accessible in the receiver side (although I have to admit I did not put a lot of effort investigating that). On the other hand, we are worried that the exchange names would provide too-low cardinality. Therefore I suggested using the routing key in combination with the exchange, but I was not aware of using it as you mention, e.g. for sharding.
> Ideally, what would you define as the best cardinality-factor for RabbitMQ - the exchange or the queue?

I'm not too experienced with other messaging systems, but AFAIK in Kafka for example, there is no advanced routing as there is in RMQ. You publish to a topic by name and that's it. I've seen some discussion around routing but it seems to be about creating sub-topics behind the scenes? (Not to diss on Kafka -- I'd love to get some replayability in some of our message queues...) Likewise, when receiving events, you subscribe to a topic by name. In both sides the names are well-known and, barring pathological configurations, low cardinality.

In RMQ on the other hand, you publish to an exchange a message with a routing key and a body (and some other optional stuff). The only well-known piece of data when **publishing** is the exchange name; as I mentioned in my first comment the routing key may be low cardinality as well, but it might not be. The exchange then decides, based on its type and its configured bindings, which queue(s) the message will end up in. You have a lot of possibilities here, which makes guessing the final queue(s) impossible unless you also know the server configuration: directly to a queue named like the routing key? to some different queue, based on the routing key? ignore the routing key and use headers instead? Anyway, when publishing you only know the exchange name and the message, so **the exchange is probably the best describer** here. The routing key's cardinality is completely unknown, it really depends on the application's architecture. User IDs in routing keys are of course realistic for sharding, but not everybody uses routing keys for sharding. You might use them for plain topic subscription, in which case the cardinality can vary from low (like log severity) to high-ish (like postal codes).

When **receiving** a message however, you ignore exchanges completely and you read from a named queue (this may get tricky for sharding btw, since shards are implemented as queues; you have to find some way to decide which consumer reads from which queue). **The queue is probably the best describer here.** Fun!

As a last note for clarifying the RMQ architecture further: queues don't belong to exchanges. Multiple exchanges can be setup to route some or all of their incoming messages to the same queue. However, exchanges and queues both belong to vhosts, which is now that I think about it an interesting piece of info to add to spans because you can have multiple exchanges or queues with the same name, as long as they live in different vhosts...

eyalkoren:
@fgabolde I forgot to thank you about your detailed explanation - it is super useful for us, thanks a lot for taking the time.

We eventually based our approach greatly based on your feedback and we will see how it goes:
- Transactions and spans will be named with exchange or queue if we have it (i.e. receiver).
- We will have a new field for the routing key and we may add a config in the future to opt in to append it to the name as well.
- In the existing `queue.name` field, we will set the queue name if available or the exchange otherwise.
- In the `context.destination.service.resource` field we want to make sure sender and receiver set the same, so we will use exchange name in both.

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/test/java/co/elastic/apm/agent/rabbitmq/RabbitMQTest.java
```diff
@@ -0,0 +1,672 @@
...
+            expectedResource);
+
+    }
+
+    private static void checkPollSpan(Span span, String queue){
+        checkSpanCommon(span,
+            "poll",
+            String.format("RabbitMQ POLL from %s", queue),
+            queue,
+            "rabbitmq");
```
eyalkoren, line 648: Should contain the exchange in the resource name

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/ChannelInstrumentation.java
```diff
@@ -0,0 +1,311 @@
...
+        public static Object[] onBasicPublish(@Advice.This Channel channel,
+                                              @Advice.Argument(0) @Nullable String exchange,
+                                              @Advice.Argument(4) @Nullable AMQP.BasicProperties properties) {
+            if (!tracer.isRunning()) {
+                return null;
+            }
+
+            Span exitSpan = createExitSpan(exchange);
+            if (exitSpan == null) {
+                // tracer disabled or ignored exchange
```
eyalkoren, line 152: [minor]
```suggestion
                // tracer disabled or ignored exchange or this is nested within another exit span
```

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/ChannelInstrumentation.java
```diff
@@ -0,0 +1,311 @@
...
+            if (properties == null) {
+                properties = new AMQP.BasicProperties();
+            }
+
+            Map<String, Object> currentHeaders = properties.getHeaders();
+            if (currentHeaders == null) {
+                currentHeaders = new HashMap<>();
+            }
+
+            HashMap<String, Object> headersWithContext = new HashMap<>(currentHeaders);
```
eyalkoren, line 179: [minor]
If `currentHeaders == null`, create only one instead of two

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/ChannelInstrumentation.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    /**
+     * Creates a messaging exit span
+     *
+     * @param exchangeOrQueue exchange or queue name
+     * @return exit span if applicable, {@literal null} otherwise
+     */
+    @Nullable
+    private static Span createExitSpan(@Nullable String exchangeOrQueue) {
```
eyalkoren, line 279: [minor]
```suggestion
    private static Span createAndActivateExitSpan(@Nullable String exchangeOrQueue) {
```

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/ChannelInstrumentation.java
```diff
@@ -0,0 +1,311 @@
...
+    private static void captureDestination(String exchange, Channel channel, Span span) {
+        Destination destination = span.getContext().getDestination();
+
+        Destination.Service service = destination.getService();
+        service.withType("messaging")
+            .withName("rabbitmq")
+            .withResource("rabbitmq");
+
+        if (!exchange.isEmpty()) {
+            // include non-default exchange name in resource
```
eyalkoren, line 303: Include default as well

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/ChannelInstrumentation.java
```diff
@@ -0,0 +1,311 @@
...
+                return null;
+            }
+
+            exitSpan.withAction("send")
+                .withName("RabbitMQ SEND to ").appendToName(normalizeExchangeName(exchange));
+
+            properties = propagateTraceContext(exitSpan, properties);
+
+            captureMessage(exchange, properties, exitSpan);
+            captureDestination(exchange, channel, exitSpan);
```
eyalkoren, line 162: Use the normalized exchange name

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/main/java/co/elastic/apm/agent/rabbitmq/ChannelInstrumentation.java
```diff
@@ -0,0 +1,311 @@
...
+            String exchange = "";
+
+            if (null != envelope) {
+                // since exchange name is only known when receiving the message, we might have to discard it
+                if (isIgnored(envelope.getExchange())) {
+                    span.requestDiscarding();
+                }
+            }
+            captureMessage(queue, properties, span);
+            captureDestination(exchange, channel, span);
```
eyalkoren, line 264: Use the normalized exchange name

docs/supported-technologies.asciidoc
```diff
@@ -356,8 +356,22 @@ If both producer and consumer are traced, then distributed tracing is supported,
...
 
+|RabbitMQ
+|3.x - 5.x
+|The agent captures RabbitMQ Message sends, consumption and polling.
+When the producer sends a message within a traced transaction, a send span will be created and the transaction resulting
+from the message consumption will be correlated to the same trace.
+When a message is consumed, a transaction is created, regardless if the send action is traced.
+When a message is polled, a span is created for the polling operation and optionally have message fields if a message
+has been returned.
+|1.19.1
```
eyalkoren, line 367:
```suggestion
|1.20.0
```

docs/supported-technologies.asciidoc
```diff
@@ -356,8 +356,22 @@ If both producer and consumer are traced, then distributed tracing is supported,
...
+
 |===
 
+For RabbitMQ:
+
+- `context.message.queue.name` field will contain queue name when using polling, exchange name otherwise.
+- `context.message.destination.resource` field will contain `rabbitmq` for the default exchange and `rabbitmq/XXX` for named exchanges
```
eyalkoren, line 374: Always with the exchange, normalized when it is the default

eyalkoren (approved):
Well done, we are there 👍 
Two minor deviations from the spec and a small bug with resource naming.
Other than that - minor suggestions.
If you disagree (meaning - I got something wrong), just ignore.

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-plugin/src/test/java/co/elastic/apm/agent/rabbitmq/RabbitMQTest.java
SylvainJuge: For polling, the exchange name is only known when there is a message, thus that would use two different names for the cases with/without a response, whereas it's the same polling operation. Also, I found it was more consistent with the "use the queue name when available, exchange name otherwise".

SylvainJuge: My bad, I read it too quickly, here it's about the *resource name*, not the *span name*. Thus we should have:
- `rabbitmq` for polling when there is no message
- `rabbitmq/${exchange}` for polling when there is a message

We do not include queue name in resource name due to potentially high cardinality.

Also, that means that a successful message is being captured with queue name, and exchange name is visible in resource name, which could help to configure filtering using either queue name and exchange name.

fgabolde:
I meant to thank you for this feature by the way!

We upgraded one of our services to 1.20 and I can see the lovely outgoing traces. Going to upgrade another and see if the propagation works.

:+1:

felixbarny:
Great to hear, thanks for the Feedback!

Be sure to let us know how it goes when updating other services 🙂

fgabolde:
It works perfectly!

(Almost perfectly. I had to tweak my Spring setup for the full effect, initially I only had the RECEIVE span and no others. For posterity: when using Spring AMQP's `SimpleMessageListenerContainer`, I don't really know what multithreading facility is used behind the scenes, but it doesn't seem to be supported by the agent. Switching to `DirectMessageListenerContainer` fixed the issue and now I have all my spans. Documentation [here](https://docs.spring.io/spring-amqp/reference/html/#container))

felixbarny:
Thanks for reporting, I've created a follow-up issue: https://github.com/elastic/apm-agent-java/issues/1635

eyalkoren:
@fgabolde #1657 is adding proper support for Spring AMQP message handling API. If possible and still relevant, please try [this snapshot](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/PR-1657/39/artifacts) with `SimpleMessageListenerContainer` to verify it works as expected.

fgabolde:
OK, thanks! I will take a look as soon as I am able.

fgabolde:
@eyalkoren The MR you linked has been merged so the Jenkins pipeline has disappeared; that's my fault for taking so long I guess. I got the latest build on master which I assume contains the relevant changes, I attached it to my RMQ-using app, and I switched back to SimpleMessageListenerContainer, but now I have no RMQ spans at all :(

I guess I did something wrong, but I'm not sure where, so I will keep looking around.

felixbarny:
The snapshot deployment failed yesterday. So depending on when you downloaded the snapshot, the changes may not be included. Try downloading again now.

fgabolde:
I just got the agent from this build https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/master/813/artifacts (yday 7:47 PM UTC), still the same thing. I can see the RMQ span in the sender application but the receiver has nothing at all, while before this change I would have a RECEIVE span at least.

I'm pretty sure I misconfigured something, but I still don't know what. I will try building the app with the Simple container and the 1.20 agent and at least confirm if I have the old behavior back (if not, the issue is definitely on my side).

eyalkoren:
@fgabolde sorry to hear this didn't work out smoothly for you 🙁 
We just released [1.22.0](https://github.com/elastic/apm-agent-java/releases/tag/v1.22.0), which contains this fix. We still consider this as a fix because @kananindzya was able to reproduce and make it work in #1657 , but if your use case is not supported properly yet, let us know and we'll try to see why.
If you can't figure it out, please provide details about your setup, the agent configurations you use and a debug log so we can look into this.

PR #1355, opened 2020-08-25
proxy support with authentication

SylvainJuge:
## What does this PR do?

- provides a way to test proxy usage with & without authentication using a squid proxy container
- currently instruments `Authenticator.register*` static methods to provide an automatic fallback (still WIP)

In a regular application, using proxy with authentication requires to setup a global `Authenticator` [as described here](https://blogs.oracle.com/wssfc/handling-proxy-server-authentication-requests-in-java).

The goal here is not to merge this as-is, but rather to discuss how to properly implement support for proxies using authentication.

## Constraints

- agent uses `java.net.HttpUrlConnection` to communicate with backend server
- `http.proxyHost` and `http.proxyPort` JVM system properties allow to set a global proxy for the whole JVM
- while `http.proxyUser` and `http.proxyPassword` are also defined, they can't be used without registering an `Authenticator` instance explicitly
- unless using Java 9 and later, there is no way to avoid the global authenticator that is set through `Authenticator.setDefault(...)`, also, `Authenticator.getDefault()` is only available on Java 9.
- once the JVM has authenticated against a proxy server, credentials are cached and reused to avoid round-trips with the proxy

## Proposal

- get rid of the jdk `Authenticator` instrumentation as it introduces extra complexity for a very minor benefit
- add a new opt-in configuration option : `proxy_authenticator` which can have values `auto` (default), `false`, `true`
- when set to `false`, feature is disabled
- when set to `true`, a global authenticator will be registered if `http.proxyUser` and `http.proxyPassword` are set
- when set to `auto`, behavior depends on Java version:
  * Java < 9: equivalent to `false`, users have to explicitly set `true` to enable proxy authentication
  * Java >= 9: a global `Authenticator` will be registered _only if application did not already set one_ (we have a getter on the global instance).

This allows to:
- avoid any extra complexity in agent code (proxies with/without authentication aren't used by most users)
- do not have any impact on applications that already register a global `Authenticator` instance
- provide a fallback option for users that do not have any `Authenticator` instance in their application (the app isn't using the proxy) without having to modify their application.
- bonus of this PR: provide tests for proxy support.

## Alternatives
- Still provide an option to register a global `Authenticator` (true|false) to make it work with java 7 and 8. With java >= 9, use reflection to set `Authenticator` on the connection.

## Checklist

### DIFF
--- a/apm-agent-core/pom.xml
+++ b/apm-agent-core/pom.xml
@@ -158,6 +158,11 @@
             <version>2.6.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/proxysupport/AuthenticatorInstrumentation.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/proxysupport/AuthenticatorInstrumentation.java
@@ -1,0 +1,174 @@
+package co.elastic.apm.agent.proxysupport;
+
+import co.elastic.apm.agent.sdk.ElasticApmInstrumentation;
+import co.elastic.apm.agent.sdk.advice.AssignTo;
+import co.elastic.apm.agent.sdk.state.CallDepth;
+import co.elastic.apm.agent.sdk.state.GlobalThreadLocal;
+import net.bytebuddy.asm.Advice;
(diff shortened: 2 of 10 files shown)


SylvainJuge:
Update after discussing with @felixbarny: 

We don't have enough data/feedback about proxy usage to make a proper decision on what is best to do, thus we will keep proxy-support it as-is for now.

Key points:
- making this work with java 9+ with a per-connection `Authenticator` sounds a good idea
- for older java versions, the global `Authenticator` is a decent fallback, but we are not (yet) required to provide one.
- merging the tests in this PR allow to future-proof proxy usage by the agent, this part will be pushed through another separate PR.

Issue #1337, opened 2020-08-13
Migrate plugins to indy plugins


felixbarny:
A meta-issue that tracks the migration

- [x] apm-apache-httpclient-plugin https://github.com/elastic/apm-agent-java/pull/1349
- [x] apm-api-plugin https://github.com/elastic/apm-agent-java/pull/1411
- [x] apm-asynchttpclient-plugin elastic/apm-agent-java#1351
- [x] apm-bootdelegation-plugin N/A - might be removed once all plugins have been migrated to indy plugins
- [x] apm-dubbo-plugin https://github.com/elastic/apm-agent-java/pull/2087
- [x] apm-error-logging-plugin https://github.com/elastic/apm-agent-java/pull/1362
- [x] apm-es-restclient-plugin #2088
- [x] apm-grails-plugin https://github.com/elastic/apm-agent-java/pull/1371
- [x] apm-grpc https://github.com/elastic/apm-agent-java/issues/1358
- [x] apm-hibernate-search-plugin https://github.com/elastic/apm-agent-java/pull/1404
- [x] apm-httpclient11-plugin
- [x] apm-httpclient-core n/a
- [x] apm-java-concurrent-plugin
- [x] apm-jaxrs-plugin https://github.com/elastic/apm-agent-java/pull/1374
- [x] apm-jaxws-plugin elastic/apm-agent-java#1383
- [x] apm-jdbc-plugin
- [x] apm-jms-plugin https://github.com/elastic/apm-agent-java/pull/1394
- [x] apm-jmx-plugin N/A
- [x] apm-jsf-plugin https://github.com/elastic/apm-agent-java/pull/1405
- [x] apm-kafka-plugin #2095 
- [x] apm-log-correlation-plugin N/A
- [x] apm-log-shipper-plugin N/A
- [x] apm-micrometer-plugin
- [x] apm-mongoclient-plugin https://github.com/elastic/apm-agent-java/pull/1369
- [x] apm-okhttp-plugin https://github.com/elastic/apm-agent-java/pull/1368
- [x] apm-opentracing-plugin #1410
- [x] apm-process-plugin
- [x] apm-profiling-plugin N/A
- [x] apm-quartz-job-plugin https://github.com/elastic/apm-agent-java/pull/1366
- [x] apm-redis-plugin #2090
- [x] apm-scala-concurrent-plugin https://github.com/elastic/apm-agent-java/pull/1367
[...]

PR #1371, opened 2020-09-01
migrate to indy grails plugin

kananindzya:
refer #1337


apm-agent-plugins/apm-grails-plugin/src/main/java/co/elastic/apm/agent/grails/GrailsTransactionNameInstrumentation.java
```diff
@@ -114,7 +107,6 @@ static void setTransactionName(@Advice.Argument(2) Object handler) {
             setName(transaction, className, methodName);
         }
 
-        @VisibleForAdvice
         public static void setName(Transaction transaction, String className, @Nullable String methodName) {
```
felixbarny, line 110:
```suggestion
        private static void setName(Transaction transaction, String className, @Nullable String methodName) {
```

felixbarny (approved):

apm-agent-plugins/apm-grails-plugin/src/main/java/co/elastic/apm/agent/grails/GrailsTransactionNameInstrumentation.java
kananindzya: with private - this method not visible:
```
.UrlMappingsInfoHandlerAdapter
java.lang.IllegalStateException: private static void co.elastic.apm.agent.grails.GrailsTransactionNameInstrumentation$HandlerAdapterAdvice.setTransactionName(java.lang.Object) is not visible to class org.grails.web.mapping.mvc.UrlMappingsInfoHandlerAdapter
	at net.bytebuddy.asm.Advice$Dispatcher$Delegating$Resolved.bind(Advice.java:8887) ~[byte-buddy-dep-1.10.12.jar:?]
```

felixbarny: `setTransactionName` has to remain public but I think `setName` can be private.

kananindzya: done

SylvainJuge (approved):

PR #1374, opened 2020-09-01
migrate to indy jaxrs plugin

kananindzya:
refer #1337
hello, `useAnnotationValueForTransactionName` returns `false` during the test. But when `JaxRsTransactionNameInstrumentation` construct `useAnnotationValueForTransactionName`  mocked with `true` value:
```
            if (useAnnotationValueForTransactionName) {
                if (pathAnnotationValue != null) {
                    transactionName = pathAnnotationValue;
                }
            }
```


felixbarny:
You'll have to use `GlobalVariables` for `useAnnotationValueForTransactionName`, as the `JaxRsTransactionNameInstrumentation` class is loaded twice: once from the system class loader and once from the `IndyPluginClassLoader`. The constructor will only be called for the class loaded by the system class loader. 
The other class will use the initial value of `false`.

kananindzya:
> `GlobalVariables`

Thanks

kananindzya:
@SylvainJuge , @felixbarny  please review this PR
Thanks

felixbarny (approved):

SylvainJuge:
There is an unexpected `java.lang.IllegalStateException: Cannot resolve type description for co.elastic.apm.agent.httpclient.AbstractHttpClientInstrumentationTest` in the tests, @felixbarny can I let you handle this ?

Issue #1375, opened 2020-09-01
Flaky tests due to connection reset when downloading dependencies


felixbarny:
## Test method

Happens before tests, in the build phase

## Test logs

```
[2020-09-01T07:08:56.951Z] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce (enforce-java) on project apm-agent-parent: Execution enforce-java of goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce failed: Plugin org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1 or one of its dependencies could not be resolved: The following artifacts could not be resolved: org.apache.maven:maven-settings:jar:3.0, org.apache.maven:maven-repository-metadata:jar:3.0, org.apache.maven:maven-model-builder:jar:3.0: Could not transfer artifact org.apache.maven:maven-settings:jar:3.0 from/to central (https://repo.maven.apache.org/maven2): Transfer failed for https://repo.maven.apache.org/maven2/org/apache/maven/maven-settings/3.0/maven-settings-3.0.jar: Connection reset -> [Help 1]
```

## Additional context

Lately, we see a lot of test failures because of `Connection reset` errors.

@elastic/observablt-robots do you know if there are network issues?

Also, why does maven even need to resolve the dependencies? Shouldn't they be cached in the image via packer cache?

SylvainJuge:
There is currently some work being done by the CI team on this, not sure about the current progress and priority though.

cachedout:
We may be able to add this to the [image cache](https://github.com/elastic/apm-agent-java/blob/master/.ci/packer_cache.sh) if it's not there already. I will investigate this.

v1v:
I'm working on something else that seems similar to this issue, and this is my finding:
- Caching is not happening since the .m2 folder is set to another location:
  - https://github.com/elastic/apm-agent-java/blob/master/Jenkinsfile#L17

That particular configuration was there since the very beginning:
- https://github.com/elastic/apm-agent-java/commit/5fccf62177f015f9d7c40fdd98805013a7cd3dd3

https://github.com/elastic/apm-agent-java/pull/1376 might fix this issue, at least the one related to the cached artifacts

v1v:
And guess what! I faced this particular issue in my above-mentioned fix:

For what I read the `-Dhttps.protocols=TLSv1.2` flag seems to be the workaround:

[...]

PR #1376, opened 2020-09-01
[CI] use default m2

v1v:
## What does this PR do?

Avoid override the .m2 location

Relates https://github.com/elastic/apm-agent-java/issues/1375


mdelapenya (approved):

felixbarny:
IIRC, the reason we did that was that the `~./m2` folder can't be cached in the packer cache as it's not in the current directory.

We don't seem to set it in the packer_cache.sh script though.

However, I've found more references to that directory in `.ci/release/Jenkinsfile` and `.ci/settings.xml`.

v1v:
Aha! this particular approach will not allow to stash/unstash the whole build for the other stages, in other words, the very first build might fetch all the dependencies which will be stashed and then used in the following stages. Therefore this particular PR breaks the current approach of building once and use the transitive artifacts later on

PR #1377, opened 2020-09-01
[CI] fix peer reset when downloading dependencies

v1v:
## What does this PR do?

Fixes the environmental issues when the dependencies are downloaded but the connection gets resetted, for such let's use the suggested workaound

Closes https://github.com/elastic/apm-agent-java/issues/1375

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -31,7 +31,7 @@ pipeline {
     issueCommentTrigger('(?i).*(?:jenkins\\W+)?run\\W+(?:the\\W+)?(?:benchmark\\W+)?tests(?:\\W+please)?.*')
   }
   parameters {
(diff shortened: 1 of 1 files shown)


felixbarny (approved):
Where did you get that workaround from? Worth a try.

v1v:
I found the workaround in the below pages from stackoverflow:

- [ref1](https://stackoverflow.com/questions/50824789/why-am-i-getting-received-fatal-alert-protocol-version-or-peer-not-authentic)
- [ref2](https://stackoverflow.com/questions/34170811/maven-connection-reset-error)

SylvainJuge (approved):

v1v:
`Integration Tests` are broken for some other reasons

PR #1391, opened 2020-09-09
Fix HttpUrlConnection instrumentation

eyalkoren:
Currently our instrumentation only looks at the `java.net.URLConnection#connected` field and not the `sun.net.www.protocol.http.HttpURLConnection#connecting` when trying to add the header.
Since the `connecting` field was only introduced in Java 8, we are doing something instead.

In addition - removing the restriction for using `ThreadLocal` within advice classes.

### DIFF
--- a/apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
+++ b/apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
@@ -28,6 +28,7 @@ import co.elastic.apm.agent.bci.TracerAwareInstrumentation;
 import co.elastic.apm.agent.http.client.HttpClientHelper;
 import co.elastic.apm.agent.impl.transaction.Span;
 import co.elastic.apm.agent.impl.transaction.TraceContext;
+import co.elastic.apm.agent.sdk.state.CallDepth;
 import co.elastic.apm.agent.sdk.weakmap.WeakMapSupplier;
 import com.blogspot.mydailyjava.weaklockfree.WeakConcurrentMap;
 import net.bytebuddy.asm.Advice;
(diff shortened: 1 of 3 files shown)


apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
```diff
@@ -51,8 +50,15 @@
 
 public abstract class HttpUrlConnectionInstrumentation extends TracerAwareInstrumentation {
 
-    @VisibleForAdvice
-    public static final WeakConcurrentMap<HttpURLConnection, Span> inFlightSpans = WeakMapSupplier.createMap();
+    private static final WeakConcurrentMap<HttpURLConnection, Span> inFlightSpans = WeakMapSupplier.createMap();
+
+    // Used instead of inspecting sun.net.www.protocol.http.HttpURLConnection.connecting which was added in Java 8
+    private static final ThreadLocal<Boolean> connecting = new ThreadLocal<>() {
```
felixbarny, line 56: Creating a subclass of `ThreadLocal` causes leaks. See also https://cwiki.apache.org/confluence/display/TOMCAT/MemoryLeakProtection#MemoryLeakProtection-customThreadLocal.

Either use no initial value or use `CallDepth` which seems to be applicable to this case.

eyalkoren: Ahh, the anonymous class...
Thanks!

felixbarny (approved):

SylvainJuge (approved):
:+1: that's so much better to not rely on JDK internal implementation !

felixbarny:
Is there a good way to test this fix?

eyalkoren:
> Is there a good way to unit test this fix?

Not a very straightforward one. The reported stack trace is:
```
java.lang.IllegalStateException: Already connected
at sun.net.www.protocol.http.HttpURLConnection.addRequestProperty(HttpURLConnection.java:3190)
at co.elastic.apm.agent.urlconnection.UrlConnectionPropertyAccessor.setHeader(UrlConnectionPropertyAccessor.java:44)
at co.elastic.apm.agent.urlconnection.UrlConnectionPropertyAccessor.setHeader(UrlConnectionPropertyAccessor.java:34)
at co.elastic.apm.agent.impl.transaction.TraceContext.propagateTraceContext(TraceContext.java:530)
at co.elastic.apm.agent.impl.transaction.AbstractSpan.propagateTraceContext(AbstractSpan.java:558)
at sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.connect(AbstractDelegateHttpsURLConnection.java:175)
at sun.net.www.protocol.https.DelegateHttpsURLConnection.connect(DelegateHttpsURLConnection.java)
at sun.net.www.protocol.http.HttpURLConnection.getOutputStream0(HttpURLConnection.java:1350)
at sun.net.www.protocol.http.HttpURLConnection.getOutputStream(HttpURLConnection.java:1325)
```
We shouldn't get to the point of trying to add headers again, unless the instrumentation of `sun.net.www.protocol.http.HttpURLConnection.getOutputStream` is disabled. I don't know exactly what happened causing this, but in any case trying to add headers to a connection while its `connecting` state is `true` is not allowed.

PR #1393, opened 2020-09-09
WIP - try to make maven downloads less flaky on CI

SylvainJuge:
Attempt to make maven downloads less flaky on CI

On master, over the last 20 builds, 4 failed due to a download dependency error with "connection reset" error messages, thus we currently have at least 20% failed builds.

## Hypotheses/Observations

- [x] trying to set a high number of retries with ` -Dmaven.wagon.http.retryHandler.count=5`
:red_circle: it failed once after only 5 builds, thus no positive impact.
- even if artifacts are present in cache, maven tries to reach remote repository, for example we have messages like `checking for updates from central`

- [x] trying to disable keep-alive
:green_circle:  slows downs the build (from 6 to 10 min), but it works

- [x] trying to use undocumented wagon setting `-Dmaven.wagon.httpconnectionManager.ttlSeconds=120`
:red_circle: :  failed at the 2cnd build

- [x] trying `-Dmaven.wagon.http.pool=false`
:green_circle: slows down the build (from 6min to 9min), but it works

- [ ] trying to change network settings, [AWS NAT drops connections after 350s](https://docs.aws.amazon.com/vpc/latest/userguide/nat-gateway-troubleshooting.html#nat-gateway-troubleshooting-timeout)
```
sudo sysctl -w net.ipv4.tcp_keepalive_time=60
sudo sysctl -w net.ipv4.tcp_keepalive_intvl=60
sudo sysctl -w net.ipv4.tcp_keepalive_probes=10
```


SylvainJuge:
Closing this PR for now:
- infrastructure changes required with implication for infra/security approval, thus harder to expect in the short term.
- temporary workaround has been applied with #1399

Issue #953, opened 2019-12-03
Check compatibility with older APM Server versions


felixbarny:
Whenever we leverage new features in the Intake API, we have to make sure we are still compatible with older APM Server versions, possibly by conditionally making use of these features, based on the APM Server version(s) we are connected to. We currently do this for allowing non-string values for `context.tags`.

* [x] allowing int for `context.request.url.port` https://github.com/elastic/apm-server/commit/af76d15f4ab07468d4b5ae70ed11d34b52014c1a#diff-b3c35663f19380d55a7b0b375e1aeb4a 
* [x] restricting `error.culprit` to 1024 chars https://github.com/elastic/apm-server/commit/a6de0b066fb181592d8979a21192411255f50623#diff-c3b216f97537a6b919b5e7c300b6ac12
* [x] make `framework.name` and `framework.version` optional https://github.com/elastic/apm-server/commit/40d48fa7272b20e039e4b6e31109162649e462ce#diff-e544f4b544de7a8f403607374cc0eacf
* [x] Allow number for `tags` https://github.com/elastic/apm-server/commit/a2e6d884ec5efe4c043d3f2d0185ddaab295191b#diff-0d2e9aedcc4779d4c198c03f7ff3febe
* [x] Additionally allow array for headers https://github.com/elastic/apm-server/commit/f0f3662a2fc96a06077b193c12fc5f4b00588eb8#diff-b3c35663f19380d55a7b0b375e1aeb4a and https://github.com/elastic/apm-server/commit/ace70ded3d966ca2d945c904371017ac589db3c2#diff-b3c35663f19380d55a7b0b375e1aeb4a
* [x] make `lineno` optional for `stacktrace` frames https://github.com/elastic/apm-server/commit/5cbcd7799465f77006c4f676e3080d6ef2e9b787#diff-6ebe279047a91cedeb9464344128e2b7
* [x] make `transaction_id` optional for `spans` https://github.com/elastic/apm-server/commit/4f74aead2adfc6df303c584b06171461f6777e0c#diff-07fc096ed421fd31226e44717413b53b

SylvainJuge:
I feel this is something that would be a good extension of #882 (check agent payloads against server protocol definition), in the way that instead of checking the "current" version, we would also check for known versions we are supposed to be compatible with.

felixbarny:
@simitt could you confirm what 
> Additionally allow array for headers

Exactly means?

[...]

PR #1395, opened 2020-09-10
Validate server schema

### LINKED ISSUE #882: Validate intake protocol against server repository specs
**Describe the bug**

Currently the agent validates JSON serialization against a copy of the server schemas stored in this project subfolder : [apm-agent-core/src/main/resources/schema](https://github.com/elastic/apm-agent-java/tree/master/apm-agent-core/src/main/resources/schema).

This allows to have schema definitions that can be fine tuned to fit the java schema validator used in tests, which makes them
- tedious to maintain, thus rarely updated outside of ad-hoc patches
- out-of-sync with backend server, which prevents detecting regressions

What we should probably do instead is:
- fetch those definitions from [backend server specs](https://github.com/elastic/apm-server/tree/master/docs/spec) at build time
- apply any required transformations to the JSON before using them
- remove those files from current repository when tests have been migrated to their server
- provide a way to select which git branch/tag/ref to select the server version.

-----

## Implementation proposal

Agent is known to currently be compatible with apm-server `6.5.0`, as of version `7.9.1` that makes ~48 potential schema versions to test against. Testing against all of them systematically would be counter-productive as not all of them imply schema changes. Thus first we have to identify which versions have relevant schema changes.

Fetching schema at build time (for example in CI) from apm-server repository is doable, but would require two things: first having a proper test suite that allows to take version as a parameter (we don't ) and second to have a proper execution of those tests against the list of versions defined above.

Currently, schema validation is applied when using the `MockTracer` class in unit tests to report any `error`, `transaction` or `span` object. That means that validation is applied implicitly by unit tests when using `MockTracer`.

On the other hand, having everything locally within the repository
- would be close to what we have for agent spec updates from elastic/apm repository, where a copy of spec update is pushed with a PR on every agent repository. We do not plan to automate this for now.
- adds numerous immutable copies of the schema, one for each version we want to cover (we should provide a script to get/update them)
- allows to deal with any schema modification if required (even manual) required to make it compatible with our json validator.
- minimal impact to current tests: we would just validate against multiple schema versions instead of one
- allows to make those tests part of regular unit-tests, thus trivial to debug with IDE, less reliance on CI that tends to be slow and suffer from other tests flaky-ness.

### Out of scope for now
- automate schema update, it will remain manual for now. We might consider doing in the future something similar to what we do with gherkin/json specs updates from apm to apm-agent-* repositories.
- test 100% server compatibility, we might still call endpoints that aren't available (for example, shipping logs).

### High level steps
- [x] provide a shell script that allows to fetch a given version of schema from `apm-server` repository, this will be used for manual updates
- [x] identify which versions have significant schema changes and store them in distinct folders (one per version).
- [x] make `MockTracer` validate against multiple schema versions instead of a single one.
- [ ] validate testing with known schema changes listed here https://github.com/elastic/apm-agent-java/issues/953

SylvainJuge:
## What does this PR do?

Fixes #882

Partially cover changes listed in https://github.com/elastic/apm-agent-java/issues/953 for known breaking changes:
- numeric `url.port` introduced in `7.0`
- tags with numeric values introduced in `6.7`

## Checklist

- [x] This is something else

## Implementation checklist

- [x] make current schema(s) reusable as-is by applying modifications on the fly (fixing relative paths & making it classpath-friendly).
- [x] test locally with all relevant versions listed in https://github.com/elastic/apm-agent-java/issues/882#issuecomment-689409219
- [ ] validation currently passes when there are extra fields. Thus we only deal with breaking changes. Given server ignores extra fields it might not worth the effort/complexity to deal with every version (at least 13 of them so far)


eyalkoren:
#1675 is the first automated PR synching current schema files. The files were added to the same directory used in this PR - `apm-agent-core/src/test/resources/apm-server-schema/current`, so we merging master will add them where expected. However, the structure is a bit different - the automatically added JSONs are only 5, incorporating sub JSONs within them.

PR #1397, opened 2020-09-14
Eliminate Java-11 HttpClient plugin class version errors

eyalkoren:
Eliminating errors such as:
```
co.elastic.apm.agent.util.DependencyInjectingServiceLoader - unable to instantiate 'co.elastic.apm.agent.httpclient.HttpClientInstrumentation', unsupported class version error: co/elastic/apm/agent/httpclient/HttpClientInstrumentation has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```
from agent logs.
These are not real errors, they only indicate that the Java 11 HttpClient plugin cannot be loaded in pre-Java-11 JVMs.


apm-agent-core/src/main/java/co/elastic/apm/agent/util/DependencyInjectingServiceLoader.java
```diff
@@ -138,7 +138,12 @@ private T instantiate(ClassLoader classLoader, String implementation) {
             String msg = String.format("unable to instantiate '%s', please check descriptor in META-INF", implementation);
             throw new ServiceConfigurationError(msg, e);
         } catch(UnsupportedClassVersionError e) {
-            logger.error(String.format("unable to instantiate '%s', unsupported class version error: %s", implementation, e.getMessage()));
+            // The Java 11 HttpClient plugin is compiled with Java 11, so this is expected
+            if (!implementation.startsWith("co.elastic.apm.agent.httpclient")) {
```
SylvainJuge, line 142: I'm not really a fan of having something specific to the httpclient, as adding other similar plugins for java11 will need to modify this file again.

Thus keeping only the log as `debug` would probably be better for now, as we don't have any reliable way to tell if a plugin/instrumentation is _expected_ to be compatible with a given version, the only thing that we can detect is which version it has been compiled as target (we could read this from bytecode).

eyalkoren: OK, I will change to `debug`. 
In my view - this is indeed an error and I want indication of doing something wrong in my plugin. We shouldn't try to load classes where they shouldn't be loaded.
I may do a separate PR to propose a different way to deal with that. It will include some refactoring, but I think it would be the right way to go.

SylvainJuge (approved):

PR #1409, opened 2020-09-18
lazy-load jdbc helper

SylvainJuge:
## What does this PR do?

JDBC instrumentation fails on some environments when active.
This appears to be a side-effect of migration to new plugin architecture introduced in version 1.18.0.
We haven't been able to reproduce this directly, but changes have been confirmed to work where it used to happen.

## Checklist

- [x] This is a bugfix

### DIFF
--- a/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/JdbcInstrumentation.java
+++ b/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/JdbcInstrumentation.java
@@ -43,7 +43,7 @@ public abstract class JdbcInstrumentation extends TracerAwareInstrumentation {
 
     @Override
     public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
-        return classLoaderCanLoadClass("java.sql.Statement");
+        return classLoaderCanLoadClass("java.sql.Statement"); // in case java.sql module is not there
     }
 
     @Override
--- a/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/JdbcInstrumentation.java
+++ b/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/JdbcInstrumentation.java
@@ -42,11 +42,6 @@ public abstract class JdbcInstrumentation extends TracerAwareInstrumentation {
     private static JdbcHelper jdbcHelper;
 
     @Override
-    public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
-        return classLoaderCanLoadClass("java.sql.Statement"); // in case java.sql module is not there
-    }
-
-    @Override
     public final Collection<String> getInstrumentationGroupNames() {
         return JDBC_GROUPS;
     }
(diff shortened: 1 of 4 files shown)


apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/JdbcInstrumentation.java
```diff
@@ -26,18 +26,37 @@
...
     private static final Collection<String> JDBC_GROUPS = Collections.singleton("jdbc");
 
-    protected static JdbcHelper jdbcHelper = new JdbcHelper();
+    @Nullable
+    private static JdbcHelper jdbcHelper;
+
+    @Override
+    public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
+        return classLoaderCanLoadClass("java.sql.Statement"); // in case java.sql module is not there
+    }
```
eyalkoren, line 47: Does this contradict the actual fix of the lazy loading? 
Meaning - the lazy loading assumes that `java.sql` classes should be available to JDBC code when it is invoked and the problem is that they are not yet in the classpath when the instrumented class is loaded. If that's the case, the tested class loader will be excluded, even though it shouldn't be,

SylvainJuge: That's a fair point, I added this one when I expected there was an issue with java modules, but those weren't in use.

The issue was that the static call to `new JdbcHelper()` triggered loading JDBC classes in the wrong classloader, which I wasn't able to reproduce, even when using the same JVM.

SylvainJuge: This extra check has been removed, and the fix is still working. Thus I've removed it in a later commit.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/JdbcInstrumentation.java
```diff
@@ -26,18 +26,32 @@
...
+    @Nullable
+    private static JdbcHelper jdbcHelper;
 
     @Override
     public final Collection<String> getInstrumentationGroupNames() {
         return JDBC_GROUPS;
     }
+
+    protected synchronized static JdbcHelper getJdbcHelper() {
+        // lazily initialize helper to prevent trying to load classes in java.sql package with the bootstrap classloader
```
felixbarny, line 50: Is that an indication that we should more strictly separate advices from element matches?

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/JdbcInstrumentation.java
```diff
@@ -26,18 +26,32 @@
...
-    protected static JdbcHelper jdbcHelper = new JdbcHelper();
+    @Nullable
+    private static JdbcHelper jdbcHelper;
 
     @Override
     public final Collection<String> getInstrumentationGroupNames() {
         return JDBC_GROUPS;
     }
+
+    protected synchronized static JdbcHelper getJdbcHelper() {
```
felixbarny, line 49: Use the double-checked locking pattern to avoid the synchronized overhead

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/JdbcInstrumentation.java
SylvainJuge: I would say yes, separation would be better. In the same category, there is `getAdviceClass()` method that returns the advice class loaded from agent/bootstrap classloader when in practice we only use class name and then load it in child-first plugin classloader. We already discussed that a bit with @eyalkoren last week.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/JdbcInstrumentation.java
SylvainJuge: fixed in last commit.

eyalkoren (approved):

PR #1416, opened 2020-09-23
Catch HttpUrlConnection Already Connected error

eyalkoren:
This reverts the change introduced in #1391 - the nesting check approach is only valid if the nested calls are not related to the same method. For example, consider this stack trace:
```
   sun.net.www.protocol.http.HttpURLConnection.connect()
   com.example.HttpUrlConnectionWrapper.connect()
```
where `com.example.HttpUrlConnectionWrapper` does not delegate `addRequestProperty` to the actual implementation, which means it must be invoked through `sun.net.www.protocol.http.HttpURLConnection` in order to add headers. The current call-depth restriction does not allow that.

Instead, this PR always tries to add headers to requests that do not contain them already, and we catch the related `IllegalStateException` indicating it is `Already connected`.


SylvainJuge:
Do we have any test that covers this case ? Or does testing with the current JDK (11) makes it always work by default ?

SylvainJuge (approved):

eyalkoren:
Do you have an idea for a proper test in mind? 
Basically, this reverts the former PR - we always try to add the headers if the were not added yet and we capture the `Already connected` Exception when thrown. In Java 8 and above this exception will be thrown.

SylvainJuge:
We should be able to write some test code that triggers the `Already
connected` exception.
Without this fix, our agent should throw an exception (but we might not yet
be able to assert for "no thrown exception").

PR #1419, opened 2020-09-24
fix micrometer invalid double values

SylvainJuge:
## What does this PR do?

Micrometer plugin reports metrics through a dedicated JSON serialization code, and does not properly handle corner cases with gauges that can have values of `Double.NaN` or `Double.NEGATIVE_INFINITY`, as a result there are cases where metric sets are sent with strings as values instead of numbers, for example `-Infinity`.

Ideally in the future if we start to have multiple plugins that report metrics, we should probably provide an internal API for this so there is no duplication of code and potential issues.

## Checklist

- [x] This is a bugfix

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -36,6 +36,7 @@ endif::[]
 * Eliminate `unsupported class version error` messages related to loading the Java 11 HttpClient plugin in pre-Java-11 JVMs {pull}1397[1397]
 * Fix rejected metric events by APM Server with response code 400 due to data validation error - sanitizing Micrometer
 metricset tag keys - {pull}1413[1413]
+* Fix invalid micrometer metrics with non-numeric values {pull}1419[1419]
 
 [[release-notes-1.x]]
 === Java Agent version 1.x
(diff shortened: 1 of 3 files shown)


felixbarny (approved):

eyalkoren (commented):
Lots of other meters' APIs return `double` values, like timers `totalTime()` and even `count()`. Can these fail us as well?
Maybe this check should be moved to `MicrometerMeterRegistrySerializer#serializeValue()` that can return a `boolean` indicating if the value was added?

SylvainJuge:
> Lots of other meters' APIs return `double` values, like timers `totalTime()` and even `count()`. Can these fail us as well?
> Maybe this check should be moved to `MicrometerMeterRegistrySerializer#serializeValue()` that can return a `boolean` indicating if the value was added?

I have checked implementation of those (at least the version we compile against for the micrometer plugin), and they default to zero, thus I assume it is mostly safe to do so. End-user provided sub-classes aren't really expected here as the instances are provided through a builder from micrometer framework.

eyalkoren:
OK, we will fix more if required then

eyalkoren (approved):

pkgonan:
@eyalkoren @felixbarny 
Hi, when do you expect release date this PR?
Perhaps 1.18.1 version.

Thank you.

SylvainJuge:
Hi @pkgonan !
The next release isn't scheduled yet, thus make sure that you have subscribed to notifications.

In the mean time, you can use the [snapshot](https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=co.elastic.apm&a=elastic-apm-agent&v=LATEST) that is built every day from master if required.

pkgonan:
@SylvainJuge 
Thank you for your reply.

PR #1425, opened 2020-10-05
Adding trace-level log info to HTTP spans

eyalkoren:
I added all in `TRACE` level. We ask for `DEBUG` logs frequently, we don't want a per-request/http-exit logging.
If you have any ideas for additional info we may be interested at - let me know.


SylvainJuge (approved):
lgtm.

[minor] do we really require the `isTraceEnabled()` call first with slf4j for those cases ?

eyalkoren:
> do we really require the isTraceEnabled() call first with slf4j for those cases ?

I used it where a `toString` would be always invoked otherwise, to spare the related (potentially multiple) allocations

PR #1437, opened 2020-10-12
WIP | bump bytebuddy + enhance debug

SylvainJuge:
This is a WIP debug build that allows to capture more information about loss of context propagation issues with HTTP clients.


eyalkoren (commented):
My opinion is that anything that creates a log entry in a per-hit basis **and does not indicate an unexpected state** should not be in `DEBUG` level, but `TRACE`.
So, while this is useful for our current debug need, I prefer if we change that before merging. WDYT?

felixbarny:
Agree, we shouldn't merge this to master as-is. This is to debug a specific problem only.

SylvainJuge:
Debug build not required anymore.

PR #1438, opened 2020-10-13
Remove JVM-metrics note

eyalkoren:


felixbarny (approved):
I'm fine with just removing it. Alternatively, we could document (or link to existing docs if available) where to find the JVM metrics in the UI.

PR #1440, opened 2020-10-14
Disable the agent when -XX:+UseG1GC is used on Java 7

### LINKED ISSUE #1435: Disallow agent usage alongside G1 GC in Java 7

When choosing to use G1 GC in Java 7 (through the `-XX:+UseG1GC` JVM parameter), the agent may expose JVM bug causing JVM crash, that was not previously affecting the traced application.

- [ ] Disable the agent if running on Java 7 `AND` the `-XX:+UseG1GC` is used. Can be done through JMX using `ManagementFactory.getRuntimeMXBean().getInputArguments()`, but preferably we can do it right on `premain` or `agentmain` and not wait until the `MBeanServer` is setup.
- [ ] Log corresponding `ERROR`
- [ ] Add to documentation - supported technologies page and troubleshooting

Java 7 using G1 GC is not **guaranteed** to crash when the Java agent is attached, so we may choose to only `WARN` and allow the agent to trace. However, the effect of crashing an app on production after not detected in staging env has severe implications for users. We may want to consider an opt-in option for that.
## Additional context

Noticed on WebLogic 12.1.2 and possibly on JBoss.

eyalkoren:
## What does this PR do?
Closes #1435 

No need to take special measures with the JMX operation - we already have it in
https://github.com/elastic/apm-agent-java/blob/842185664456651f1d6a99602d21a7c052ddf111/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/ProcessFactory.java#L91
which is executed during initialization.
Therefore, I modified the print of JVM arguments to be done at `INFO` level.

I also restored the ability to test servlet containers with `-javaagent` even if they are tested through runtime attachment by default (only applied to Tomcat for now) in order to test manually.

## Checklist

- [x] Updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc) - under breaking changes
- [x] Add a unit test for the discovery of the JVM argument
- [x] Test manually that agent is not started when this configuration is detected, both as runtime attach and as `javaagent`
- [x] Test manually that other JVM info dependent issues are not affected 
- [x] Test manually that using this JMX operations works on J9 (and capture related errors anyway)
- [x] Document in the troubleshooting area
- [x] Document as a caveat related to Java 7 support (in supported technologies)

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
@@ -48,6 +48,7 @@ import co.elastic.apm.agent.report.ReporterConfiguration;
 import co.elastic.apm.agent.sdk.weakmap.WeakMapSupplier;
 import co.elastic.apm.agent.util.DependencyInjectingServiceLoader;
 import co.elastic.apm.agent.util.ExecutorUtils;
+import co.elastic.apm.agent.util.JvmRuntimeInfo;
 import com.blogspot.mydailyjava.weaklockfree.WeakConcurrentMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -496,6 +497,13 @@ public class ElasticApmTracer implements Tracer {
         }
     }
 
+    /**
+     * Starts the tracer. Depending on several environment and setup parameters, the tracer may be started synchronously
+     * on this thread, or its start may be delayed and invoked on a dedicated thread.
+     *
+     * @param premain true if this tracer is attached through the {@code premain()} method (i.e. using the `javaagent`
+     *                jvm parameter); false otherwise
+     */
     public synchronized void start(boolean premain) {
         long delayInitMs = getConfig(CoreConfiguration.class).getDelayInitMs();
         if (premain && shouldDelayOnPremain()) {
@@ -509,9 +517,7 @@ public class ElasticApmTracer implements Tracer {
     }
 
     private boolean shouldDelayOnPremain() {
-        String javaVersion = System.getProperty("java.version");
-        return javaVersion != null &&
-            (javaVersion.startsWith("1.7") || javaVersion.startsWith("1.8")) &&
+        return JvmRuntimeInfo.getMajorVersion() <= 8 &&
             ClassLoader.getSystemClassLoader().getResource("org/apache/catalina/startup/Bootstrap.class") != null;
     }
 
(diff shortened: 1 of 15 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/util/JvmRuntimeInfo.java
```diff
@@ -0,0 +1,201 @@
...
+                // the exact version that fixes that error is currently not known
+                // presumably, service refresh 5 (build 2.8) fixes the issue
+                return !"2.8".equals(javaVmVersion);
+            default:
+                return true;
+        }
+    }
+
+    /**
+     * NOTE: THIS METHOD REQUIRES JMX OPERATION, THEREFORE IS SHOULD ONLY BE USED WITHIN OR AFTER THE TRACER HAVE
```
SylvainJuge, line 170: Could we check for tracer state and throw an exception if triggered before tracer has been started ?

SylvainJuge (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/util/JvmRuntimeInfo.java
eyalkoren: I didn't want to do any JMX query at initialization time because of my JULI deadlock trauma (this kicks all kind of initializations). 
However, @felixbarny feels it is safe enough to do this simple JMX query very early on, as long as we don't try to `getPlatformMBeanSerever`. It certainly simplifies things a lot (no need to uninstrument classes just after they are done being instrumented). So we'll go with that. I will test and merge the change soon

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -539,6 +546,11 @@ private synchronized void startSync() {
             logger.warn("Trying to start an already initialized agent");
             return;
         }
+        if (!JvmRuntimeInfo.isJvmConfigurationSupported()) {
+            logger.error("JVM configuration is not supported. The agent will not be started and classes are un-instrumented.");
```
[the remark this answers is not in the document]
felixbarny: Forgot to submit this yesterday^ 🙄

apm-agent-core/src/main/java/co/elastic/apm/agent/util/JvmRuntimeInfo.java
felixbarny: I'd also recommend adding a way to disable the bootstrap checks, for example with a system property, in case it does cause problems. It can't be a config option as these get initialized later.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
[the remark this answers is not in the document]
eyalkoren: Yes, testing the changes now. 
The reason I assume the damage is not already done is the fact that reports we got described a crash that happens some time after agent is loaded, maybe only happen on JITed code 🤷‍♂️ .

felixbarny (approved):

PR #1443, opened 2020-10-19
Async Profiler - add support for AArch64

### LINKED ISSUE #1439: profiler aarch64 support
As reported [in our forum](https://discuss.elastic.co/t/aarch64-apm-java-client-support/251714/5).

It seems that trying to use async-profiler on aarch64 architecture fails with the following message:
```
Caused by: java.lang.UnsatisfiedLinkError: /usr/bin/libasyncProfiler-linux-arm-672c6dfb97228a24cd9de46d11a995ed.so: /usr/bin/libasyncProfiler-linux-arm-672c6dfb97228a24cd9de46d11a995ed.so: wrong ELF class: ELFCLASS32 (Possible cause: can't load ARM .so on a AARCH64 platform)
```

However, async-profiler should support aarch64 as it's part of the [supported platforms](https://github.com/jvm-profiling-tools/async-profiler#supported-platforms)

This should be reproducible by enabling the profiler on an aarch64 machine with `ELASTIC_APM_PROFILING_INFERRED_SPANS_ENABLED=true`

eyalkoren:
Closes elastic/apm-agent-java#1439


apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/asyncprofiler/AsyncProfiler.java
```diff
@@ -102,8 +102,10 @@ private static String getLibraryFileName() {
         String os = System.getProperty("os.name").toLowerCase();
         String arch = System.getProperty("os.arch").toLowerCase();
         if (os.contains("linux")) {
-            if (arch.contains("arm") || arch.contains("aarch")) {
+            if (arch.contains("arm")) {
```
felixbarny, line 105: Does this binary work for both 64 and 32-bit arm?

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/asyncprofiler/AsyncProfiler.java
```diff
@@ -102,8 +102,10 @@ private static String getLibraryFileName() {
...
                 return "libasyncProfiler-linux-arm";
+            } else if (arch.contains("aarch")) {
+                return "libasyncProfiler-linux-aarch64";
```
felixbarny, line 108: Does this binary work for both 64 and 32-bit aarch? Is there even a 32-bit version? Throw exception if not 64 bit?

eyalkoren: It's 64-specific. I believe the `os.arch` for `aarch32` should be `arm`. 
@apangin can you confirm?

apangin: Right, `os.arch` for 32-bit ARM is `arm`, for 64-bit ARM is `aarch64`.
There was even a [fix](http://mail.openjdk.java.net/pipermail/aarch32-port-dev/2016-October/000514.html) to make `os.arch` return `arm` on systems that report `aarch32`.

eyalkoren: Thanks for confirming! Seems like it should be good enough then, but do you think it's better to look for `aarch64`? Can there be a AArch64 OS where the `os.arch` contains `aarch` but not `aarch64`?

apangin: IMO, doesn't matter. You may look for `aarch` or `aarch64` - I expect this to be the same set of systems.

eyalkoren: OK, good enough for me - let's leave that as is. We can change in the future if we get reports this doesn't support values in specific systems.

eyalkoren: > There was even a fix to make os.arch return arm on systems that report aarch32

Even though this fix is old, I made it more specific so that we do not fail on such old JDKs.

felixbarny (approved):

PR #1449, opened 2020-10-21
feature added setLabel methods to Span, Transaction of public api

### LINKED ISSUE #1446: Public API - deprecate addLabel in favor of setLabel

Most other agents use `setLabel` (or `SetLabel` or similar), we should align.

Add three new `setLabel` methods to the API and deprecate `addLabel`

Allowing both

kananindzya:
closes elastic/apm-agent-java#1446


apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
```diff
@@ -162,10 +194,10 @@
      *
      * @param key   The label key.
      * @param value The label value.
-     * @since 1.5.0, APM Server 6.7
+     * @since 1.9.0, APM Server 6.7
```
SylvainJuge, line 197: [question] should it be `1.19` here for the `@since` javadoc tag instead ?

SylvainJuge (approved):
LGTM, I just have one question regarding the `@since` javadoc tag value.

apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
eyalkoren: Of course, good catch, thanks!

felixbarny:
The APM getting started docs also mention `addLabel` so they need to be updated as well: https://www.elastic.co/guide/en/apm/get-started/current/metadata.html#labels-fields

PR #1459, opened 2020-10-27
Setting context class loader of agent threads to null

eyalkoren:
Setting context class loader of agent threads explicitly to `null` to avoid Tomcat's warnings about possible memory leaks


SylvainJuge (approved):
LGTM, can't really hurt as it's only our threads, and we know it might be a potential issue.

PR #1467, opened 2020-11-02
Agent load testing

cachedout:
## Summary

This adds support for load-testing the Java agent.

## Details

This is a new pipeline which introduces the ability to provision any version of the Java agent, instrument a test application with it, and then apply load to that application in order to test the stability and performance characteristics of the instrumented application over time.

When launching the pipeline, users are presented with the following:

pipe

As you can see, it allows the user to select the version of the agent to deploy, along with the JDK to use. You can also select the duration of the test. Load is [generated via Locust](https://locust.io) and if you wish to specify particulars about how load generation should be conducted, you can do so.

The tests runs on bare-metal, with both load-generation and the test application residing on separate bare-metal machines. 

success

At the conclusion of the test run, a file is produced which shows the performance of the test application as instrumented with [JFR](https://docs.oracle.com/javacomponents/jmc-5-4/jfr-runtime-guide/about.htm#JFRUH170). 

If you wish, you may also enable the `metrics collection` checkbox, which periodically collects system metrics directly from the operating system. (This is useful if you don't wish to rely on JFR or you simple want the perspective from the OS itself instead of from inside the Java process.)

## Deployment

This PR may be merged at any time, but it will not be ready for use until the Bandstand application is also deployed. (This is a separate application developed in conjunction with this one that eases the burden of orchestrating multiple bare-metal machines and handles service discovery, etc.) It is not being released as an OSS application and as such, is not linked from this PR.

## Future enhancements and caveats

Presently, this relies on a dedicated machine to receive requests from the agent. Nothing is done with this data, and we would like to find a way to run these tests without requiring a dedicated APM server or perhaps making a version of APM Server which does not require Elasticsearch to run. This is a topic for a follow-up discussion.

Additionally, we would like to make the orchestration application something fully managed in the lifetime of the pipeline instead of being a persistent service. This isn't urgent, however, and can be done later.

Finally, we are interested in potentially gathering metrics from a few other sources which can be displayed with the Elastic Stack. Specifically, it would be nice to include an option to monitor the host with Metricbeat and it would also be nice to find a way to collect and display the load generation metrics. Presently, this requires just looking at the logs and can clearly be improved.


.ci/load/README.md
```diff
@@ -0,0 +1,61 @@
...
+## Orchestration
+
+Because of the nature of these tests, they require the use of a dedicated orchestration layer. For this project, that orchestrator is called Bandstand. It is an internal project to Elastic.
+
+Bandstand performs two primary functions.
+
+1. *Service discovery* The pipeline utilizes a load-generation machine and an application server which run in parallel to each other. These machines need to know how to find each other and Bandstand gives them that information.
+2. *Orchestration* We need to have a system which tells various services when to start an stop based on the state of other services. For example, we can't start load-generation until we are assured that the application is up and in a coherent state. By using an independent orchestrator, we can allow each service to report on its own state. This avoid a number of otherwise difficult-to-maintain dependencies between service state through the lifetime of the load generation.
+
+While the orchestration layer is currently relatively lightweight and simple, it is built to be able to easily extended for more sophisticated needs, such as multiple test runs inside a single test exuection.
```
mdelapenya, line 20:
```suggestion
While the orchestration layer is currently relatively lightweight and simple, it is built to be able to be easily extended for more sophisticated needs, such as multiple test runs inside a single test execution.
```

.ci/load/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Because the [test application](https://projects.spring.io/spring-petclinic/) is instrumented with [JFR](https://docs.oracle.com/javacomponents/jmc-5-4/jfr-runtime-guide/about.htm#JFRUH170), it is necessary to have a copy of the [JDK Mission Control application](https://www.baeldung.com/java-flight-recorder-monitoring#3-visualize-data) which is distributed with the JDK.
+
+The results of the JFR instrumentation are available after a test run is complete by navigating to Jenkins and viewing the job results. A link to `Artifacts` will bring you to a page which shows the files generated by the test run. Download `src/spring-petclinic/flight.jfr` and then view its contents with JDK MIssion Control to examine the moment-by-moment profile of the application under test.
+
+To view the load generation output, see the console output in Jenkins for the `Load Generation` step.
+
+## Hacking
+
+*Caution*: The following requires access to libraries which are internal to Elastic. Do not attempt to follow these instructions of you are an open-source user
```
mdelapenya, line 50:
```suggestion
*Caution*: The following requires access to libraries which are internal to Elastic. Do not attempt to follow these instructions if you are an open-source user
```

.ci/load/README.md
```diff
@@ -0,0 +1,61 @@
...
+of the APM Java Agent. Instead, contact the core development team with questions.
+
+To set up a local development environment for executing load tests and for developing the test framework, perform the following steps:
+
+1. Checkout the Bandstand application and build a docker container from it tagged as `bandstand`.
+2. Checkout the [fork of the APM Pipeline library](https://github.com/cachedout/apm-pipeline-library-1/tree/perf) which contains the necessary modifications to provision a load-testing environment locally.
+3. In the `apm-pipeline-library` checkout from step 3, create a soft link from `local/jenkins_home` to the your checkout of the `apm-java-agent` codebase.
```
mdelapenya, line 57:
```suggestion
3. In the `apm-pipeline-library` checkout from step 2, create a soft link from `local/jenkins_home` to the your checkout of the `apm-java-agent` codebase.
```

.ci/load/scripts/param_gen/gen_params.py
```diff
@@ -0,0 +1,126 @@
...
+"""
+
+import requests
+import argparse
+import github
+from packaging.version import parse as parse_version
+
+CATALOG_URL = 'https://jvm-catalog.elastic.co'
+SUPPORTED_JDKS = ['oracle', 'openjdk', 'adoptopenjdk', 'amazon', 'jdk', 'zulu']
+MIN_JDK_VERSION = '11'
```
felixbarny, line 54: We'll also need to test against Java 7 and 8. Does Spring PetClinic event support such old versions?

mdelapenya (approved):
Just a few typos, but LGTM!

.ci/load/scripts/locustfile.py
```diff
@@ -0,0 +1,13 @@
+from locust import HttpUser, task, between
+
+
+class QuickstartUser(HttpUser):
+    wait_time = between(1, 2)
```
felixbarny, line 5: Can we make it executing requests at a fixed rate in order to avoid coordinated omission problems?

.ci/load/README.md
```diff
@@ -0,0 +1,61 @@
+# APM Java Agent Load Geneator
+
+This directory contains configuration files for load-testing the APM Java Agent.
+
+## Load Tests
+
+Load tests are run using two bare-metal workers. One worker is dedicated to the role of load generation, which other worker is dedicated to the role of hosting the application under test, which is instrumented by the APM Java Agent.
+
+Load generation is performed using [Locust](https://locust.io/) and the application which is instrumented by the APM Java Agent and placed under load is [Spring Petclinic](https://projects.spring.io/spring-petclinic/).
```
felixbarny, line 9: Is it possible to view the Locust results? If so, how?

felixbarny (commented):
Looking great! Let's try to merge this in soon so we can give it a try and iterate from there.

.ci/load/README.md
cachedout: Yes. It's not super-clean because you have to basically just navigate to the Blue Ocean UI in Jenkins and then scroll to the end of the console log for the load generation step, but the output is there. It will look like this:

loc

I would like to find a way to present this as an artifact of the test results so it can be easily downloaded but I just made this a future TODO. Let me know if you'd like me to prioritize improving this experience because it's definitely possible.

.ci/load/scripts/locustfile.py
cachedout: Most certainly! I will make a change to the PR to modify the defaults.

(Note that this pipeline also supports passing in your own Locust configuration file so you can design a test however you see fit. Obviously, I started with something super-basic but I'm thinking that we could also design "scenarios" which pre-package various load test configurations as well, but that's more of a long-term roadmap item.)

.ci/load/scripts/param_gen/gen_params.py
cachedout: Good question. I might need a little help from your team to make this determination. I see Java 8 referenced here, though: https://github.com/spring-projects/spring-petclinic#prerequisites

.ci/load/README.md
felixbarny: sounds like the right tradeoff for an MVP

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,179 @@
...
...acle-11+15-linux', 'oracle-11+16-linux', 'oracle-11+17-linux', 'oracle-11+18-linux', 'oracle-11+19-linux', 'oracle-11+20-linux', 'oracle-11+21-linux', 'oracle-11+22-linux', 'oracle-11+23-linux', 'oracle-11+24-linux', 'oracle-11+25-linux', 'oracle-11+26-linux', 'oracle-11+27-linux', 'oracle-11+28-linux', 'oracle-11+5-linux', 'oracle-11.0.2+7-linux', 'oracle-11.0.2+9-linux', 'oracle-11.0.3+12-linux', 'oracle-11.0.4+10-linux', 'oracle-11.0.5+10-linux', 'oracle-11.0.6+8-linux', 'oracle-12+33-linux', 'oracle-12.0.1+12-linux', 'oracle-12.0.2+10-linux', 'oracle-13+33-linux', 'oracle-13.0.1+9-linux', 'oracle-13.0.2+8-linux', 'zulu-11.0.1-linux', 'zulu-11.0.2-linux', 'zulu-11.0.3-linux', 'zulu-11.0.4-linux', 'zulu-11.0.5-linux', 'zulu-11.0.6-linux', 'zulu-11.0.7-linux', 'zulu-11.0.8-linux', 'zulu-11.0.9-linux', 'zulu-12-linux', 'zulu-12.0.0-linux', 'zulu-12.0.1-linux', 'zulu-12.0.2-linux', 'zulu-13-linux', 'zulu-13.0.0-linux', 'zulu-13.0.1-linux', 'zulu-13.0.2-linux', 'zulu-13.0.3-linux', 'zulu-13.0.4-linux', 'zulu-13.0.5-linux', 'zulu-14-linux', 'zulu-14.0.0-linux', 'zulu-14.0.1-linux', 'zulu-14.0.2-linux', 'zulu-15.0.0-linux', 'zulu-15.0.1-linux'], name: "jvm_version", description: "JVM")
```
v1v, line 29: OMG!

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,179 @@
...
+
+    stages {
+        stage('Pre-flight'){
+            steps {
+                echo 'Getting authentication information from Vault'
+                withSecretVault(secret: 'secret/apm-team/ci/bandstand', user_var_name: 'APP_TOKEN_TYPE', pass_var_name: 'APP_TOKEN'){
+                    script {
+                        sh(script: ".ci/load/scripts/start.sh", returnStdout: true)
+                        env.LOCUST_RUN_TIME = params.duration
+                        env.LOCUST_USERS = params.concurrent_requests
```
v1v, line 48: You can set them in the top-level environment closure

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,179 @@
...
+        // End script auto-generation
+    }
+
+    stages {
+        stage('Pre-flight'){
+            steps {
+                echo 'Getting authentication information from Vault'
+                withSecretVault(secret: 'secret/apm-team/ci/bandstand', user_var_name: 'APP_TOKEN_TYPE', pass_var_name: 'APP_TOKEN'){
+                    script {
+                        sh(script: ".ci/load/scripts/start.sh", returnStdout: true)
```
v1v, line 46: Is there any reason to use the returnStdout?

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,179 @@
...
+        APP_BASE_DIR = "src/${APP}"
+        METRICS_BASE_DIR="metrics/"
+        AGENT_BASE_DIR = "agent/"
+        ORCH_URL = 'obs-load-orch.app.elastic.co:8000'
+        // Set below for local development
+        // ORCH_URL='10.0.2.2:8000'
+        DEBUG_MODE = '0' // set to '0' for production
+        // This is a placeholder server. This will change to a dummy APM Server when it is available.
+        APM_SERVER_URL = 'https://2a2bd0e2806a47e5996eeeec6d22e6df.apm.eu-west-3.aws.elastic-cloud.com:443'
+
```
v1v, line 16: If this is a MBP, then I'd suggest to also move the params to env variables to bypass an existing issue with the very first builds.


```suggestion
        LOCUST_RUN_TIME = "$(params.duration}"
        LOCUST_USERS = "$(params.concurrent_requests}"
```

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,179 @@
...
+                    agent { label 'metal' }
+                    steps {
+                            withSecretVault(secret: 'secret/apm-team/ci/bandstand', user_var_name: 'APP_TOKEN_TYPE', pass_var_name: 'APP_TOKEN'){
+                                echo 'Preparing load generation..'
+                                script {
+                                    if (params.locustfile) {
+                                        echo 'Using user-supplied plan for load-generation with Locust'
+                                        sh script: "echo \"${params.locustfile}\">.ci/load/scripts/locustfile.py"
+                                    }
+                                }
```
v1v, line 66:
```suggestion
                                whenTrue(params.locustfile) {
                                    echo 'Using user-supplied plan for load-generation with Locust'
                                    sh script: "echo \"${params.locustfile}\">.ci/load/scripts/locustfile.py"
                                }
```

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,179 @@
...
+                                script {
+                                    if (params.agent_config) {
+                                        echo 'Writing user-supplied agent configuration'
+                                        dir("${AGENT_BASE_DIR}") {
+                                            sh script: "echo \"${params.agent_config}\">custom_config.cfg"
+                                        }
+                                    }
+                                }
```
v1v, line 110:
```suggestion
                                whenTrue(params.agent_config) {
                                    echo 'Writing user-supplied agent configuration'
                                    dir("${AGENT_BASE_DIR}") {
                                        sh script: "echo \"${params.agent_config}\">custom_config.cfg"
                                     }
                                }
```

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,179 @@
...
+                    stages{
+                        stage('Provision Java') {
+                            steps {
+                                echo "Provisioning Java version: ${params.jvm_version}"
+                                script {
+                                    def jdk_home = sh(script: ".ci/load/scripts/fetch_sdk.sh ${params.jvm_version}", returnStdout: true).trim()
+                                    env.JAVA_HOME = jdk_home
+                                    env.JAVACMD = "${env.JAVA_HOME}/bin/java"
+                                    env.PATH = "${env.JAVA_HOME}/bin:$PATH"
+                                }
```
v1v, line 82:
```suggestion
                                setEnvVar('JAVA_HOME', sh(script: ".ci/load/scripts/fetch_sdk.sh ${params.jvm_version}", returnStdout: true).trim())
                                setEnvVar('JAVACMD', "${env.JAVA_HOME}/bin/java")
                                setEnvVar('PATH', "${env.JAVA_HOME}/bin:$PATH")
```

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,179 @@
...
+        stage('Pre-flight'){
+            steps {
+                echo 'Getting authentication information from Vault'
+                withSecretVault(secret: 'secret/apm-team/ci/bandstand', user_var_name: 'APP_TOKEN_TYPE', pass_var_name: 'APP_TOKEN'){
+                    script {
+                        sh(script: ".ci/load/scripts/start.sh", returnStdout: true)
+                        env.LOCUST_RUN_TIME = params.duration
+                        env.LOCUST_USERS = params.concurrent_requests
+                        env.SESSION_TOKEN = sh(script: ".ci/load/scripts/start.sh", returnStdout: true).trim()
+                    }
```
v1v, line 50: The below will help to get rid of the required script closure 

```suggestion
                    sh(script: ".ci/load/scripts/start.sh", returnStdout: true)
                    setEnvVar('SESSION_TOKEN', sh(script: ".ci/load/scripts/start.sh", returnStdout: true).trim())
```

.ci/load/Jenkinsfile
cachedout: Hah! These are auto-generated via magic! See [here](https://github.com/elastic/apm-agent-java/pull/1467/files#diff-42da74e42e514299c22a473ebd48a9c7bfa9df1ccda27192752a4f723614c945).

.ci/load/Jenkinsfile
cachedout: Not really. I had it on for development but it should be safe to turn off for production. Will update. Good catch!

.ci/load/Jenkinsfile
cachedout: Ooh, I didn't know about `whenTrue`. That's handy! Thanks!

v1v:
Is JJBB required?

cachedout:
> Is JJBB required?

Yes, it is. I was going to do a follow-up PR with those changes.

.ci/load/scripts/param_gen/gen_params.py
cachedout: @felixbarny Also, with this, we are allowed only 250 entries in a drop-down list. If I set this down to 7 as the minimum, we're just over the limit. I could potentially refactor this to try and work around this but before I do, I should ask if there are any of these flavors you clearly won't be needing?

.ci/load/Jenkinsfile
cachedout: This won't be a MBP so I think we're OK here.

.ci/load/Jenkinsfile
cachedout: @v1v Setting this actually broke the pipeline. :(

```
Preparing load generation..— Print Message<1s

[2020-11-03T08:22:31.234Z] Preparing load generation..

Scripts not permitted to use method groovy.lang.GroovyObject invokeMethod java.lang.String java.lang.Object (org.jenkinsci.plugins.workflow.cps.CpsClosure2 whenTrue java.lang.String org.jenkinsci.plugins.workflow.cps.CpsClosure2)
```

cachedout: I went ahead and wrapped this in a `Boolean.valueOf()` and things are fine again. 👍

felixbarny:
To add to the wishlist for a future iteration, it would be great to be able to compare two runs with different agent configurations and otherwise exact same settings. The most common use case would be to determine the overhead of the agent or a particular agent feature.

.ci/load/scripts/param_gen/gen_params.py
felixbarny: Does removing `'amazon'` from `SUPPORTED_JDKS` bring us under the limit? If not, also exclude `'jdk'`.

cachedout:
> To add to the wishlist for a future iteration, it would be great to be able to compare two runs with different agent configurations and otherwise exact same settings.

@felixbarny Absolutely! I'd like do this along with the (already requested) ability to do multiple runs for each scenario and then have the ability to also have it output an average out each set of runs for each scenario. (Of course it would still output individual runs as well.)

.ci/load/scripts/locustfile.py
cachedout: Updated. Some additional reading on this for folks who might be interested:

https://github.com/locustio/locust/issues/646
https://github.com/locustio/locust/pull/1118

Down the road, there's no reason we couldn't explore supporting different types of load-generation tools if we really wanted to, but IMHO we should see how far this one gets us for the time being. I'm open to swapping it out if it doesn't meet our needs, though.

.ci/load/scripts/locustfile.py
```diff
@@ -0,0 +1,13 @@
+from locust import HttpUser, task, constant
+
+
+class QuickstartUser(HttpUser):
+    wait_time = constant(1)
```
felixbarny, line 5: IIUC, `constant` is still vulnerable to coordinated omission as it waits one second after the requests has been completed. By using `constant_pacing`, we can ensure a constant RPS, regardless of how long it takes to serve the request.

An example of when this can be problematic: Imagine there's a 1-minute stop-the-world JVM pause due to a major garbage collection. During that time, we'd effectively "look away" by waiting for the current requests to execute before executing others. Only by executing pre-defined request/sec rate, no matter how the system under test responds, we can fully include the effect of the long stop-the-world pause in the test results.

cachedout: Ah, OK. I'll change this to `constant_pacing`.

.ci/load/Jenkinsfile
v1v: Sorry about the error, glad you solved :)

.ci/load/scripts/param_gen/gen_params.py
cachedout: @felixbarny Hrm. Sadly, removing both `amazon` and `jdk` only gets us to 266. If we just took the most recent patch release of any minor would that be sufficient for you or do you need that level of granularity?

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,170 @@
...
+        // End script auto-generation
+    }
+
+    stages {
+        stage('Pre-flight'){
+            steps {
+                echo 'Getting authentication information from Vault'
+                withSecretVault(secret: 'secret/apm-team/ci/bandstand', user_var_name: 'APP_TOKEN_TYPE', pass_var_name: 'APP_TOKEN'){
+                    sh(script: ".ci/load/scripts/start.sh", returnStdout: true)
+                    setEnvVar('SESSION_TOKEN', sh(script: ".ci/load/scripts/start.sh", returnStdout: true).trim())
```
kuisathaverat, line 48: Should we mask this SESSION_TOKEN?

kuisathaverat, line 48: Is correct to call twice the same script?

.ci/load/scripts/param_gen/gen_params.py
felixbarny: Unfortunately, we do need that level of granularity so that we can try to reproduce specific issues such as JVM crashes. It also helps us to bisect the JVM update version where a particular JVM bug has been fixed that causes issues in combination with the APM agent. 
Another possibility would be to just list the Zulu builds, as they seem to offer the most exhaustive list of update versions, and to add a plain-text input for things other than Zulu.
Or we could add different dropdowns for each vendor. By default, no JDK would be selected. If more than one checkbox has a non-default value, or if all have a non-default value we can consider that an input error.

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,170 @@
...
+        APP = 'spring-petclinic'
+        APP_BASE_DIR = "src/${APP}"
+        METRICS_BASE_DIR="metrics/"
+        AGENT_BASE_DIR = "agent/"
+        ORCH_URL = 'obs-load-orch.app.elastic.co:8000'
+        // Set below for local development
+        // ORCH_URL='10.0.2.2:8000'
+        DEBUG_MODE = '0' // set to '0' for production
+        // This is a placeholder server. This will change to a dummy APM Server when it is available.
+        APM_SERVER_URL = 'https://2a2bd0e2806a47e5996eeeec6d22e6df.apm.eu-west-3.aws.elastic-cloud.com:443'
```
kuisathaverat, line 15: we usually put the APM URL and the APM token in a Vault secret

.ci/load/scripts/param_gen/gen_params.py
cachedout: Sounds good. Let me play with some options and see what I can come up with.

.ci/load/Jenkinsfile
cachedout: It's not a secure token. All security is handled via the APP_TOKEN which is a shared secret. The session token is really just an identifier for the session so it shouldn't need to be obscured.

cachedout: Gah. That snuck back in via a bad merge. Will remove. Thanks.

.ci/load/Jenkinsfile
cachedout: The token is in a secret but I left the URL public. I'll change it.

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,170 @@
...
...acle-11+15-linux', 'oracle-11+16-linux', 'oracle-11+17-linux', 'oracle-11+18-linux', 'oracle-11+19-linux', 'oracle-11+20-linux', 'oracle-11+21-linux', 'oracle-11+22-linux', 'oracle-11+23-linux', 'oracle-11+24-linux', 'oracle-11+25-linux', 'oracle-11+26-linux', 'oracle-11+27-linux', 'oracle-11+28-linux', 'oracle-11+5-linux', 'oracle-11.0.2+7-linux', 'oracle-11.0.2+9-linux', 'oracle-11.0.3+12-linux', 'oracle-11.0.4+10-linux', 'oracle-11.0.5+10-linux', 'oracle-11.0.6+8-linux', 'oracle-12+33-linux', 'oracle-12.0.1+12-linux', 'oracle-12.0.2+10-linux', 'oracle-13+33-linux', 'oracle-13.0.1+9-linux', 'oracle-13.0.2+8-linux', 'zulu-11.0.1-linux', 'zulu-11.0.2-linux', 'zulu-11.0.3-linux', 'zulu-11.0.4-linux', 'zulu-11.0.5-linux', 'zulu-11.0.6-linux', 'zulu-11.0.7-linux', 'zulu-11.0.8-linux', 'zulu-11.0.9-linux', 'zulu-12-linux', 'zulu-12.0.0-linux', 'zulu-12.0.1-linux', 'zulu-12.0.2-linux', 'zulu-13-linux', 'zulu-13.0.0-linux', 'zulu-13.0.1-linux', 'zulu-13.0.2-linux', 'zulu-13.0.3-linux', 'zulu-13.0.4-linux', 'zulu-13.0.5-linux', 'zulu-14-linux', 'zulu-14.0.0-linux', 'zulu-14.0.1-linux', 'zulu-14.0.2-linux', 'zulu-15.0.0-linux', 'zulu-15.0.1-linux'], name: "jvm_version", description: "JVM")
```
kuisathaverat, line 31: I think we can make a step in the library that collects all the version from https://jvm-catalog.elastic.co/jdks

cachedout: This PR includes a script which does that collection already, which is how this snippet generated, but it isn't dynamic.

.ci/load/Jenkinsfile
kuisathaverat: I do not think so, it is a TOKEN to coordinate the process, probably we can use BUILD_ID instead is a unique identified per build.

cachedout: Using something like the BUILD_ID would require some changes to the process as it is currently designed. At present, a unique SESSION_ID is generated by the orchestrator and passed back to Jenkins which then uses it for the lifetime of the pipeline. I agree that we could change this to let Jenkins tell the orchestrator which BUILD_ID it wants to use, but I'm not clear on what advantage that would have. Could you please clarify?

kuisathaverat: there is no much benefice to use BUILD_ID, only to not have to generate it, so I think is fine to generate the SESSION_ID

.ci/load/Jenkinsfile
kuisathaverat: I saw the script, I was thinking more about not to have to update that list ever, so a groovy step in the library can grab this list from the URL and short the names, or read that list from some file that is generated and stored daily in the library repo.

kuisathaverat (approved):

.ci/load/Jenkinsfile
cachedout: Gotcha. Cool. I agree that this would be a good follow-up. Having this list be dynamic would be much better in the long run.

.ci/load/scripts/param_gen/gen_params.py
cachedout: There were some duplicates in the list that I removed. We're good now. 👍

cachedout:
@felixbarny I think all the review feedback has been addressed at this point so if you'd like to give this a final review and 👍 we should be able to get this in and start testing.

.ci/load/Jenkinsfile
felixbarny: does not seem to be resolved yet? not sure if that's a nice-to-have or a requirement @kuisathaverat?

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,169 @@
...
+        // ORCH_URL='10.0.2.2:8000'
+        DEBUG_MODE = '0' // set to '0' for production
+        // This is a placeholder server. This will change to a dummy APM Server when it is available.
+        APM_SERVER_URL = 'https://2a2bd0e2806a47e5996eeeec6d22e6df.apm.eu-west-3.aws.elastic-cloud.com:443'
+        LOCUST_RUN_TIME = "${params.duration}"
+        LOCUST_USERS = "${params.concurrent_requests}"
+
+    }
+    options {
+        timeout(time: 3, unit: 'HOURS')
```
felixbarny, line 21: [minor] for long running stability tests, we might want to have longer or no timeouts

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,169 @@
...
+                                whenTrue(Boolean.valueOf(params.locustfile)) {
+                                    echo 'Using user-supplied plan for load-generation with Locust'
+                                    sh script: "echo \"${params.locustfile}\">.ci/load/scripts/locustfile.py"
+                                }
+                                sh(script: ".ci/load/scripts/load_agent.sh")
+                            }
+                    }
+                }
+                stage('Test application') {
+                    agent { label 'metal' }
```
felixbarny, line 67: are the workers guaranteed to be the same run-to-run? how many cpu cores do they have?

.ci/load/Jenkinsfile
```diff
@@ -0,0 +1,169 @@
...
+                                    gitCheckout(
+                                        basedir: "apm-agent-java",
+                                        branch: 'master',
+                                        repo: "https://github.com/elastic/${REPO}.git",
+                                        credentialsId: 'f6c7695a-671e-4f4f-a331-acdce44ff9ba',
+                                        shallow: false
+                                    )
+                                    dir("apm-agent-java"){
+                                        echo 'Switching to requested version'
+                                        sh(script: "git checkout v${params.apm_version}")
```
felixbarny, line 90: a likely feature request in the future will be to  checkout specific revisions or master

felixbarny (approved):

.ci/load/Jenkinsfile
cachedout: How about I set this to 72 hours for now which should be more than long enough and then we can re-visit it if we find a need for multi-day tests? Let me know if that sounds all right.

felixbarny: sgtm

.ci/load/Jenkinsfile
cachedout: This is an important discussion point. Thanks for bringing it up!

The tl;dr here is that they _should_ be the same but an absolute guarantee is challenging. For example, the ES Performance Testing team has a fleet of machines but have discovered over time that there are going to be variances even when they try hard to avoid them. SSDs in arrays fail, machines aren't tagged the same way by the provider but aren't physically identical, etc, etc.

So, where does that leave us? I think the best thing to do here is to be cautious about comparing results _between_ runs but that we continue to enhance this pipeline to support scenarios where we can run multiple invocations of a single test scenario multiple times on what we can guarantee to be the same machine(s) and maybe even some sort of comparative logic as well. (So, run scenario A and the scenario B on the same machine and output the results.)

I'm also going to file an issue in the infra repo to try to get an audit underway so we can know a bit more about what divergence we do have currently. As mentioned earlier, it should be a _lot_ but it may be some.

.ci/load/Jenkinsfile
cachedout: Aye. I bumped that (in my head) to an upcoming iteration but it should be do-able, no problem.

.ci/load/Jenkinsfile
cachedout: This is my mistake. I accidentally resolved this thread. I'll update this as well.

v1v (approved):

.ci/load/Jenkinsfile
cachedout: Some additional data here:

We have four workers right now, all of which vary in subtle ways. I propose that we make the following changes:

1. Pin the stage that runs the application to the same worker every time. We can do this by using the `benchmark` label, which currently is only assigned to one machine. That machine has the following specs:

*  Ubuntu 18.04
* 6CPUs 
* 64 GB 

2. We keep the load-generation stage marked as `metal` which will allow it to float between the other bare-metal machines which have slightly varying specs. However, there's not much reason to believe that they vary enough to modify the behavior of the load-generation script, which doesn't consume a great deal of resources at present.

3. We decide on a plan to order some additional machines which we can put into the `benchmark` pool which will ensure better consistency going forward.

(I will link backward from a ticket which provides more info, so we don't link from a public repo into a private one.)

Let me know how this sounds @felixbarny and @v1v and if you give a 👍 I will make the necessary change.

felixbarny: 👍

.ci/load/Jenkinsfile
cachedout: (Very delayed follow-up) It looks like this is [now handled and that hashes can be passed in](https://github.com/elastic/apm-agent-java/blob/master/.ci/load/Jenkinsfile#L31).

.ci/load/Jenkinsfile
cachedout: (Super-late follow-up) This ended up just getting turned into a free-form text field and seems to be just fine, so I don't see a need to build anything fancier for this.

PR #1465, opened 2020-11-01
Override IndyBootstrapDispatcher class module

eyalkoren:
## What does this PR do?
Fixes #1415 
Definitely hackish, but seems to solve the problem... 
The `java.lang.Class#getModule` method is the only non-native thing that seems to access the `java.lang.Class#module` field (both in OpenJDK and OpenJ9), so this may be good enough.

## Checklist

- [ ] This is an enhancement of existing features, or a new feature in existing plugins
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] I have made corresponding changes to the documentation
- [x] This is a bugfix
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [ ] I have added tests that would fail without this fix
- [ ] This is a new plugin
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [ ] My code follows the [style guidelines of this project](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#java-language-formatting-guidelines)
  - [ ] I have made corresponding changes to the documentation
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#testing) pass locally with my changes
  - [ ] I have updated [supported-technologies.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/docs/supported-technologies.asciidoc)
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] Added an instrumentation plugin? Describe how you made sure that old, non-supported versions are not instrumented by accident.
- [ ] This is something else
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)

PR #1468, opened 2020-11-02
Override IndyBootstrapDispatcher class module

### LINKED ISSUE #1415: Fix support for Java modules
Context - reported in [our forum](https://discuss.elastic.co/t/illegalaccesserror-module-access/249452):
```
java.lang.IllegalAccessError: Module 'java.base' no access to: package 'java.lang' because module 'java.base' can't read module 'unnamed module @61b4e073'
	at java.base/java.lang.invoke.MethodHandle.sendResolveMethodHandle(MethodHandle.java:1237)
	at java.base/java.lang.invoke.MethodHandle.getCPMethodHandleAt(Native Method)
	at java.base/java.lang.invoke.MethodHandle.resolveInvokeDynamic(MethodHandle.java:1070)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor.schedule(ScheduledThreadPoolExecutor.java:556)
	at co.elastic.apm.agent.profiler.SamplingProfiler.run(SamplingProfiler.java:342)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:836)
```
The `SamplingProfiler` is scheduled for execution by `java.util.concurrent.ScheduledThreadPoolExecutor`, which is loaded by the `java.base` module. For some reason, although it explicitly implements `Runnable`, an invokedynamic resolution is done, causing this failure.

[The unnamed module](http://openjdk.java.net/projects/jigsaw/spec/sotms/#the-unnamed-module) should have all it's packages exported automatically, however this doesn't mean they can be accessed by named modules.

eyalkoren:
## What does this PR do?

Closes #1415 
Supersedes #1465 

## Checklist

- [x] This is a bugfix
  - [ ] ~~I have added tests that would fail without this fix~~ - we currently only have integration tests for J9 with Java 8
  - [x] I added a unit test for the module override functionality 
  - [x] I tested manually that the fix works on the `adoptopenjdk/openjdk11-openj9:jdk-11.0.1.13-alpine-slim` docker image with Spring PetClinic
  - [x] I tested manually that the fix works on the `adoptopenjdk/openjdk9-openj9:jdk-9.0.4.12_openj9-0.9.0-alpine` docker image with Spring PetClinic


apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
```diff
@@ -196,14 +203,40 @@ public static Method getIndyBootstrapMethod() {
...
+            indyBootstrapDispatcherClass = Class.forName(INDY_BOOTSTRAP_CLASS_NAME, false, null);
         } catch (ClassNotFoundException e) {
             byte[] bootstrapClass = IOUtils.readToBytes(ClassLoader.getSystemClassLoader().getResourceAsStream(INDY_BOOTSTRAP_RESOURCE));
             ClassInjector.UsingUnsafe.ofBootLoader().injectRaw(Collections.singletonMap(INDY_BOOTSTRAP_CLASS_NAME, bootstrapClass));
+            indyBootstrapDispatcherClass = Class.forName(INDY_BOOTSTRAP_CLASS_NAME, false, null);
+        }
+
+        if (JvmRuntimeInfo.getMajorVersion() >= 9 && JvmRuntimeInfo.isJ9VM()) {
+            try {
+                logger.info("Overriding IndyBootstrapDispatcher class's module to java.base module");
```
SylvainJuge, line 218: [minor] make this log message specific to J9 JVM as it won't happen with other JVMs

apm-agent-core/src/test/java/co/elastic/apm/agent/bci/IndyBootstrapTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+import co.elastic.apm.agent.AbstractInstrumentationTest;
+import net.bytebuddy.dynamic.loading.ClassInjector;
+import org.junit.jupiter.api.Test;
+import org.stagemonitor.util.IOUtils;
+
+import java.io.InputStream;
+import java.util.Collections;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
SylvainJuge, line 35: [minor] use assertJ assertions instead of Junit ones

apm-agent-core/src/test/java/co/elastic/apm/agent/bci/IndyBootstrapTest.java
```diff
@@ -0,0 +1,53 @@
...
+    void testSetJavaBaseModule() throws Throwable {
+        Module javaBaseModule = Class.class.getModule();
+        assertNotEquals(IndyBootstrapTest.class.getModule(), javaBaseModule);
+
+        // In order to test this functionality, IndyBootstrapDispatcherModuleSetter needs to be loaded from the Boot CL
+        InputStream classFileAsStream = IndyBootstrapDispatcherModuleSetter.class.getResourceAsStream("IndyBootstrapDispatcherModuleSetter.class");
+        byte[] bootstrapClass = IOUtils.readToBytes(classFileAsStream);
+        ClassInjector.UsingUnsafe.ofBootLoader().injectRaw(Collections.singletonMap(IndyBootstrapDispatcherModuleSetter.class.getName(), bootstrapClass));
+
+        IndyBootstrap.setJavaBaseModule(IndyBootstrapTest.class);
```
SylvainJuge, line 50: [question] Would it be possible to have an exception thrown with an explicit message when this method is called and `IndyBootstrapDispatcherModuleSetter` hasn't been loaded from the right classloader ? If yes, that might help to diagnose any unexpected state in the field.

SylvainJuge (approved):
LGTM, only a few minor comments/questions.

apm-agent-core/src/test/java/co/elastic/apm/agent/bci/IndyBootstrapTest.java
eyalkoren: This can't happen - the `setJavaBaseModule` method tries to load it from the bootstrap CL explicitly, so failing to inject it there will result with `ClassNotFoundException`.
As you can see here, I _do_ load it with the test's class loader, but its single method is never invoked on any class that is not loaded correctly.

PR #1469, opened 2020-11-02
Fix attach external config

SylvainJuge:
## What does this PR do?

When using [remote attach](https://www.elastic.co/guide/en/apm/agent/java/master/setup-attach-cli.html), if there is any external configuration provided through `config_file` option, this external configuration file is ignored.

The expected behavior here should be to:
- load the external configuration first
- allow to override configuration items from the external configuration file with entries in the `--config` provided through CLI.

## Checklist

- [x] This is a bugfix

### DIFF
--- a/apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
+++ b/apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
@@ -121,21 +121,7 @@ public class ElasticApmAttacher {
             // when an external configuration file is used, we have to load it first
             String externalConfig = configuration.get("config_file");
             if (null != externalConfig) {
-                FileInputStream stream = null;
-                try {
-                    stream = new FileInputStream(externalConfig);
-                    properties.load(stream);
-                } catch (IOException e) {
-                    e.printStackTrace();
-                } finally {
-                    if (stream != null) {
-                        try {
-                            stream.close();
-                        } catch (IOException e) {
-                            // silenty ignored
-                        }
-                    }
-                }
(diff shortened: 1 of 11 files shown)


apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
```diff
@@ -130,6 +137,24 @@ static File createTempProperties(Map<String, String> configuration) {
         return tempFile;
     }
 
+    private static void loadExternalProperties(Properties properties, String externalProperties) {
+        FileInputStream stream = null;
+        try {
```
felixbarny, line 142: use try-with-resources

felixbarny (approved):
just a minor suggestion, otherwise LGTM

apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
SylvainJuge: Thanks, I always forget that one !

SylvainJuge:
I have updated documentation for the most common case (remote attach with CLI), but in practice it will also make it possible to use `elasticapm.properties` either in classpath or next to agent jar to include a `config_file` entry, in which case we might not properly define/enforce ordering and priority.

I think it's unlikely enough that we can wait until it becomes an issue before dealing with that.

SylvainJuge:
I've just realized that the initial implemented ordering wasn't consistent with what was documented, thus I've made `config_file` option have higher priority than configuration provided through runtime attach parameters.

PR #1477, opened 2020-11-06
implement fix for hp-ux version parsing

### LINKED ISSUE #1474: Support for Java 8 in HP-UX
I have a Hp-UX server withe Weblogic 12 that I am trying to instrument with the java agent elastic-apm-agent-1.18.1.jar. It seems that the numbering of the HP-UX releases of the java package is different from the official one so Java 1.8.0.17 corresponds to 1.8.0.202.
I receive the following errror when the instrumentation activates:
Failed to start agent - JVM version not supported: 1.8.0.17-hp-ux Java HotSpot(TM) 64-Bit Server VM 25.17-b1

Is it possible to disable this validation check? Is an HP-UV environment supported at all?

SylvainJuge:
## What does this PR do?

Fixes #1474 

Add an extra runtime property `elastic.apm.disable_bootstrap_checks` that allows to bypass JVM version check when it's value is set to `true`.

## Checklist

- [x] This is a bugfix
  - [x] I have updated [supported JVM version docs](https://www.elastic.co/guide/en/apm/agent/java/current/supported-technologies-details.html#supported-java-versions)
  - [x] I have documented the `elastic.apm.disable_bootstrap_checks` flag

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/AgentMain.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/AgentMain.java
@@ -92,7 +92,7 @@ public class AgentMain {
                 msgTemplate = "WARNING : JVM version unknown or not supported, safety check disabled - %s %s %s";
             } else {
                 doDisable = true;
-                msgTemplate = "Failed to start agent - JVM version not supported: %s %s %s";
+                msgTemplate = "Failed to start agent - JVM version not supported: %s %s %s. To override Java version verification, set the "elastic.apm.disable_bootstrap_checks" System property to "true".";
             }
 
             System.err.println(String.format(msgTemplate,
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/AgentMain.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/AgentMain.java
@@ -92,7 +92,7 @@ public class AgentMain {
                 msgTemplate = "WARNING : JVM version unknown or not supported, safety check disabled - %s %s %s";
             } else {
                 doDisable = true;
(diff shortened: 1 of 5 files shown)


felixbarny (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/AgentMain.java
```diff
@@ -82,9 +82,26 @@ public synchronized static void init(String agentArguments, Instrumentation inst
...
+            String msgTemplate;
+
+            boolean doDisable;
+            if (Boolean.parseBoolean(System.getProperty("elastic.apm.disable_bootstrap_checks"))) {
+                // safety check disabled, warn end user that it might
+                doDisable = false;
+                msgTemplate = "WARNING : JVM version unknown or not supported, safety check disabled - %s %s %s";
+            } else {
+                doDisable = true;
+                msgTemplate = "Failed to start agent - JVM version not supported: %s %s %s";
```
eyalkoren, line 95:
```suggestion
                msgTemplate = "Failed to start agent - JVM version not supported: %s %s %s. To override Java version verification, set the "elastic.apm.disable_bootstrap_checks" System property to "true".";
```

eyalkoren (commented):
Nice.
We need to document the `disable_bootstrap_checks` flag somehow.
See suggestion to do that when aborting at startup. 

In addition, I think we can at least add it to the `NOTE` message we have at the [supported Java version section](https://www.elastic.co/guide/en/apm/agent/java/current/supported-technologies-details.html#supported-java-versions).
This section needs to be updated to include these new JVM limitations as well.

eyalkoren (approved):

PR #1478, opened 2020-11-06
Test public api usage in integration tests

SylvainJuge:
## What does this PR do?

Adds a few API usages in Servlet test applicaiton, which could also be used example for API usages.

## Checklist

- [x] This is something else

### DIFF
--- a/integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/AbstractServletContainerIntegrationTest.java
+++ b/integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/AbstractServletContainerIntegrationTest.java
@@ -154,6 +154,7 @@ public abstract class AbstractServletContainerIntegrationTest {
             .withEnv("ELASTIC_APM_TRACE_METHODS", "public @@javax.enterprise.context.NormalScope co.elastic.*")
             .withEnv("ELASTIC_APM_DISABLED_INSTRUMENTATIONS", "") // enable all instrumentations for integration tests
             .withEnv("ELASTIC_APM_PROFILING_SPANS_ENABLED", "true")
+            .withEnv("ELASTIC_APM_APPLICATION_PACKAGES", "co.elastic.webapp") // allows to use API annotations
             .withLogConsumer(new StandardOutLogConsumer().withPrefix(containerName))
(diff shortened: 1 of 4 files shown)


integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/AbstractServletContainerIntegrationTest.java
```diff
@@ -154,6 +154,7 @@ protected AbstractServletContainerIntegrationTest(GenericContainer<?> servletCon
             .withEnv("ELASTIC_APM_TRACE_METHODS", "public @@javax.enterprise.context.NormalScope co.elastic.*")
             .withEnv("ELASTIC_APM_DISABLED_INSTRUMENTATIONS", "") // enable all instrumentations for integration tests
             .withEnv("ELASTIC_APM_PROFILING_SPANS_ENABLED", "true")
+            .withEnv("ELASTIC_APM_APPLICATION_PACKAGES", "co.elastic") // allows to use API annotations, we have to use a broad package due to multiple apps
```
felixbarny, line 157: just out of curiosity: why would `co.elastic.webapp` not work?

felixbarny (approved):

integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/AbstractServletContainerIntegrationTest.java
SylvainJuge: Other applications were in other sub-packages, for example `co.elastic.apm.spring.boot` or `co.elastic.apm.soap`, and setting it to the `webapp` sub-package breaks the `soap` test application as it makes soap transactions ignored (`co.elastic.apm.soap.HelloWorldServiceImpl` is annotated with `@javax.jws.WebService`).

There is a small difference here when the `application_packages` is not set, the `@javax.jws.WebService` is properly instrumented, whereas methods annotated with `@co.elastic.apm.api.CaptureSpan` are not.

I think it might be worth making it consistent and make `@CaptureSpan` and other annotations instrumentation work by default (even if it makes things slower), but that should be fixed in another PR.

PR #1479, opened 2020-11-06
Move main load test pipeline to linux worker

cachedout:
## What does this PR do?

Runs the orchestration layer, which manages the load-generation and the application, on a regular linux worker instead of the jenkins master itself. This should get us around needing to install things like `jq` on the Jenkins masters, as shown here: https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-load/2/console

### DIFF
--- a/.ci/load/Jenkinsfile
+++ b/.ci/load/Jenkinsfile
@@ -1,6 +1,6 @@
 // For documentation on this pipeline, please see the README.md in this directory
 pipeline {
-    agent any
+    agent { label 'linux && immutable' }
     environment {
         REPO = 'apm-agent-java'
         APP = 'spring-petclinic'


v1v (changes requested):
This is not a MBP therefore it requires to checkout the source code probably in this tage? https://github.com/elastic/apm-agent-java/pull/1479/files#diff-7b74176d56dcf2d85162c7e15a233c3115eca7999fc6744a9b3e59bda5a2861cR43

cachedout:
@v1v I don't understand what you mean by this. Why wouldn't merging this to master and then re-running the pipeline pick up the changes? They should be pulled directly from the master branch, so I'm confused about why this wouldn't work.

v1v (approved):
I actually said something not right 🙏

PR #1483, opened 2020-11-08
Avoid inter-thread context propagation for JMS polling transactions

eyalkoren:
## What does this PR do?
Fixes an issue reported in [the forum](https://discuss.elastic.co/t/invalid-transactions-grouping-with-activemq/252678).

## Checklist

- [ ] This is a bugfix

### DIFF
--- a/apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageConsumerInstrumentation.java
+++ b/apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageConsumerInstrumentation.java
@@ -228,6 +228,10 @@ public abstract class JmsMessageConsumerInstrumentation extends BaseJmsInstrumen
                     if (messageHandlingTransaction != null) {
                         messageHandlingTransaction.withType(MESSAGE_HANDLING)
                             .withName(RECEIVE_NAME_PREFIX)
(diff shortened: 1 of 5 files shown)


apm-agent-plugins/apm-jms-plugin/src/main/java/co/elastic/apm/agent/jms/JmsMessageConsumerInstrumentation.java
```diff
@@ -227,7 +227,8 @@ public static void afterReceive(@Advice.Origin Class<?> clazz,
                     Transaction messageHandlingTransaction = helper.startJmsTransaction(message, clazz);
                     if (messageHandlingTransaction != null) {
                         messageHandlingTransaction.withType(MESSAGE_HANDLING)
-                            .withName(RECEIVE_NAME_PREFIX);
+                            .withName(RECEIVE_NAME_PREFIX)
+                            .disableAsyncPropagation();
```
felixbarny, line 231: [minor] add a comment on why that's necessary there

felixbarny (approved):

eyalkoren:
Actually....
I rethought this and I want to try a different approach.
If we don't allow inter-thread context propagation when handling a message, we will lose all async executions on the message handling flow. 
Still, the only thread that should `end` a `message-handling` transaction should be the thread that created it.
So I will try this out.

PR #1484, opened 2020-11-09
Using type pool cache only when applying instrumentation for the entire JVM

eyalkoren:
Fixes `type-cache-pool-cleaner` thread leak when calling `ensureInstrumented`


felixbarny (approved):
LGTM

Please add a short description on which issue that fixes.

PR #1495, opened 2020-11-09
End message-handling transactions only on creating thread

eyalkoren:
## What does this PR do?
Fixes an issue reported in [the forum](https://discuss.elastic.co/t/invalid-transactions-grouping-with-activemq/252678).
Supersedes #1483 

## Checklist

- [ ] This is a bugfix


eyalkoren:
I just realized there is a potential bug in this approach with the following scenario:
1. Thread A starts a message-handling transaction called T1 when a message is received at the end of the `receive` invocation 
2. Thread A submits a task for another poll and go on to handle the message
3. Thread B picks this task and due to the concurrency plugin, it activates transaction T1
4. Thread B invokes the `receive` method, seeing there is an active transaction it did not activate, so it deactivates it
5. Thread B exits the `receive` invocation, starting a new transaction - T2 based on the received message
6. The task execution ends, so the concurrency plugin deactivates T2 on Thread B, which already issues a warning in the log for deactivating a span that is not the one it activated
7. There is no thread that can `end` T2

There are multiple ways to fix that, but all of which introduce some new risks and complications that stem from the fact that something deactivates something it did not activate.

So, we will find a Spring-JMS-specific solution

PR #1496, opened 2020-11-10
Excluding Spring-JMS specific Runnable from the concurrent plugin

eyalkoren:
## What does this PR do?
Yet another stab on fixing the issue reported in [the forum](https://discuss.elastic.co/t/invalid-transactions-grouping-with-activemq/252678), without losing context propagation for JMS-poll-transactions altogether and with minimal effect on anything else.
Supersedes #1483.

## Checklist

- [x] This is a bugfix
  - [x] I have adjusted the current Spring JMS test so that it would fail without this fix


felixbarny (approved):
The overall approach LGTM, but tests are failing. Is it related to this PR?

eyalkoren:
> tests are failing. Is it related to this PR?

It must be, but probably related to the concurrency plugin dependency. Checking...

eyalkoren:
Apparently what happened is that since a transaction is created and activated before each test starts, then when registering a `MessageListener`, this is done on a task that runs on a different thread and due to the concurrency plugin, the transaction is activated on this thread. Then, when `onMessge` is invoked, there is an active transaction on that thread, so the receive transaction is not created.

I split the Spring test into a separate module that has the dependencies on Spring JMS and on the concurrency plugin.

eyalkoren:
And now the `pom` thing 🤦 
Fixing...

felixbarny (approved):

PR #1508, opened 2020-11-12
remove useless allocation in TraceState.copyOf()

### LINKED ISSUE #1503: Allocation rate increase after PR 1384
The https://github.com/elastic/apm-agent-java/pull/1384 PR introduced a performance regression for allocation rate, about extra 50 bytes per transactions are now allocated.

image

SylvainJuge:
## What does this PR do?

Fixes #1503 

## Checklist

- [x] This is a bugfix
  - [x] ~I have added tests that would fail without this fix~ tested locally with async-profiler


eyalkoren (changes requested):
Leaking a member of a pooled object to another pooled object is very problematic.

I think we should avoid adding the `header` field to the `tracestate` list (in `TraceState#set()`) and adjust the `TraceState#toTextHeader()` method (as well as the `TraceState.TextTracestateAppender#join()` method) to append it only when required.
`TraceState#copyFrom()` already takes care of copying the `header` value.

One additional thing to fix is resetting `header` in `TraceState#resetState`.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/sampling/ConstantSampler.java
```diff
@@ -36,12 +37,15 @@
 
     private final boolean decision;
     private final double rate;
+
     private final String rateString;
```
eyalkoren, line 41: We can remove this one I guess
```suggestion
```

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/sampling/ConstantSampler.java
```diff
@@ -36,12 +37,15 @@
...
+    private final String traceStateHeader;
 
     private ConstantSampler(boolean decision) {
         this.decision = decision;
-        rate = decision ? 1.0d : 0.0d;
-        rateString = Double.toString(rate);
+        this.rate = decision ? 1.0d : 0.0d;
+        this.rateString = Double.toString(rate);
```
eyalkoren, line 47:
```suggestion
```

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/sampling/ProbabilitySampler.java
```diff
@@ -56,16 +56,16 @@
...
-        lowerBound = -higherBound;
+        this.higherBound = (long) (Long.MAX_VALUE * samplingRate);
+        this.lowerBound = -higherBound;
         this.sampleRate = samplingRate;
-        rateString = Double.toString(samplingRate);
+        this.traceStateHeader = TraceState.getHeaderValue(samplingRate);
     }
 
-    public static Sampler of(double samplingRate) {
+    public static Sampler of(double samplingRate) { // use 'Double' here
```
eyalkoren, line 68: I didn't get this comment

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/sampling/Sampler.java
```diff
@@ -57,8 +57,10 @@
      */
     double getSampleRate();
 
+    // While header is not related to sampler itself, putting this here allows to reuse the same
+    // String instance as long as the sample rate does not change to minimize allocation
```
eyalkoren, line 61: 👍  Can be added to the javadoc

eyalkoren (approved):
LGTM.
Few minor comment.

PR #1518, opened 2020-11-19
Add JVM params to load testing

SylvainJuge:
- add the ability to add some extra JVM parameters for load testing
- add missing requirements for param generation
- make `--gh-token` parameter mandatory otherwise list of agent releases is empty


.ci/load/Jenkinsfile
```diff
@@ -25,10 +25,11 @@ pipeline {
...
+        string(name: "jvm_options", defaultValue: "", description: "Extra JVM options")
         string(name: "concurrent_requests", defaultValue: "100", description: "The number of concurrent requests to test with")
-        string(name: "duration", defaultValue: "10m", description: "Test duration. Specify with unit. Ex: 30m, 1h")
+        string(name: "duration", defaultValue: "10", description: "Test duration in minutes. Max: 280")
```
cachedout, line 32: Ah. We need to change this in `gen_params.py` to be `string(name: "duration", defaultValue: "10m", description: "Test duration. Specify with unit. Ex: 30m, 1h")` and then regenerate this snippet. Otherwise, we will break the option that's passed to Locust, which requires the unit to be passed in.

Let me know if you'd like to me to just make a PR against yours for this change for this. I'm happy to. :)

cachedout (changes requested):

.ci/load/Jenkinsfile
SylvainJuge: fixed in the last commit.

cachedout (approved):

PR #1520, opened 2020-11-20
[load-testing] Always build agent with Java 11

cachedout:
## What does this PR do?
When running the load tests, we always want to build the agent with Java 11.

Here is some sample output showing the variables being set:

pre-post

### DIFF
--- a/.ci/load/Jenkinsfile
+++ b/.ci/load/Jenkinsfile
@@ -71,6 +71,7 @@ pipeline {
                         stage('Provision Java') {
                             steps {
                                 echo "Provisioning Java version: ${params.jvm_version}"
+                                setEnvVar('ORIG_PATH', "$PATH")
                                 setEnvVar('JAVA_HOME', sh(script: ".ci/load/scripts/fetch_sdk.sh ${params.jvm_version}", returnStdout: true).trim())
                                 setEnvVar('JAVACMD', "${env.JAVA_HOME}/bin/java")
                                 setEnvVar('PATH', "${env.JAVA_HOME}/bin:$PATH")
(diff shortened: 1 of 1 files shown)


.ci/load/Jenkinsfile
```diff
@@ -91,7 +92,14 @@ pipeline {
                                         echo 'Switching to requested version'
                                         sh(script: "git checkout v${params.apm_version}")
                                         echo 'Building agent'
-                                        sh(script: './mvnw  clean install -DskipTests=true -Dmaven.javadoc.skip=true')
+                                        withEnv(["JAVA_HOME=${env.HUDSON_HOME}/.java/java11"]){
+                                            withEnv(["JAVACMD=${env.JAVA_HOME}/bin/java", "PATH=${env.JAVA_HOME}/bin:$ORIG_PATH"]){
```
v1v, line 96: I didn't understand the trick of nester withEnv steps, but I just tested and it seems there is no other way.

Regarding the `PATH` variable, can we skip `ORIG_PATH` and use `"JAVA_HOME+PATH=${env.JAVA_HOME}/bin"`? See [with-env](https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#withenv-set-environment-variables) for a better explanation of what I mean :)

v1v (approved):
With a nit comment

SylvainJuge (approved):

.ci/load/Jenkinsfile
cachedout: Here I'd still prefer to use `ORIG_PATH` because I want to ensure that we have a clean path without Java explicitly defined anywhere on it. If we prepend to the current path as you suggest, we'll be prepending to what we previously set in L77.

PR #1526, opened 2020-11-23
Fix JSF error in Jetty 9.4 and upgrading Byte Buddy to 1.10.18

eyalkoren:
Fixing recent failures in Jetty 9.4 tests ([example](https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/job/master/653/)).

It seems that with the [latest Jetty 9.4 release](https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-server/9.4.34.v20201102) there is a bytecode transformation error when instrumenting `javax.faces.webapp.FacesServlet`, which fails instrumentation for an entire batch of classes, causing a seemingly non-related symptom.

Upgrading the `myfaces` dependency of the JSF app to `2.3.2` works around the problem.
A separate issue related to the BCI was opened in BB repo - https://github.com/raphw/byte-buddy/issues/979.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,7 +106,9 @@
         <version.spring>5.0.15.RELEASE</version.spring>
         <version.jetty-server>9.4.11.v20180605</version.jetty-server>
         <version.json-schema-validator>0.1.19</version.json-schema-validator>
+        <!-- Byte Buddy and ASM must be kept in sync -->
         <version.byte-buddy>1.10.18</version.byte-buddy>
+        <version.asm>9.0</version.asm>
 
         <!-- used both for plugin & annotations dependency -->
         <version.animal-sniffer>1.17</version.animal-sniffer>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -106,7 +106,7 @@
         <version.spring>5.0.15.RELEASE</version.spring>
         <version.jetty-server>9.4.11.v20180605</version.jetty-server>
         <version.json-schema-validator>0.1.19</version.json-schema-validator>
-        <version.byte-buddy>1.10.12</version.byte-buddy>
+        <version.byte-buddy>1.10.18</version.byte-buddy>
```
SylvainJuge, line 109: [minor] maybe adding another property for the asm-tree dependency just next to this one + with a small comment so we don't forget to keep those two in sync.

SylvainJuge (approved):

PR #1530, opened 2020-11-24
[cache][ci] use cached m2 repository

v1v:
## What does this PR do?

The build stage is the one preparing the build artifacts that are later on used in the following stages, so, in order to honour the .m2 cache, the build stage does the following:

- If the `.m2` cached folder in the CI worker exists then copy the content to the `$WORKSPACE/.m2`.  <--- **This is the hack**
- maven build
- the whole workspace is archived to be consumed later on in the UTs, Javadoc, Smoke Tests and so on stages.

~~Speed up the stash/unstash with our v2 that uses Google Storage. Ideally from 2 minutes to just a ?? seconds:~~

## Tests
- [x] UTs
- [x] ITs
- [x] Smoke tests
- [x] Benchmarks
- [x] Javadoc

## Screenshots

- Prepare the .m2 during the build stage

image

## Iterations

- Use absolute path where the .m2 repository is hosted in the CI workers.

It did not work since the built cached artifacts were hosted in another CI workers.
 
- Stash/Unstash

Reverted 156a41efc04f0ad1636ceb9bbf28e58dbc54d3e3 since there was no much improvement, so let's keep it simple

|currently|proposal|
|---------|--------|
|image|image|
|image|image|

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,7 @@ pipeline {
     REPO = 'apm-agent-java'
     BASE_DIR = "src/github.com/elastic/${env.REPO}"
     NOTIFY_TO = credentials('notify-to')
-    JOB_GCS_BUCKET = credentials('gcs-bucket')
+    JOB_GCS_BUCKET = 'apm-ci-temp'
     DOCKERHUB_SECRET = 'secret/apm-team/ci/elastic-observability-dockerhub'
     ELASTIC_DOCKER_SECRET = 'secret/apm-team/ci/docker-registry/prod'
     CODECOV_SECRET = 'secret/apm-team/ci/apm-agent-java-codecov'
@@ -16,6 +16,7 @@ pipeline {
     ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
     MAVEN_CONFIG = '-Dmaven.repo.local=.m2'
     OPBEANS_REPO = 'opbeans-java'
+    JOB_GCS_CREDENTIALS = 'apm-ci-gcs-plugin'
   }
   options {
     timeout(time: 1, unit: 'HOURS')
@@ -87,7 +88,7 @@ pipeline {
                 sh label: 'mvn install', script: "./mvnw clean install -DskipTests=true -Dmaven.javadoc.skip=true"
                 sh label: 'mvn license', script: "./mvnw license:aggregate-third-party-report -Dlicense.excludedGroups=^co\\.elastic\\."
               }
-              stash allowEmpty: true, name: 'build', useDefaultExcludes: false
+              stashV2(name: 'build', bucket: "${JOB_GCS_BUCKET}", credentialsId: "${JOB_GCS_CREDENTIALS}")
               archiveArtifacts allowEmptyArchive: true,
                 artifacts: "${BASE_DIR}/elastic-apm-agent/target/elastic-apm-agent-*.jar,${BASE_DIR}/apm-agent-attach/target/apm-agent-attach-*.jar,\
                       ${BASE_DIR}/target/site/aggregate-third-party-report.html",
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -79,6 +79,10 @@ pipeline {
             withGithubNotify(context: 'Build', tab: 'artifacts') {
               deleteDir()
               unstash 'source'
+              // prepare m2 repository with the existing dependencies
+              whenTrue(fileExists('/var/lib/jenkins/.m2/repository')) {
```
felixbarny, line 83: where is that dir coming from? from the packer cache?

v1v: that's the default home directory for the linux CI Workers, in fact, they don't have a HOME env variable but that's where all the cached artifacts are stored.

kuisathaverat (approved):

cachedout (approved):

eyalkoren (approved):
Together with your description, looks awesome.
A thought: if we can rely on this as a constant location, can we set this as the local maven repository instead of copying?
Also, were you able to measure the difference in timing with and without this caching?

v1v:
> A thought: if we can rely on this as a constant location, can we set this as the local maven repository instead of copying?

I gave a try, but unfortunately the smoke-tests requires the cached artifacts for the given build, since it consumes them. For instance -> [log](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp%2FPR-1530/detail/PR-1530/2/pipeline#step-170-log-479) when using the local maven repository.

> Also, were you able to measure the difference in timing with and without this caching?

I ran a build similar to the master branch and it took 44 minutes (master took 45 minutes)

image

image

I don't see major improvements regarding the overall build time, though I didn't take in consideration the allocation time for the workers and whether some of the static workers were used at that time.

eyalkoren:
I see, it makes sense to separate external maven artifacts and build-process artifacts and if IIUC this is what you are doing.

> I ran a build similar to the master branch and it took 44 minutes (master took 45 minutes)

Which indication we have that caching worked? It's better to compare only the build stage between two runs, which is the stage that is primarily affected by this.

v1v:
> I see, it makes sense to separate external maven artifacts and build-process artifacts and if IIUC this is what you are doing.

Sort of, I could not find a way in maven to configure the folder with the external maven artifacts and the folder with the generated build process artifacts.

So I ended up to copy the cached .m2 folder to the workspace and set the maven m2 flag to point out there.

The issue was with the stages that consume those artifacts that they don't use the `target` folder but the `m2` folder.

> Which indication we have that caching worked? 

No reset connections as we got in the master branch. Neither any misconfiguration failures

Jenkinsfile
felixbarny: This copies a cached folder to this build. But how does the `/var/lib/jenkins/.m2/repository` folder get populated? Wouldn't it always be empty?

v1v: `/var/lib/jenkins/.m2/repository` gets populated every night when the daily images run the `packer_cache.sh`. It should not be empty, but it it was then it will not cause any issues at all

felixbarny: got it, thanks. A minor suggestion: we could add a `du -hs .m2` so that we can verify whether this directory is non-empty.

v1v: Do you mean temporarily? 

it was not empty when I tested it out:

see [copy step](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp%2FPR-1530/detail/PR-1530/8/pipeline/98#step-104-log-1) that took 15 seconds.

v1v: https://github.com/elastic/apm-agent-java/pull/1530/commits/21c44bac8ebe3365b5b07be4cc3382c4be3416b1 is there now

felixbarny: I guess it doesn't hurt to just stay in there

felixbarny (approved):

felixbarny:
Is there a way to do something similar to the internal deploy snapshot and release jobs?

PR #1531, opened 2020-11-24
Fix NPE when async ES client is nested within Hibernate search

### LINKED ISSUE #1528: Null Pointer Exception shaded.weaklockfree.weakconcurrentmap.put
hi guys,
sorry if will not be perfct its my first post here..... Today i started the last snapshot of the java agent: 1.19.0 because i run more than 1 webapp in my tomcat and was the only way to get it run.
i receive several WARN in my log about a NullPoint Exception:
java.lang.nullpointerexception at co.elastic.apm.agent.shaded.weaklockfree.weakconcurrentmap.put(weakconcurrentmap.java:161)

Looking around i saw it was supposed to be fixed here: https://github.com/elastic/apm-agent-java/issues/1309
my enviroment is:
JAVA 1.8.0_161-b12
TOMCAT 8.0.52
java agent: 1.19.0

best regards
Luca

eyalkoren:
Closes #1528 
When ES client is used within Hibernate search, the helper code produces a `null` span as we don't allow nested exit spans.

### DIFF
--- a/apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-5_6/src/main/java/co/elastic/apm/agent/es/restclient/v5_6/ElasticsearchClientAsyncInstrumentation.java
+++ b/apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-5_6/src/main/java/co/elastic/apm/agent/es/restclient/v5_6/ElasticsearchClientAsyncInstrumentation.java
@@ -79,6 +79,7 @@ public class ElasticsearchClientAsyncInstrumentation extends ElasticsearchRestCl
     public static class ElasticsearchRestClientAsyncAdvice {
         @VisibleForAdvice
         public static final GlobalThreadLocal<Span> spanTls = GlobalThreadLocal.get(ElasticsearchRestClientAsyncAdvice.class, "spanTls");
+
         @AssignTo.Argument(5)
         @Advice.OnMethodEnter(suppress = Throwable.class)
         public static ResponseListener onBeforeExecute(@Advice.Argument(0) String method,
@@ -89,8 +90,8 @@ public class ElasticsearchClientAsyncInstrumentation extends ElasticsearchRestCl
             ElasticsearchRestClientInstrumentationHelper<HttpEntity, Response, ResponseListener> helper = esClientInstrHelperManager.getForClassLoaderOfClass(Response.class);
             if (helper != null) {
                 Span span = helper.createClientSpan(method, endpoint, entity);
-                spanTls.set(span);
                 if (span != null) {
+                    spanTls.set(span);
                     return helper.<ResponseListener>wrapResponseListener(responseListener, span);
                 }
             }
--- a/apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-5_6/src/main/java/co/elastic/apm/agent/es/restclient/v5_6/ElasticsearchClientAsyncInstrumentation.java
+++ b/apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-5_6/src/main/java/co/elastic/apm/agent/es/restclient/v5_6/ElasticsearchClientAsyncInstrumentation.java
@@ -24,13 +24,11 @@
  */
 package co.elastic.apm.agent.es.restclient.v5_6;
 
-import co.elastic.apm.agent.bci.VisibleForAdvice;
 import co.elastic.apm.agent.es.restclient.ElasticsearchRestClientInstrumentation;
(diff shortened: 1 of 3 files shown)


apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-5_6/src/main/java/co/elastic/apm/agent/es/restclient/v5_6/ElasticsearchClientAsyncInstrumentation.java
```diff
@@ -79,6 +79,7 @@ public ElasticsearchClientAsyncInstrumentation(ElasticApmTracer tracer) {
     public static class ElasticsearchRestClientAsyncAdvice {
         @VisibleForAdvice
         public static final GlobalThreadLocal<Span> spanTls = GlobalThreadLocal.get(ElasticsearchRestClientAsyncAdvice.class, "spanTls");
+
         @AssignTo.Argument(5)
         @Advice.OnMethodEnter(suppress = Throwable.class)
         public static ResponseListener onBeforeExecute(@Advice.Argument(0) String method,
```
felixbarny, line 85: [minor] The best practice would be to return an `Object[]` instead of using `GlobalThreadLocal`

felixbarny (approved):

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-5_6/src/main/java/co/elastic/apm/agent/es/restclient/v5_6/ElasticsearchClientAsyncInstrumentation.java
eyalkoren: How would that look like? We would use `Object[]` as the `@Advice.Enter` type in the exit advice?
Let's not do that now, it means either adding allocation for the Object array or complication of reusing such (probably through another `GlobalThreadLocal` 🙂  )

felixbarny: > How would that look like? We would use Object[] as the @Advice.Enter type in the exit advice?

Exactly.

> adding allocation for the Object array

Adding something to a Global/DetachedThreadLocal also allocates a `WeakKey` when calling `WeakConcurrentMap#put`. The JIT should be able to easily replace the `Object[]` allocation with stack allocations once it inlines the indy-dispatched advices.

eyalkoren: > Adding something to a Global/DetachedThreadLocal also allocates a WeakKey when calling `WeakConcurrentMap#put`.

That's right, you only added key caching for lookups!

> The JIT should be able to easily replace the Object[] allocation with stack allocations once it inlines the indy-dispatched advices.

OK, as long as we don't measure through microbenchmarks without sufficient warmup 😜 

I will change that then

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-5_6/src/main/java/co/elastic/apm/agent/es/restclient/v5_6/ElasticsearchClientAsyncInstrumentation.java
```diff
@@ -77,32 +75,37 @@ public ElasticsearchClientAsyncInstrumentation(ElasticApmTracer tracer) {
...
             ElasticsearchRestClientInstrumentationHelper<HttpEntity, Response, ResponseListener> helper = esClientInstrHelperManager.getForClassLoaderOfClass(Response.class);
             if (helper != null) {
                 Span span = helper.createClientSpan(method, endpoint, entity);
-                spanTls.set(span);
                 if (span != null) {
-                    return helper.<ResponseListener>wrapResponseListener(responseListener, span);
+                    Object[] ret = new Object[2];
+                    ret[0] = span;
+                    ret[1] = helper.<ResponseListener>wrapResponseListener(responseListener, span);
+                    return ret;
```
felixbarny, line 94:
```suggestion
                    return new Object[]{span, helper.<ResponseListener>wrapResponseListener(responseListener, span)};
```

feel free to ignore if you like yours better

felixbarny (approved):

PR #1538, opened 2020-11-25
Fix formatting for trace_methods_duration_threshold

xco-sk:
## What does this PR do?
The list of conditions for `trace_methods_duration_threshold` is not formatted properly, resulting in an inline display of list items. This PR fixes the issue and shows the list of conditions in the correct way.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -16,8 +16,6 @@ endif::[]
 
 [float]
 ===== Bug fixes
-* Fix formatting for trace_methods_duration_threshold in documentation
-
 ////
 
 === Unreleased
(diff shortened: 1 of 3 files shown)


eyalkoren:
Thanks for reporting!
Note that the `configuration.asciidoc` is created automatically based on our configuration code.
So, the way to fix that would be to add a couple of newlines to [the config source code](https://github.com/elastic/apm-agent-java/blob/master/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java#L476-L493) and then run the `ConfigurationExporterTest`.
No need for a `CHANGELOG` entry in this case.
Let me know if you prefer that I do that.

eyalkoren:
You would need to run `ConfigurationExporterTest` as well and push the modified `configuration.asciidoc`

xco-sk:
@eyalkoren thanks for the info, fixed the description in ConfigurationExporter. I also executed the ConfigurationExporterTest, so the asciidoc file is now generated from source.

PR #1541, opened 2020-11-26
[load test] build test app with recent JDK

SylvainJuge:
We currently have 2 JDKs involved in load tests:
- one (Java11) that is used to build the agent
- one for the test application, which version is provided through a parameter that is used to build & run the test application.

When trying to run test on old Java7 versions, like 7u111, we have maven that is unable to download dependencies due to TLS/certificate issues.

Thus, in this case we need to:
- use Java 8 JDK to run maven, test application will still target and be compatible with Java 7.
- for more recent versions, we will build and run the test application with the selected JDK.

### DIFF
--- a/.ci/load/Jenkinsfile
+++ b/.ci/load/Jenkinsfile
@@ -14,6 +14,10 @@ pipeline {
         LOCUST_RUN_TIME = "${params.duration}"
         LOCUST_USERS = "${params.concurrent_requests}"
         LOCUST_IGNORE_ERRORS = "${params.ignore_application_errors}"
+        //
+        JAVA_HOME_BUILD_AGENT = "${env.HUDSON_HOME}/.java/java11" // agent should always be built with Java 11
+        JAVA_HOME_BUILD_APP = '' // set at a later state depending on JDK
+        JDK_MAJOR_VERSION = '' // set at a later state depending on JDK
     }
     options {
         timeout(time: 72, unit: 'HOURS')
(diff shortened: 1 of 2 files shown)


.ci/load/Jenkinsfile
```diff
@@ -75,6 +79,15 @@ pipeline {
                                 setEnvVar('JAVA_HOME', sh(script: ".ci/load/scripts/fetch_sdk.sh ${params.jvm_version}", returnStdout: true).trim())
                                 setEnvVar('JAVACMD', "${env.JAVA_HOME}/bin/java")
                                 setEnvVar('PATH', "${env.JAVA_HOME}/bin:$PATH")
+
+                                def jdk_major = sh(script: ".ci/load/scripts/java_major.sh ${params.jvm_version}", returnStdout: true).trim())
+                                def app_build_jdk = "${env.JAVA_HOME}"
+                                if(jdk_major == '7'){
+                                    // build application with a more recent JDK that don't have TLS issues
+                                    app_build_jdk = sh(script: ".ci/load/scripts/fetch_sdk.sh zulu-8.0.272-linux", returnStdout: true).trim()
```
cachedout, line 87: I am worried about this line because I think it has the potential to overwrite the JDK that we download here on L79.

SylvainJuge: do you think of any better way to handle this ? We could maybe have a `java7` folder alongside the `java11` that is used to build the agent.

cachedout: I wonder if we could just generate have the script generate a randomly generated directory name so we can be (fairly) sure we're avoiding collisions? So, we'd modify the shell script to return `java-7-mYRand0mStr`? Either way, so long as it's unique, I think we'll be fine.

PR #1550, opened 2020-12-02
Fix load testing build with multiple JDKs

SylvainJuge:
Fixes the current issues when trying to build with multiple JDK versions:

As a bonus, we now only need to set the `JAVA_HOME` environment variable when invoking maven directly in the `mvnw` invocation command.

### DIFF
--- a/.ci/load/Jenkinsfile
+++ b/.ci/load/Jenkinsfile
@@ -15,6 +15,7 @@ pipeline {
         LOCUST_USERS = "${params.concurrent_requests}"
         LOCUST_IGNORE_ERRORS = "${params.ignore_application_errors}"
         //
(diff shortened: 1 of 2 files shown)


.ci/load/Jenkinsfile
```diff
@@ -103,17 +90,11 @@ pipeline {
                                         shallow: false
                                     )
                                     dir("apm-agent-java"){
-                                        echo 'Switching to requested version'
+                                        echo 'Switch to requested agent version = {params.apm_version}'
```
cachedout, line 93:
```suggestion
                                        echo "Switch to requested agent version = ${params.apm_version}"
```

cachedout (approved):
This looks like it should work pretty well outside of some minor fix-ups. Nicely done!

PR #1551, opened 2020-12-03
added support for adding span destination service info to public api

### LINKED ISSUE #1521: Support for adding span destination via public API
Currently, there's no way to add the required context fields for the Service Map in the public API.
```
span.destination.service.resource
span.destination.service.name
span.destination.service.type
```

kananindzya:
closes #1521


apm-agent-api/src/main/java/co/elastic/apm/api/AbstractSpanImpl.java
```diff
@@ -151,4 +151,40 @@ public void injectTraceHeaders(HeaderInjector headerInjector) {
     private void doInjectTraceHeaders(MethodHandle addHeader, HeaderInjector headerInjector) {
         // co.elastic.apm.agent.plugin.api.AbstractSpanInstrumentation.InjectTraceHeadersInstrumentation
     }
+
+    @Nonnull
+    @Override
+    public Span withDestinationServiceResource(String resource) {
```
felixbarny, line 157: Hm, I think we're hitting some limits with how we modelled spans and transactions here. These methods are only relevant for spans, not transactions. But in the public API, transaction extends span. We should think about introducing a `BaseSpan` that contains methods that are relevant for both spans and transactions. But that would likely break binary compatibility.

apm-agent-api/src/main/java/co/elastic/apm/api/AbstractSpanImpl.java
```diff
@@ -151,4 +151,40 @@ public void injectTraceHeaders(HeaderInjector headerInjector) {
...
+        Object span = doAppendDestinationServiceResource(resource);
+        return span != null ? new SpanImpl(span) : NoopSpan.INSTANCE;
```
felixbarny, line 159: why not just return `this`?

kananindzya: fixed

apm-agent-api/src/main/java/co/elastic/apm/api/AbstractSpanImpl.java
kananindzya: hi @felixbarny , please review changes

eyalkoren:
@kananindzya I am closing this as I believe this is superseded by #1788 .
If I am wrong, please feel free to reopen.

PR #1556, opened 2020-12-07
[CI] Download maven dependencies firstly

v1v:
## What does this PR do?

* Use [go-offline](https://maven.apache.org/plugins/maven-dependency-plugin/go-offline-mojo.html) prebuild to fetch all the dependencies.
* Set `maven.wagon.httpconnectionManager.ttlSeconds` to 25 seconds and `maven.wagon.http.retryHandler.count` to 3 retries.

## Why

I still see issues related to the connection reset :/ 

## Further details

The documentation for `maven.wagon.httpconnectionManager.ttlSeconds` is available in the source code:
* https://github.com/apache/maven-wagon/blob/wagon-3.4.1/wagon-providers/wagon-http-shared/src/main/java/org/apache/maven/wagon/shared/http/AbstractHttpClientWagon.java#L297-L305

Documentation for `maven.wagon.http.retryHandler.count` is in https://maven.apache.org/wagon/wagon-providers/wagon-http/

This is similar to the change made in Pulsar: apache/pulsar#8386

### DIFF
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 source /usr/local/bin/bash_standard_lib.sh
 
+JAVA_HOME=$HOME/.java/java11 \
+  ./mvnw dependency:go-offline --fail-never -q -B
+
 JAVA_HOME=$HOME/.java/java11 ./mvnw clean verify \
   --fail-never -q -B \
   -Dmaven.javadoc.skip=true \
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -9,7 +9,7 @@ JAVA_HOME=$HOME/.java/java11 ./mvnw clean verify \
   -Dmaven.javadoc.skip=true \
   -Dhttps.protocols=TLSv1.2 \
   -Dmaven.wagon.http.retryHandler.count=3 \
-  -Dhttp.keepAlive=false \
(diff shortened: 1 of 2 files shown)


cachedout (approved):

eyalkoren (approved):

eyalkoren:
Ahh, it fails at self artifacts that are only created at `mvn install` - https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/PR-1556/1/pipeline

v1v:
Good catch, I enabled the `--fail-never` in the packer-cache.sh script, but I did not enable this in the script. I'll do it now

eyalkoren:
Maybe you can use [`excludeGroupIds`](https://maven.apache.org/plugins/maven-dependency-plugin/go-offline-mojo.html#excludeGroupIds) to exclude resolving `co.elastic.apm` dependencies?

v1v:
Unfortunately `excludeGroupIds` does not get honoured for a multimodule maven project. Might be related to [MDEP-516](https://issues.apache.org/jira/browse/MDEP-516)

I've tried locally with:
- `./mvnw -U org.apache.maven.plugins:maven-dependency-plugin:3.1.2:go-offline -DexcludeGroupIds=co.elastic.apm`
- Even I added some changes in the pom.xml but same behaviour 

So, the alternative is to run `go-offline` with `never fail` once and might reduce the number of times we hit the issue :/

PR #1560, opened 2020-12-08
[load testing] make it work with all JDK catalog

SylvainJuge:
- avoid JDK re-download when possible + alternate to version parsing (wasn't compatible with amz corretto JVMs)
- capture application log
- avoid full checkout on every stage
- stash/unstash binaries + provide them as artifacts for easier download


.ci/load/Jenkinsfile
```diff
@@ -22,6 +22,7 @@ pipeline {
         timestamps()
         ansiColor('xterm')
         durabilityHint('PERFORMANCE_OPTIMIZED')
+        skipDefaultCheckout() // don't re-checkout at avery stage
```
cachedout, line 25: 👍 This is a really good idea!

cachedout (approved):

.ci/load/Jenkinsfile
SylvainJuge: I still need to test with production setup, but I think that not doing this makes the "clean and wipe workspace" also applied on every single stage and only once at the beginning of the pipeline execution.

PR #1586, opened 2020-12-22
remove non-Multi-Branch-Pipeline load testing job

SylvainJuge:
## What does this PR do?

This removes the non-Multi-Branch-Pipeline load testing job that isn't used anymore, the MBP variant allows better flexibility as it allows to use another branch than `master`.


kuisathaverat (approved):
I've open an issue to request infra to remove the Job from the instance also.

PR #1592, opened 2020-12-28
CallTree recycling

eyalkoren:
## What does this PR do?
Initial attempt to fix a sampling profiler's memory leak reported at [the forum](https://discuss.elastic.co/t/apm-consuming-lot-of-heap-memory/259571).


felixbarny:
Another idea that's probably worth implementing in addition to this: only hold soft (not weak) references to call tree **roots**. This doesn't fix the root cause but mitigates the severeness of the symptom. Basically another safety net.

eyalkoren:
> Another idea that's probably worth implementing in addition to this: only hold soft (not weak) references to call tree roots. This doesn't fix the root cause but mitigates the severeness of the symptom. Basically another safety net.

Good idea! I had a thought in the same direction with regard to the `profiledThreads` map - what happens if the application kills and replaces threads within a pool frequently and the configurations dictates continuous profiling (which is the default)? Then we will accumulate call trees in the map and `clearProfiledThreads` won't be called unless there is an error during profiling. 
I only focus in `profiledThreads` map, because it is the only thing I found that holds roots, besides the root pool, which I don't assume you referred to (if so, this should be done generically through the object pooling infrastructure).

One more thought I have is to connect the `SamplingProfiler` (actually the `ProfilingFactory`) to the circuit breaker mechanism - we only need to clear everything in `pause` because the profiler is noop when the tracer is `PAUSED`.

I will do some more additional cosmetic recycling changes to the recycling flows.

eyalkoren:
Now that I started implementing the cache, I saw that it we would need to allocate a `SoftReference` every time we start a new stack analysis. So it feels odd to pool root objects and still allocate every time... 
I will think about it some more. One option is to keep entries in `profiledThreads` when finishing stack analysis and keep track of their usage (number of iterations since last used or timestamp since last used etc.) and expunge them when we decide they are stale, but that may be an overkill.
For now I am adding the other stuff.

eyalkoren:
I added a deliberate memory leak through `profiledThreads` and verified that the circuit breaker is cleaning this leak as expected:
Screen Shot 2021-01-03 at 9 13 55

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/CallTree.java
```diff
@@ -436,7 +436,16 @@ private void fillStackTrace(List<StackFrame> stackTrace) {
...
+    /**
+     * Recycles this subtree to the provided pool recursively.
+     * Note that this method ends by recycling {@code this} node (i.e. - this subtree root), which means that
+     * <b>the caller of this method should make sure that no reference to this object is held anywhere</b>.
+     * <p>ALSO NOTE: MAKE SURE NOT TO CALL THIS METHOD FOR {@link CallTree.Root} INSTANCES.</p>
+     *
+     * @param pool the pool to which all subtree nodes are to be recycled
+     */
+    public final void recycle(ObjectPool<CallTree> pool) {
+        assert !(this instanceof Root);
```
SylvainJuge, line 448: why not throw directly an `IllegalArgumentException` when called on a root node instead of an assertion that might be disabled ?

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/CallTree.java
```diff
@@ -735,6 +754,8 @@ public void resetState() {
             super.resetState();
             activeSpan = null;
             activationTimestamp = -1;
+            // todo - should we reset activeSpanSerialized?
```
SylvainJuge, line 757: leftover TODO ?

SylvainJuge (commented):
[question] would it make sense to have all tree nodes keep a reference to their respective pools (root nodes and others) so that recycling would always be done to the right pool ?

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/CallTree.java
eyalkoren: I thought of avoiding the `instanceof` check overhead in production. Normally I think it is not something that would be specific for a unique corner case, our tests should find it if we try to recycle improperly

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/CallTree.java
eyalkoren: Well, half leftover...
Didn't want to look into it now, but wanted the reminder. But I will see if we need to something with it or leave it as is

SylvainJuge (approved):

PR #1593, opened 2020-12-28
Add bootstrap checks

felixbarny:
## What does this PR do?
Adds the following bootstrap checks:
- if `-Xverify:none` or `-noverify` flags are set, the agent will log: `WARNING: -Xverify:none and -noverify are not supported by Elastic APM Agent. In future versions, the Elastic APM Agent will not start when these flags are set, unless the system property ‘elastic.apm.disable_bootstrap_checks’ is set to ‘true’.` The intention is to disable the agent a couple of releases following the release that includes this PR.
- if G1GC is used with Java 8 lower than update 181, the agent will issue a warning message, saying that a related issue that is known to cause crashes was fixed in 181.

Just to emphasize that crashes caused by disabled verification is not a hypothetical problem:
People observe crashes in real applications and ask questions on Stack Overflow:
https://stackoverflow.com/questions/40321205/jvm-crash-problematic-frame-canonicalizerdo-if
https://stackoverflow.com/questions/58855063/jvm-exception-access-violation-crash-in-spring-boot-application

A good blog post on the topic:
https://blogs.oracle.com/buck/never-disable-bytecode-verification-in-a-production-system

And a thread by Alexey Shipilev:
https://twitter.com/shipilev/status/1134422174163505152

## Checklist

- [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc) - should be mentioned as a breaking change
- [ ] I have updated the supported Java versions section in supported technologies page to indicate those restrictions
- [ ] I have opened an issue with a proper future milestone to disable agent when `-Xverify:none` or `-noverify` flags are detected


eyalkoren (commented):
Excellent, thanks @felixbarny!

This means we prefer that the app fails to start when `-Xverify:none` or `-noverify` are set, to make sure the user is aware that there may be crashes we cannot foresee if they choose to run the agent with bootstrap checks disabled. 
It makes sense, though it may be problematic for some users. 
However, if we do think this is a good practice, should we do the same for Java 7 being started without the `CompileCommand` setting (assuming that we verify it is indeed required)?

felixbarny:
> This means we prefer that the app fails to start when `-Xverify:none` or `-noverify` are set

No, it just means we won't start the agent. The application itself will still run but without the agent.

> However, if we do think this is a good practice, should we do the same for Java 7 being started without the CompileCommand setting (assuming that we verify it is indeed required)?

I was thinking the same. If we don't find another workaround, such as delaying the instrumentation on Java 7, we could add a bootstrap check for that.

eyalkoren:
> No, it just means we won't start the agent. The application itself will still run but without the agent.

Of course, the agent 🤦 

> If we don't find another workaround, such as delaying the instrumentation on Java 7, we could add a bootstrap check for that.

👍  I will try this.

eyalkoren:
Let's add a check for `-XX:CompileCommand=exclude,java.lang.invoke.LambdaForm*::*`.
This PR should include additions to documentation and release notes. I will add checkboxes to the checklist.

apm-agent-premain/src/main/java/co/elastic/apm/agent/premain/AgentMain.java
```diff
@@ -130,12 +108,13 @@ public synchronized static void init(String agentArguments, Instrumentation inst
...
-        int majorVersion = JvmRuntimeInfo.getMajorVersion();
+        JvmRuntimeInfo runtimeInfo = JvmRuntimeInfo.ofCurrentVM();
+        int majorVersion = runtimeInfo.getMajorVersion();
         return
             (majorVersion == 7) ||
             // In case bootstrap checks were disabled
-            (majorVersion == 8 && JvmRuntimeInfo.isHpUx() && JvmRuntimeInfo.getUpdateVersion() < 2) ||
-            (majorVersion == 8 && JvmRuntimeInfo.isHotSpot() && JvmRuntimeInfo.getUpdateVersion() < 40);
+            (majorVersion == 8 && runtimeInfo.isHotSpot() && runtimeInfo.getUpdateVersion() < 2) ||
+            (majorVersion == 8 && runtimeInfo.isHotSpot() && runtimeInfo.getUpdateVersion() < 40);
```
SylvainJuge, line 117: [question] so the current state here is that those JVM versions are not supported out of the box in `JavaVersionBootstrapCheck`, but just in case the bootstraps checks are disabled we still apply the startup delay.
Would it make sense to have all those per-version checks all done in `JavaVersionBootstrapCheck` so that everything that is related to JVM version is within the same class ? (Might be enough as static method though).

apm-agent-premain/src/main/java/co/elastic/apm/agent/premain/JavaVersionBootstrapCheck.java
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Checks if a given version of the JVM is likely supported by this agent.
+     * <br>
+     * Supports values provided before and after https://openjdk.java.net/jeps/223, in case parsing fails due to an
+     * unknown version format, we assume it's supported, thus this method might return false positives, but never false
+     * negatives.
+     *
+     * @return true if the version is supported, false otherwise
+     */
+    public boolean isJavaVersionSupported() {
```
SylvainJuge, line 59: [minor] could probably be made package-private.

SylvainJuge (approved):

Issue #1583, opened 2020-12-21
JIT Compilation error with Oracle JDK 7u80


SylvainJuge:
## Describe the bug

When running with Oracle 7u80 JDK, after running for a while with agent enabled, the JVM process crash with the following message in the crash report:

```
# A fatal error has been detected by the Java Runtime Environment:
#
#  Internal Error (safepoint.cpp:318), pid=819438, tid=140189423261440
#  guarantee(PageArmed == 0) failed: invariant
#
# JRE version: Java(TM) SE Runtime Environment (7.0_80-b15) (build 1.7.0_80-b15)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (24.80-b11 mixed mode linux-amd64 compressed oops)
```

The stack trace is not directly related to the agent, and **async-profiler is not enabled**
```
Current thread (0x00007f806c1ba800):  VMThread [stack: 0x00007f8064b66000,0x00007f8064c66000] [id=819449]

Stack: [0x00007f8064b66000,0x00007f8064c66000],  sp=0x00007f8064c646d0,  free space=1017k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
V  [libjvm.so+0x9a320a]  VMError::report_and_die()+0x2ea
V  [libjvm.so+0x498da2]  report_vm_error(char const*, int, char const*, char const*)+0x62
V  [libjvm.so+0x8b03b4]  SafepointSynchronize::begin()+0x824
V  [libjvm.so+0x9aa96d]  VMThread::loop()+0x4bd
V  [libjvm.so+0x9aaaf0]  VMThread::run()+0x70
V  [libjvm.so+0x8238c8]  java_start(Thread*)+0x108
```

The last compilation events at the time of the crash indicate that `co.elastic.apm.agent.servlet.ServletApiAdvice::onEnterServletService` has just been compiled by the JIT compiler
```
Compilation events (10 events):
Event: 213.351 Thread 0x00007f806c85d000 3311   !         org.springframework.web.servlet.tags.UrlTag::replaceUriTemplateParams (215 bytes)
Event: 213.351 Thread 0x00007f806c85a800 3312   !         org.springframework.web.servlet.tags.UrlTag::createQueryString (181 bytes)
Event: 213.397 Thread 0x00007f806c85a800 nmethod 3312 0x00007f80699b2490 code [0x00007f80699b26a0, 0x00007f80699b3798]
[...]

PR #1594, opened 2020-12-28
Adding ability to delay agent initialization on startup

eyalkoren:
Adding the ability to delay instrumentation and do that on the background during or after app initialization.
The delay includes the loading of the `IndyBootstrapDispatcher` class, so to avoid loading `CallSite` on startup.

#### Checklist
- [x] Add ability to configure delay through System property
- [x] Auto-apply a fix delay on Java 7 and early Java 8 updates
- [x] Test fix on JVMs where it is easily reproducible, namely `zulu-7.0.285-linux` and `oracle-7u80-linux`
- [x] Add documentation
- [x] Add to CHANGELOG


eyalkoren:
This delay in instrumentation on startup did not prove to be a proper solution for the collection of Java 7 crashes reported through #1583.

I did multiple tests with zulu-7.0.285-linux that without any setting crashes within one minute.
Some attempts with the default 2 sec delay allowed the JVM to run without crashing, however, the agent seized to send any type of event within one minute. Since there is no crash log, it's tough to know exactly what happened. 
In other cases, the crash did occur even with the delayed instrumentation - [hs_err_pid2392037.log](https://github.com/elastic/apm-agent-java/files/5749807/hs_err_pid2392037.log).

Trying to increase the delay in instrumentation to 100 seconds (using the `delay_instrumentation` config) only postponed the crash - [hs_err_pid2393447.log](https://github.com/elastic/apm-agent-java/files/5749775/hs_err_pid2393447.log).

The only consistent solution so far is the compilation exclusion setting: `-XX:CompileCommand=exclude,java.lang.invoke.LambdaForm*::*`.

eyalkoren:
@elastic/apm-agent-java this ability to delay the initiation of instrumentation did not prove to solve the `CallTree` compilation crash. Would we like to have this capability still - configurable, hidden, disabled (noop) by default?

SylvainJuge:
There isn't much extra complexity added here, and I can see at least two relevant use-cases:
- future-proofing: work-around for not-yet-known similar issues that we had with Tomcat (we really don't know if and when it might happen)
- support: having an extra option to delay instrumentation in order to help debugging a support case

eyalkoren:
Closes #1583

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -129,12 +129,13 @@
         .tags("added[1.11.0]")
         .build();
 
-    private final ConfigurationOption<TimeDuration> delayInit = TimeDurationValueConverter.durationOption("ms")
```
SylvainJuge, line 132: Is that on purpose that we have two distinct properties ? I'm not sure to get why we would need to split delay for tracer init and the premain execution.
- `delay_tracer_start` with alias `delay_tracer_start_ms` which is internal but documented in `CoreConfiguration`
- `delay_agent_premain_ms` property which does not rely on configuration to avoid any dependency.

apm-agent-premain/src/main/java/co/elastic/apm/agent/premain/AgentMain.java
```diff
@@ -101,13 +99,72 @@ public synchronized static void init(String agentArguments, Instrumentation inst
...
+    }
 
-        try {
-            // workaround for classloader deadlock https://bugs.openjdk.java.net/browse/JDK-8194653
-            FileSystems.getDefault();
+    /**
+     * Returns whether agent initialization should be delayed when occurring through the {@code premain} route.
+     * This works around a JVM bug (https://bugs.openjdk.java.net/browse/JDK-8041920) causing JIT fatal error if
+     * agent code causes the loading of MethodHandles prior to JIT compiler initialization.
+     * @return {@code true} for any Java 7 and early Java 8 HotSpot JVMs, {@code false} for all others
```
SylvainJuge, line 130: Removing that constraint on early Java 8 early versions would be really nice !! Do we need to make load-testing able to run tests with forks to validate this ? We could use a temporary branch on the project itself as a work-around.

docs/troubleshooting.asciidoc
```diff
@@ -18,6 +18,17 @@ IMPORTANT: If you do so, *please attach your debug logs* so that we can analyze
...
+[float]
+[[trouble-shooting-additional-agent]]
+=== Running alongside other agents
+Like many other Java agents, our agent instruments classes by injecting bytecode into them on runtime. Our bytecode
+instrumentation guarantees to produce only valid bytecode and to never change the class's "schema". If you are using
+other Java agents in addition to ours that have the same guarantees, they should not interfere with each other.
+However, some Java agents do not conform to these guarantees, in which case there may be issues with using them in
+parallel to ours. If you encounter errors, one of the first things to do is to remove any additional agent in order to
+check whether this is indeed a case of agent mismatch. We may still find a way to make them work together, but this
+information would be cruicial for our ability to assist, so make sure to include it when reporting such issues.
```
SylvainJuge, line 30: :heart_eyes: I love being able to link to this paragraph whenever there are multiple agents. Should we do the same for `-Xverify:none` too ?

SylvainJuge (approved):
LGTM, just a minor question on why we have two properties.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
eyalkoren: Yes, it is intentional, not saying it is the right thing 🙂 
Ideally, we wouldn't recommend anyone using the premain delay, unless we really have to. This is the absolute delay in all agent stuff, including all class loading and linkage. The disadvantages of using it are the need to reinstrument loaded classes (instead of applying instrumentation during class loading), which may have more side effects, as well as the more restricted configuration option (only as System property). We just had a customer lately getting strange JMX behaviour with specific WebLogic MBeans that was overcome through the delay in tracer start. For a large deployment, it might be less user-friendly to enforce changes in the startup scripts all across instead of applying the change through the agent config file.

I see your concern with having multiple properties dealing with delays, but since they are both hidden, it will mainly be our burden and not our users', so not as bad.

However, it got me thinking I will now break it for this user. Do you think we should:
-  just keep the old name?
-  use the new name and aliases for the old ones?
-  keep this change and notify this user, as it is internal and not likely widely used?

apm-agent-premain/src/main/java/co/elastic/apm/agent/premain/AgentMain.java
eyalkoren: Indeed, I thought saw to, but @felixbarny pointed out to me that we already had this restriction before moving to `invokedynamic` (although we used MethodHandles before). I think that using so old Java 8 updates is less common than using Java 7, so let's keep this restriction and this handles cases where some IS trying to use our agent on these old JVMs and disables the bootstrap checks

docs/troubleshooting.asciidoc
eyalkoren: We decided to be more blunt and restrictive with disabled BC verification - it will be handled through https://github.com/elastic/apm-agent-java/pull/1593. The current plan is to issue a warning when we encounter that in a few versions later disable ourselves when it is used, enforcing the user to disable the bootstrap checks.

PR #1597, opened 2021-01-05
null-safe weak set/map + detached thread local

SylvainJuge:
## What does this PR do?

The `WeakConcurrentMap` implementation throws `NullPointerException` in the following cases:
- trying to retrieve entry with a `null` key, or test if `null` key entry exists
- trying to store a `null` value

While trying to perform any of the above operations is a bug and should be avoided, we can't ensure that any instrumentation plugin will try to, causing errors and verbose log.

This PR provides a wrapper implementation for WeakConcurrentMap that will improve current implementation:
- avoids throwing NPE + provide no-op implementation when trying to perform any of the above operations with null keys/values
- logs a single line in log with `WARN` level
- provides full error stack trace in `DEBUG` level to allow easy identification of the root cause.

Few caveats:
- we could not easily wrap the map with our own type as there are numerous (~22) usages of `co.elastic.apm.agent.sdk.weakmap.WeakMapSupplier#createMap` 

## Checklist

- [x] This is a bugfix
  - [x] manual testing with `log_level=debug` with packaged agent
  - [x] manual testing with `log_level=warn` with packaged agent


apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/weakmap/NullSafeWeakConcurrentMap.java
```diff
@@ -0,0 +1,129 @@
...
+     * checks if key or value is {@literal null}
+     *
+     * @param v key or value
+     * @return {@literal true} if key is non-null, {@literal false} if null
+     */
+    private static <T> boolean isNull(@Nullable T v) {
+        if (null != v) {
+            return false;
+        }
+        String msg = "trying to use null key or value";
```
eyalkoren, line 121: We should probably have separate message for key and value, for easier analysis in the next bug

apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/weakmap/WeakMapSupplier.java
```diff
@@ -36,12 +36,13 @@
...
 
     public static <K, V> WeakConcurrentMap<K, V> createMap() {
-        WeakConcurrentMap<K, V> result = new WeakConcurrentMap<>(false);
+        WeakConcurrentMap<K, V> result = new NullSafeWeakConcurrentMap<>(false);
         registeredMaps.add(result);
         return result;
     }
 
     public static <V> WeakConcurrentSet<V> createSet() {
+        // note: we don't have a NullSafeWeakConcurrentSet here because we can't change the underlying map impl.
```
eyalkoren, line 45: Why do we need to change the underlying map and not extend and override set API - `add`, `contains`, `remove` etc.?

eyalkoren (commented):
There are two usages of `WeakConcurrentMap` that deliberately not use the factory (so they can use a dedicated cleaner thread) - let's replace those as well.

Also, worth looking into the already reported stack trace and see if we can find what's wrong:
```
2020-12-22 12:17:40,576 ERROR [stderr] (default task-1) java.lang.NullPointerException
2020-12-22 12:17:40,577 ERROR [stderr] (default task-1) at co.elastic.apm.agent.shaded.weaklockfree.WeakConcurrentMap.containsKey(WeakConcurrentMap.java:146)
2020-12-22 12:17:40,579 ERROR [stderr] (default task-1) at co.elastic.apm.agent.process.ProcessHelper.doStartProcess(ProcessHelper.java:66)
2020-12-22 12:17:40,580 ERROR [stderr] (default task-1) at co.elastic.apm.agent.process.ProcessHelper.startProcess(ProcessHelper.java:51)
2020-12-22 12:17:40,580 ERROR [stderr] (default task-1) at co.elastic.apm.agent.process.ProcessStartInstrumentation$ProcessBuilderStartAdvice.onExit(ProcessStartInstrumentation.java:77)
2020-12-22 12:17:40,581 ERROR [stderr] (default task-1) at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1071)
2020-12-22 12:17:40,581 ERROR [stderr] (default task-1) at java.base/java.lang.Runtime.exec(Runtime.java:591)
2020-12-22 12:17:40,582 ERROR [stderr] (default task-1) at java.base/java.lang.Runtime.exec(Runtime.java:450)
...
```

eyalkoren:
Let's also handle the NPEs coming from the `WeakConcurrentMap` used by `com.blogspot.mydailyjava.weaklockfree.DetachedThreadLocal` (see #1598) - we don't use it directly, we extend it with `GlobalThreadLocal`, so I think all we need to do is override `set()`, because keys are guaranteed to be non null, but take another look.

apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/weakmap/WeakMapSupplier.java
SylvainJuge: The constructor does not allow to change the single field, thus that would mean re-implementing the whole class, while it's only about ~120 lines I'm not sure it's really worth it for now.

apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/weakmap/NullCheck.java
```diff
@@ -0,0 +1,38 @@
...
+        return true;
+    }
+
+    public static <T> boolean isNullKey(@Nullable T key){
+        return true;
+    }
+
+    public static <T> boolean isNullValue(@Nullable T value) {
+        return true;
+    }
```
eyalkoren, line 37: ❓

SylvainJuge: sorry, forgot to update those, it's not really surprising that test fails with that :-)

eyalkoren (approved):

PR #1453, opened 2020-10-24
fetch cloud metada - common logic

kananindzya:
refer elastic/apm-agent-java#1264 
Note: we may have problems if a new version is used in the cloud with old versions of apm-server
Note2: I didn't add cloud.account.name field in serialization, because this field value is not available in the selected providers (aws, gcp, azure).

PR #1599, opened 2021-01-06
Cloud metadata

### LINKED ISSUE #1264: Collect cloud metadata
Implementing https://github.com/elastic/apm/issues/256 and https://github.com/elastic/apm/issues/289.
Reference implementation in [Python agent](https://github.com/elastic/apm-agent-python/pull/826).

eyalkoren:
## What does this PR do?
Closes #1264
Extending the work started at #1453 

## Checklist
  - [x] [Spec](https://github.com/elastic/apm/blob/master/specs/agents/metadata.md#cloud-provider-metadata) is implemented
  - [x] Updating [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [x] Unit tests for raw metadata deserialization logic
  - [x] Unit tests for metadata serialization logic
  - [x] Unit tests for the new async metadata creation mechanism with all `CLOUD_PROVIDER` options
  - [x] manual tests on each cloud with all `CLOUD_PROVIDER` options
  - [x] manual tests on each cloud with invalid URL or improper input parsing - for future changes in cloud APIs
  - [x] manual test with older APM server


apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ApmServerConfigurationSource.java
```diff
@@ -160,10 +157,15 @@ String fetchConfig(final ConfigurationRegistry configurationRegistry) {
             return null;
         }
         try {
+            payloadSerializer.blockUntilReady();
```
felixbarny, line 160: Not saying this doesn't work but did you consider providing a `Future<MetaData>` in the constructor and blocking on `Future#get` here and call `DslJsonSerializer#serializeMetadata(MetaData metaData)`? That way, we could keep state management out of the `DslJsonSerializer`.

eyalkoren: Of course, this is how I started, but there were several reasons I changed that: the first is that nobody really needs the metadata object, it is spread it around just so it can be handed to serialization. 
In addition, I wanted to create a central serialization for it, meaning that `MetaData` (or something else) needs to serialize and hold the serialized data. This would mean that `MetaData` (or something else) now either create and use a `PayloadSerializer`, or I extract the metadata serialization from the rest, but that proved to be too painful and risky for this PR. Both options looked bad when I tried them.
Lastly, I don't think it makes sense to block during serialization itself, for example- you better only start the HTTP request when ready to write to it. So I thought it would be better if the serialization interface provides a good indication for that, instead of relying on `Future#get` (this is not the only usage of it).

Bottom line - there are many options to do that and I played with several ending with this. Given the multiple dependencies we had on `MetaData` and the time I wanted to spend on this, this seems as the least ugly.

felixbarny: Thanks for explaining the background

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/MetaDataTest.java
```diff
@@ -0,0 +1,149 @@
...
+            metaDataFuture.get(0, TimeUnit.MILLISECONDS);
+        } catch (TimeoutException e) {
+            timeoutException = e;
+        }
+        if (provider != GCP && provider != currentCloudProvider) {
+            // The GCP fails quickly on DNS lookup
+            assertThat(timeoutException).isInstanceOf(TimeoutException.class);
+        }
+        // In AWS we may need two timeouts - one for the API token and one for the metadata itself
+        long timeout = (long) (coreConfiguration.geCloudMetadataDiscoveryTimeoutMs() * ((provider == AWS) ? 2.5 : 1.5));
```
SylvainJuge, line 98: [question] would there be a way to avoid static timeouts as those might imply test flakyness ?

apm-agent-core/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SylvainJuge, line 1: [question] why do we have to use `logback-test.xml` here an not `logback.xml` ?

apm-agent-plugins/apm-kafka-plugin/apm-kafka-headers-plugin/src/test/java/co/elastic/apm/agent/kafka/KafkaClientVersionsIT.java
```diff
@@ -30,6 +30,7 @@
 import org.junit.runners.Parameterized;
 
 import java.util.Arrays;
+import java.util.List;
```
SylvainJuge, line 33: [minor] likely unused import

apm-agent-plugins/apm-log-shipper-plugin/src/main/java/co/elastic/apm/agent/log/shipper/ApmServerLogShipper.java
```diff
@@ -51,12 +50,12 @@
...
 
-    public ApmServerLogShipper(ApmServerClient apmServerClient, ReporterConfiguration reporterConfiguration, MetaData metaData, PayloadSerializer payloadSerializer) {
-        super(reporterConfiguration, metaData, payloadSerializer, apmServerClient);
+    public ApmServerLogShipper(ApmServerClient apmServerClient, ReporterConfiguration reporterConfiguration, PayloadSerializer payloadSerializer) {
+        super(reporterConfiguration, payloadSerializer, apmServerClient);
     }
 
     @Override
-    public boolean onLineAvailable(TailableFile tailableFile, byte[] line, int offset, int length, boolean eol) throws IOException {
+    public boolean onLineAvailable(TailableFile tailableFile, byte[] line, int offset, int length, boolean eol) throws Exception {
```
SylvainJuge, line 58: [question] why change to the broad `Exception` here ? If it's a way to make uses of this API to having to handle potential exceptions, that would be worth to add the the method Javadoc to be explicit.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/CloudMetadataProvider.java
```diff
@@ -0,0 +1,401 @@
...
+            }
+        }
+        String awsMetadataUrl = "http://169.254.169.254/latest/dynamic/instance-identity/document";
+        Map<String, String> documentHeaders = null;
+        if (token != null) {
+            documentHeaders = new HashMap<>(1);
+            documentHeaders.put("X-aws-ec2-metadata-token", token);
+        }
+        String metadata = executeRequest(awsMetadataUrl, "GET", documentHeaders, queryTimeoutMs);
+        logger.debug("AWS metadata retrieved");
```
SylvainJuge, line 202: maybe output JSON for easy debugging

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/CloudMetadataProvider.java
```diff
@@ -0,0 +1,401 @@
...
+                                    // Expected - trial and error method
+                                }
+                                return gcpInfo;
+                            }
+                        });
+                    } finally {
+                        executor.shutdown();
+                    }
+
+                    long futureTimeout = queryTimeoutMs + 200;
```
SylvainJuge, line 138: [question] why do we add a fixed extra timeout for the future, how do we know that 200ms is fine ?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/CloudMetadataProvider.java
```diff
@@ -0,0 +1,401 @@
...
+                }
+                case AZURE: {
+                    try {
+                        cloudProviderInfo = getAzureMetadata(queryTimeoutMs);
+                    } catch (Exception e) {
+                        unexpectedError = e;
+                    }
+                    break;
+                }
+                case AUTO: {
```
SylvainJuge, line 104: [minor] maybe move this large block into a separate method as it's already nested in a switch statement.

SylvainJuge (commented):
Nice! we can definitely see that having an asynchronous fetching of metadata is not without complications.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ApmServerConfigurationSource.java
eyalkoren: I don't mind, but I think it is weird to explain why something was chosen over something else you wouldn't necessarily consider as a default option. Imagine you only see this code now for the first time- `MetaData` can be easily an internal serialization thing not known to the outer world (it is not package private only not to broaden the refactoring of this PR even more). In such case, this is the interface of the serializer- it needs time to fully initialize, no matter why, so users of it should be aware. I made efforts to emphasize this in the API as much as possible, including javadoc and an Exception which is part of the API.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
```diff
@@ -132,7 +141,7 @@ public void flush() throws IOException {
     }
 
     @Override
-    public void flush() throws IOException {
+    public void fullFlush() throws IOException {
```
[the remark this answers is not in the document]
eyalkoren: It is explained in the interface, but I will copy to here as well - if you missed it others will to 🙂

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/PayloadSerializer.java
```diff
@@ -71,4 +102,10 @@
     JsonWriter getJsonWriter();
 
     void writeBytes(byte[] bytes, int len);
+
+    class UninitializedException extends Exception {
```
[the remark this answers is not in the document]
eyalkoren: The main purpose of this type is to make anyone calling methods that throws it aware of the requirement for initialization, and the javadoc explains that. I prefer you notice this in compilation and for that it needs to be a checked Exception.

> it's not a recoverable

Not sure why - it may affect only the first send/serialization and not the subsequent ones.

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/MetaDataTest.java
eyalkoren: I don't think this implies flakiness, it is a **timeout** with a big safety cushion - discovery of AWS is very quick on AWS and almost exactly 2 seconds elsewhere, with Azure it would be quick on Azure and about 1 second elsewhere. It was my test confirming that all metadata discovery queries occur in parallel. If we see it is flaky, it's extremely easy to just change the timeout to 5 seconds, but a test with a timeout is necessary as we want to make sure nobody breaks the metadata discovery timeouts and let them just execute for too long.

apm-agent-core/src/test/resources/logback-test.xml
eyalkoren: We don't have to, but it is very specific for tests (the schema validation library) and it works...

apm-agent-plugins/apm-log-shipper-plugin/src/main/java/co/elastic/apm/agent/log/shipper/ApmServerLogShipper.java
eyalkoren: Because it felt inappropriate to add `UninitializedException` (serializer-specific) to the `FileChangeListener` interface...
Would you prefer that? Or something else?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/CloudMetadataProvider.java
eyalkoren: I did that and got very big output, including SSH keys 😬 ...

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/CloudMetadataProvider.java
eyalkoren: > why do we add a fixed extra timeout for the future

We want to limit this blocking somehow, but if I use the same timeout used for the HTTP query, that would be too tight, it needs to be longer.

> how do we know that 200ms is fine ?

We don't know, we live on the edge 😛

SylvainJuge (approved):

PR #1607, opened 2021-01-08
[CI] Run hourly to look for head changes

v1v:
## What does this PR do?

Automatically trigger any builds for any PRs that got a change in their head target from once a week to every hour if required
This will not update the PR but the workspace where the build happens in the CI. In other words, the build merges the PR with its Target branch before running anything.

It has been configured to run hourly to avoid any issues with the GitHub API quota, we could stress a bit more to run every 30 minutes, but let's see if this approach is somehow interesting to you. 

## Why

This will ensure PRs are always ready to merge, even if they have not been updated with the latest changes.

### DIFF
--- a/.ci/jobs/apm-agent-java-mbp.yml
+++ b/.ci/jobs/apm-agent-java-mbp.yml
@@ -42,3 +42,4 @@
             timeout: '15'
             use-author: true
             wipe-workspace: 'True'
+    periodic-folder-trigger: 1h


cachedout:
I worry a bit about this behavior. Unless it's _very_ clear what is happening, I can foresee a scenario like this:

1. PR is submitted and tests pass.
2. [time passes]
3. The branch against which the PR has been submitted is updated
4. Tests automatically (re)-run against the PR.
5. Due to a change in the upstream branch, tests fail.

At this point, if a developer goes to try to reproduce this problem, it seems natural to me that they would try to replicate it _given the state of their branch when it was submitted_, instead of reproducing what is actually being tests, which is their changes merged with HEAD on the branch.

If we can make it absolutely clear that the test failure happened against the changes which were merged with the head of the branch and give a clear indication to the developer against the actual test environment (their changes + HEAD), then they can replicate that state locally. Otherwise, we may be setting developers up for a very painful debugging experience.

Hopefully I explained this concern well, but if not I'm happy to jump on a call.

Also, if the Java devs are OK with this then it seems perfectly fine and I have no objection at all. I just wanted to specifically call out the behavior and what _could_ happen if we aren't careful just so there aren't any misunderstandings. Thanks!

cachedout:
Follow-up. @v1v explained to me that this behavior is actually on by default (which I wasn't aware of) and that this change simply makes the period between checks more aggressive. That makes this change much more desirable and I think we should merge it. 👍

kuisathaverat (approved):
LGTM, we will consume 30-50 API calls with the scan is not too much

cachedout (approved):

PR #1633, opened 2021-01-27
Fallback to TLS and quietly catch initialization error

eyalkoren:
## What does this PR do?
Supports FIPS mode by defaulting to `TLS` if `SSL` is not supported and by quietly catching failure to create a trust-all `SocketFactory` (assuming that in this mode, APM Server certification verification is required).

Context: reported in our [forum](https://discuss.elastic.co/t/apm-agent-is-not-working-on-fips-enabled-java-application/261815).

This implementation should not change behaviour of any existing deployment when the agent is upgraded.

## Checklist

- [x] This is a bugfix
  - [x] The fix was verified by the user who reported it

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -33,6 +33,7 @@ configured through the <<config-cloud-provider, `cloud_provider`>> config option
 [float]
 ===== Bug fixes
 * Fixing crashes observed in Java 7 at sporadic timing by applying a few seconds delay on bootstrap - {pull}1594[#1594]
+* Fallback to using "TLS" `SSLContext` when "SSL" is not available - {pull}1633[#1633]
 
 [[release-notes-1.x]]
 === Java Agent version 1.x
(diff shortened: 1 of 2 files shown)


SylvainJuge (approved):
LGTM, there should not have any side-effect for the current behavior, only to make it work in FIPS mode. Double checking with user that reported the issue would be required though.

SylvainJuge:
Just one thing that could be improved a bit after the [last comment](https://discuss.elastic.co/t/apm-agent-is-not-working-on-fips-enabled-java-application/261815/7), we could probably log the warning message in WARN + message from exception, and then log the full exception stack only in DEBUG, thus that would mean just a single WARN message when using in FIPS mode. Otherwise that would leave a confusing stack trace in logs whenever the fallback is used.

eyalkoren:
Yes, this is definitely a leftover warning, I intend it to be only `info` message if there's a failure to create the trust-all factory, as it is only a problem if the user turns off server cert verification. Only if this is the case, we will issue a warning at the first connection attempt.
I will change that and add the stack trace with `debug` level as you suggest.

SylvainJuge (approved):

PR #1648, opened 2021-02-02
Avoid sending null fields in the cloud metadata

eyalkoren:
## What does this PR do?
Fixes a problem reported in the [forum](https://discuss.elastic.co/t/apm-agent-is-not-working-on-fips-enabled-java-application/261815/8) - following the recent addition of (#1599), the agent may fail to work with APM Server 7.9.x, which doesn't allow `null` values, if one of the cloud metadata fields is not populated. 
This was fixed in 7.10, but since we did not release yet, we would prevent it in the agent side by not serilalizing cloud metadata `null` values.

## Checklist

- [x] This is a bugfix
  - [ ] ~~I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)~~ - No need, the cloud metadata discovery feature was not released yet
  - [x] I have reproduced the problem and verified the fix works as expected


apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
```diff
@@ -517,42 +517,47 @@ private static void serializeSystem(final SystemInfo system, final StringBuilder
...
         jw.writeByte(OBJECT_END);
     }
 
+    private static void serializeNameAndIdField(@Nullable CloudProviderInfo.NameAndIdField nameAndIdField, String fieldName,
+                                                StringBuilder replaceBuilder, JsonWriter jw) {
+        if (nameAndIdField != null && !nameAndIdField.isEmpty()) {
+            writeFieldName(fieldName, jw);
+            jw.writeByte(JsonWriter.OBJECT_START);
+            boolean idWritten = false;
+            @Nullable String id = nameAndIdField.getId();
```
SylvainJuge, line 542: [minor] I don't think that `@Nullable` on local variable is relevant here.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/CloudMetadataProvider.java
```diff
@@ -370,10 +372,12 @@ static CloudProviderInfo deserializeAzureMetadata(@Nullable String rawMetadata)
 
         CloudProviderInfo cloudProviderInfo = new CloudProviderInfo("azure");
         cloudProviderInfo.setAccount(new CloudProviderInfo.ProviderAccount(subscriptionId));
-        cloudProviderInfo.setInstance(new CloudProviderInfo.ProviderInstance(vmId, vmName));
-        cloudProviderInfo.setProject(new CloudProviderInfo.ProviderProject(resourceGroupName));
+        cloudProviderInfo.setInstance(new CloudProviderInfo.NameAndIdField(vmName, vmId));
```
SylvainJuge, line 375: [minor] we could maybe make syntax a bit lighter here we by having `cloudProviderInfo.setInstance(String vmName, String vmId)`, so the setters would be mostly using Strings as parameters and the getters would return the built `NameAndIdField`.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/CloudProviderInfo.java
```diff
@@ -106,49 +107,32 @@ public ProviderMachine getMachine() {
         return machine;
     }
 
-    public void setMachine(ProviderMachine machine) {
+    public void setMachine(@Nullable ProviderMachine machine) {
         this.machine = machine;
     }
 
-    public static class ProviderAccount {
```
SylvainJuge, line 113: +1 on having less single use classes with identical fields.

SylvainJuge (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/CloudMetadataProvider.java
eyalkoren: I started applying this suggestion, but even if it made the syntax "lighter", it made the API more complex, as for each of these you need to support both textual and numeric ID, which means duplicating setters...
So I left it as is.

PR #1510, opened 2020-11-13
changed server_urls to server_url

kananindzya:
closes #1507 

@felixbarny hi, please review this PR

PR #1650, opened 2021-02-04
Server url config

### LINKED ISSUE #1507: [META 359] Consolidate server_url/server_urls option
Rename `server_urls` to `server_url` (see https://github.com/elastic/apm/issues/359) in a backwards compatible way.

Adapt documentation to only describe the `server_url` option.

eyalkoren:
## What does this PR do?
Closes #1507 
Closes #1510 (this PR continues based on it)
We are adding a `server_url` config **in addition** to the `server_urls` config. The new one will only accept a single URL, thus it will be aligned with other agents, and it will take precedence over the old one.

### Checklist:
- [x] `server_urls` config documentation notifies that this option is not aligned with other agents
- [x] Replace all examples in documentation to only use `server_url`
- [x] Replace the CI secret for the secret token to use `ELASTIC_APM_SERVER_URL`
- [x] Replace all usages in tests
- [x] Log an `info` message if both config options are set that only the plural form will be ignored
- [x] Add to CHANGELOG in the breaking changes section - removing the option to set empty `server_urls` and introducing the `disable_send` config instead
- [x] I tested manually different configurations and configuration change scenarios


felixbarny (approved):

apm-agent-core/src/test/java/co/elastic/apm/agent/report/ApmServerClientTest.java
```diff
@@ -243,23 +246,61 @@ public void testSimulateConcurrentConnectionError() {
...
+    public void testServerUrlSettingOverridesServerUrls() throws IOException {
+        URL tempUrl = new URL("http", "localhost", 9999, "");
+        config.save("server_url", tempUrl.toString(), SpyConfiguration.CONFIG_SOURCE_NAME);
+        List<URL> updatedServerUrls = apmServerClient.getServerUrls();
         assertThat(updatedServerUrls).isEqualTo(List.of(tempUrl));
     }
 
+    @Test
+    public void testDisableSend() {
+        // We have to go through that because the disable_send config is non-dynamic
```
SylvainJuge, line 284: Is there anything that would prevent this configuration from being dynamic ? If yes, then having a follow-up PR would be nice to prevent making this one larger.

SylvainJuge (approved):

apm-agent-core/src/test/java/co/elastic/apm/agent/report/ApmServerClientTest.java
eyalkoren: Making the `disable_send` config dynamic would be very easy, and even make its testing easier than it is now 🙂 
The reasons I thought it doesn't make sense to make it dynamic:
- when set to `true`, we are not polling for central config, which will make its dynamic nature quite impaired and confusing - it can be changed from `false` to `true` through central config, but not vice versa...
- the use case for this config is very specific -  you would normally use it when you don't want to setup APM server, but do want agent to trace things, for whatever reason. In such corner cases, I can't see a dynamic use for it. We already have the `recording` config which should be used in other cases.

PR #1655, opened 2021-02-10
Fix sample less than precision

SylvainJuge:
## What does this PR do?

This fixes a rounding error for less than precision but not zero values of sample_rate that were rounded zero, which is not consistent with the spec: (https://github.com/elastic/apm/blob/master/specs/agents/tracing-sampling.md#transaction-sampling)

> Values greater than `0` but less than `0.0001` should be rounded to `0.0001`.

Thanks @trentm  for finding this !

## Checklist

- [x] This is a bugfix


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceState.java
```diff
@@ -94,7 +100,11 @@ public void addTextHeader(String headerValue) {
                     if (0 <= value && value <= 1.0) {
                         // ensure proper rounding of sample rate to minimize storage
                         // even if configuration should not allow this, any upstream value might require rounding
-                        double rounded = Math.round(value * 10000d) / 10000d;
+                        double rounded = Math.round(value * PRECISION) / PRECISION;
+                        if (value > 0 && rounded == 0) {
+                            // avoid rounding to zero
+                            rounded = 1d / PRECISION;
+                        }
```
eyalkoren, line 108: Looks like a duplication. Can you use `RoundedDoubleConverter` to do that, preferably using the same hard coded precision used in `CoreConfiguration` for the `sample_rate` config?

SylvainJuge: Good point, fixed in the last commit.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -166,7 +167,7 @@
             "You must use the query bar to filter for a specific environment in versions prior to 7.2.")
         .build();
 
-    private final ConfigurationOption<Double> sampleRate = RoundedDoubleConverter.withPrecision(4)
+    private final ConfigurationOption<Double> sampleRate = RoundedDoubleConverter.withPrecision(SAMPLE_RATE_PRECISION_DIGITS)
```
eyalkoren, line 170: [minor] since the entire logic now resides within `RoundedDoubleConverter`, there is no need for this constant.

eyalkoren (approved):
The logic is common now, so 👍 

If you want, you can hide the precision even further so there are no numeric constants like `CoreConfiguration#SAMPLE_RATE_PRECISION_DIGITS` and `TraceState#PRECISION` (and eliminate duplication even further with the decimal power calculation): for example, you can add a `FOUR_DIGIT_PRECISION` public static constant instance to `RoundedDoubleConverter` and use it both in `TraceState` and `CoreConfiguration`.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
SylvainJuge: Good point, I've added a public static factory method to hide this, we always have ways to reuse with higher/lower precision if required (was already tested thus no harm to keep it).

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -58,6 +58,7 @@
     public static final String SERVICE_NAME = "service_name";
     public static final String SERVICE_NODE_NAME = "service_node_name";
     public static final String SAMPLE_RATE = "transaction_sample_rate";
+    public static final int SAMPLE_RATE_PRECISION_DIGITS = 4;
```
eyalkoren, line 61: Leftover?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceState.java
```diff
@@ -24,16 +24,21 @@
...
 package co.elastic.apm.agent.impl.transaction;
 
+import co.elastic.apm.agent.configuration.converter.RoundedDoubleConverter;
 import co.elastic.apm.agent.objectpool.Recyclable;
 
 import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.List;
 
+import static co.elastic.apm.agent.configuration.CoreConfiguration.SAMPLE_RATE_PRECISION_DIGITS;
```
eyalkoren, line 34: Leftover?

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/converter/RoundedDoubleConverter.java
```diff
@@ -24,36 +24,53 @@
...
-    private final int precisionDigits;
+    public static final int DEFAULT_PRECISION = 4;
+    private static final RoundedDoubleConverter DEFAULT_INSTANCE = new RoundedDoubleConverter(DEFAULT_PRECISION);
+
+    private final double precisionFactor;
+    private final NumberFormat numberFormat;
 
-    public static ConfigurationOption.ConfigurationOptionBuilder<Double> withPrecision(int precisionDigits) {
-        return ConfigurationOption.builder(new RoundedDoubleConverter(precisionDigits), Double.class);
+    public static RoundedDoubleConverter withDefaultPrecision() {
```
eyalkoren, line 41: 👍

eyalkoren (approved):

PR #1659, opened 2021-02-11
ci: use the releaseNotification step

v1v:
### What

Use https://github.com/elastic/apm-pipeline-library/pull/976 to simplify the body manipulation
Enable email/slack communications when a tag release has created

### Actions

- [x] Need the above-mentioned PR and also a release ✅ 
- [ ] email address?

### DIFF
--- a/.ci/release/Jenkinsfile
+++ b/.ci/release/Jenkinsfile
@@ -14,7 +14,7 @@ pipeline {
     JAVA_HOME = "${env.HUDSON_HOME}/.java/java10"
     PATH = "${env.JAVA_HOME}/bin:${env.PATH}"
     SLACK_CHANNEL = '#apm-agent-java'
-    NOTIFY_TO = 'TBD'
+    NOTIFY_TO = 'build-apm+apm-agent-java@elastic.co'
   }
   options {
     timeout(time: 3, unit: 'HOURS')
--- a/.ci/release/Jenkinsfile
+++ b/.ci/release/Jenkinsfile
@@ -13,6 +13,8 @@ pipeline {
     HOME = "${env.WORKSPACE}"
     JAVA_HOME = "${env.HUDSON_HOME}/.java/java10"
     PATH = "${env.JAVA_HOME}/bin:${env.PATH}"
+    SLACK_CHANNEL = '#apm-agent-java'
+    NOTIFY_TO = 'TBD'
   }
   options {
     timeout(time: 3, unit: 'HOURS')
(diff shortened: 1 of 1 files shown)


SylvainJuge:
@v1v do you think we could review & merge this as-is, or does it still needs few changes before it's ready ?

As it's release we might not be able to test it properly until the next release, but given it's only about notifications we should not expect breaking changes here.

.ci/release/Jenkinsfile
```diff
@@ -13,6 +13,8 @@ pipeline {
     HOME = "${env.WORKSPACE}"
     JAVA_HOME = "${env.HUDSON_HOME}/.java/java10"
     PATH = "${env.JAVA_HOME}/bin:${env.PATH}"
+    SLACK_CHANNEL = '#apm-agent-java'
+    NOTIFY_TO = 'TBD'
```
v1v, line 17: what email?

kuisathaverat (approved):

SylvainJuge (approved):

.ci/release/Jenkinsfile
```diff
@@ -238,8 +242,27 @@ pipeline {
               sh(label: "Push Docker image", script: "./scripts/jenkins/push_docker.sh")
             }
           }
+          post {
+            success {
+              notifyStatus(slackStatus: 'good', subject: "[${env.REPO}] Release published", body: "Great news, the release has been done successfully. (<${env.RUN_DISPLAY_URL}|Open>).")
```
cachedout, line 247:
```suggestion
              notifyStatus(slackStatus: 'good', subject: "[${env.REPO}] Release published", body: "Great news! The release has completed successfully. (<${env.RUN_DISPLAY_URL}|Open>).")
```

cachedout (approved):
👍

.ci/release/Jenkinsfile
v1v: @elastic/apm-agent-java , can you clarify what email do you want to receive it from? And whether it can be exposed here? Otherwise, I can ask the infra team to create a credentials entry with the email address if it helps

SylvainJuge: As suggested by @felixbarny, using the same thing as other agents do seems the best option:
We do not (yet) have email alias that I'm aware of though, +1 on creating one to make it consistent with others if required.

https://github.com/elastic/apm-agent-nodejs/blob/master/.ci/Jenkinsfile#L10

v1v: That's the generic one, so we can follow up if the generic one is not sufficient, then
```suggestion
    NOTIFY_TO = credentials('notify-to')
```

felixbarny: 🐻

v1v: 🤣 
I raised a PR to configure the credentials in the `internal-ci`, just waiting for the review and deployment.

v1v: Infra team suggested that we don't need to mask the email:


So I'll add the filter `apm-agent-java`

```suggestion
    NOTIFY_TO = 'build-apm+apm-agent-java@elastic.co'
```

kuisathaverat (approved):

PR #1660, opened 2021-02-14
External plugin loading fix

eyalkoren:
## What does this PR do?
Fixes an error in the loading of external plugins, as reported in [the forum](https://discuss.elastic.co/t/trying-external-plugin-test-on-real-setup-fails-with-cannot-resolve-type-description-for-co-elastic-apm-agent-bci-classloading-lookupexposer/260923): the parent of `ExternalPluginClassLoader`s is the agent class loader, i.e. the class loader that actually loads `co.elastic.apm.agent.bci.ElasticApmAgent`, which is the bootstrap CL. While this is sufficient for loading agent classes, it is insufficient for locating class files through `java.lang.ClassLoader#getResourceAsStream` as the agent jar is only appended to bootstrap class loader search, but not to resource lookup. 
The proposed fix is to leave the parent CL as is, but use a compound `ClassFileLocator` when creating the Indy plugin CL for external plugins.

## Checklist

- [x] This is a bugfix
  - [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc) - both bug fix and newly exposed capability
  - [x] I have added tests that would fail without this fix - requires a real integration test that loads a full agent.
  - [x] The disabled test is replaced with a valid test
  - [x] The external plugins capability is exposed through the `plugins_dir` configuration


apm-agent-core/src/test/java/co/elastic/apm/agent/bci/InstrumentationTest.java
```diff
@@ -104,7 +105,12 @@ void testIntercept() {
         assertThat(interceptMe()).isEqualTo("intercepted");
     }
 
+    /*
+     * todo: this test is invalid because the agent CL here is the App CL, whereas in reality it would be the bootstrap CL
```
felixbarny, line 109: Is that test failing now or does it just not test what it should test?

felixbarny (approved):

apm-agent-core/src/test/java/co/elastic/apm/agent/bci/InstrumentationTest.java
eyalkoren: It is failing now because I [didn't allow](https://github.com/elastic/apm-agent-java/pull/1660/files#diff-2ec63f545539ff7d65f52ef01087a18c7cdd9eb7130de156fb412ae6fe9e580eR324) not having the agent jar. I can make it less strict, but I don't like having testing-specific logic in the production code...

felixbarny: 👌

integration-tests/external-plugin-test/external-plugin-common/pom.xml
```diff
@@ -0,0 +1,18 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>external-plugin-test</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.21.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>external-plugin-common</artifactId>
```
SylvainJuge, line 12: [question] I'm not sure about the "external-plugin-common" naming here, as it's just an internal module of the test application that we have to ship as a regular jar to allow the external plugin to have it as a dependency.

Do you think we could have something that makes it more obvious that it's a module of the application and not the external plugin ? Using a very generic `app-XXX` with `XXX=internal` for example might prevent having to explain what modules does what.

integration-tests/external-plugin-test/external-plugin/pom.xml
```diff
@@ -0,0 +1,106 @@
...
+            <type>test-jar</type>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>external-plugin-common</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
```
SylvainJuge, line 74: [minor] Here would be a good place to document that this dependency is on what we want to instrument, thus we have to have it as a compile-time dependency but NOT shade and embed it as target application classes will used instead.

Also, shouldn't the scope be `provided` here ?

integration-tests/external-plugin-test/external-plugin/src/test/java/co/elastic/apm/plugin/test/PluginInstrumentationTest.java
```diff
@@ -0,0 +1,80 @@
...
+import co.elastic.apm.agent.AbstractInstrumentationTest;
+import co.elastic.apm.agent.impl.transaction.Transaction;
+import org.junit.jupiter.api.Test;
+
+import java.util.Objects;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * todo: explain what this does- only a unit test for quick sanity tests of the instrumentation. It is not a full
```
SylvainJuge, line 36: [minor] leftover todo, but relevant prose

SylvainJuge (approved):

integration-tests/external-plugin-test/external-plugin-common/pom.xml
eyalkoren: Not sure what you mean. It is required in both other modules, not only the test app. The unit test (`PluginInstrumentationTest`) also uses it.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
```diff
@@ -307,23 +316,34 @@ public static ConstantCallSite bootstrap(MethodHandles.Lookup lookup,
...
                 pluginClasses,
                 adviceClassLoader,
+                classFileLocator,
                 isAnnotatedWith(named(GlobalState.class.getName()))
                     // no plugin CL necessary as all types are available form bootstrap CL
                     // also, this plugin is used as a dependency in other plugins
                     .or(nameStartsWith("co.elastic.apm.agent.concurrent")));
             Class<?> adviceInPluginCL = pluginClassLoader.loadClass(adviceClassName);
-            Class<LookupExposer> lookupExposer = (Class<LookupExposer>) pluginClassLoader.loadClass("co.elastic.apm.agent.bci.classloading.LookupExposer");
+            Class<LookupExposer> lookupExposer = (Class<LookupExposer>) pluginClassLoader.loadClass(LOOKUP_EXPOSER_CLASS_NAME);
```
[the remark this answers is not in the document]
eyalkoren: I think the todo above will cover that

integration-tests/external-plugin-test/external-plugin-common/pom.xml
eyalkoren: Renamed to `plugin-instrumentation-target`

PR #1664, opened 2021-02-15
try improve OOTB experience with intelij

SylvainJuge:
## What does this PR do?

Micrometer module requires to compile and target java 8, thus the `maven.compiler.target` property was set to target Java 8.
However, when Intelij imports the module, Java 8 is picked as module language level, thus the tests (which happen to rely on Java 9 API) can't be properly compiled.

We now have an `InteliJ` profile that is automatically enabled and sets the `maven.compiler.target` property to `11`, BUT this value is not used when the module overrides `maven.compiler.target`. Setting `maven.compiler.source` to the same value in the profile allows to work-around this behavior.

Also, the behavior wasn't 100% consistent when importing the whole project and refreshing the module settings from maven, this PR attempts to fix both in order to not have any errors in the IDE by default.

## Checklist

- [x] manual test on project import
- [x] manual test on existing project re-import (Ctrl-Shift-O)


eyalkoren (approved):
❤️

Issue #1665, opened 2021-02-15
[CI] find a workaround to the maven dependencies download failures


kuisathaverat:
We have a really recurrent issue downloading maven dependencies, we have to find a definitive workaround.

```
[2021-02-14T09:13:38.564Z] [ERROR] Failed to execute goal pl.project13.maven:git-commit-id-plugin:4.0.2:revision (get-the-git-infos) on project elastic-apm-agent: Execution get-the-git-infos of goal pl.project13.maven:git-commit-id-plugin:4.0.2:revision failed: Plugin pl.project13.maven:git-commit-id-plugin:4.0.2 or one of its dependencies could not be resolved: Failed to collect dependencies at pl.project13.maven:git-commit-id-plugin:jar:4.0.2 -> pl.project13.maven:git-commit-id-plugin-core:jar:4.0.2: Failed to read artifact descriptor for pl.project13.maven:git-commit-id-plugin-core:jar:4.0.2: Could not transfer artifact pl.project13.maven:git-commit-id-plugin-core:pom:4.0.2 from/to central (https://repo.maven.apache.org/maven2): Transfer failed for https://repo.maven.apache.org/maven2/pl/project13/maven/git-commit-id-plugin-core/4.0.2/git-commit-id-plugin-core-4.0.2.pom: Connection reset -> [Help 1]
```

PR #1666, opened 2021-02-15
[CI] Retry dependencies download

kuisathaverat:
## What does this PR do?

* It adds the comment `/test` to trigger a build from GitHub
* It retries the build 5 times with a timeout of 10 seconds between times, this force to download the dependencies correctly, in the worst case the build fails for other reasons, it increase the build time on less than 5 min

closes https://github.com/elastic/apm-agent-java/issues/1665


.ci/packer_cache.sh
```diff
@@ -1,14 +1,15 @@
 #!/usr/bin/env bash
-source /usr/local/bin/bash_standard_lib.sh
+set +e
```
v1v, line 2: I almost comment that `--fail-never` was removed 👍

v1v (approved):

Jenkinsfile
```diff
@@ -81,12 +81,13 @@ pipeline {
...
-                sh label: 'Prepare .m2 cached folder', returnStatus: true, script: 'cp -rf /var/lib/jenkins/.m2/repository .m2'
+                sh label: 'Prepare .m2 cached folder', returnStatus: true, script: 'cp -Rf /var/lib/jenkins/.m2/repository ${HOME}/.m2'
                 sh label: 'Size .m2', returnStatus: true, script: 'du -hs .m2'
               }
               dir("${BASE_DIR}"){
-                sh label: 'mvn dependencies', script: './mvnw -q dependency:go-offline --fail-never'
-                sh label: 'mvn install', script: "./mvnw clean install -DskipTests=true -Dmaven.javadoc.skip=true"
+                retryWithSleep(retries: 5, seconds: 10) {
+                  sh label: 'mvn install', script: "./mvnw clean install -DskipTests=true -Dmaven.javadoc.skip=true"
+                }
```
v1v, line 90: This might cause some issues with the GitHub PR comment since it will report any of those failed iterations :/  See https://github.com/elastic/apm-agent-java/pull/1666#issuecomment-779167310

kuisathaverat: use `retry` on the command line will hide that we have an issue, this implementation is a patch to resolve the issue, but it is not a definitive solution. In a follow up we will move to our own maven repo and check if that resolves the issue.

v1v: I meant to decouple the dependencies downloading/caching from the install goal.

Otherwise there will be some failed steps in the GitHub comment, and that's what I think it might not be ideal since it will cause some confusion.

Since the issues seems to be related with the CI environment when fetching all the artifacts from the maven central. I'd rather prefer to split the `stage('Build')` stage in:
- fetch dependencies for the CI.
- Run the mvn install

`".ci/packer_cache.sh"` might not fail, but the `"./mvnw clean install` might do it and that's a genuine reason.

What do you think the below implementation?

```
    ## Prepare the CI worker with all the cached dependencies 
    sh label: 'fetch dependencies', script: ".ci/packer_cache.sh"
    ## Run actually the maven goal to be validated in this stage. 
    sh label: 'mvn install', script: "./mvnw clean install -DskipTests=true -Dmaven.javadoc.skip=true"
```


I'm not sure if `package` should be replaced with `install`

kuisathaverat: maybe the message should not send the errors if the build is success

v1v: Good point, I'll open a GitHub issues about it to discuss it further

SylvainJuge (approved):
LGTM. Can we validate this without merging on master ?

.ci/packer_cache.sh
```diff
@@ -1,14 +1,15 @@
...
 
-JAVA_HOME=$HOME/.java/java11 \
-  ./mvnw dependency:go-offline --fail-never -q -B
+source /usr/local/bin/bash_standard_lib.sh
 
-JAVA_HOME=$HOME/.java/java11 ./mvnw clean verify \
-  --fail-never -q -B \
+retry 3 JAVA_HOME=$HOME/.java/java11 \
```
felixbarny, line 6: Do we have any idea about the root cause of the connection issues? It never happens locally. Could it be an OS or network setup issue?

kuisathaverat: since the issue is easy to replicate in a local environment with an empty .m2 repository, the most probable cause is the implementation of the retry system for downloads that maven have, many downloads at the same time, and the service of maven central, we are not the only people that face the issue.

Jenkinsfile
```diff
@@ -81,12 +81,13 @@ pipeline {
...
               whenTrue(fileExists('/var/lib/jenkins/.m2/repository')) {
-                sh label: 'Prepare .m2 cached folder', returnStatus: true, script: 'cp -rf /var/lib/jenkins/.m2/repository .m2'
+                sh label: 'Prepare .m2 cached folder', returnStatus: true, script: 'cp -Rf /var/lib/jenkins/.m2/repository ${HOME}/.m2'
                 sh label: 'Size .m2', returnStatus: true, script: 'du -hs .m2'
               }
               dir("${BASE_DIR}"){
-                sh label: 'mvn dependencies', script: './mvnw -q dependency:go-offline --fail-never'
-                sh label: 'mvn install', script: "./mvnw clean install -DskipTests=true -Dmaven.javadoc.skip=true"
+                retryWithSleep(retries: 5, seconds: 10) {
+                  sh label: 'mvn install', script: "./mvnw clean install -DskipTests=true -Dmaven.javadoc.skip=true"
```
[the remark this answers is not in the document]
kuisathaverat: based on the experience the wagon settings are not really useful. The only strategy that workaround the issue is to retry the `maven` command again.

Jenkinsfile
v1v: There are some cases that might also create some confusion, let me explain:

Given a PR that failed in the `test` stage and got 4 iterations for the `mvn install` then the GitHub PR comment will contain:
* The test failure details
* The failed steps:
  * 4 steps for the `mvn install` command
  * 1 step for the `mvn test` command

So the `mvn install` steps are not important to be highlighted since the 5th retry worked in the end. But the GitHub PR comment will report them anyway.

What do you think to redesign this PR to avoid the above scenario? The previous comment could help 

I

PR #1668, opened 2021-02-17
[load-testing] silence tomcat to avoid verbose logs

SylvainJuge:
We observed an increasing system memory usage during previous executions, which might be related to having a very large log file for the application (multiple dozen of gigabytes). Fixing this log verbosity would allow to:
1. keep log size at a reasonable signal/noise level, so we don't have to store it on external storage
2. ensure that this observation was only related to a large log file and not the agent activity.

Also, those errors aren't really relevant in the log file and the agent is expected to catch them, it is safe to ignore them when running the application.

## Checks

- [ ] run with this change + agent enabled = increasing system memory usage ?
  - If no, it was related to large log file, thus :smile: 
  - if yes then test without the agent thus run another test :grimacing: 
- [ ] run with this change + agent disabled = do we observe the same without the agent ?
  - if yes: then it's not related to the agent :smile: 
  - if no: then need further investigation :grimacing:


felixbarny:
Maybe we should also adjust the load script to hit the error endpoint less frequently or not at all. The simpler the load test script, the less jitter and the more reliable and actionable the results. I don't think we should attempt to mimic a real user in the load tests.

SylvainJuge:
> Maybe we should also adjust the load script to hit the error endpoint less frequently or not at all. The simpler the load test script, the less jitter and the more reliable and actionable the results. I don't think we should attempt to mimic a real user in the load tests.

That's exactly what we aim to cover with the different `benchmark configurations` defined in https://github.com/elastic/observability-dev/issues/1248, we should probably not put lots of errors in the baseline configuration, however, we might have one such configuration that tries to reproduce the behavior of a very chatty application.

SylvainJuge:
Also, it seems that the exceptions are still reported to the app log file, I'll need to investigate on how to properly remove those.

SylvainJuge:
started a 10h run after properly removing the verbose logging of errors: https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-load-mbp/view/change-requests/job/PR-1668/5/

SylvainJuge:
update: the application log is not verbose anymore, however there was still a small increase in system memory (test was less than 2h though due to JVM crash, thus might not be enough to conclude).

SylvainJuge:
Update: after running another execution with 10+ hours we have a flat system memory usage, thus it's very highly correlated with having a large log file.

PR #1687, opened 2021-03-08
Fix Nullpointer risk

Edimz:
I'm getting a Nullpointer here since my implementation does not contain the field "connected". Maybe reflection can't find it in my superclass.

## What does this PR do?
I added a declaring type of "connected" to prevent nullpointer exception when not using the field in a subclass.

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
+++ b/apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
@@ -82,7 +82,7 @@ public abstract class HttpUrlConnectionInstrumentation extends TracerAwareInstru
                 return null;
             }
             Span span = inFlightSpans.get(thiz);
-            if (span == null && !Boolean.TRUE.equals(connected)) {
+            if (span == null && !connected) {
                 final URL url = thiz.getURL();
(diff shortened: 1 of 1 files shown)


eyalkoren:
Thanks for reporting and suggesting a fix!
The `connected` variable is primitive, so I can't see how this fix prevents this NPE. I assume the error is thrown when Byte Buddy fails to assign a value for it because it can't find it in the class.
Please provide the entire NPE message and stack trace.

> my implementation does not contain "connected"

Can you elaborate - what is "your" implementation? Is it some custom implementation of `java.net.HttpURLConnection`? Can you share this code?

SylvainJuge:
Hi @Edimz which version of the JDK was used when you had `NullPointerException` triggered without this change ?
While providing a unit test to reproduce/validate that change might be tricky, being able to reproduce it once locally with the exact same JDK version might be enough here.

Edimz:
@eyalkoren, you are correct, my quick fix was not good, I changed to

```
@Advice.FieldValue(value = "connected",
                                       declaringType = URLConnection.class) boolean connected

```
And that made the trick, tho I got a new Nullpointer that I do not know how to fix, yet.

Edimz:
@SylvainJuge I just tested with another fix (correct one) but still get Nullpointer on another parameter (thiz.getURL() is null), I think I'll disable inferred spans until I've debugged this thoroughly.

eyalkoren:
@Edimz we will be glad if you can answer our questions:
- What JDK did are you using when seeing this error?
- What is the exact class that extends `java.net.HttpURLConnection` that is related to the error?
- Can you provide a stack trace of this error?

Please either answer those so that we can follow up, or close this issue if it is irrelevant anymore.
Thanks!

eyalkoren:
Closing as we got no feedback. If required, we can always reopen.

Issue #1673, opened 2021-02-19
Unexpected UnsupportedOperationException with ForkJoinTask instrumentation


SylvainJuge:
An ` UnsupportedOperationException` is thrown during ForkJoinTask instrumentation.

JVM: OpenJDK 11.0.8 (AdoptOpenJdk).
Found while executing instrumentation unit tests in the IDE.

The class `java.util.concurrent.ForkJoinTask` seems to be redefined at least twice with a different number of attributes, which trigger the exception.

By abusing some debugger watches, I managed to capture the following class dumps:
- before instrumentation
- after instrumentation

[class-dump.zip](https://github.com/elastic/apm-agent-java/files/6009987/class-dump.zip)

It seems that our instrumentation adds a class attribute, whereas we should not modify the class structure.

<details>
<summary>Full exception stack trace</summary>

```
2021-02-19 13:28:22,023 [main] WARN  co.elastic.apm.agent.bci.ElasticApmAgent - Error while redefining classes Error invoking java.lang.instrument.Instrumentation#retransformClasses
2021-02-19 13:28:22,025 [main] DEBUG co.elastic.apm.agent.bci.ElasticApmAgent - Error invoking java.lang.instrument.Instrumentation#retransformClasses
java.lang.IllegalStateException: Error invoking java.lang.instrument.Instrumentation#retransformClasses
	at net.bytebuddy.agent.builder.AgentBuilder$RedefinitionStrategy$Dispatcher$ForJava6CapableVm.retransformClasses(AgentBuilder.java:6985) ~[byte-buddy-dep-1.10.20.jar:?]
	at net.bytebuddy.agent.builder.AgentBuilder$RedefinitionStrategy$Collector$ForRetransformation.doApply(AgentBuilder.java:7256) ~[byte-buddy-dep-1.10.20.jar:?]
	at net.bytebuddy.agent.builder.AgentBuilder$RedefinitionStrategy$Collector.apply(AgentBuilder.java:7101) ~[byte-buddy-dep-1.10.20.jar:?]
	at net.bytebuddy.agent.builder.AgentBuilder$RedefinitionStrategy.apply(AgentBuilder.java:4937) ~[byte-buddy-dep-1.10.20.jar:?]
	at net.bytebuddy.agent.builder.AgentBuilder$Default.doInstall(AgentBuilder.java:9574) ~[byte-buddy-dep-1.10.20.jar:?]
[...]

PR #1689, opened 2021-03-08
Add InteliJ debugger work-around

SylvainJuge:
## What does this PR do?

Fixes https://github.com/elastic/apm-agent-java/issues/1673

Idea debugger agent is attached to the JVM and instruments `ForkJoinTask` by modifying the class structure when the `Debug Async Traces` feature is enabled (see related issue).

While this agent declares itself as being able to re-transform classes, it just skips applying bytecode transformation the class is being re-defined, while this optimization makes sense for most applications, it does not when agents are being attached and debugged.

The observed behavior is consistent with [agent code on github](https://github.com/JetBrains/intellij-community/blob/master/java/debugger/debugger-agent/src/com/intellij/rt/debugger/agent/CaptureAgent.java#L168) and the de-compiled agent version shipped with 2020.3

As a result, it makes our agent not able to get the modified bytecode by InteliJ agent, and we try to re-define it with the original class structure, which can't work.
This instrumentation allows to make the InteliJ agent re-generate it's modified version when re-defining classes, which allows our agent to properly instrument `ForkJoinTask`.

## Checklist

- [x] Test locally (was tested on #1305 PR which triggered the warning on agent startup). 
- ~~Find a good way to ensure this advice is applied early and before `ForkJoinTaskInstrumentation`~~
- [x] File an issue or open a PR in Idea repository with a proposed change, maybe adding an extra parameter to avoid this optimization


felixbarny:
Have you considered creating a PR for the IntelliJ agent that removes `&& classBeingRedefined == null` from the if statement? It seems that would fix the issue. The comment says they don't support retransform but what they actually do is to disallow other agents to retransform, too.

SylvainJuge:
Yes, that's exactly my plan here, (and it's the last item in the checklist).

apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/BaseInstrumentation.java
```diff
@@ -0,0 +1,17 @@
+package co.elastic.apm.agent.concurrent;
+
+import co.elastic.apm.agent.bci.TracerAwareInstrumentation;
+
+public abstract class BaseInstrumentation extends TracerAwareInstrumentation {
+
+    static {
+        if (Boolean.parseBoolean(System.getProperty("intellij.debug.agent"))) {
```
felixbarny, line 8:
```suggestion
        if (Boolean.getBoolean("intellij.debug.agent")) {
```

felixbarny (approved):

eyalkoren (approved):
Awesome solution for warning only when this instrumentation is applied!

SylvainJuge:
Opened PR on Idea debugger: https://github.com/JetBrains/intellij-community/pull/1528 for a proper fix. In the meantime we have this work-around.

PR #1691, opened 2021-03-08
Adding documentation about using agent only for metrics

eyalkoren:

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -426,7 +426,9 @@ If you want to dynamically change the status of the agent, use <<config-recordin
 [[config-instrument]]
 ==== `instrument` (added[1.0.0,Changing this value at runtime is possible since version 1.15.0])
 
-A boolean specifying if the agent should instrument the application to collect performance metrics for the app. When set to false, Elastic APM will not affect your application at all.
+A boolean specifying if the agent should instrument the application to collect traces for the app.
+ When set to `false`, Elastic APM will not apply bytecode instrumentation, thus it would have no effect on 
+your application, however it would still collect and send metrics to APM Server.
 
 NOTE: Both active and instrument needs to be true for instrumentation to be running.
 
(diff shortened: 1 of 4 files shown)


docs/configuration.asciidoc
```diff
@@ -426,7 +426,9 @@ If you want to dynamically change the status of the agent, use <<config-recordin
 [[config-instrument]]
 ==== `instrument` (added[1.0.0,Changing this value at runtime is possible since version 1.15.0])
 
-A boolean specifying if the agent should instrument the application to collect performance metrics for the app. When set to false, Elastic APM will not affect your application at all.
+A boolean specifying if the agent should instrument the application to collect traces for the app.
+ When set to `false`, Elastic APM will not apply bytecode instrumentation, thus it would have no effect on
```
felixbarny, line 430: Technically, the agent still does some bytecode instrumentation but only for things related to manual APIs, such as the public API, the OT bridge, and Micrometer.

docs/metrics.asciidoc
```diff
@@ -503,3 +504,14 @@ Fields:
...
+
+[float]
+[[metrics-only-mode]]
+=== Using the agent only for metrics collection
+
+There are cases where you would want to use the agent only to collect and ship metrics, without tracing any Java code.
+This may be useful, for example, when you already have an APM Server set up, and you want to ingest all metrics listed
+above from a JVM that runs something other than your application into Elasticsearch.
+
+In such cases, you may set the <<config-instrument, `instrument`>> config option to `false`. By doing so, agent will
```
felixbarny, line 516:
```suggestion
In such cases, you may set the <<config-instrument, `instrument`>> config option to `false`. By doing so, the agent will
```

docs/metrics.asciidoc
```diff
@@ -503,3 +504,14 @@ Fields:
...
 * `${name}`: The value of `functionCounter.count()` (the delta when using `CountingMode.STEP`).
 --
+
+[float]
+[[metrics-only-mode]]
+=== Using the agent only for metrics collection
+
+There are cases where you would want to use the agent only to collect and ship metrics, without tracing any Java code.
+This may be useful, for example, when you already have an APM Server set up, and you want to ingest all metrics listed
+above from a JVM that runs something other than your application into Elasticsearch.
```
felixbarny, line 514: [minor] I find that example a bit confusing. Especially this bit:

> runs something other than your application into Elasticsearch

3rd party applications may also be traced with the agent and you may just want metrics from a 1st party app.

Maybe just remove that sentence?

felixbarny (approved):
LGTM, just a minor/optional comment

docs/metrics.asciidoc
eyalkoren: OK

PR #1697, opened 2021-03-12
Vert.x Web instrumentation

AlexanderWert:
Provides Instrumentation for Vertx.x Web / HttpServer and Context propagation into Vert.x event engine.


apm-agent-plugins/apm-vertx/apm-vertx-common/src/main/java/co/elastic/apm/agent/vertx/NettyByteTransfer.java
```diff
@@ -0,0 +1,60 @@
...
+    private static NettyByteTransfer INSTANCE;
+    private final ByteBuf byteBuf;
+
+    private NettyByteTransfer(ByteBuf byteBuf) {
+        this.byteBuf = byteBuf;
+    }
+
+    public static NettyByteTransfer get(ByteBuf byteBuf) {
+        if (INSTANCE == null) {
+            INSTANCE = new NettyByteTransfer(byteBuf);
```
felixbarny, line 45: Unless I'm missing something, this stores the `ByteBuf` from the first request as a singleton which prevents the capturing from other requests.
I'd suggest changing the interface so that you don't need instance variables, instead the ByteBuf would be a parameter to the `transfer` and `readableBytes` methods.

apm-agent-plugins/apm-vertx/apm-vertx-common/src/main/java/co/elastic/apm/agent/vertx/NettyByteTransfer.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package co.elastic.apm.agent.vertx;
+
+import co.elastic.apm.agent.util.IOUtils;
+import io.netty.buffer.ByteBuf;
+
+import javax.annotation.Nullable;
+import java.nio.ByteBuffer;
+
+public class NettyByteTransfer implements IOUtils.BufferTransfer {
```
felixbarny, line 33:
```suggestion
public class NettyByteTransfer implements IOUtils.BufferTransfer<ByteBuf> {
```

apm-agent-plugins/apm-vertx/apm-vertx-common/src/main/java/co/elastic/apm/agent/vertx/NettyByteTransfer.java
```diff
@@ -0,0 +1,60 @@
...
+    public static NettyByteTransfer get(ByteBuf byteBuf) {
+        if (INSTANCE == null) {
+            INSTANCE = new NettyByteTransfer(byteBuf);
+        }
+
+        return INSTANCE;
+    }
+
+    @Override
+    public void transfer(ByteBuffer buffer) {
```
felixbarny, line 52:
```suggestion
    public void transfer(ByteBuf src, ByteBuffer dest) {
```

apm-agent-plugins/apm-vertx/apm-vertx-common/src/main/java/co/elastic/apm/agent/vertx/NettyByteTransfer.java
```diff
@@ -0,0 +1,60 @@
...
+        byteBuf.readBytes(buffer);
+    }
+
+    @Override
+    public int readableBytes() {
```
felixbarny, line 57:
```suggestion
    public int readableBytes(ByteBuf buf) {
```

apm-agent-plugins/apm-vertx/apm-vertx-3_6-plugin/src/main/java/co/elastic/apm/agent/vertx_3_6/helper/VertxWebHelper.java
```diff
@@ -0,0 +1,102 @@
...
+
+        return transaction;
+    }
+
+    private HttpServerRequest unwrapHttpServerRequest(HttpServerRequest request) {
+        try {
+            // reflection needed to unwrap the original request
+            if (request.getClass().getName().equals("io.vertx.ext.web.impl.HttpServerRequestWrapper")) {
+                Field delegateField = request.getClass().getDeclaredField("delegate");
+                delegateField.setAccessible(true);
```
felixbarny, line 84: That fails when module encapsulation is enforced and causes a warning on Java 9+. Any alternatives to that which don't require unwrapping?

felixbarny (commented):
A first pass on this. Looks pretty good so far 👏

apm-agent-plugins/apm-vertx/apm-vertx-common/src/main/java/co/elastic/apm/agent/vertx/NettyByteTransfer.java
AlexanderWert: 🤦 yes, you're absolutely right. Just extracted this functionality into a Vertx-specific IOUtils that extends the common IOUtils

apm-agent-plugins/apm-vertx/apm-vertx-3_6-plugin/src/main/java/co/elastic/apm/agent/vertx_3_6/helper/VertxWebHelper.java
AlexanderWert: Solved this with additional instrumentation. Though it still feels like a hacky workaround.

apm-agent-plugins/apm-vertx/apm-vertx-3_6-plugin/src/main/java/co/elastic/apm/agent/vertx_3_6/wrapper/GenericHandlerWrapper.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class GenericHandlerWrapper<T> implements Handler<T> {
+
+    private final Handler<T> actualHandler;
+    private final AbstractSpan parentSpan;
+
+    public GenericHandlerWrapper(AbstractSpan parentSpan, Handler<T> actualHandler) {
+        this.parentSpan = parentSpan;
+        this.actualHandler = actualHandler;
+    }
```
felixbarny, line 38: To guard against situations where the parent has ended and is recycled before the `handle` method starts, we'll have to increment the reference count here and decrement it after activating the parent span.

There may be other places where this is relevant.

```suggestion
        parentSpan.incrementReferences();
    }
```

apm-agent-plugins/apm-vertx/apm-vertx-3_6-plugin/src/main/java/co/elastic/apm/agent/vertx_3_6/wrapper/GenericHandlerWrapper.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Override
+    public void handle(T event) {
+        parentSpan.activate();
```
felixbarny, line 42:
```suggestion
        parentSpan.activate();
        parentSpan.decrementReferences();
```

apm-agent-plugins/apm-vertx/apm-vertx-4-plugin/src/main/java/co/elastic/apm/agent/vertx_4/VertxWebInstrumentation.java
```diff
@@ -0,0 +1,146 @@
...
+    public Collection<String> getInstrumentationGroupNames() {
+        return Arrays.asList("vertx", "vertx-web");
+    }
+
+    /**
+     * Instruments {@link io.vertx.core.impl.VertxFactory#createTracer}}
+     * <p>
+     * Instruments VertxFactory to return our own Tracer implementation.
+     */
+    public static class VertxFactoryInstrumentation extends VertxWebInstrumentation {
```
felixbarny, line 75: It looks like this instrumentation depends on being executed while Vert.x initializes and thus doesn't work with runtime attachment.
I have an idea for an alternative, though. You could instrument `io.vertx.core.impl.ContextInternal#tracer` and return a constant noop tracer if the application didn't register a tracer. Then, instrument implementations of `io.vertx.core.spi.tracing.VertxTracer`. Instead of relying on the propagation of the return value from `receiveRequest` to `sendResponse`, use the transaction object from the `Context` in `sendResponse` that you're already adding in `co.elastic.apm.agent.vertx_4.VertxWebHelper#startOrGetTransaction`.

As the `co.elastic.apm.agent.vertx_4.tracer.ElasticVertxTracer` doesn't activate the transaction, I suppose that you rely on `co.elastic.apm.agent.vertx_4.wrapper.RoutingContextHandlerWrapper` to activate the transaction when a route gets executed?

apm-agent-plugins/apm-vertx/apm-vertx-common/src/main/java/co/elastic/apm/agent/vertx/MultiMapHeadersGetterSetter.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public String getFirstHeader(String headerName, MultiMap carrier) {
+        return carrier.get(headerName);
+    }
+
+    @Override
+    public <S> void forEach(String headerName, MultiMap carrier, S state, HeaderConsumer<String, S> consumer) {
+        for(String headerValue : carrier.getAll(headerName)){
+            consumer.accept(headerValue, state);
+        }
```
felixbarny, line 44: To avoid allocating an iterator
```suggestion
        List<String> values = carrier.getAll(headerName);
        for (int i = 0, size = values.size(); i < size; i++) {
            consumer.accept(values.get(i), state);
        }
```

apm-agent-plugins/apm-vertx/apm-vertx-4-plugin/src/main/java/co/elastic/apm/agent/vertx_4/VertxEventLoopInstrumentation.java
```diff
@@ -0,0 +1,135 @@
...
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+
+public abstract class VertxEventLoopInstrumentation extends VertxWebInstrumentation {
```
felixbarny, line 43: This misses the case where `io.vertx.core.Context#runOnContext` or `io.vertx.core.Context#executeBlocking)` is called directly, as `VertxImpl` delegates to `Context`.

This can be reproduced in test when changing `routingContext.vertx().executeBlocking` to `routingContext.vertx().getOrCreateContext().executeBlocking`.

apm-agent-plugins/apm-vertx/apm-vertx-common/src/main/java/co/elastic/apm/agent/vertx/AbstractRoutingContextHandlerWrapper.java
```diff
@@ -0,0 +1,59 @@
...
+
+import co.elastic.apm.agent.impl.transaction.Transaction;
+import io.vertx.core.Handler;
+import io.vertx.ext.web.RoutingContext;
+
+public abstract class AbstractRoutingContextHandlerWrapper implements Handler<RoutingContext> {
+
+    private final Handler<RoutingContext> actualHandler;
+
+    protected AbstractRoutingContextHandlerWrapper(Handler<RoutingContext> handler) {
```
felixbarny, line 35: I think the way the routing context handlers are wrapped also doesn't work well with runtime attachment as the instrumentation is applied only when the routes are initialized.

A fairly easy workaround is to instrument `io.vertx.ext.web.impl.RouterImpl#handle` which is stable across versions. As an added benefit, the request is not wrapped, so you don't need to use magic to unwrap it.

apm-agent-plugins/apm-vertx/apm-vertx-4-plugin/src/main/java/co/elastic/apm/agent/vertx_4/VertxWebInstrumentation.java
AlexanderWert: Good point, resolved this as proposed.

apm-agent-plugins/apm-vertx/apm-vertx-common/src/main/java/co/elastic/apm/agent/vertx/MultiMapHeadersGetterSetter.java
AlexanderWert: Done.

apm-agent-plugins/apm-vertx/apm-vertx-4-plugin/src/main/java/co/elastic/apm/agent/vertx_4/VertxEventLoopInstrumentation.java
AlexanderWert: Resolved through instrumenting the Context directly instead of the VertxImpl

apm-agent-plugins/apm-vertx/apm-vertx-common/src/main/java/co/elastic/apm/agent/vertx/AbstractRoutingContextHandlerWrapper.java
AlexanderWert: io.vertx.ext.web.impl.RouterImpl#handle is a bad place, because there the RoutingContext does not exist yet. But resolved it through instrumenting RouteImpl#handleContext / RouteState#handleContext

apm-agent-plugins/apm-vertx/apm-vertx4-plugin/src/main/java/co/elastic/apm/agent/vertx/v4/web/NoopVertxTracer.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package co.elastic.apm.agent.vertx.v4.web;
+
+import io.vertx.core.Context;
+import io.vertx.core.spi.tracing.SpanKind;
+import io.vertx.core.spi.tracing.TagExtractor;
+import io.vertx.core.spi.tracing.VertxTracer;
+import io.vertx.core.tracing.TracingPolicy;
+
+class NoopVertxTracer implements VertxTracer<Object, Object> {
```
SylvainJuge, line 33: [minor] We could simply reuse the `io.vertx.core.spi.tracing.VertxTracer#NOOP` constant instead

apm-agent-plugins/apm-vertx/apm-vertx4-plugin/src/main/java/co/elastic/apm/agent/vertx/v4/web/WebInstrumentation.java
```diff
@@ -0,0 +1,278 @@
...
+    public static abstract class VertxTracerInstrumentation extends WebInstrumentation {
+
+        @Override
+        public final ElementMatcher<? super NamedElement> getTypeMatcherPreFilter() {
+            return nameContains("Tracer");
+        }
+
+        @Override
+        public final ElementMatcher<? super TypeDescription> getTypeMatcher() {
+            return hasSuperType(named("io.vertx.core.spi.tracing.VertxTracer"))
```
SylvainJuge, line 104: [question] Does it means we instrument any existing tracer if there is any and if there is none we inject our own no-op which comes not-so-no-op anymore once it's instrumented ?

SylvainJuge (approved):
LGTM

apm-agent-plugins/apm-vertx/apm-vertx-common/src/main/java/co/elastic/apm/agent/vertx/AbstractHttpTransactionHelper.java
```diff
@@ -0,0 +1,260 @@
...
+                }
+            }
+        }
+    }
+
+    protected boolean hasBody(@Nullable String contentTypeHeader, String method) {
+        return METHODS_WITH_BODY.contains(method) && contentTypeHeader != null;
+    }
+
+    public void applyDefaultTransactionName(String method, String pathFirstPart, @Nullable String pathSecondPart, Transaction transaction, int priorityOffset) {
```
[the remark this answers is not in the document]
AlexanderWert: 👍

apm-agent-plugins/apm-vertx/apm-vertx4-plugin/src/main/java/co/elastic/apm/agent/vertx/v4/web/NoopVertxTracer.java
AlexanderWert: Good point, missed there is a NOOP tracer.

apm-agent-plugins/apm-vertx/apm-vertx4-plugin/src/main/java/co/elastic/apm/agent/vertx/v4/web/WebInstrumentation.java
AlexanderWert: exactly

SylvainJuge: In that case if you have time to add a minor comment to explain that a bit and make it more obvious that would really be nice. I'm not sure that I'll remember that trick in a few months :-).

apm-agent-plugins/apm-vertx/apm-vertx4-plugin/src/main/java/co/elastic/apm/agent/vertx/v4/web/NoopVertxTracer.java
AlexanderWert: Tried to do the same with the `io.vertx.core.spi.tracing.VertxTracer#NOOP` tracer. Unfortunately, `NOOP` is returning null as the trace object so that the corresponding `sendResponse` method of the traces isn't called at all. 
--> We need to use our own NoopVertxTracer!

PR #1700, opened 2021-03-15
De-dot Micrometer metric names

felixbarny:
## What does this PR do?

This is a breaking change. However, it seems to be a necessary one:
The consequences could be quite severe
(mapping conflicts may only be fixable by deleting the index).
Also, the micrometer integration is still in beta so breaking changes
are probably ok.

The alternatives, such as flattened fields have downsides, such as
lack of Kibana support and worse performance.
The Elasticsearch team is not considering flattened fields the
endgame and are investigating other alternatives.

With that, I don't see the need for a config option to opt-out of the
de-doting yet.

relates to https://github.com/elastic/apm/issues/347

## Checklist

- [x] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-plugins/apm-micrometer-plugin/src/main/java/co/elastic/apm/agent/micrometer/MicrometerMeterRegistrySerializer.java
+++ b/apm-agent-plugins/apm-micrometer-plugin/src/main/java/co/elastic/apm/agent/micrometer/MicrometerMeterRegistrySerializer.java
@@ -96,26 +96,26 @@ public class MicrometerMeterRegistrySerializer {
                     Meter meter = meters.get(i);
                     if (meter instanceof Timer) {
                         Timer timer = (Timer) meter;
-                        hasValue = serializeTimer(jw, timer.getId(), timer.count(), timer.totalTime(TimeUnit.MICROSECONDS), hasValue);
+                        hasValue = serializeTimer(jw, timer.getId(), timer.count(), timer.totalTime(TimeUnit.MICROSECONDS), hasValue, replaceBuilder);
                     } else if (meter instanceof FunctionTimer) {
                         FunctionTimer timer = (FunctionTimer) meter;
-                        hasValue = serializeTimer(jw, timer.getId(), (long) timer.count(), timer.totalTime(TimeUnit.MICROSECONDS), hasValue);
+                        hasValue = serializeTimer(jw, timer.getId(), (long) timer.count(), timer.totalTime(TimeUnit.MICROSECONDS), hasValue, replaceBuilder);
                     } else if (meter instanceof LongTaskTimer) {
                         LongTaskTimer timer = (LongTaskTimer) meter;
-                        hasValue = serializeTimer(jw, timer.getId(), timer.activeTasks(), timer.duration(TimeUnit.MICROSECONDS), hasValue);
+                        hasValue = serializeTimer(jw, timer.getId(), timer.activeTasks(), timer.duration(TimeUnit.MICROSECONDS), hasValue, replaceBuilder);
                     } else if (meter instanceof DistributionSummary) {
                         DistributionSummary timer = (DistributionSummary) meter;
-                        hasValue = serializeDistributionSummary(jw, timer.getId(), timer.count(), timer.totalAmount(), hasValue);
+                        hasValue = serializeDistributionSummary(jw, timer.getId(), timer.count(), timer.totalAmount(), hasValue, replaceBuilder);
                     } else if (meter instanceof Gauge) {
                         Gauge gauge = (Gauge) meter;
-                        hasValue = serializeValue(gauge.getId(), gauge.value(), hasValue, jw);
+                        hasValue = serializeValue(gauge.getId(), gauge.value(), hasValue, jw, replaceBuilder);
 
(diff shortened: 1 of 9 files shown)


cyrille-leclerc:
Thanks @felixbarny, I'm excited to see this alignment across our products. Did you implement a configuration flag for backward compatibility?

felixbarny:
I've added a `dedot_custom_metrics` flag. The docs contain a warning that explains the potential for mapping conflicts. While we wouldn't recommend setting the flag to `false`, it may help to mitigate backward compatibility issues.

apm-agent-plugins/apm-micrometer-plugin/src/main/java/co/elastic/apm/agent/micrometer/MicrometerMeterRegistrySerializer.java
```diff
@@ -199,28 +210,40 @@ private static void serializeValue(Meter.Id id, String suffix, long value, JsonW
...
 
-    private static void serializeValueStart(String key, String suffix, JsonWriter jw) {
-        jw.writeByte(JsonWriter.QUOTE);
-        jw.writeAscii(key);
-        jw.writeAscii(suffix);
-        jw.writeByte(JsonWriter.QUOTE);
+    private static void serializeValueStart(String key, String suffix, JsonWriter jw, StringBuilder replaceBuilder, boolean dedotMetricName) {
+        replaceBuilder.setLength(0);
+        if (dedotMetricName) {
+            DslJsonSerializer.sanitizeLabelKey(key, replaceBuilder);
```
SylvainJuge, line 235: [minor] maybe rename method as it's now not only used for label keys, but more keys in general.

cyrille-leclerc:
Thanks @felixbarny . I'm not familiar enough with asciidoc and I didn't see `De-dotting be disabled via <<config-dedot-custom-metrics, ``dedot_custom_metrics``>>`

apm-agent-plugins/apm-micrometer-plugin/src/test/java/co/elastic/apm/agent/micrometer/MicrometerMetricsReporterTest.java
```diff
@@ -118,7 +119,32 @@ void testDisabledMetrics() {
         JsonNode metricSet = getSingleMetricSet();
         assertThat(metricSet.get("metricset").get("tags").get("foo").textValue()).isEqualTo("bar");
         assertThat(metricSet.get("metricset").get("samples")).hasSize(1);
-        assertThat(metricSet.get("metricset").get("samples").get("root.metric.include").get("value").doubleValue()).isEqualTo(42);
+        assertThat(metricSet.get("metricset").get("samples").get("root_metric_include").get("value").doubleValue()).isEqualTo(42);
+    }
+
+    @Test
+    void testDedotMetricName() {
+        meterRegistry.counter("foo.bar").increment(42);
```
SylvainJuge, line 127: [minor] maybe add extra assertion to ensure `MetricsConfiguration#isDedotCustomMetrics` returns true by default.

SylvainJuge (approved):

felixbarny:
@cyrille-leclerc The [comment from CI](https://github.com/elastic/apm-agent-java/pull/1700#issuecomment-799237388) contains a handy link to the docs preview: https://apm-agent-java_1700.docs-preview.app.elstc.co/guide/en/apm/agent/java/master/config-metrics.html

Screen Shot 2021-03-15 at 15 17 03

felixbarny:
@cyrille-leclerc: I think I misinterpreted your last statement. I just implemented that flag after you asked for it.

lucabelluccini:
Hello @felixbarny - I wanted to propose the feature flag but you did in the meanwhile 👏 

I would be tempted to switch its default to `false`: we do not have a breaking change and we keep getting mapping rejections only for users of such metrics which contain a `.` in the field name and for which they have the `scalar/object` problem.
If instead the way forward is de-dotting all the metrics in the future, then we can switch to `true`.

felixbarny:
Hey Luca, I was thinking of that but the current behavior (`dedot_custom_metrics=false`) is actually quite dangerous. It affects all Spring Boot users. By default, Spring Boot exports jdbc connection pool metrics that result in a mapping conflict. I think it's better to be safe than sorry and de-dot by default, which is what Metricbeat does, too.

I know that this is a breaking change and thus can be painful for some users. Those users can use the flag to opt-out (and live in constant danger of mapping conflicts). Note that the Micrometer integration is still in beta which is why I think introducing a breaking change is reasonable.

cyrille-leclerc:
Thanks @lucabelluccini . As @felixbarny explained, we looked at the pros and cons and we prefer to introduce this backward incompatible change on this beta version with a configuration param to go back to the previous mode: 
* Pros
   * Dedotted works out of the box with most popular use cases such as Spring Boot
   * Behaviour aligned with metricbeat 
* Cons
   * backward compatibility problem mitigated by the configuration flag to re enable the old behaviour. The beta phase is intended to support backward incompatible changes if we unfortunately have to introduce some.

lucabelluccini:
Totally agree and I didn't recall it was in `beta`. 👏 Thank you for addressing this!

PR #1671, opened 2021-02-19
fixes with tests on windows

kananindzya:
## Checklist

- [ ] This is an enhancement of existing features, or a new feature in existing plugins
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] I have made corresponding changes to the documentation
- [ ] This is a bugfix
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [ ] I have added tests that would fail without this fix
- [ ] This is a new plugin
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [ ] My code follows the [style guidelines of this project](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#java-language-formatting-guidelines)
  - [ ] I have made corresponding changes to the documentation
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#testing) pass locally with my changes
  - [ ] I have updated [supported-technologies.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/docs/supported-technologies.asciidoc)
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] Added an instrumentation plugin? Describe how you made sure that old, non-supported versions are not instrumented by accident.
- [ ] This is something else
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)

```
1. MetaDataTest#testCloudProvider_SingleProvider fails for AWS provider with:
2021-02-19 10:15:07,972 [elastic-apm-metadata] WARN  co.elastic.apm.agent.impl.CloudMetadataProvider - AWS is defined as the cloud_provider, but no metadata was found where expected
java.net.SocketException: Network is unreachable: connect
	at java.net.PlainSocketImpl.waitForConnect(Native Method) ~[?:?]
	at java.net.PlainSocketImpl.socketConnect(PlainSocketImpl.java:107) ~[?:?]
[...]

PR #1703, opened 2021-03-16
[CI] new stages for running the compile and test maven goals on Windows

v1v:
## What does this PR do?

Add a new stage in the CI pipeline to run the `build` and `test` maven goals on Windows.

This implementation won't increase the build time since it will run in parallel in the `Tests` meta-stage.


eyalkoren (commented):
@v1v this way everything that we split to steps on Linux happen as a single step on Windows, which means we lose parallelism and the builds will take much longer.
I thought of two alternatives:
1. Do the following stages for in parallel on Windows as we do for Linux: `Initializing`, `Checkout`, `Build` and `Tests` (same sub-stages)
2. Apply the same pipeline in parallel on a Windows worker

WDYT?

Jenkinsfile
```diff
@@ -31,6 +32,7 @@ pipeline {
     issueCommentTrigger('(?i)(.*(?:jenkins\\W+)?run\\W+(?:the\\W+)?(?:benchmark\\W+)?tests(?:\\W+please)?|/test).*')
   }
   parameters {
+    string(name: 'JAVA_VERSION', defaultValue: 'java11', description: 'What Java version?')
```
v1v, line 35: A parameter helper to set a different java version if required.

Jenkinsfile
```diff
@@ -16,6 +16,7 @@ pipeline {
     ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
     MAVEN_CONFIG = '-Dmaven.repo.local=.m2'
     OPBEANS_REPO = 'opbeans-java'
+    JAVA_VERSION = "${params.JAVA_VERSION}"
```
v1v, line 19: This is required when using params for the very first build, since the params is not loaded correctly for the very first build.

Jenkinsfile
```diff
@@ -44,7 +46,7 @@ pipeline {
       options { skipDefaultCheckout() }
       environment {
         HOME = "${env.WORKSPACE}"
-        JAVA_HOME = "${env.HUDSON_HOME}/.java/java11"
+        JAVA_HOME = "${env.HUDSON_HOME}/.java/${env.JAVA_VERSION}"
```
v1v, line 49: Avoid hardcoded values

v1v:
> @v1v this way everything that we split to steps on Linux happen as a single step on Windows, which means we lose parallelism and the builds will take much longer.

The new stage runs in parallel among the other stages. As far as I see, the longest stages are `Smoke Tests`, those take about 30 minutes.

IIUC, the linux build and test stages take about 8 + 15 minutes =  25 minutes. So with the new windows stage I'd expect an execution time of ~25 minutes that still is less that the smoke test execution times.

To illustrate what I mean, If I use [this](https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/job/master/804/) build from the master branch

```
Start - Checkout  - Build   - Tests ...
         (20 sec)  (7min)        | 
                                 |- Benchmarks      (8 min)
                                 |- Javadoc.        (6 min)
                                 |- Smoke Tests 01. (41 min)
                                 |- Smoke Tests 02. (37 min)
                                 |- Unit Tests      (15 min)
                                 |- Window           ** 23 min **
```

v1v:
Existing issues with the tests:
- temporary files cannot be deleted since they are used by another process
```
	Suppressed: java.nio.file.FileSystemException: C:\Users\JENKIN~1.PAC\AppData\Local\Temp\junit2172759383544760768\agent.json: The process cannot access the file because it is being used by another process.
```
- docker environment

```
 java.lang.IllegalStateException: Could not find a valid Docker environment. Please see logs and check configuration
```


@eyalkoren , do you happen to know if the above is something related to the project itself?

Jenkinsfile
```diff
@@ -142,6 +144,45 @@ pipeline {
             }
           }
         }
+        stage('Windows') {
```
v1v, line 147: This stage will run in parallel with the other ones. For clarity, I split it in two substages:
- build
- test

In order to help with compilation and test errors.

eyalkoren:
This is what happens in the Linux stages: 
- `Build`: `mvnw clean install -DskipTests=true -Dmaven.javadoc.skip=true` - compile and package artifacts, without running tests or creating javadoc
- `Tests`:
  - `Unit Tests`: `mvnw test` - run only unit tests based on the stashed state from the `Build` stage
  - `Smoke Tests 01`: run half of the integration tests
  - `Smoke Tests 02`: run the other half of the integration tests (theoretically, I think that in practice both steps run all integration tests, but we need to look into this)

When you run `mvnw clean install` you run all of the above sequentially in one stage. 
Your latest change separates the build and test stages, but they are still executed sequentially **and it will only run unit tests, not integration tests**.

So, back to what I suggested above: add `Windows Build` as a parallel stage to `Build` (which can now be called `Build Linux`) and add three parallel stages to `Tests` - `Windows Unit Tests`, `Windows Smoke Tests 1` and `Windows Smoke Tests 2`. 
I didn't expect unstashing on Windows a state that was stashed from a Linux stage would work, but as long as it works with the `source` stash, we can keep one `Checkout` stage. 
This is of course only a suggestion and you may find better alternative, as long as it doesn't take longer and all tests are executed. Also, it makes more sense the `Windows Build` is a stage of `Build` and not of `Tests`.

Lastly - don't worry about test failures at the moment. We don't expect them to pass until we merge #1671 . The current goal is to have this pipeline set up. We will worry about test failures later.

v1v:
Thanks Eyal for your details.

I won't run the `clean install` goal but split in two stages, similar to the one for the `linux`.

Let me clarify the reason for the existing multi-stage windows within the `tests` meta-stage:

1) The reason to run the build and test on windows in sequential stages is for one reason:
- decouple the build failures from test failures

2) The reason to run the windows meta-stage within the `Tests` stages is related to faster feedback. Let me explain

If we use the approach of a meta-build stage (for linux and windows) and a meta-tests stage as below:

```
  checkout - Build  -     Tests 
             |- linux       |- Unit test  (use the worker from the checkout)
             |- windows     |- smoke test 1
                            |- smoke test 2
                            |- Windows test
                            |- Benchmarks
                            |- Javadoc
```

There will be some implications:

a) windows build stage will require:
  - a new ephemeral windows worker (2 minute average)
  - stash the workspace (~1min and 30sec average)

Therefore the build time for the meta-build stage will be increased from 7 mins to 10 mins approx.

b) windows test stage will require:
  - a new ephemeral windows worker (2 minutes average)
  - unstash the workspace (30 seconds).


If we use the approach I'm suggesting then:

```
Start - Checkout  - Build   - Tests ...
         (20 sec)  (7min)        | 
                                 |- Benchmarks      (8 min)
                                 |- Javadoc.        (6 min)
                                 |- Smoke Tests 01. (41 min)
                                 |- Smoke Tests 02. (37 min)
                                 |- Unit Tests      (15 min)
                                 |- Window           ** 23 min **
                                      |- Build   (7 min).  - Unit Tests        (15 min)
```

There will be only one Windows request, aka only 2 minutes to be provisioned and the unstashed workspace will a few seconds.


> So, back to what I suggested above: add Windows Build as a parallel stage to Build (which can now be called Build Linux) and add three parallel stages to Tests - Windows Unit Tests, Windows Smoke Tests 1 and Windows Smoke Tests 2.

I was not aware that we needed Smoke tests too. We might need to set the requirements then I can work on iterations, if you don't mind :)

eyalkoren:
Just had a talk with @v1v, here's the summary:
We will keep this PR as is, as it provides 90% of the value we are looking for, which is being aware when new PRs break the build or unit tests on Windows, without extending the build time and without requiring too much resources. 
Integration tests are less likely to be affected frequently, but it would be nice to get those running on additional OS (and JVM implementations) occasionally, so for that we will have a separate pipeline running weekly, just doing `mvnw clean verify`.

cachedout:
@eyalkoren I think we got a bit stalled here. We have this in our 2-week iteration but it's `blocked` and I reading back through the discussion here, it looks like there's an agreement to merge this as it sits and then perhaps schedule a follow-up for the smoke tests?

If that aligns with your current understand, perhaps the next step would be to set an review to `Approved` (if you have no other issues) and we can get this merged and moving forward?  @v1v is on holidays this week but it would be great if we could get this in and ready for any follow-up work he might need to do on his return.

Thanks!

cc: @kseniia-kolpakova

SylvainJuge:
Update on the Java agent team: @jackshirazi made the tests work on Windows :tada: , thus the next step is ensuring that they are properly run into CI.
The current state is that docker-related tests are failing (https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/job/PR-1703/98/testReport/), this is the first test using docker, thus not specific to the test itself.
The current assumption is that the way that docker is installed on the Windows images is not compatible with our tests, which requires further investigation.

jackshirazi:
Turns out to be testcontainers don't yet support docker on Windows Server 2019 -  https://www.testcontainers.org/supported_docker_environment/windows/ 
issue is open at https://github.com/testcontainers/testcontainers-java/issues/2960

Jenkinsfile
```diff
@@ -143,6 +145,47 @@ pipeline {
...
+          agent { label 'windows-2019-docker-immutable' }
```
mdelapenya, line 149: @jackshirazi this is the label that we should update to `windows-10` once https://github.com/elastic/infra/pull/31021 is there.

As @v1v is coming back next week, we can discuss about the worker types then

cachedout: @jackshirazi and @eyalkoren 

It appears that [we can't proceed with Docker on Windows 10 as originally planned](https://github.com/elastic/infra/issues/31282#issuecomment-973112476). We'll need to consider other approaches.

jackshirazi: thanks. looks like we'll have to go with windows server and disable the testcontainer tests for windows. Will discuss this in the team

felixbarny:
Do we need both this PR and 
* https://github.com/elastic/apm-agent-java/pull/1705?

SylvainJuge:
FYI that will likely conflict with the changes I'm trying to add in #2400.

v1v:
> Do we need both this PR.

It depends, if it's needed to run on a PR basis or on a weekly basis. If both, then we might need both, if the latter then we can close this one in favour of the other one.

I don't recall the reason, but likely to illustrate that both requirements could be fulfil. 

What do we do?

SylvainJuge:
Yes, I think that both #1705 and this PR #1703 (or at least parts of each) would be required.
Ideally we need to have:
- the ability to run windows tests on a weekly basis to detect regressions in `master`
- the ability to opt-in windows tests on a PR basis


The way Windows tests are implemented is through executing `mvn clean install` in a separate step, so keeping it as an optional step controlled through a pipeline parameter seems a good option (and we can add enabling this on PRs with a label like with #2400).

In order to move this forward I think the next steps are:
- [ ] finish work on #2400 first as it will introduce conflicts
- [ ] update this PR after #2400 is merged, optionally enable windows steps through PR labels, and maybe move them in a parallel block next to JDK Compatibility tests (which are quite similar in nature: they both rebuild the project + execute unit tests with a different OS/JDK)
- [ ] merge this PR
- [ ] follow-up work on #1705 to ensure that windows tests are executed weekly, either through scheduling a single "test everything" job execution, or through dedicated jobs: one for JDK compatibility, one for Windows compatibility.

SylvainJuge:
Next steps before "ready to merge"
- [ ] update with `main`
- [ ] test optional windows steps activation through parameters (one-shot, not persistent)
- [ ] test optional windows steps activation through PR tags (persistent)
- [ ] some refactoring: format & cleanup, execute Windows build in parallel

Jenkinsfile
```diff
@@ -124,23 +130,58 @@ pipeline {
...
+          options { skipDefaultCheckout() }
+          environment {
+            JAVA_HOME = "C:\\Users\\jenkins\\.java\\${env.JAVA_VERSION}"
+            PATH = "${env.JAVA_HOME}\\bin;${env.PATH}"
+          }
+          when {
+            beforeAgent true
+            anyOf {
+              expression { return params.windows_ci }
```
v1v, line 156: This won't trigger the build in PRs or branches. Let's add the `branches` one to be mandatory, although the nightly builds/weekly could drive this too, what are you thoughts?

```suggestion
              expression { return params.windows_ci }
              not { changeRequest() }
```

SylvainJuge: We don´t need to have the Windows parts executed on every merge/run on `main` branch, thus here keeping only opt-in activation is fine.

The plan is to execute it with this parameter enabled through the weekly schedule in #1705 (which might still require some work due to the recent changes).

Jenkinsfile
```diff
@@ -92,13 +98,13 @@ pipeline {
       }
       environment {
         HOME = "${env.WORKSPACE}"
-        JAVA_HOME = "${env.HUDSON_HOME}/.java/java11"
+        JAVA_HOME = "${env.HUDSON_HOME}/.java/${env.JAVA_VERSION}"
         PATH = "${env.JAVA_HOME}/bin:${env.PATH}"
```
kuisathaverat, line 102: this path env var affects to a window stage, the `:` causes an issue

v1v:
Good catch Ivan! With all the refactoring, we missing that we cannot use the global env variable with the PATH in the meta-stage but in each individual stage. @SylvainJuge , that's the only way we know to support Windows and Linux, as I've seen you doing changes in this PR, do you want to do the change ? or should I do it?

SylvainJuge (approved):
I think that we are good to go here:
- the Windows build passes properly
- there Windows tests currently fail, but only due to a "proper test failure" (thus likely an agent/test bug on windows), we will deal with that with separate PRs.

Merging this will unblock #1705

PR #1704, opened 2021-03-17
Avoid runtime attach user conflicts

SylvainJuge:
## What does this PR do?

Append a per-user value to the unpacked jar file that is used for runtime attach
in order to prevent conflicts when distinct users try to attach with the same
agent version.

## Checklist

- [x] This is a bugfix


apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java
```diff
@@ -202,7 +202,12 @@ private static File getAgentJarFile() {
                     throw new IllegalStateException("Agent jar not found");
                 }
                 String hash = md5Hash(ElasticApmAttacher.class.getResourceAsStream("/elastic-apm-agent.jar"));
-                File tempAgentJar = new File(System.getProperty("java.io.tmpdir"), "elastic-apm-agent-" + hash + ".jar");
+
+                // we have to include current user name as multiple copies of the same agent could be attached
+                // to multiple JVMs, each running under a different user. Also, we have to make it path-friendly.
+                String user = md5Hash(System.getProperty("user.name"));
+
+                File tempAgentJar = new File(System.getProperty("java.io.tmpdir"), String.format("elastic-apm-agent-%s-%s.jar", user, hash));
```
felixbarny, line 210: Alternatively, when creating the temp file, could we set the file permissions to be more permissive?

felixbarny: See also `java.nio.file.attribute.PosixFilePermissions`

SylvainJuge: Yes, that could work here also. But it would create an implicit "shared ownership", thus it's probably better to keep them distinct because if temporary files created by one user get deleted it might have impacts on another process for another user.

felixbarny: The same can happen for different processes of the same user. In Unix file systems, it's fine to delete a file that's currently in use by another process because as long as the process has a file handle for that file, it won't actually be deleted.

felixbarny (approved):

PR #1715, opened 2021-03-25
log readonly msg errors only in debug

SylvainJuge:
## What does this PR do?

Some JMS implementations might use read-only messages during processing, which throws `MessageNotWriteableException` when the agent tries to write context-propagation headers to it.

This change allows to log those only at `debug` level, which leaves the default logging level uncluttered with those irrelevant log messages.

## Checklist

- [x] This is a bugfix


apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsMessagePropertyAccessor.java
```diff
@@ -78,6 +79,8 @@ public void setHeader(String headerName, String headerValue, Message message) {
         headerName = jmsifyHeaderName(headerName);
```
felixbarny, line 79: IIRC, @eyalkoren mentioned that in this specific issue, the header might have already been set.
By checking for the presence of the header, we may prevent the exception.

```suggestion
        headerName = jmsifyHeaderName(headerName);
        if (getFirstHeader(headerName, message) != null) {
            return;
        }
```

SylvainJuge: good suggestion, I've added a few test for that also.

eyalkoren (approved):

PR #1717, opened 2021-03-26
Fix thread pool method matchers

felixbarny:
There's not a reported problem with those but the matchers can miss methods with covariant return types


SylvainJuge:
I've opened https://github.com/felixbarny/apm-agent-java/pull/4 with a few other places where this might be an issue.

SylvainJuge (approved):

PR #1718, opened 2021-03-29
feat: use metricbeat to grab host metrics

kuisathaverat:
## What does this PR do?

It uses a new pipeline step to report host metrics using Metricbeat to our load-test ESS cluster. This helps to analyze the test results and the host metrics.

### DIFF
--- a/.ci/load/Jenkinsfile
+++ b/.ci/load/Jenkinsfile
@@ -197,8 +197,8 @@ pipeline {
 
                         // force kill any remaining app instance to ensure app can start
                         sh(script: ".ci/load/scripts/app.sh stopApp")
-
-                        script {
+                        metricbeat(es_secret: 'secret/apm-team/ci/load-test-cloud-elastic'){
+                          script {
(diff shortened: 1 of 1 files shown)


v1v (approved):

mdelapenya (approved):

SylvainJuge:
Thanks for working on this.

Where does the metrics are being sent ? Are they pushed to the same cluster we use for running the APM server that we use to send metrics to ? Having everything in the same place would allow to check everything worked as expected (which was part of the initial request).

kuisathaverat:
> Thanks for working on this.
> 
> Where does the metrics are being sent ? Are they pushed to the same cluster we use for running the APM server that we use to send metrics to ? Having everything in the same place would allow to check everything worked as expected (which was part of the initial request).

The metrics will send to the same cluster and accessible by checking the metrics plugins, so I think we can correlate the traces and the metrics. Probably we have to make some tunning on ECS values, we will see it after a few executions.

cachedout (approved):

SylvainJuge (approved):

PR #1730, opened 2021-04-01
Fix another JMS readonly exception

SylvainJuge:
## What does this PR do?

Improves previous fix #1715 as there was another direct writing to JMS message not covered.

## Checklist

- [x] This is a bugfix


apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelper.java
```diff
@@ -129,39 +131,35 @@ public Span startJmsSendSpan(Destination destination, Message message) {
...
-                span.getContext().getDestination().getService()
-                    .withName("jms")
-                    .withResource("jms")
-                    .withType(MESSAGING_TYPE);
-                if (destinationName != null) {
-                    span.getContext().getDestination().getService().getResource().append("/").append(destinationName);
-                    span.withName("JMS SEND to ");
-                    addDestinationDetails(null, destination, destinationName, span);
-                    if (isDestinationNameComputed) {
-                        message.setStringProperty(JMS_DESTINATION_NAME_PROPERTY, destinationName);
```
SylvainJuge, line 144: the offending writing call was right here.

apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsMessagePropertyAccessor.java
```diff
@@ -83,9 +83,9 @@ public void setHeader(String headerName, String headerValue, Message message) {
         try {
             message.setStringProperty(headerName, headerValue);
         } catch (MessageNotWriteableException e) {
-            logger.debug("Failed to set JMS message property due to read-only message", e);
+            logger.debug("Failed to set JMS message property {} due to read-only message", headerName, e);
         } catch (JMSException e) {
-            logger.warn("Failed to set JMS message property. Distributed tracing cannot work without that.");
+            logger.warn("Failed to set JMS message property {}.", headerName);
```
SylvainJuge, line 88: here the name of the property will make it obvious to know the case and if it's related to distributed tracing or not.
Now this is also used to add the destination name for the receiver side.

apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsInstrumentationHelper.java
```diff
@@ -49,6 +49,8 @@
 
 public class JmsInstrumentationHelper {
 
+    private static final JmsMessagePropertyAccessor MSG_ACCESSOR = JmsMessagePropertyAccessor.instance();
```
eyalkoren, line 52: [minor] It's already a singleton, I don't think it makes sense to hold a reference to it

apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsMessagePropertyAccessor.java
eyalkoren: It's not for us, it's for the user who doesn't necessarily knows this property is related to distributed tracing. In addition, this message is still valid - if we can't set a property, distributed tracing will not work. We can say "Distributed tracing _**may**_ not work..." instead.

eyalkoren (approved):

PR #1731, opened 2021-04-01
Adds circuit breaker to app deployment in load pipeline

cachedout:
## What does this PR do?

This adds a circuit breaker to the application deployment loop so that if no satisfactory response is returned after 300 iterations (5 minutes), the script fails and exits with a return code of 1. 

This exit code should then bubble up to the pipeline and cause it to terminate.

Closes https://github.com/elastic/observability-robots/issues/305


.ci/load/scripts/load_agent.sh
```diff
@@ -86,11 +87,15 @@ function appIsReady() {
 }
 
 function waitForApp() {
+    counter=1
     while :
     do
         if appIsReady; then 
             break 
+        elif [ $counter -gt $POLL_TIMEOUT ]; then
```
SylvainJuge, line 95: [minor] maybe echo a small message to explain why we are giving up

SylvainJuge (approved):

.ci/load/scripts/load_agent.sh
cachedout: Good suggestion. Thanks! I have added an explanatory message.

PR #1657, opened 2021-02-11
support Spring AMQP's SimpleMessageListenerContainer

kananindzya:
closes #1635 

## Checklist

- [X] This is an enhancement of existing features, or a new feature in existing plugins
  - [X] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [X] I have added tests that prove my fix is effective or that my feature works
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] I have made corresponding changes to the documentation
- [ ] This is a bugfix
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [ ] I have added tests that would fail without this fix
- [ ] This is a new plugin
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [ ] My code follows the [style guidelines of this project](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#java-language-formatting-guidelines)
  - [ ] I have made corresponding changes to the documentation
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#testing) pass locally with my changes
  - [ ] I have updated [supported-technologies.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/docs/supported-technologies.asciidoc)
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] Added an instrumentation plugin? Describe how you made sure that old, non-supported versions are not instrumented by accident.
- [ ] This is something else
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)

PR #1736, opened 2021-04-01
Loading Advice classes lazily when required

eyalkoren:
One step in the direction of complete separation between Advice and Instrumentation classes.
Next step is extract Advice code from all Instrumentations to different classes and it would also make sense to put all advices in separate source files as well.

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
@@ -125,7 +125,13 @@ public class ElasticApmAgent {
     private static final WeakConcurrentMap<Class<?>, Set<Collection<Class<? extends ElasticApmInstrumentation>>>> dynamicallyInstrumentedClasses = WeakMapSupplier.createMap();
     @Nullable
     private static File agentJarFile;
-    private static final Map<String, ClassLoader> pluginClassLoaderByAdviceClass = new ConcurrentHashMap<>();
+
+    /**
+     * A mapping from advice class name to the class loader that loaded the corresponding instrumentation.
+     * We need this in order to locate the advice class file. This implies that the advice class needs to be collocated
+     * with the corresponding instrumentation class.
+     */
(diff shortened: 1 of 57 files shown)


felixbarny (approved):
LGTM

felixbarny:
> and it would also make sense to put all advices in separate source files as well.

Personally, I kind of like instrumentations to be colocated with instrumentations so that it's clearer which method an advice instruments. Do you fear eager linking if it's a static inner class? I don't think it should be an issue as they're compiled into separate class files.

eyalkoren:
> Do you fear eager linking if it's a static inner class? I don't think it should be an issue as they're compiled into separate class files.

No, I agree, there shouldn't be any linkage error due to inner classes. This is just a matter of source code.
It is very misleading to place sources of two classes at the same file, where it is important to load them in different times and with different dependencies. I would even say the ideal encapsulation would be to separate them to two different modules, where the instrumentation class cannot have a dependency in the advice class and the library classes, but I proposed a middle-way that will not require a very tough refactoring. What we have with this PR is only the [javadoc warning](https://github.com/elastic/apm-agent-java/pull/1736/files#diff-4d130dca8badaf3c00f2282a5751067e55ee9c47f9154595f3733e316fb70175R148-R161), but I think it's much easier to miss improper dependencies this way. An example: https://github.com/elastic/apm-agent-java/pull/1657/commits/c86ecb758089722693e620f5cdb7a0c2c11b41ee

felixbarny:
> I think it should fix a problem reported as failing our AsyncServletTest when running on Windows

IINM, it turned out not to be a different linking behavior under Windows but a bug in the package scanning, which has been fixed via https://github.com/elastic/apm-agent-java/pull/1671/commits/c66bdd478847a4f2d276feccf0b95d5dc84a1ea7, right?

Not saying that this change isn't useful, just trying to connect the dots.

eyalkoren:
> IINM, it turned out not to be a different linking behavior under Windows but a bug in the package scanning, which has been fixed via c66bdd4, right?

Right.

PR #1737, opened 2021-04-01
Added Support for JBoss Logging to the log correlation plugin

### LINKED ISSUE #1734: Log correlation should support JBoss Logging
Wildfly uses JBoss Logging, which is not supported right now.

`MdcActivationListener` should support `org.jboss.logging.MDC`

tobiasstadler:
## What does this PR do?
Fixes #1734

## Checklist
  - [x] I have added info about JBoss logging to the [relevant section](https://www.elastic.co/guide/en/apm/agent/java/current/supported-technologies-details.html#supported-logging-frameworks) in the supported technologies page
- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -60,6 +60,7 @@ By default, the VM arguments are not printed, but only when the `-a`/`--list-vma
 Even when matching on the main class name or on system properties,
 ** Checks the Java version before attaching to avoid attachment on unsupported JVMs.
 * Cassandra instrumentation - {pull}1712[#1712]
+* Log correlation supports JBoss Logging - {pull}1737[#1737]
 
 [float]
 ===== Bug fixes
--- a/apm-agent-plugins/apm-log-correlation-plugin/pom.xml
+++ b/apm-agent-plugins/apm-log-correlation-plugin/pom.xml
@@ -36,6 +36,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging</artifactId>
+            <version>3.4.1.Final</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>apm-error-logging-plugin</artifactId>
             <version>${project.version}</version>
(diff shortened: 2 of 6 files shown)


SylvainJuge (approved):
LGTM, took the opportunity to make test slightly more generic.

eyalkoren:
@tobiasstadler Thanks! ❤️ 

I tested with `3.0.0.GA` and all works as expected (only requires switching to use `getMap().clear()` because the `clear()` API was not yet available) and added an entry to the [supported technologies page](https://www.elastic.co/guide/en/apm/agent/java/current/supported-technologies-details.html#supported-logging-frameworks) accordingly. Please let me know if you want to change something.

Also, I am not familiar enough with this logging library - can it be used in conjugation with slf4j or other? If so, did you verify it behaves as expected in such configuration?

tobiasstadler:
@eyalkoren Thank you for your support.

JBoss Logging itself is just a facade (similar to SLF4J) and needs a backend. The difference to SLFJ is, that some backends (JBoss LogManager, SLF4J, log4j, log4j2, java.until.logging) are packaged together with the api.

I tested it with JBoss LogManager. If the backend is SL4J (or another logging implementation supported by the agent) the MDC is set twice (once using the JBoss Logging api and once using the SLF4J api). but I don‘t think this is a problem as the values are the same. Using SLF4J together with the log4j2 backend should behave in the same way. Do you wan‘t me to test it?

I don‘t think using `getMap().clear()` is a good idea, as e.g. the SLF4J backend calls `org.slf4j. MDC.getCopyOfContextMap()`. AFAIK clearing this map does not affect the original MDC map.

eyalkoren:
> I tested it with JBoss LogManager. If the backend is SL4J (or another logging implementation supported by the agent) the MDC is set twice (once using the JBoss Logging api and once using the SLF4J api). but I don‘t think this is a problem as the values are the same. Using SLF4J together with the log4j2 backend should behave in the same way. Do you wan‘t me to test it?

What happens when it is set twice? If the second overrides the first, then we are OK. 

> I don‘t think using getMap().clear() is a good idea, as e.g. the SLF4J backend calls org.slf4j. MDC.getCopyOfContextMap(). AFAIK clearing this map does not affect the original MDC map.

I don't actually use it in our production code, it was only for testing 3.0.0.GA that couldn't compile with `clear()`. Since tests passed, I was good with that and didn't bother to look into the implementation. Now the tests are reverted to use `clear()` (which is apparently not required, but I don't mind leaving it).

tobiasstadler:
Yes, second one overwrites the first one.

PR #1746, opened 2021-04-06
Guarding from null host in Apache legacy HTTP client

eyalkoren:
## What does this PR do?
Fixing a reported issue:
```
java.lang.NullPointerException
at co.elastic.apm.agent.httpclient.LegacyApacheHttpClientInstrumentation$LegacyApacheHttpClientAdvice.onBeforeExecute(LegacyApacheHttpClientInstrumentation.java:69)
at org.apache.http.impl.client.DefaultRequestDirector.execute(DefaultRequestDirector.java:367)
at org.apache.http.impl.client.AbstractHttpClient.doExecute(AbstractHttpClient.java:835)
at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:83)
at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:108)
at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:56)
...
```

If the `HttpRequest` argument in the instrumented method is `null`, we should not get to the line producing the NPE, so I just added a null-check for the `host` argument. If it is `null`, we can still call `HttpClientHelper#startHttpClientSpan()` with a `null` hostname and it will take the hostname from the URI instead.

### DIFF
--- a/apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/LegacyApacheHttpClientInstrumentation.java
+++ b/apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/LegacyApacheHttpClientInstrumentation.java
@@ -56,7 +56,7 @@ public class LegacyApacheHttpClientInstrumentation extends BaseApacheHttpClientI
     public static class LegacyApacheHttpClientAdvice {
         @Nullable
         @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
-        public static Object onBeforeExecute(@Advice.Argument(0) HttpHost host,
+        public static Object onBeforeExecute(@Advice.Argument(0) @Nullable HttpHost host,
                                              @Advice.Argument(1) HttpRequest request) {
             final AbstractSpan<?> parent = tracer.getActive();
             if (parent == null) {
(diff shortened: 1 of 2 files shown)


apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/LegacyApacheHttpClientInstrumentation.java
```diff
@@ -56,7 +56,7 @@
     public static class LegacyApacheHttpClientAdvice {
         @Nullable
         @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
-        public static Object onBeforeExecute(@Advice.Argument(0) HttpHost host,
+        public static Object onBeforeExecute(@Advice.Argument(0) @Nullable HttpHost host,
                                              @Advice.Argument(1) HttpRequest request) {
```
SylvainJuge, line 60: why not annotate `HttpRequest` with `@Nullable` for consistency, while there is an `instanceof` check, it would make reading it slightly easier.

SylvainJuge (approved):

apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/LegacyApacheHttpClientInstrumentation.java
eyalkoren: Consistency shouldn't be a consideration here, only whether it can be `null`. Do you have a reason to think it may be?

SylvainJuge: Not really, let's keep it this way then.

PR #1747, opened 2021-04-06
CI: enable stable tag for successful builds in the CI

v1v:
## What does this PR do?

Tag the latest stable commit in the master branch.

## Why

This should reduce the instabilities in the internal CI instance when deploying the daily snapshots.


Jenkinsfile
```diff
@@ -282,6 +282,21 @@ pipeline {
...
+      when {
+        branch 'master'
+      }
+      steps {
+        deleteDir()
+        unstash 'source'
+        dir("${BASE_DIR}"){
+          setupAPMGitEmail(global: false)
+          sh(label: "checkout ${BRANCH_NAME} branch", script: "git checkout -f '${BRANCH_NAME}'")
+          gitCreateTag(tag: 'stable', tagArgs: '--force', pushArgs: '--force')
```
v1v, line 296: We use something similar for the apm-pipeline-library:

- https://github.com/elastic/apm-pipeline-library/blob/c8f9802cda896d92701a67504c66b0f26fda9169/.ci/Jenkinsfile#L229-L241

eyalkoren: So with `tagArgs: '--force', pushArgs: '--force'` it would remove the former `stable` tag and set a new one?

v1v: exactly, it does override the existing tag if exists.

```
       -f, --force
           Replace an existing tag with the given name (instead of failing)
```

We use the same approach to override the `current` tag in the jenkins shared library (this is the way we can deliver new feature without the need to raise PRs in each consumer)

Jenkinsfile
```diff
@@ -282,6 +282,21 @@ pipeline {
         githubNotify(context: "${env.GITHUB_CHECK_ITS_NAME}", description: "${env.GITHUB_CHECK_ITS_NAME} ...", status: 'PENDING', targetUrl: "${env.JENKINS_URL}search/?q=${env.ITS_PIPELINE.replaceAll('/','+')}")
       }
     }
+    stage('Stable') {
+      options { skipDefaultCheckout() }
+      when {
+        branch 'master'
```
eyalkoren, line 288: Let's remark this, just to test that the tag gets created.

Jenkinsfile
v1v: For instance:

image

Jenkinsfile
v1v: Do you mean to change the stage name?

```
stage('Create stable tag') {
```

v1v: Otherwise, if something bad happened, the build will fail

cachedout (approved):

mdelapenya (approved):

Jenkinsfile
eyalkoren: No, I meant to allow this to execute on any branch (like this one) to see that the tag is created and replaced properly after a commit

v1v: Gotcha, though, I'll add some changes and revert them afterwards

v1v: https://github.com/elastic/apm-agent-java/pull/1747/commits/afc5b7a9816d7c42f56cd16819b98c44beca118f is the commit with the changes:
- run immediately after the checkout
- run even on a PR basis.

v1v: Build [#3](https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/job/PR-1747/3/) created the tag:
- https://github.com/elastic/apm-agent-java/tree/stable

image

Build [#4](https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/job/PR-1747/4/) forced a retag

image


**NOTE**: bear in mind, the commit history might differ between branches and PRs, since there is no a rebase on branches but on PRs.

eyalkoren: OK, let's restore and see how it works. Thanks!

eyalkoren (approved):
Approving (the proper pipeline step 🙂 )

v1v:
I'll merge this PR, failures are not related to this change

PR #1748, opened 2021-04-06
Put memory limits on test docker containers

SylvainJuge:
## What does this PR do?

In attempt to reduce test flakyness, we should constraint docker containers available memory.
The values are arbitrary, and should be low enough to always fit our test infrastructure and development machines.

- When the limit is set too low, the container startup fails with `Container crashed with out-of-memory (OOMKilled)` message.
- When the limit is not set, the container might use the host system memory as "available memory", which is the case by default when running elasticsearch image that starts with close to 16gb heap on a 32gb system, even if less than 16gb is actually available when starting it.
- Good heuristic when running tests: total system memory usage should stay more or less flat and not jump from ~50% to 99% usage when running the `*IT` tests through `mvn verify`.

-----
Also attempts to minimize `Container startup failed` failures which seem related to a docker bug that is making most containers relying on log parsing to detect proper startup to fail.
- https://github.com/moby/moby/issues/41820
- related testcontainers issue https://github.com/testcontainers/testcontainers-java/issues/3613

Update: assertion on issues above have been removed, as it seems the outcome is still random and might still work as expected.
We work-around those for now by using an older CI image `debian-9`, this constraint could be removed once we have docker 20.10.6 available.

Update on Update: the `debian-9` tag has been removed, hopefully making it ready for when the related infra change is available, that allows to merge to master and already provide extra reliability.


eyalkoren (approved):
Nice!
Let's reduce logs verbosity as well 🙂 
Since you already went over all those tests using containers, can you rename all to end with the `IT` suffix? Who knows, maybe this convention can become handy if we decide not to always run these tests...

SylvainJuge:
We can't really rename them all with `IT` nor consider those using Testcontainers to be integration tests, as some of them are unit tests. For example, the RabbitMQ plugin is a good example of this where common instrumentation is tested with one RabbitMQ version and is then tested with multiple versions in integration tests.

However, I think once we have migrated to Junit5, using appropriate groups/tags to allow running subsets of tests would be relevant.

eyalkoren:
> For example, the RabbitMQ plugin is a good example of this where common instrumentation is tested with one RabbitMQ version and is then tested with multiple versions in integration tests.

I don't understand why. There is no rigid definition of what integration test is, but let's assume it means a test that spins and communicates with another process that runs some complex third party software (not mock APM server for that matter). Then if a test is doing that, it gets this suffix, otherwise it doesn't. In fact, the RabbitMQ test was lately renamed to reflect that, and its two subclasses are suffixed with IT as well.

PR #1758, opened 2021-04-14
Avoid stack frame re-computation + prevent indirect classloading

SylvainJuge:
## What does this PR do?

Attempt to avoid some thread locking issues that might be related to indirect classloading:

- avoid using classloading to resolve types when using runtime instrumentation
- avoid re-computing bytecode stack frames when not needed (depends on not-yet-released bytebuddy https://github.com/raphw/byte-buddy/commit/de70e4a0eb6530c0a3445b88ae1ac620d98e21e1).
## Checklist

- [x] This is a bugfix


SylvainJuge:
Currently pending on bytebuddy `1.11.0` release (relying on a locally built snapshot).

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/PatchBytecodeVersionTo51Transformer.java
```diff
@@ -48,12 +49,31 @@
...
                                      TypePool typePool, FieldList<FieldDescription.InDefinedShape> fieldList, MethodList<?> methodList, int writerFlags, int readerFlags) {
+
+                if (isAtLeastJava7(typeDescription)) {
+                    // in this case we won't prevent recomputing stack frames, but at least we avoid extra visitor wrapping
+                    // this happens for example with db2 JDBC driver
+                    return classVisitor;
+                }
+
                 return new ClassVisitor(Opcodes.ASM7, classVisitor) {
                     private boolean patchVersion;
```
eyalkoren, line 78: At this point we already know `version < Opcodes.V1_7`, so this flag and related checks are not required anymore in this `ClassVisitor`, right?

SylvainJuge: Good point, I've removed this extra check.

One very minor case might be when the class file version is not yet known when the `transform` method is executed but is known when the `AsmVisitorWrapper.wrap` method is being executed. That would be very unlikely though.

eyalkoren: I didn't get that - we are talking about a check within `wrap()` and then checks within the `visit` methods. Do we have a reason to believe the version we get in `wrap()` can somehow be different from the one the `visit()` method gets?

My actual suggestion was to keep your check and remove the ones in the `visit` methods

eyalkoren (approved):

PR #1763, opened 2021-04-15
Basic support for Apache Struts 2 (#1762)

### LINKED ISSUE #1762: (Basic) Support for Apache Struts 2
Apache Struts 2 is built on the Servlet API, so the transaction are named by the ServletApiAdvice.

The transactions should be named after the action that is called by the Struts framework, e.g MyAction#execute (similar to JAX-RS or Spring MVC)

tobiasstadler:
## What does this PR do?
Fixes #1762

## Checklist
- [ ] This is a bugfix
- [x] This is a new plugin
- [ ] This is something else


apm-agent-plugins/apm-struts-plugin/src/main/java/co/elastic/apm/agent/struts/Struts2TransactionNameInstrumentation.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static java.util.Collections.singletonList;
+import static net.bytebuddy.matcher.ElementMatchers.hasSuperType;
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public class Struts2TransactionNameInstrumentation extends ElasticApmInstrumentation {
+
+    @Override
+    public final ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return hasSuperType(named("com.opensymphony.xwork2.ActionProxy"));
```
SylvainJuge, line 42: `hasSuperType` is quite expensive matching, please override `getTypeMatcherPreFilter` to limit the set of classes that will be searched. If you look for other usages of this type matcher they almost always use the pre-filter.

apm-agent-plugins/pom.xml
```diff
@@ -58,6 +58,7 @@
         <module>apm-rabbitmq</module>
         <module>apm-okhttp-test</module>
         <module>apm-cassandra</module>
+        <module>apm-struts-plugin</module>
```
SylvainJuge, line 61: [question] would it make sense to have a separate module for struts1 and struts2 ? If so, should we rather have `struts2` in the module name ?

SylvainJuge (commented):
That's definitely awesome, thanks !

apm-agent-plugins/apm-struts-plugin/src/main/java/co/elastic/apm/agent/struts/Struts2TransactionNameInstrumentation.java
tobiasstadler: The default implementation is located in org.apache.struts2.factory, so I used that package as a filter. The downside ist that instrumenting custom implementations of the ActionProxy class (outside that package) doesn't work any longer. But as I don't think that feature is used that much, that is probably ok.

SylvainJuge: In that case, using a similar strategy like we have in `co.elastic.apm.agent.pluginapi.CaptureSpanInstrumentation#getTypeMatcher`, so you can include the application classes and the already known implementations, but instead do that in the `getTypeMatcherPreFilter` method.

```
    @Override
        return isInAnyPackage(config.getApplicationPackages(), ElementMatchers.<NamedElement>none())
            .and(not(isProxy()))
            .and(declaresMethod(getMethodMatcher()));
    }
```

tobiasstadler: so:
```
    @Override
        return named("org.apache.struts2.factory.StrutsActionProxy")
            .or(isInAnyPackage(config.getApplicationPackages(), none()));
    }

    @Override
    public final ElementMatcher<? super TypeDescription> getTypeMatcher() {
        return hasSuperType(named("com.opensymphony.xwork2.ActionProxy"));
    }
```
?

tobiasstadler: @SylvainJuge is the above ok?

felixbarny (approved):

apm-agent-plugins/apm-struts-plugin/src/main/java/co/elastic/apm/agent/struts/Struts2TransactionNameAdvice.java
```diff
@@ -0,0 +1,57 @@
...
+import co.elastic.apm.agent.impl.transaction.Transaction;
+import co.elastic.apm.agent.util.VersionUtils;
+import com.opensymphony.xwork2.ActionProxy;
+import net.bytebuddy.asm.Advice;
+
+import static co.elastic.apm.agent.impl.transaction.AbstractSpan.PRIO_HIGH_LEVEL_FRAMEWORK;
+
+public class Struts2TransactionNameAdvice {
+
+    private static final String FRAMEWORK_NAME = "Struts 2";
```
tobiasstadler, line 37: @SylvainJuge Should this be only `Struts` as one can find the version the the framework version field? What do you think?

SylvainJuge: Yes, `Struts` should be enough here, we have the full version number provided if we need to.

apm-agent-plugins/apm-struts-plugin/src/main/java/co/elastic/apm/agent/struts/Struts2TransactionNameInstrumentation.java
SylvainJuge: Sorry, I gave you to the wrong example here:

Doing something similar to 
`co.elastic.apm.agent.jaxrs.JaxRsTransactionNameInstrumentation#getTypeMatcherPreFilter` is a better strategy as it will work (albeit slow) if `application_packages` is not set.

Thus in your case it could be like this:

```
    @Override
        return named("org.apache.struts2.factory.StrutsActionProxy")
            .or(isInAnyPackage(config.getApplicationPackages(), any()));
    }

    @Override
    public final ElementMatcher<? super TypeDescription> getTypeMatcher() {
        return hasSuperType(named("com.opensymphony.xwork2.ActionProxy"));
    }
```

apm-agent-plugins/apm-struts-plugin/src/main/java/co/elastic/apm/agent/struts/Struts2TransactionNameInstrumentation.java
```diff
@@ -0,0 +1,76 @@
...
+    private final StacktraceConfiguration config;
+
+    public Struts2TransactionNameInstrumentation(ElasticApmTracer tracer) {
+        config = tracer.getConfig(StacktraceConfiguration.class);
+    }
+
+    @Override
+    public ElementMatcher<? super NamedElement> getTypeMatcherPreFilter() {
+        return named("org.apache.struts2.factory.StrutsActionProxy")
+            .or(isInAnyPackage(config.getApplicationPackages(), ElementMatchers.<NamedElement>any()));
```
felixbarny, line 54: That's a really expensive matcher. Especially because the application_packages setting is optional and this matcher runs even if struts is not on the classpath.

The pre-filter is intended to return a fast executing filter that may result in false negatives. For example `nameContains("ActionProxy")` limits the number of type descriptions Byte Buddy has to load from disk considerably and there's a good chance that custom implementations are still matched. There's a flag to turn off all pre-matchers in case they exclude classes that should be implemented.

tobiasstadler: I am fine with `nameContains("ActionProxy")`. Should I change it?

felixbarny: Let's see what @SylvainJuge says but I'd say yes.

SylvainJuge: Yes, I agree with @felixbarny here, we could simply have it with `nameContains("ActionProxy")` in `getTypeMatcherPreFilter` and avoid relying on `application_packages`:
- there are only a few cases where `StrutsActionProxy` class is extended (heuristic being search on public code samples in github).
- `application_packages` is often not set or set improperly, thus avoiding relying on it is better.

tobiasstadler: Done

SylvainJuge (approved):

tobiasstadler:
Thank You!

PR #1765, opened 2021-04-15
Added a integration test for Quarkus

tobiasstadler:
## What does this PR do?
I added a integration test for Quarkus with undertow as servlet container.

## Checklist
- [ ] This is a bugfix
- [ ] This is a new plugin
- [x] This is something else


integration-tests/quarkus/quarkus-jaxrs-base/src/test/java/co/elastic/apm/quarkus/jaxrs/AbstractQuarkusJaxRSTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    private static ConfigurationRegistry config;
+    private static MockReporter reporter;
+
+    @BeforeAll
+    public static void beforeAll() {
+        MockTracer.MockInstrumentationSetup mockInstrumentationSetup = MockTracer.createMockInstrumentationSetup();
+        config = mockInstrumentationSetup.getConfig();
+        when(config.getConfig(ReporterConfiguration.class).isReportSynchronously()).thenReturn(true);
```
eyalkoren, line 61: This is irrelevant when using the mock reporter

integration-tests/quarkus/quarkus-jaxrs-undertow/src/test/resources/application.properties
```diff
@@ -0,0 +1 @@
+quarkus.class-loading.parent-first-artifacts=org.ow2.asm:asm
```
eyalkoren, line 1: Is this required because this is tested as a unit test, where ASM is not relocated?
If you try to attach the built agent separately, would it work without this setting?
If so, I think it implies we need better use a real integration test if possible.

eyalkoren (commented):
Looks like a great foundation for Quarkus testing, thanks! Let's add it.

Given the way it is currently written, this module doesn't seem to fit in this integration tests module. All other tests in this location rely on"real" integration tests where a separate JVM is started to run the application and the agent artifact is attached to it. I suggest one of two options:
1. turn this into a real integration test (preferred)
2. move this to be a plugin (even though currently no additional instrumentation is required)

Based on my debugging, your app is traced thanks to our `FilterChain` instrumentation which captures the `io.undertow.servlet.handlers.FilterHandler.FilterChainImpl`.
IIUC, it can also run without any `Servlet` dependency, only relying on Vert.x. How difficult would it be to add a test for that, see that our existing Vert.x tracing capabilities?

And lastly, this tests our ability to capture traces. Are there any other Quarkus features/concepts we should trace? For example - is there any concept of registering tasks that may not be supported with our context propagation mechanism? Or are there special Quarkus clients that we can test?

tobiasstadler:
@eyalkoren AFAIK the Spring Boot ITs aren't executed in a separate JVM, but I will look how to execute the test in a separat JVM.

Actually I have an IT for the Vert.X integration, but the test fails. Quarkus uses `org.jboss.threads.EnhancedQueueExecutor` to dispatch the request to RESTEASY (see `io.quarkus.resteasy.runtime.standalone.VertxRequestHandler` and `io.quarkus.runtime.ExecutorRecorder#createExecutor`), but `org.jboss.threads.EnhancedQueueExecutor` is not instrumented by co.elastic.apm.agent.concurrent.ExecutorInstrumentation.

eyalkoren:
> Actually I have an IT for the Vert.X integration, but the test fails...

What happens if you enable the instrumentation of this executor by commenting out  https://github.com/elastic/apm-agent-java/blob/f9267ff7fc7df619cb2272b8790fde618055c8f6/apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/ExecutorInstrumentation.java#L87 ?

tobiasstadler:
The test passes, with:
```
diff --git a/apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/ExecutorInstrumentation.java b/apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/ExecutorInstrumentation.java
index 030bcb7c..4ac307e7 100644
--- a/apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/ExecutorInstrumentation.java
+++ b/apm-agent-plugins/apm-java-concurrent-plugin/src/main/java/co/elastic/apm/agent/concurrent/ExecutorInstrumentation.java
@@ -84,7 +84,6 @@ public abstract class ExecutorInstrumentation extends TracerAwareInstrumentation
             .and(not(named("org.apache.felix.resolver.ResolverImpl$DumbExecutor")))
             .and(not(nameContains("jetty")))
             .and(not(nameContains("tomcat")))
-            .and(not(nameContains("jboss")))
             .and(not(nameContains("undertow")))
             .and(not(nameContains("netty")))
             .and(not(nameContains("vertx")))
diff --git a/apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/agent/jaxrs/JaxRsTransactionNameInstrumentation.java b/apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/agent/jaxrs/JaxRsTransactionNameInstrumentation.java
index fe8103c7..0a57b04e 100644
--- a/apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/agent/jaxrs/JaxRsTransactionNameInstrumentation.java
+++ b/apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/agent/jaxrs/JaxRsTransactionNameInstrumentation.java
@@ -142,7 +142,7 @@ public class JaxRsTransactionNameInstrumentation extends TracerAwareInstrumentat
                         transactionName = pathAnnotationValue;
                     }
                 }
-                transaction.withName(transactionName, PRIO_HIGH_LEVEL_FRAMEWORK, false);
+                transaction.withName(transactionName, PRIO_HIGH_LEVEL_FRAMEWORK, true);
                 transaction.setFrameworkName("JAX-RS");
                 transaction.setFrameworkVersion(VersionUtils.getVersion(javax.ws.rs.GET.class, "javax.ws.rs", "javax.ws.rs-api"));
             }
```

I had to change `JaxRsTransactionNameInstrumentation` also, because both `JaxRsTransactionNameInstrumentation` and `co.elastic.apm.agent.vertx.AbstractVertxWebHelper` set the name of the transaction using `PRIO_HIGH_LEVEL_FRAMEWORK`.

eyalkoren:
Awesome, thanks! 
So let us see how we handle the excluded class. It's a similar problem to the one in https://github.com/elastic/apm-agent-java/pull/1920, although not sure we can/want-to treat both the same.

In the mean time, can you add this test, together with the changes that enabled it (but with a big all-capital TODO to change the `jboss` package commenting out)?

eyalkoren:
The JAX-RS transaction naming change fails a test in `JaxRsTransactionNameInstrumentationTest`. Conceptually, I think it's more correct to lower the priority used by the Vert.x instrumentation, as it is a lower-level framework.

integration-tests/quarkus/quarkus-jaxrs-undertow/src/test/resources/application.properties
tobiasstadler: yes, this was needed because ASM is not relocated

eyalkoren:
@tobiasstadler do you want to follow up on this?
I brought it up-to-date, but probably there are some things to fix and dependency versions to update. 
Let me know if you want to proceed with this.

tobiasstadler:
Sorry, but I don‘ want to continue. We can close it if you want.

JonasKunz:

integration-tests/quarkus/quarkus-jaxrs-base/src/main/resources/META-INF/beans.xml
SylvainJuge, line 1: [question] is that empty file required ? If yes than a small xml comment within it might help us not forget about it.

apm-agent-plugins/apm-vertx/apm-vertx-common/src/main/java/co/elastic/apm/agent/vertx/AbstractVertxWebHelper.java
```diff
@@ -78,7 +78,7 @@ protected void setRouteBasedTransactionName(Transaction<?> transaction, RoutingC
         if (!webConfiguration.isUsePathAsName()) {
             String path = routingContext.currentRoute().getPath();
             if (path != null) {
-                StringBuilder transactionName = transaction.getAndOverrideName(AbstractSpan.PRIO_HIGH_LEVEL_FRAMEWORK);
+                StringBuilder transactionName = transaction.getAndOverrideName(AbstractSpan.PRIO_HIGH_LEVEL_FRAMEWORK - 1);
```
SylvainJuge, line 81: [question] why did this priority needed to be changed ? Does it means without this the quarkus transaction name is not the one we expect ?

JonasKunz: This is required to make the instrumentation work correctly with RestEasy: RestEasy uses JAX-RS annotations, which we already instrument. In the case of quarkus it however collides with the Vert.x router instrumentation, because both have the same priority, but the router comes first.
I think it makes sense to lower the vert.x router prio in general, because it can be expected that the router is actually used a tool to implement higher-level frameworks (such as JAX-RS).

integration-tests/quarkus/quarkus-jaxrs-base/src/main/resources/META-INF/beans.xml
JonasKunz: I checked it, it indeed is required. Quarkus ignores the contents but looks for the presence of the file to perform the context setup.

CHANGELOG.asciidoc
```diff
@@ -41,6 +41,7 @@ communication - {pull}2996[#2996]
 * Add the <<config-long-field-max-length>> config to enable capturing larger values for specific fields - {pull}3027[#3027]
 * Provide fallback correlation when `ecs-logging-java` is used - {pull}3064[#3064]
 * Added separate Java 8 build with updated log4j2 - {pull}3076[#3076]
+* Added tests for Quarkus / RestEasy, adjusted vert.x router transaction name priority - {pull}1765[#1765]
```
SylvainJuge, line 44: This will need to be updated as there was a release since this PR was last updated.

SylvainJuge (approved):

JonasKunz:
Thanks @tobiasstadler for your contribution!
Closes #2981 .

JonasKunz:

PR #1777, opened 2021-04-20
Make non-default settings observable even in INFO

lucabelluccini:
## What does this PR do?

It is a proposal to log the settings which are not at default as INFO instead of DEBUG.

## Checklist

- [x] This is something else


eyalkoren:
That's a good idea! We work too hard to get real configurations sometimes. I don't think it should be too verbose.
@SylvainJuge any objection?

eyalkoren (approved):

SylvainJuge (approved):
That's definitely a very good idea ! :shipit:

PR #1780, opened 2021-04-22
Remove single-package plugin restriction

SylvainJuge:
## What does this PR do?

Currently, all plugins instrumentation and advice classes need to be stored within a single sub-package of `co.elastic.apm.agent`.

The current implementation has a few limitations:
- sharing classes between multiple modules requires to use the same package, hence limiting/complexifying naming (Hibernate search is a good example of that).
- migration to Indy dispatcher for plugins like `redis` or `elasticsearch` that have many variants and sub-packages is tedious because they have multiple packages levels.

### Proposal
- simplify the constraint for internal plugins: should only be sub-packages of `co.elastic.apm.agent`.
- load all classes within sub-packages in the same plugin classloader, for example all those will be loaded in the same
  - `co.elastic.apm.agent.aaa`
  - `co.elastic.apm.agent.aaa.bbb`
  - `co.elastic.apm.agent.aaa.ccc`
  
 This allows to have a "plugin root" package (here `aaa`) that contains common classes used by both `aaa.bbb` and `aaa.ccc`.

## Checklist

- [x] This is something else


felixbarny (approved):

felixbarny:
> Currently, all plugins instrumentation and advice classes need to be stored within a single sub-package of co.elastic.apm.agent.

For the record, that's not quite true: There can be any number of sub-packages, but the `ElasticApmInstrumentation` classes need to be in `co.elastic.apm.agent.${pluginPackage}` and must not be in a sub-package.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
```diff
@@ -370,12 +376,14 @@ public static ConstantCallSite bootstrap(MethodHandles.Lookup lookup,
...
-        pluginClasses = classesByPackage.get(packageName);
+        if (!adviceClassName.startsWith(EMBEDDED_PLUGINS_PACKAGE_PREFIX)) {
+            throw new IllegalArgumentException("invalid advice class location : " + adviceClassName);
+        }
+        String pluginPackage = adviceClassName.substring(0, adviceClassName.indexOf('.', EMBEDDED_PLUGINS_PACKAGE_PREFIX.length()));
+        List<String> pluginClasses = classesByPackage.get(pluginPackage);
         if (pluginClasses == null) {
-            classesByPackage.putIfAbsent(packageName, PackageScanner.getClassNames(packageName, adviceClassLoader));
-            pluginClasses = classesByPackage.get(packageName);
+            classesByPackage.putIfAbsent(pluginPackage, PackageScanner.getClassNames(pluginPackage, adviceClassLoader));
```
eyalkoren, line 385: Wouldn't this scan ALL agent classes?

felixbarny: I don't think so:

```java
String adviceClassName = "co.elastic.apm.agent.plugin.foo.PluginInstrumentation";
adviceClassName.substring(0, adviceClassName.indexOf('.', "co.elastic.apm.agent.".length()));
```

Returns `co.elastic.apm.agent.plugin`, so it would scan all classes in this package and sub-package.

eyalkoren: Ahh, of course! Read too quickly, missed the "from-index" search

PR #1785, opened 2021-04-25
relativize method adds / to path(ZipPath implementation of Path inter…

kananindzya:
## What does this PR do?
`relativize` method of `java.nio.file.Path` implementation `jdk.nio.zipfs.ZipPath` adds '/' :
```
        while (dotdots > 0) {
            result[pos++] = (byte)'.';
            result[pos++] = (byte)'.';
            if (pos < len)       // no tailing slash at the end
                result[pos++] = (byte)'/';
            dotdots--;
        }
```
So `replace` method not replaced '/' with '.' value. 
https://gist.github.com/kananindzya/f1ad4bfd978c9ed08436481b5ef86cb0
This bug reproduced only on windows, because on linux systems, `File.separatorChar` returns '/' value.

## Checklist

- [X] This is a bugfix
  - [X] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/util/PackageScanner.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/util/PackageScanner.java
@@ -91,7 +91,9 @@ public class PackageScanner {
             @Override
             public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
                 if (file.toString().endsWith(".class")) {
-                    classNames.add(basePackage + "." + basePath.relativize(file).toString().replace(File.separatorChar, '.').replace('/', '.').replace(".class", ""));
+                    // We need to escape both the filesystem-specific separator and the explicit `/` separator that may be added by the relativize() implementation
+                    String classNameSuffix = basePath.relativize(file).toString().replace(File.separatorChar, '.').replace('/', '.').replace(".class", "");
+                    classNames.add(basePackage + "." + classNameSuffix);
                 }
                 return FileVisitResult.CONTINUE;
             }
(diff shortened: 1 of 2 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/util/PackageScanner.java
```diff
@@ -91,7 +91,7 @@
             @Override
             public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
                 if (file.toString().endsWith(".class")) {
-                    classNames.add(basePackage + "." + basePath.relativize(file).toString().replace(File.separatorChar, '.').replace(".class", ""));
+                    classNames.add(basePackage + "." + basePath.relativize(file).toString().replace(File.separatorChar, '.').replace('/', '.').replace(".class", ""));
```
eyalkoren, line 94: Excellent. I would add a comment, so better add a variable for that. Something like:
```suggestion
                    // We need to escape both the filesystem-specific separator and the explicit `/` separator that may be added by the relativize() implementation
                    String classNameSuffix = basePath.relativize(file).toString().replace(File.separatorChar, '.').replace('/', '.').replace(".class", "");
                    classNames.add(basePackage + "." + classNameSuffix);
```

eyalkoren (changes requested):

eyalkoren (approved):

eyalkoren:
@kananindzya thanks a lot for this fix!!
When you build the agent and try it through `-javaagent` attachment on your Windows system, does it work as expected?

kananindzya:
> @kananindzya thanks a lot for this fix!!
> When you build the agent and try it through `-javaagent` attachment on your Windows system, does it work as expected?

Yes. after agent build - it works as expected, without errors. Details: 
https://gist.github.com/kananindzya/9d1ee0af08dd161f53e02becde25ff80

PR #1792, opened 2021-04-28
Remove gRPC Helper class from Instrumentation

SylvainJuge:
## What does this PR do?

Removes `GrpcHelper` class loading from `*Instrumentation` class in order to prevent eager-loading of gRPC types in the bootstrap classloader. In practice, it only worked thanks to lazy-loading of types that made `GrpcHelper` class being loaded in the proper class-loader.

## Checklist

- [x] This is something else

### DIFF
--- a/apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/ChannelInstrumentation.java
+++ b/apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/ChannelInstrumentation.java
@@ -34,15 +34,14 @@ import net.bytebuddy.description.NamedElement;
 import net.bytebuddy.description.method.MethodDescription;
 import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.matcher.ElementMatcher;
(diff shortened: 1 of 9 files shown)


eyalkoren (approved):

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/ChannelInstrumentation.java
```diff
@@ -75,15 +74,15 @@
     public static Object onEnter(@Advice.This Channel channel,
                                  @Advice.Argument(0) MethodDescriptor<?, ?> method) {
 
-        return helper.startSpan(tracer.getActive(), method, channel.authority());
+        return GrpcHelper.getInstance().startSpan(tracer.getActive(), method, channel.authority());
```
felixbarny, line 77: An alternative that also works nicely is to create a separate static inner class for the advice and to store the helper as in a private static final variable. That has two advantages: No risk of eagerly loaded method parameters (such as ClientCall) as the advice class will not be loaded when the method matchers execute, and avoiding a static method call to get the helper instance.

That's definitely not a blocker for this PR.

Issue #1617, opened 2021-01-14
[META 373] Extend ECS logging documentation 


AlexanderWert:
Extend the ECS logging documentation to cover ECS formatters replacement (if applicable) as described in https://github.com/elastic/apm/issues/373

eyalkoren:
We have it documented as the recommended way in the [Log Correlation guide](https://www.elastic.co/guide/en/apm/agent/java/current/log-correlation.html#log-correlation), documented in details in the [configuration section](https://www.elastic.co/guide/en/apm/agent/java/current/config-logging.html#config-log-ecs-reformatting) and mentioned in the list of [supported logging frameworks](https://www.elastic.co/guide/en/apm/agent/java/current/supported-technologies-details.html#supported-logging-frameworks).
Any suggestion where else we need to add something?

PR #1793, opened 2021-04-28
Add ECS-reformatting OVERRIDE

eyalkoren:
## What does this PR do?
Adds the ability to `OVERRIDE` original logs with ECS-formatted events. As opposed to `SHADE` and `REPLACE` options, this is not limited only to log files, but to additional logging options. Most importantly, it allows to reformat System.out logs to ECS, thus enabling native Docker/k8s support.

Mechanism specifications:
- Supports dynamic configuration of the `log_ecs_reformatting` config
- Guarantees that each log event is handled exactly once
- ECS shade/replace appenders should be created lazily only if and when configured, in order to avoid the creation of the `ecs.json` file where not required
- Advice calls from the logging framework should be very efficient, unless when the configuration changes and resources need to be created per appender.
- A config option allows to disable override/shade based on the underlying framework formatter type (where formatter is framework-specific, e.g `Layout` in log4j or `Encoder` in Logback)

## Checklist

- [x] I have added tests that prove the specifications above are maintained
- [x] I have tested manually with log4j1, log4j2 and Logback
- [x] I have implemented the new config option
- [x] I have updated the [Log Onboarding spec](https://github.com/elastic/apm/blob/master/specs/agents/log-onboarding.md) with required changes/additions (e.g. new config option).


apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2AppenderGetLayoutAdvice.java
```diff
@@ -0,0 +1,55 @@
...
+
+import javax.annotation.Nullable;
+
+public class Log4j2AppenderGetLayoutAdvice {
+
+    @SuppressWarnings({"unused"})
+    @Nullable
+    @AssignTo.Return
+    @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class, inline = false)
+    public static Layout<?> shadeLoggingEvent(@Advice.This(typing = Assigner.Typing.DYNAMIC) Appender thisAppender,
```
felixbarny, line 41: Consider having an allow list for Layouts that we override. For example, only `org.apache.logging.log4j.core.layout.PatternLayout` so that we don't override the layout when an appender has a limited set of layouts it's compatible with. For example, using the SyslogAppender with EcsLayout would not make sense and break things.

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
```diff
@@ -103,21 +111,46 @@ public void stopShading(A originalAppender) {
...
-    public boolean shouldSkipAppend(A appender) {
-        return loggingConfiguration.getLogEcsReformatting() == LogEcsReformatting.REPLACE && getOrCreateShadeAppenderFor(appender) != null;
+    public boolean onAppendEnter(A appender) {
+        if (callDepth.isNestedCallAndIncrement()) {
+            // If this is a nested call, never skip, as it means that the skipping decision was already made in the
+            // outermost invocation
+            return false;
+        }
+
+        A shadeAppender = getOrCreateShadeAppenderFor(appender);
```
felixbarny, line 131: This creates a shade appender even if the reformatting=off. Is that intended?

apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2LogShadingHelper.java
```diff
@@ -95,6 +73,41 @@ protected String getAppenderName(AbstractOutputStreamAppender<?> appender) {
             .setStackTraceAsArray(false)
             .build();
 
+        Appender ret = null;
+
+        if (isOverrideConfigured()) {
```
felixbarny, line 78: Is it an issue when the config is dynamically changed at runtime? The value returned by this method will still be the same cached value. For example, when changing from `override` to `replace`, the map will contain a `EcsLayoutHolder`.

apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2AppenderGetLayoutAdvice.java
eyalkoren: See the issue comment for possible enhancements. Don't you think it's more user-friendly to filter based on appender class name? I thought of adding such a config, but having per-framework default exclusions

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogShadingHelper.java
eyalkoren: See the issue comment for options of making it dynamic. It won't stay this way for sure. But I thought of an approach of creating always and look at configuration every time an append is called. The main problem with that may be that the file will be created already when creating and starting the appender. It's on my list to look into.

apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2LogShadingHelper.java
eyalkoren: Yes, the dynamic handling is not sorted out yet, it is on my todo list, including proper testing. I proposed two approaches for that at the initial comment, would love to get your input on that

apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2AppenderGetLayoutAdvice.java
eyalkoren: Agreed to add a config based on layout/encoder type

apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2AppenderGetLayoutAdvice.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.annotation.Nullable;
+import java.io.Serializable;
+
+/**
+ * The Log4j2 {@link Appender} does not expose a {@code setLayout()} API that allows us to override logging events.
+ * However, it exposes an {@link Appender#getLayout()} API <b>that is always used when events are logged</b>.
+ * Therefore, by instrumenting this method and replacing the returned {@link Layout}, we can implement the
+ * {@link co.elastic.apm.agent.logging.LogEcsReformatting#OVERRIDE OVERRIDE} use case.
+ */
+public class Log4j2AppenderGetLayoutAdvice {
```
felixbarny, line 42: I think we should agree on where to put advices. Either as static inner classes or as stand-alone classes.

I'd prefer the static inner class approach a lot. It makes instrumentations more cohesive as it's easier to grasp where an advice is applied to.

apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2AppenderGetLayoutAdvice.java
```diff
@@ -0,0 +1,61 @@
...
+    @AssignTo.Return
+    @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class, inline = false)
+    public static Layout<?> shadeLoggingEvent(@Advice.This(typing = Assigner.Typing.DYNAMIC) Appender thisAppender,
+                                              @Advice.Return @Nullable Layout<?> originalLayout) {
+
+        if (originalLayout == null) {
+            // Effectively disables instrumentation to all database appenders
+            return null;
+        }
+        Layout<? extends Serializable> ecsLayout = Log4J2EcsReformattingHelper.instance().getEcsOverridingFormatterFor(thisAppender);
```
felixbarny, line 55: Add a static field for the helper.

apm-agent-plugins/apm-log-shader-plugin/apm-logback-plugin/apm-logback-plugin-impl/src/main/java/co/elastic/apm/agent/logback/LogbackAppenderAppendAdvice.java
```diff
@@ -36,19 +35,19 @@
     @Advice.OnMethodEnter(suppress = Throwable.class, skipOn = Advice.OnNonDefaultValue.class, inline = false)
     public static boolean shadeAndSkipIfOverrideEnabled(@Advice.Argument(value = 0, typing = Assigner.Typing.DYNAMIC) final Object eventObject,
                                                         @Advice.This(typing = Assigner.Typing.DYNAMIC) OutputStreamAppender<ILoggingEvent> thisAppender) {
-        return thisAppender instanceof FileAppender && eventObject instanceof ILoggingEvent &&
-            LogbackLogShadingHelper.instance().shouldSkipAppend((FileAppender<ILoggingEvent>) thisAppender);
+
+        return eventObject instanceof ILoggingEvent && LogbackEcsReformattingHelper.instance().onAppendEnter(thisAppender);
```
felixbarny, line 39: It's a pitty that we can't use `@Local` in indy advices. Propagating the config value from the enter to the exit advice that way would be neat. But ofc. the ThreadLocal also works fine.

apm-agent-plugins/apm-log-shader-plugin/apm-logback-plugin/apm-logback-plugin-impl/src/main/java/co/elastic/apm/agent/logback/LogbackEcsReformattingHelper.java
```diff
@@ -0,0 +1,137 @@
...
+import ch.qos.logback.core.rolling.FixedWindowRollingPolicy;
+import ch.qos.logback.core.rolling.RollingFileAppender;
+import ch.qos.logback.core.rolling.SizeBasedTriggeringPolicy;
+import co.elastic.apm.agent.log.shader.AbstractEcsReformattingHelper;
+import co.elastic.apm.agent.log.shader.Utils;
+import co.elastic.logging.logback.EcsEncoder;
+
+import javax.annotation.Nullable;
+
+class LogbackEcsReformattingHelper extends AbstractEcsReformattingHelper<OutputStreamAppender<ILoggingEvent>, Encoder<ILoggingEvent>> {
```
felixbarny, line 41: I really like how straightforward the implementations now are.

felixbarny (approved):
LGTM. Just some minor stylistic comments.

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractEcsReformattingHelper.java
```diff
@@ -0,0 +1,447 @@
...
+    private static final Object NULL_FORMATTER = new Object();
+
+    private static final CallDepth callDepth = CallDepth.get(AbstractEcsReformattingHelper.class);
+
+    /**
+     * A mapping between original appender and the corresponding ECS-appender.
+     * Used when {@link LoggingConfiguration#logEcsReformatting log_ecs_reformatting} is set to
+     * {@link LogEcsReformatting#SHADE SHADE} or {@link LogEcsReformatting#REPLACE REPLACE}.
+     */
+    private static final WeakConcurrentMap<Object, Object> originalAppender2ecsAppender = WeakMapSupplier.createMap();
```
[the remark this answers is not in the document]
eyalkoren: A habit 🙂 
This guarantees there are no duplication, no matter where and how many helpers are created. Conceptually, everything here could be static, but with the maps this may actually have an effect if not.
Any reason to make them non-static? Are you worried from higher contention or something?

apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2AppenderGetLayoutAdvice.java
eyalkoren: I agree it's much more readable this way, the thing is that it's much easier to get it wrong and make the instrumentation independent on library types. I fixed quite a few of those lately, and I think a main reason for that is allowing something in the `import` section, that is not allowed somewhere in the class.

apm-agent-plugins/apm-log-shader-plugin/apm-logback-plugin/apm-logback-plugin-impl/src/main/java/co/elastic/apm/agent/logback/LogbackAppenderAppendAdvice.java
eyalkoren: It wouldn't be enough in this case though, because in log4j2 we need to use it in a separate advice as well

apm-agent-plugins/apm-log-shader-plugin/apm-logback-plugin/apm-logback-plugin-impl/src/main/java/co/elastic/apm/agent/logback/LogbackEcsReformattingHelper.java
eyalkoren: Thanks! 
The abstract helper contains some complexity, but not as bad as I feared and I am also satisfied with how library helpers ended up being all about library-specific-ECS-logging and the advices are minimal and almost identical.

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractEcsReformattingHelper.java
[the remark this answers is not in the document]
felixbarny: If they weren't static, we could use generics `WeakConcurrentMap<A, A>`/`WeakConcurrentMap<A, R>` which could make things a bit more readable. As we already have a static singleton instance, I was just a bit confused why the maps need to be shared across the different implementations.
That's not a blocker though.

apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2AppenderGetLayoutAdvice.java
felixbarny: (also applies to other similar places)

apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2AppenderGetLayoutAdvice.java
felixbarny: Are you thinking about automatically validating the imports? Or are you saying that it's just easier for the reviewer to verify that there are no invalid imports?

I just had an idea how we could verify that the instrumentation doesn't use invalid classes. In `AbstractInstrumentationTest`, we could create an isolated child class loader (parent=bootstrap) that loads each instrumentation class and calls the `get*Matcher` methods. That would result in linkage errors if library types are referenced.

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractEcsReformattingHelper.java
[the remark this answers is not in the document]
eyalkoren: It's a good point, but since I use dummy formatter and appender to optimize lookups that can't or shouldn't result with real implementations, switching to generic maps means adding the complication of creating these dummies in the helper subclasses. I like the fact that it is hidden from the sub-helpers, but I can try this out

apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2AppenderGetLayoutAdvice.java
eyalkoren: Sure

apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2AppenderGetLayoutAdvice.java
eyalkoren: > Are you thinking about automatically validating the imports? Or are you saying that it's just easier for the reviewer to verify that there are no invalid imports?

Auto-validating the imports means looking at source, so probably not. My experience shows that it's easier for the author to add a reference to something that is already at the imports in the IDE, and easier for the reviewer to miss.

> I just had an idea how we could verify that the instrumentation doesn't use invalid classes. In AbstractInstrumentationTest, we could create an isolated child class loader (parent=bootstrap) that loads each instrumentation class and calls the get*Matcher methods. That would result in linkage errors if library types are referenced.

Good idea! You mean a separate URL CL that knows the agent jar and is the child of the boot CL (since it's a unit test, we cannot rely on the boot CL to load agent classes)? 
The only thing is that we will still rely only on tested JVMs, and may be exposed to differences in linkage in non tested ones, but should capture such issues very quickly in most cases.

AlexanderWert:
Closes https://github.com/elastic/apm-agent-java/issues/1617

PR #1795, opened 2021-04-28
Es client scenario cancellable query

### LINKED ISSUE #1686: Add a ES-client test scenario of cancelled query
We need to add a test for a long-executing ES client query that is cancelled through the relatively new `org.elasticsearch.client.Cancellable` API to make sure such spans are ended properly. If not, we should add the proper instrumentation

kananindzya:
## What does this PR do?
closes  #1686

- [x] This is something else


eyalkoren (approved):
@kananindzya thanks for adding this test!
I modified the test a bit, but more importantly, after discussing it internally quite a bit, we decided that the right outcome for this kind of spans is `UNKNOW`, so I added a small bit to the actual instrumentation. 
Currently, it is set to `FAILURE` because when you cancel, the `ActionListener#onFailure()` is invoked.

PR #1808, opened 2021-05-10
Avoid tracestate error with sample rate provided through headers

SylvainJuge:
## What does this PR do?

When using OpenTracing API, there are some code paths that could lead to setting transaction sample rate multiple times.

Because we try to minimize the amount of parsing & memory allocation involved, we only allowed to set the sample rate once using either:
- `co.elastic.apm.agent.impl.transaction.TraceState#set` for root transaction with value from sampler
- `co.elastic.apm.agent.impl.transaction.TraceState#addTextHeader` for headers parsing

This PR changes the existing behavior (throw an exception) to:
- ignore when trying to set value more than once with an identical value.
- throw an exception only when trying to use a different value.
- only try to set sample rate from sampler only when it's not already set.

## Checklist

- [x] This is a bugfix

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
@@ -427,7 +427,12 @@ public class TraceContext implements Recyclable {
         transactionId.copyFrom(id);
         if (sampler.isSampled(traceId)) {
             flags = FLAG_RECORDED;
-            traceState.set(sampler.getSampleRate(), sampler.getTraceStateHeader());
+
+            if (Double.isNaN(traceState.getSampleRate())) {
+                // sample rate could already have been set through headers
+                // we have to make sure to only use sampler rate as fallback only.
+                traceState.set(sampler.getSampleRate(), sampler.getTraceStateHeader());
+            }
         }
         clock.init();
         onMutation();
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
@@ -427,12 +427,7 @@ public class TraceContext implements Recyclable {
         transactionId.copyFrom(id);
         if (sampler.isSampled(traceId)) {
             flags = FLAG_RECORDED;
-
-            if (Double.isNaN(traceState.getSampleRate())) {
-                // sample rate could already have been set through headers
-                // we have to make sure to only use sampler rate as fallback only.
-                traceState.set(sampler.getSampleRate(), sampler.getTraceStateHeader());
-            }
+            traceState.set(sampler.getSampleRate(), sampler.getTraceStateHeader());
         }
         clock.init();
         onMutation();
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceState.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceState.java
@@ -109,7 +109,7 @@ public class TraceState implements Recyclable {
                             // we don't minimize allocation as re-writing should be an exception
                             headerValue = rewriteBuffer.toString();
                         }
-                        if (checkSampleRate(rounded)) {
+                        if (updateSampleRateCheck(rounded)) {
                             sampleRate = rounded;
                         }
                     }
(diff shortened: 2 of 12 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
```diff
@@ -427,7 +427,12 @@ public void asRootSpan(Sampler sampler) {
         transactionId.copyFrom(id);
         if (sampler.isSampled(traceId)) {
             flags = FLAG_RECORDED;
-            traceState.set(sampler.getSampleRate(), sampler.getTraceStateHeader());
+
+            if (Double.isNaN(traceState.getSampleRate())) {
```
eyalkoren, line 431: I would still do this check inside `TraceState#set()`. In fact, it already does it. Since you do this check here, you won't get the behaviour you want, e.g. throw an Exception when trying to override with a different value, or am I missing something?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
```diff
@@ -427,7 +427,12 @@ public void asRootSpan(Sampler sampler) {
...
+                // sample rate could already have been set through headers
+                // we have to make sure to only use sampler rate as fallback only.
```
eyalkoren, line 433: How come it could have been set through headers? Isn't `asRootSpan()` only invoked when no trace context headers are received (i.e. root)?

eyalkoren (commented):
While this seems OK, I need to understand what we are missing here - the assumption is that `TraceState#set()` can only be invoked if a the transaction is started as a root, meaning - without `traceparent` header. And as far as I can see, `addTextHeader ()` should only be invoked if there is a valid `traceparent` header. Any attempt to create a child transaction when there is already an active transaction should raise a red flag, right?
As for multiple attempts to set the sample rate based on the `tracestate` header - I can see how this may happen, and we should guard from that and treat it as illegal state or something.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
SylvainJuge: Yes, you are right, this extra check just avoids one possible buggy case by just attempting to ignore setting sample rate if it's already set. Thus removing it here (and only keep the one within `TraceState`) makes sense.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
SylvainJuge: 1. yes, first wins is a sensible strategy here.
2. logging with `debug` is fine as it's an exceptional case that would only be investigated.
3. yes, we should only propagate valid header and de-duplicate.

Even if we should not call `TraceState#set` twice, having a consistent behavior with `TraceState.addTextHeader()` would be nice, thus logging both with `debug` seems the right option.

SylvainJuge:
After discussing this with @eyalkoren this morning, we agreed on the following behavior:
- trying to call `TraceState#set` multiple times is an error and likely a bug in the agent, we should throw an exception
- trying to call `TraceState#set` once a valid header has been parsed is also an error and should throw an exception
- multiple calls to `TraceState#addTextHeader` are expected due to multi-valued headers (valid for HTTP).
- if there are multiple values of sample rate in header, then we only take the 1st and ignore others but issue a warning in log (1).
- if there is any invalid header value, we rewrite the header to avoid propagating an invalid header
- if there is any rounding difference, then we also rewrite the header

(1) while the agent just corrects what could be an invalid upstream tracestate header, we have to log it as it could be an indicator of another bug with an upstream service also instrumented by one of our agents.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceState.java
```diff
@@ -72,68 +76,106 @@ public void copyFrom(TraceState other) {
...
+            // ensure proper rounding of sample rate to minimize storage
+            // even if configuration should not allow this, any upstream value might require rounding
+            double rounded = DOUBLE_CONVERTER.round(value);
+            if (rounded != value) {
+                // value needs to be re-written due to rounding
+                headerValue = rewriteRoundedHeader(headerValue, valueStart, valueEnd, rounded);
+                sampleRate = rounded;
+            } else if (!Double.isNaN(sampleRate)) {
+                log.warn("sample rate already set to {}, trying to set it to {} through header will be ignored", sampleRate, value);
+                headerValue = rewriteRemoveInvalidHeader(headerValue, vendorStart, vendorEnd);
```
eyalkoren, line 138: Logically, shouldn't we check that as early as possible? The line above suggests we may have overridden the sample-rate already.

eyalkoren (approved):
Small comment/suggestion,  but other than that LGTM

PR #1811, opened 2021-05-11
Fix vertx test for 4.x

SylvainJuge:
## What does this PR do?

Fixes Vert.x tests that rely on exact version matching.

## Checklist

- [x] This is something else


apm-agent-plugins/apm-vertx/apm-vertx3-plugin/src/test/java/co/elastic/apm/agent/vertx/v3/VertxServerTest.java
```diff
@@ -50,7 +50,7 @@ protected boolean useSSL() {
     }
 
     @Override
-    protected String expectedVertxVersion() {
-        return "3.6";
+    protected int getMajorVersion() {
+        return 3;
```
AlexanderWert, line 54: so, we would not check for the minor version in this case?
just wondering if this may become a problem. But actually shouldn't be

AlexanderWert (approved):
LGTM, I'm fine with skipping testing the minor version match

PR #1812, opened 2021-05-12
Shared JSON spec for span types/subtypes

SylvainJuge:
## Description

Creates and enforces JSON spec for all the span types/subtypes that are used in the Java agent

This is a preliminary step to having a shared spec applied to all APM agents.

Related PR for all agents : https://github.com/elastic/apm/pull/443

## Checklist

- [x] gather the list of all values currently reported by tests for the Java agent to make test pass.
- [x] decide how to handle values for test only
  - sharing those across agent does not make sense
  - applying a wildcard or prefix-based approach hardcoded before the check could help (for example do not validate every `foo.*` type.
  - Solution: Allow manual opt-out on a per-test-method basis
- [x] decide how to handle `null` (not provided) or `""` (empty) sub-types:
  - should we make them optional on some known types ? making them optional on every type ?
  - ~~Solution (1): normalize to `null` instead of empty and treat `sub-type` as optional~~
  - Solution (2): normalize to `null` instead of empty and allow `null` on some types (not all of them).
- [x] decide how to handle "open ended" span subtypes:
  - `process` type that uses binary name as sub-type (or find a proper way to fix it).
  - Solution: remove `subtype` that was using binary name, it's already provided in `name`
  - `template` type that uses template engine name as sub-type (and is also very platform-specific)
  - Solution: enforce only the `type` and not `subtype` as it's really platform-specific
 - [x] decide what to do with OpenTracing & OpenTelemetry bridges
   - remap `cache.redis` to `db.redis` for consistency
   - use `custom` as default span type instead of `unknown`

### DIFF
--- a/apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
+++ b/apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
@@ -219,35 +219,37 @@ public class MockReporter implements Reporter {
         if (checkStrictSpanType) {
             JsonNode typeJson = getMandatoryJson(SPAN_TYPES_SPEC, type, String.format("span type '%s' is not allowed by the spec", type));
 
-            String typeComment = getOptionalComment(typeJson);
-
-            JsonNode jsonOptionalSubtype = typeJson.get("optional_subtype");
-            boolean optionalSubtype = jsonOptionalSubtype != null && jsonOptionalSubtype.isBoolean() && jsonOptionalSubtype.asBoolean();
+            boolean allowNullSubtype = getBooleanJson(typeJson, "allow_null_subtype");
+            boolean allowUnlistedSubtype = getBooleanJson(typeJson, "allow_unlisted_subtype");
 
             String subType = span.getSubtype();
-
-            JsonNode subTypesJson = typeJson.get("subtypes");
-
-            if (subTypesJson == null || subTypesJson.isEmpty()) {
-                assertThat(subType)
-                    .describedAs("span type '%s' does not allow for subtypes", type)
-                    .isNull();
-            } else if (subType == null) {
-                // span does not have a sub-type, make sure that it's only when spec allows for it
-                assertThat(optionalSubtype)
-                    .describedAs("span type '%s' subtype is not optional by the spec (optional_subtype=false)", type)
+            if (null == subType) {
+                assertThat(allowNullSubtype)
+                    .describedAs("span type '%s' requires non-null subtype (allow_null_subtype=false)")
                     .isTrue();
             } else {
-                assertThat(subType)
-                    .describedAs("span subtype is required by the spec for type '%s'", type)
-                    .isNotNull();
-
-                // we have a sub-type, make sure that the sub-type matches the spec
-                getMandatoryJson(subTypesJson, subType, String.format("span subtype '%s' is now allowed by the spec for type '%s' (%s)", subType, type, typeComment));
+                if (!allowUnlistedSubtype) {
+                    JsonNode subTypesJson = typeJson.get("subtypes");
+                    if (subTypesJson != null && !subTypesJson.isEmpty()) {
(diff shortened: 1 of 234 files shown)


SylvainJuge:
See https://github.com/elastic/apm/pull/443 PR for adding this to shared specifications.

apm-agent-plugins/apm-process-plugin/src/main/java/co/elastic/apm/agent/process/ProcessHelper.java
```diff
@@ -72,8 +72,6 @@ void doStartProcess(@Nonnull AbstractSpan<?> parentContext, @Nonnull Process pro
 
         Span span = parentContext.createSpan()
             .withType("process")
-            .withSubtype(binaryName)
-            .withAction("execute")
```
eyalkoren, line 76: Why? Risk of too high cardinality?

eyalkoren (commented):
LGTM, few minor comments and one question about the spec itself - for a given type, is `null` subtype allowed when the spec defines a list of subtypes?

apm-agent-plugins/apm-process-plugin/src/main/java/co/elastic/apm/agent/process/ProcessHelper.java
SylvainJuge: Yes, while the total number of executables would still be quite limited in practice, there is no extra value to have this as sub-type as everything is already included in the name. Also, there would be no other action but `execute`.

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
```diff
@@ -164,27 +189,73 @@ public synchronized void report(Span span) {
...
+            .describedAs("span type is mandatory")
+            .isNotNull();
+
+        if (checkStrictSpanType) {
+            JsonNode typeJson = getMandatoryJson(SPAN_TYPES_SPEC, type, String.format("span type '%s' is not allowed by the spec", type));
+
+            String typeComment = getOptionalComment(typeJson);
+
+            JsonNode jsonOptionalSubtype = typeJson.get("optional_subtype");
+            boolean optionalSubtype = jsonOptionalSubtype != null && jsonOptionalSubtype.isBoolean() && jsonOptionalSubtype.asBoolean();
```
eyalkoren, line 225: `jsonOptionalSubtype.isBoolean()` should be asserted, not used as a valid initialization for `optionalSubtype`. 
In other words, `optional_subtype` should be either `null` or a string that can be parsed into a valid `boolean`, otherwise this should probably fail.

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
```diff
@@ -164,27 +189,73 @@ public synchronized void report(Span span) {
...
+
+            String subType = span.getSubtype();
+
+            JsonNode subTypesJson = typeJson.get("subtypes");
+
+            if (subTypesJson == null) {
+                assertThat(subType)
+                    .describedAs("span type '%s' does not allow for subtypes", type)
+                    .isNull();
```
eyalkoren, line 234: Unless `optionalSubtype == true`, right?

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
```diff
@@ -164,27 +189,73 @@ public synchronized void report(Span span) {
...
+            } else if (subType == null) {
+                // span does not have a sub-type, make sure that it's only when spec allows for it
+                assertThat(optionalSubtype)
+                    .describedAs("span type '%s' subtype is not optional by the spec (optional_subtype=false)", type)
+                    .isTrue();
+            } else {
+                assertThat(subType)
+                    .describedAs("span subtype is required by the spec for type '%s'", type)
+                    .isNotNull();
```
eyalkoren, line 243: [minor] This assertion is redundant (`subType` is not `null` here)

eyalkoren (approved):
LGTM! One minor suggestion

PR #1815, opened 2021-05-16
Basic support for remote EJB invocations on a WildFly application server

### LINKED ISSUE #1814: Support for remote EJB invocation on WildFly
We have some applications that communicate via remote EJB invocations.

It would be nice if an invocation of a remote EJB creates a span on the client and a child transaction on the server.

None.

## Additional context
We are using WildFly, so support for WildFly is sufficient for us.

tobiasstadler:
## What does this PR do?
Fixes #1814 for WildFly

## Checklist
- [ ] This is a bugfix
- [x] This is a new plugin
- [ ] This is something else


tobiasstadler:

tobiasstadler:
@SylvainJuge May you could have a look please?

tobiasstadler:
@felixbarny Maybe you could have a look please?

SylvainJuge:
Hi @tobiasstadler ,

Supporting EJBs is not in our short-term roadmap for now, thus we don´t think that it would be beneficial to just add partial support for a single server implementation. From experience instrumenting EJBs is full of caveats and corner cases, thus it's definitely a huge development effort to commit to this task.

Few things on top of my mind:
- EJB2 and EJB3 support
- EJBs over RMI : context propagation is definitely tricky and might depend on the wire protocol
- plenty of application servers implementations to test : Jboss, Wildfly, Websphere, Weblogic, ...

However, since it already works for your use-case, it would be probably a very good candidate for an external plugin, which means it won't be part of the agent main codebase, but could be loaded at runtime to enable it on your setup.
There are a few things to consider here:
- modifications required into the `core` part of the agent would still be required (there are a few of them)
- building an external plugin is not yet well documented, but we have an example in [integration-tests/external-plugin-test](https://github.com/elastic/apm-agent-java/tree/master/integration-tests/external-plugin-test)

What do you think about this ?

tobiasstadler:
I am totally fine with building an external plugin (actually I already did this for something else). The only problem is that I need to instrument `org.jboss.as.ejb3.remote.AssociationImpl`. This is currently not possible because `org.jboss.as` is excluded in `classes_excluded_from_instrumentation_default` (Of course I could change the setting, but this negatively affects maintainability and performance). I implemented `CustomElementMatchers# anyMatchExcept` as workaround. Should I create another PR with this change only?

tobiasstadler:
@SylvainJuge I converted my code to an external plugin locally and it seems to work except that I can not set the framework name using the public api (not so important). Also I am unsure if `span.setDestinationService/setDestinationAddress` is equivalent to `span.getContext().getDestination(). withAddress/withPort/withResource/withName`

SylvainJuge:
Thanks !

FYI I'll be off for a couple of weeks, thus I'll likely handle that when I get back, or @eyalkoren might also take a look at it (but no guarantee).

tobiasstadler:
I will create an external Plugin.

Thank you for your support!

tobiasstadler:
FYI, you can find the plugin here: https://github.com/tobiasstadler/apm-wildfly-ejb-plugin

PR #1819, opened 2021-05-18
Fix runtime attach for current user

SylvainJuge:
## What does this PR do?

When using runtime attach as `root` user, it seems that in some circumstances the attach process hangs likely due to a `sudo` interactive input. In this case, the target JVM also runs as `root`, thus trying to use `sudo` is not even required at all.

When this happens with agent `1.23.0`, the following stack trace is visible in a thread dump:

```
   java.lang.Thread.State: WAITING (on object monitor)
        at java.lang.Object.wait(java.base@11.0.9.1/Native Method)
        - waiting on <0x00000000c65e0ab0> (a java.lang.ProcessImpl)
        at java.lang.Object.wait(java.base@11.0.9.1/Object.java:328)
        at java.lang.ProcessImpl.waitFor(java.base@11.0.9.1/ProcessImpl.java:495)
        - waiting to re-lock in wait() <0x00000000c65e0ab0> (a java.lang.ProcessImpl)
        at co.elastic.apm.attach.UserRegistry$User.canSwitchToUser(UserRegistry.java:153)
        at co.elastic.apm.attach.UserRegistry$User.of(UserRegistry.java:144)
        at co.elastic.apm.attach.UserRegistry$User.access$100(UserRegistry.java:130)
        at co.elastic.apm.attach.UserRegistry.get(UserRegistry.java:125)
        at co.elastic.apm.attach.UserRegistry.getCurrentUser(UserRegistry.java:90)
        at co.elastic.apm.attach.AgentAttacher.<init>(AgentAttacher.java:73)
        at co.elastic.apm.attach.AgentAttacher.main(AgentAttacher.java:131)
```

What this PR does (at least as time of writing) is just making sure that we do not use `sudo` whenever the current user is already the one from the target JVM. Fixing the blocking behavior will be a second step.

## Checklist

- [x] ensure that we redirect IO in case the underlying command use stdin/stderr (would have allowed to debug this easily).
- [x] ~~adding a timeout when waiting for `sudo` execution to prevent such blocking behavior + properly log the issue~~
    --> not doable as `waitFor` with timeout variant has been added in Java 8.
- [x] This is a bugfix

### DIFF
--- a/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
+++ b/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
@@ -138,10 +138,12 @@ public class UserRegistry {
 
         private static User of(String username) {
             try {
+                // should always able to switch your yourself
+                boolean canSwitch = getCurrentUserName().equals(username);
                 if (Platform.isWindows()) {
-                    return new User(username, false);
+                    return new User(username, canSwitch);
                 } else {
-                    return new User(username, canSwitchToUser(username));
+                    return new User(username, canSwitch || canSwitchToUser(username));
                 }
             } catch (Exception e) {
                 throw new RuntimeException(e);
--- a/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
+++ b/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
@@ -31,6 +31,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
```diff
@@ -138,10 +138,12 @@ private User(String username, boolean canSwitchToUser) {
 
         private static User of(String username) {
             try {
+                // should always able to switch your yourself
+                boolean canSwitch = getCurrentUserName().equals(username);
```
felixbarny, line 142: consider doing that check in `canSwitchToUser`

eyalkoren (approved):

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
```diff
@@ -185,13 +193,28 @@ public ProcessBuilder runAs(List<String> cmd) {
             if (!canSwitchToUser) {
                 throw new IllegalStateException(String.format("Cannot run as user %s", username));
             }
+            if (!username.equals(getCurrentUserName())) {
```
felixbarny, line 196: [minor] Can be simplified to `if (!isCurrentUser()) {`

apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/UserRegistryTest.java
```diff
@@ -53,4 +53,16 @@ void testTempDir() throws Exception {
         assertThat(UserRegistry.getAllUsersMacOs().getAllTempDirs()).contains(System.getProperty("java.io.tmpdir"));
     }
 
+    @Test
+    @DisabledOnOs(OS.WINDOWS)
+    void testCurrentUserCanSwitchToHimself() {
```
felixbarny, line 58: ~`Himself`~ -> `Themselves`, or just `Self`

PR #1820, opened 2021-05-18
Use Connection#getCatalog to set context.db.instance in jdbc instrumentation

### LINKED ISSUE #1800: Add database/schema name to JDBC spans
We are using multiple databases during one request. At the moment we cannot really distinguish which db span belongs to which database.

`java.sql.Connection` provides `getCatalog()` and `getSchema()` which can be used to provide additional information for JDBC spans.

## Additional context
We are using PostgreSQL and some of the databases are hosted on the same server, thus we cannot use the `destination.*`

tobiasstadler:
## What does this PR do?
Partly fixes #1800 

## Checklist
- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else


tobiasstadler:
@felixbarny or @SylvainJuge Would you mind having a look? Thank You!

felixbarny:
Hey Tobias, sorry for the radio silence. We haven't forgotten about you but are busy finishing some things for the next release. I'm off next week but I hope someone from the team can find some time to look at your PRs.

tobiasstadler:
Thank you for the update. Have a nice week off!

SylvainJuge:
I added an implementation of the idea I suggested in https://github.com/elastic/apm-agent-java/issues/1800#issuecomment-853138393.

There is one caveat though, is that we have a metadata connection cache on per-connection basis (see [getConnectionMetaData)](https://github.com/elastic/apm-agent-java/blob/a837d43aa3fa189c19ae5dadaa4e0e28ae52d97a/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/JdbcHelper.java#L154)
If we capture the "instance" field when creating the cache entry, it means we will keep the same for the lifetime of the connection.

So, for example if you connect first to a server without targeting any database, for example to create a new DB, there won't be any catalog or schema for that connection, but if then the application reuses the connection to connect to target a specific catalog/schema and calls `setSchema` or `setCatalog`, then we would not report this change. This would be easily solved by capturing calls to those methods though.

apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/AbstractJdbcInstrumentationTest.java
```diff
@@ -408,6 +408,7 @@ private Span assertSpanRecorded(String rawSql, boolean preparedStatement, long e
         Db db = span.getContext().getDb();
         assertThat(db.getStatement()).isEqualTo(rawSql);
         DatabaseMetaData metaData = connection.getMetaData();
+        assertThat(db.getInstance()).isEqualToIgnoringCase(connection.getCatalog());
```
SylvainJuge, line 411: [question] why should we need to ignore case here ? Given the instrumentation also calls `getCatalog` the values must be identical.

SylvainJuge (approved):

apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/AbstractJdbcInstrumentationTest.java
tobiasstadler: The surrounding asserts also used isEqualToIgnoringCase, although it is not necessary (e.g. `assertThat(db.getType()).isEqualToIgnoringCase("sql")`). So I did the same.

PR #1821, opened 2021-05-19
Give precedence to w3c traceparent header

felixbarny:
## What does this PR do?

Implements https://github.com/elastic/apm/pull/442

## Checklist


SylvainJuge (approved):
LGTM, do we have a few examples when this would be problematic to be a bit more clear why this change was required ? (probably @AlexanderWert might have an idea on this one also).

AlexanderWert:
Whenever there is something in the middle (let's say a reverse proxy) of two services being monitored by our agent. And the proxy uses the W3C headers to create own spans (--> thus sends modified traceparent header to the downstream service), but **passes our proprietary header as is**. In this case the downstream service has a conflict, especially if it prefers the proprietary header over the W3C header, resulting in wrong parent child relationship.

PR #1822, opened 2021-05-21
Add Javalin plugin

spinscale:
## What does this PR do?

This adds a javalin plugin which creates spans for handlers. This is what it looks like in the UI

Screenshot 2021-05-21 at 14 16 53

Javalin is based on handlers. A handler can be a before or after handler or what to be executed for a certain HTTP endpoint or a combination of those. A handler is basically a functional interface, so that one can also add lambdas for a handler. Those had to be wrapped into a handler to create spans out of them, which is done with another instrumentation.

TODO: There are two left-over todos in the code. First I would like to get rid of the reflection, second the naming of span types and subtypes. I don't know if there are any standards for this I need to adhere to.

## Checklist

- [ ] This is a bugfix
- [x] This is a new plugin
- [ ] This is something else


docs/supported-technologies.asciidoc
```diff
@@ -142,6 +142,11 @@ rendering
 |The transactions are named based on your action (`ActionClass#actionMethod`).
 |1.24.0
 
+|Javalin
+|3.x
+|
+|1.24.0
```
eyalkoren, line 148:
```suggestion
|1.25.0
```

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinHandlerLambdaInstrumentation.java
```diff
@@ -0,0 +1,99 @@
...
+
+import co.elastic.apm.agent.bci.TracerAwareInstrumentation;
+import co.elastic.apm.agent.sdk.advice.AssignTo;
+import io.javalin.http.Context;
+import io.javalin.http.Handler;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.jetbrains.annotations.NotNull;
```
eyalkoren, line 35:
```suggestion
```
Add a `package-info.java` instead - see [example](https://github.com/elastic/apm-agent-java/blob/master/apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/package-info.java

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,167 @@
...
+        @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
+        public static Object setSpanAndTransactionName(@Advice.This Handler handler,
+                                                       @Advice.Argument(0) Object ctxObject) throws Exception {
+            final Transaction transaction = tracer.currentTransaction();
+            if (transaction == null) {
+                return null;
+            }
+            Context ctx = ((Context) ctxObject);
+
+            // TODO FIXME I don't think this is the right way, I am sure there is a fancy bytecode trick that I am not aware off
```
eyalkoren, line 90: Yes, there is! You can use `@Advice.FieldValue`. See [example](https://github.com/elastic/apm-agent-java/blob/master/apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientInstrumentation.java#L58)

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,167 @@
...
+            // transaction name like the HTTP path
+            if (handlerType.isHttpMethod()) {
+                transaction.setFrameworkName(FRAMEWORK_NAME);
+                transaction.setFrameworkVersion(VersionUtils.getVersion(Handler.class, "io.javalin", "javalin"));
+                // no need for anonymous handler class names in the transaction
+                if (handlerClassName.startsWith("co.elastic.apm.agent.javalin.JavalinHandlerLambdaInstrumentation$WrappingHandler")
+                    || handlerClassName.startsWith("io.javalin.http.JavalinServlet")) {
+                    setName(transaction, handlerType.name() + " " + ctx.endpointHandlerPath());
+                } else {
+                    setName(transaction, handlerType.name() + " " + ctx.endpointHandlerPath() + " " + handlerClassName);
```
eyalkoren, line 109: Don't use these `+` operators for names, as this allocates unnecessary objects.
This misses the purpose of our `co.elastic.apm.agent.impl.transaction.AbstractSpan#appendToName()` API. Your `setName()` method can get all required name elements and build it through appends.

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,167 @@
...
+            // create own span for all handlers including after/before
+            final Span span = startSpan(handlerType.name(), ctx.matchedPath(), ctx.method(), ctx.url());
+            return span;
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class, inline = false)
+        public static void onAfterExecute(@Advice.Enter @Nullable Object spanObj,
+                                          @Advice.Argument(0) Object ctxObject,
+                                          @Advice.Thrown @Nullable Throwable t) {
+            if (spanObj instanceof Span) {
```
eyalkoren, line 127: I don't think this check is necessary. It should be enough to check `if (spanObj != null)`

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,167 @@
...
+                Context ctx = ((Context) ctxObject);
+                final Span span = (Span) spanObj;
+                span.deactivate();
```
eyalkoren, line 130: This seems wrong if a response future is set. In such case, you should probably `deactivate` as part of `whenComplete`. Either way, `deactivate` just before you `end`.

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,167 @@
...
+
+        private static Span startSpan(String handlerTypeName, String matchedPath, String method, String url) {
+            final AbstractSpan<?> parent = tracer.getActive();
+            if (parent == null) {
+                return null;
+            }
+
+            Span span = parent.createSpan().activate();
+            // TODO figure out correct type and subtype or leave out?
+            span.withType("web")
```
eyalkoren, line 151: We are soon to have a [spec](https://github.com/elastic/apm/pull/443/files#diff-b09b7eae35b49561ec883168d53a9b112c63632aee03d2977ccf81f300bc614e) for allowed types and subtypes and since this is not part of it, the test will fail.
If you can't find something relevant in the spec, you may propose adding this.

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,167 @@
...
+
+            final HandlerType handlerType = (HandlerType) handlerTypeField.get(ctx);
+            final String handlerClassName = handler.getClass().getName();
+
+            // transaction name like the HTTP path
+            if (handlerType.isHttpMethod()) {
+                transaction.setFrameworkName(FRAMEWORK_NAME);
+                transaction.setFrameworkVersion(VersionUtils.getVersion(Handler.class, "io.javalin", "javalin"));
+                // no need for anonymous handler class names in the transaction
+                if (handlerClassName.startsWith("co.elastic.apm.agent.javalin.JavalinHandlerLambdaInstrumentation$WrappingHandler")
```
eyalkoren, line 105: [minor]:
```suggestion
                if (handlerClassName.equals("co.elastic.apm.agent.javalin.JavalinHandlerLambdaInstrumentation$WrappingHandler")
```

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinHandlerLambdaInstrumentation.java
```diff
@@ -0,0 +1,99 @@
...
+    // Javalin
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return named("io.javalin.Javalin");
+    }
+
+    // public Javalin addHandler(@NotNull HandlerType handlerType, @NotNull String path, @NotNull Handler handler, @NotNull Set<Role> roles) {
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return named("addHandler")
```
eyalkoren, line 56: Add a test for lambda handlers added through this API

apm-agent-plugins/apm-javalin-plugin/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+            <artifactId>jetty-webapp</artifactId>
+            <version>9.4.40.v20210413</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-servlet</artifactId>
+            <version>9.4.40.v20210413</version>
+            <scope>provided</scope>
+        </dependency>
```
eyalkoren, line 43: Are all these `provided` dependencies indeed required? All I found in non-tests are dependencies in `io.javalin.http` classes. Try to reduce dependencies of this scope to minimum.

eyalkoren (commented):
@spinscale Thanks for the PR!
You would first need to merge master and resolve the conflicts.
Then, you would need to remove the usage of lambda from instrumentation, so that compilation doesn't fail.

Then we can see how it builds and tests.
In addition, please either remove to `README` or (better) - add some description of the plugin.

Other comments are inlined.

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
eyalkoren: @SylvainJuge WDYT would be best?

SylvainJuge: That's a very good question !

The other example we have for such internal span types is the one for `servlet` / `request-dispatcher`.
If we follow the same pattern we could pick `javalin` / `javalin-handler`.

Would there exist other span subtypes ?

apm-agent-plugins/apm-javalin-plugin/pom.xml
spinscale: If this is left out, an older version of those dependencies is used (as there is a jetty dependency somewhere else), that does not belong to the current javalin version resulting in class cast exception.

Any other idea how to solve this?

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
spinscale: not in this case currently, will go with `javalin/javalin-handler` then

apm-agent-plugins/apm-javalin-plugin/pom.xml
SylvainJuge: There was some changes related to jetty issues with the webflux plugin (and few other plugins), that was merged last week to `master`. Thus if you update to `master` you should now be able to use any jetty version within a plugin without seeing class loading/cast exceptions.

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
spinscale: I saw `Advice.FieldValue`, however this seems to be about the class that is being instrumented, which in this case would be the `Handler` functional interface. I need to access a field value within the argument of the instrumented class.

Am I wrong on that one and there is a way? :-)

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,167 @@
...
+import static net.bytebuddy.matcher.ElementMatchers.not;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+
+public class JavalinInstrumentation extends TracerAwareInstrumentation {
+
+    private static final String FRAMEWORK_NAME = "Javalin";
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return hasSuperType(named("io.javalin.http.Handler")).and(not(isInterface()));
```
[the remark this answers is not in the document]
spinscale: Hm. I don't have any good idea to limit this TBH, as this is just a functional interface and I have a hard time assuming that everyone suffixes/prefixes their javalin class with `Handler` - I saw in [JaxRsTransactionNameInstrumentation](https://github.com/elastic/apm-agent-java/blob/master/apm-agent-plugins/apm-jaxrs-plugin/src/main/java/co/elastic/apm/agent/jaxrs/JaxRsTransactionNameInstrumentation.java#L70-L75) that might help if application packages are specified. Should I go that route?

apm-agent-plugins/apm-javalin-plugin/pom.xml
spinscale: removing those dependencies still results in 

```
java.lang.IncompatibleClassChangeError: Found interface org.eclipse.jetty.http.pathmap.PathSpec, but class was expected
        at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:448) ~[jetty-servlet-9.4.11.v20180605.jar:9.4.11.v20180605]
        at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:1594) ~[jetty-server-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:186) ~[jetty-server-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1350) ~[jetty-server-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141) ~[jetty-server-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.server.handler.StatisticsHandler.handle(StatisticsHandler.java:179) ~[jetty-server-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:127) ~[jetty-server-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.server.Server.handle(Server.java:516) ~[jetty-server-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.server.HttpChannel.lambda$handle$1(HttpChannel.java:388) ~[jetty-server-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.server.HttpChannel.dispatch(HttpChannel.java:633) ~[jetty-server-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:380) [jetty-server-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:277) [jetty-server-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:311) [jetty-io-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105) [jetty-io-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.io.ChannelEndPoint$1.run(ChannelEndPoint.java:104) [jetty-io-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.runTask(EatWhatYouKill.java:336) [jetty-util-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.doProduce(EatWhatYouKill.java:313) [jetty-util-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.tryProduce(EatWhatYouKill.java:171) [jetty-util-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.run(EatWhatYouKill.java:129) [jetty-util-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread.run(ReservedThreadExecutor.java:383) [jetty-util-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:882) [jetty-util-9.4.40.v20210413.jar:9.4.40.v20210413]
        at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1036) [jetty-util-9.4.40.v20210413.jar:9.4.40.v20210413]
        at java.lang.Thread.run(Thread.java:834) [?:?]
```

The version `jetty-servlet-9.4.11.v20180605.jar:9.4.11.v20180605` is still the old one after merging the current master

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,167 @@
...
+                final Span span = (Span) spanObj;
+                span.deactivate();
+
+                final CompletableFuture<?> responseFuture = ctx.resultFuture();
+                if (responseFuture == null) {
+                    // sync request
+                    span.captureException(t).end();
+                } else {
+                    // future was set in the handler, so we need to end the span only on future completion
+                    responseFuture.whenComplete((o, futureThrowable) -> span.captureException(futureThrowable).end());
```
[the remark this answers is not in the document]
spinscale: javalin has a minimum requirement of 8, so I changed that in the module as well

spinscale:
thx for your comments, did a first run at them and left two more questions, where I don't have a good answer/enough agent knowledge...

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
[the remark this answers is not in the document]
eyalkoren: Overriding `co.elastic.apm.agent.sdk.ElasticApmInstrumentation#getClassLoaderMatcher` to restrict only to class loaders that can load the `io.javalin.http.Handler` interface is already a great improvement. It relies on the fact that implementations must be loaded from a class loader to which the interface is visible.

The `application_packages` is a good option if all `Handler` implementations are expected to be such. However, it doesn't fit very well with our aim to get a good zero-configuration experience.

Maybe the best option is to use the [`ensureInstrumented`](https://github.com/elastic/apm-agent-java/blob/master/apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/DynamicTransformer.java#L47) capability, which dynamically instruments classes if they are not yet instrumented, but it requires that you instrument the place where the handler is submitted for execution. Your `addHandler` instrumentation would only cover agent attachment at startup, and not runtime attachment. If you find some `submit(Handler handler)` or `invoke(Handler handler)` API to instrument, that would be good.

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
eyalkoren: Ohh sorry, you are right, it is the `Context` field you are after. 
Let me see if I can suggest something

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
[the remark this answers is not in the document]
eyalkoren: I just looked at the code, it looks like `addHandler` is not some type of registration, it is called for every handling, is that correct? If so, it is a good candidate to do `ensureInstrumented`

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
eyalkoren: OK, if I am correct and `addHandler` is what gets called every time a `Handler` is submitted to `handle`, then you can use a global **weakly referenced** mapping between a `Handler` **instance** and the corresponding type. So, when `addHandler` is called, you map the `HandlerType` (as the map value) to the `Handler` instance (as the map key) and when `handle` ends, you clear this entry from the map.
That is preferable over getting the field value through reflection. In general, reflection is problematic because it may affect the class loading order, something that Byte Buddy does a lot to prevent. Also, relying on API tends to be more stable than relying on internal stuff, like a field.

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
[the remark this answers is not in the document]
eyalkoren: I think it should be OK - we should catch the `UnsupportedClassVersionError` thrown when failing to load the instrumentation class and log it in `debug` level.

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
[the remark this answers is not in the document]
spinscale: You may be confusing the two intrumentations here:

1. `JavalinHandlerLambdaInstrumentation` is instrumented against `io.javalin.Javalin.addHandler()` and is only called when a new handler is added in order to wrap anonymous lambdas
2. `JavalinInstrumentation` instruments `io.javalin.http.Context.handle()` and is invoked on every HTTP request, so this would indeed be a good candidate for ensureInstrumented instrumentation? However how much sense does this make, if the lambda wrapping is not applied, so it would only apply for non-lambda anonymous handlers in that case?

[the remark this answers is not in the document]
eyalkoren: No, I am not confusing, but I think it is the time for a chat, rather than this back and forth 🙂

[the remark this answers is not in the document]
eyalkoren: Chat summary - we'll start with the class loader matcher only and that should be OK

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
eyalkoren: Comment above is irrelevant as the `addHandler` is indeed only part of the setup flow, not part of the request handling flow.

spinscale:
I opened a PR against javalin to expose the handlertype so we don't have to use reflection (got already merged in master, will be merged in the 3x release branch as well). I'll wait for the next javalin release and then finish this PR, so we don't need to go with reflection at all.

Thanks so much for all the help, much appreciated!

eyalkoren:
If Javalin is using semver for versioning, you may want to use something like `co.elastic.apm.agent.bci.bytebuddy.CustomElementMatchers#implementationVersionLte()` through `co.elastic.apm.agent.httpclient.ApacheHttpAsyncClientRedirectInstrumentation#getProtectionDomainPostFilter()`, to enforce this version restriction (instead of getting `NoSuchMethodError` or the like).
You can either negate `implementationVersionLte()` or implement `implementationVersionGte()` based on it.

spinscale:
As the latest Javalin version has been released with my patch I just resumed work here and removed all the reflection code plus adding a version check... the build fails with

```
17:12:31  [ERROR] Failed to execute goal org.codehaus.mojo:animal-sniffer-maven-plugin:1.17:check (signature-check) on project elastic-apm-agent: Execution signature-check of goal org.codehaus.mojo:animal-sniffer-maven-plugin:1.17:check failed. UnsupportedOperationException -> [Help 1]
```

and I am not sure what to do about it... any help appreciated :-)

apm-agent-plugins/apm-javalin-plugin/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+        <groupId>co.elastic.apm</groupId>
+        <version>1.24.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>apm-javalin-plugin</artifactId>
+    <name>${project.groupId}:${project.artifactId}</name>
+
+    <properties>
+        <apm-agent-parent.base.dir>${project.basedir}/../..</apm-agent-parent.base.dir>
+        <maven.compiler.target>11</maven.compiler.target>
```
eyalkoren, line 16: Why compiling with Java 11 and not 8?

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,163 @@
...
+                return null;
+            }
+
+            // create own span for all handlers including after/before
+            return startSpan(handlerClassName, ctx.handlerType(), isLambdaHandler, ctx.matchedPath(), ctx.method(), ctx.url());
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class, inline = false)
+        public static void onAfterExecute(@Advice.Enter @Nullable Object spanObj,
+                                          @Advice.Argument(0) Object ctxObject,
```
eyalkoren, line 126: Use `io.javalin.http.Context` explicitly

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,163 @@
...
+                return null;
+            }
+            Context ctx = ((Context) ctxObject);
+
+            final String handlerClassName = handler.getClass().getName();
+            final boolean isLambdaHandler = handlerClassName.equals("co.elastic.apm.agent.javalin.JavalinHandlerLambdaInstrumentation$WrappingHandler");
+
+            // transaction name gets only set if we are dealing with a HTTP method processing, not before/after handlers
+            if (ctx.handlerType().isHttpMethod()) {
+                final StringBuilder name = transaction.getAndOverrideName(PRIO_HIGH_LEVEL_FRAMEWORK);
```
eyalkoren, line 100: Shouldn't you use:
```suggestion
                final StringBuilder name = transaction.getAndOverrideName(PRIO_HIGH_LEVEL_FRAMEWORK, false);
```
So that you don't override the name for every handler?

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,163 @@
...
+
+            Span span = parent.createSpan().activate();
+            span.appendToName(handlerType.name()).appendToName(" ").appendToName(matchedPath).appendToName(" ");
+            if (isLambdaHandler) {
+                span.appendToName("<Lambda>");
+            } else {
+                span.appendToName(handlerClassName);
+            }
+
+            span.getContext().getHttp().withUrl(url).withMethod(method);
```
eyalkoren, line 158: This context info is only relevant for HTTP client spans, not here.

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,163 @@
...
+            }
+        }
+
+        private static Span startSpan(String handlerClassName, HandlerType handlerType, boolean isLambdaHandler, String matchedPath, String method, String url) {
+            final AbstractSpan<?> parent = tracer.getActive();
+            if (parent == null) {
+                return null;
+            }
+
+            Span span = parent.createSpan().activate();
```
eyalkoren, line 150: [minor] I prefer to see the `activate` in the `onMethodEnter` advice, it's easier to follow the state with regard to deactivation in `onMethodExit`.

eyalkoren (changes requested):

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,163 @@
...
+    @Override
+    public String getAdviceClassName() {
+        return "co.elastic.apm.agent.javalin.JavalinInstrumentation$HandlerAdapterAdvice";
+    }
+
+    public static class HandlerAdapterAdvice {
+
+        @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
+        public static Object setSpanAndTransactionName(@Advice.This Handler handler,
+                                                       @Advice.Argument(0) Object ctxObject) throws Exception {
```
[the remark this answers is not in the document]
spinscale: done

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
spinscale: changed this method to only set the span name so all the code remains in the advice

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
spinscale: removed

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,163 @@
...
+    @Override
+    public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
+        return classLoaderCanLoadClass("io.javalin.http.Handler");
+    }
+
+    // ctx.handlerType() was introduced in Javalin 3.13.8 and above, so instrument only in that case
+    @Override
+    public ElementMatcher.Junction<ProtectionDomain> getProtectionDomainPostFilter() {
+        return implementationVersionLte("3.13.7");
+    }
```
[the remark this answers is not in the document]
spinscale: Tried to come up with an implementation using method handles, maybe you can advice if this this a useful one.. I am not too proud of using the tri-state of unboxing a `Boolean`, but this way I don't have to use sth like an AtomicReference... improvements welcome :-)

also, do I need to add that code to the second advice as well?

Also, the `JavalinHandlerLambdaInstrumentation` also needs the implementation of the `getClassLoaderMatcher` method of the other advice, correct?

eyalkoren (approved):
@spinscale See the PR I pushed - it is safer (because the advice is not compiled with the potentially non-existing `handlerType()` method) and more efficient (because lookup is done only once per advice class).

In addition,  I used the `java.lang.Class#getSimpleName()` rather than the fully qualified name (`java.lang.Class#getName()`) for transaction and span names, but I would suggest removing them altogether from the names, we normally use the HTTP method and path OR the controller/class name.
However, this is up to you as you are a real client for this instrumentation 😄 

With this, this PR is approved, but please verify that it works as expected in a real app. I also recommend adding a test with multiple handlers.

spinscale:
did one final change to properly support Javalin 3 and Javalin 4 (which is currently in alpha stage but will be released some time in the future). Check out some Kibana APM screenshots from an app

Screenshot 2021-07-07 at 15 54 47

and some spans (including a before handler)

image

Looks good to me!

eyalkoren:
That indeed looks really good @spinscale, well done! 👏 
Do the stack traces of the handler spans provide enough info to identify the handler class?

We still have this checkbox to tick - `Added an instrumentation plugin? Describe how you made sure that old, non-supported versions are not instrumented by accident.`, and indeed we just made a relevant change in the last commits, so if you can do a quick sanity check with Javalin < 3.13.8, that would conclude efforts on this PR.

felixbarny:
What I noticed in the last screenshot is that there's a type `custom`. I suppose that's the span from the `Handler`? I'd suggest setting `.withType("app").withSubtype("handler")`. The value `app` is a special value for the breakdown calculation that is also used for the transaction itself.

spinscale:
the above `withSubType` suggestion requires a change in the `span_types.json` file... is this ok to do? Would it make sense then to rename `handler` to `javalin` just like the span type definitions for the `db` type, which refers to the underlying data store?

spinscale:
There is one more thing that can be fixed... in case of an older version there should not be any instrumentation by the `JavalinHandlerLambdaInstrumentation`. With the most recent refactoring of the `getHandlerType()` method, this requires a context, which is not available in that specific instrumentation, as it happens when a new handler is registered, where as the context is only created when a HTTP request is processed.

What would be a proper workaround for this? Using another `MethodHandles.lookup()` in the other instrumentation?

SylvainJuge (approved):

SylvainJuge:
> the above `withSubType` suggestion requires a change in the `span_types.json` file... is this ok to do? Would it make sense then to rename `handler` to `javalin` just like the span type definitions for the `db` type, which refers to the underlying data store?

I think that sticking to the rather generic `app` / `handler` is probably better for now, we are working to align this spec across agents.

apm-agent-plugins/apm-javalin-plugin/src/main/java/co/elastic/apm/agent/javalin/JavalinInstrumentation.java
```diff
@@ -0,0 +1,189 @@
...
+
+            // transaction name gets only set if we are dealing with a HTTP method processing, not before/after handlers
+            if (handlerType.isHttpMethod()) {
+                final StringBuilder name = transaction.getAndOverrideName(PRIO_HIGH_LEVEL_FRAMEWORK, false);
+                if (name != null) {
+                    transaction.setFrameworkName(FRAMEWORK_NAME);
+                    transaction.setFrameworkVersion(VersionUtils.getVersion(Handler.class, "io.javalin", "javalin"));
+                    transaction.withType("request");
+
+                    name.append(handlerType.name()).append(" ").append(ctx.endpointHandlerPath());
```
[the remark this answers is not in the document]
spinscale: I figured that this means it will only called once within the HTTP handler instead of within any before/after HTTP handlers as well, thus this placement

apm-agent-plugins/apm-javalin-plugin/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.javalin</groupId>
+            <artifactId>javalin</artifactId>
+            <version>3.13.9</version>
+            <scope>provided</scope>
+        </dependency>
+        <!-- this has to override the global jetty dependency, and needs to match the javalin version -->
+        <dependency>
```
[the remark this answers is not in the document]
spinscale: done

spinscale:
changed type/subtype to `app/handler`. Then the only left over point is what is the best way to ensure, that the lambda instrumentation does not happen on older versions... adding another check in that instrumentation or changing the existing one to cater for both cases

eyalkoren:
> Then the only left over point is what is the best way to ensure, that the lambda instrumentation does not happen on older versions... adding another check in that instrumentation or changing the existing one to cater for both cases

No, we don't want to make this lookup unless the context class is already loaded, because that would affect the order of class loading. As you can see in how I did it - I am relying on the already loaded context class.

One option is to let the lambda instrumentation wrap handlers even if not supported. I think that is not that bad and if someone finds it really problematic, they can always disable the Javalin instrumentation, or upgrade.
Besides the small overhead, do you have any specific concern with wrapping in older versions?

spinscale:
If that is an option, I'm fine. I just read

> `Added an instrumentation plugin? Describe how you made sure that old, non-supported versions are not instrumented by accident.`

that this is not an option. If we're good this way, I'm good

eyalkoren:
It's not ideal, but I think we can live with that. This checkbox is mostly about - how did you verify that everything works as expected when attaching the agent to a non-supported version?
Did you verify? On which version?

spinscale:
Running the tests against an older versions, shows that the Javalin app still works (as the assertions pass), however the span names do not exist and those assertions fail in the test.

In addition I ran a small sample app with an older version and the app worked without instrumentation.

PR #1824, opened 2021-05-22
Vert.x WebClient instrumentation

AlexanderWert:
Extended Vert.x instrumentation with instrumentation for Vert.x Web Client.

Notes:
- Didn't use the tracer approach for version 4 as this does not allow to handle instrumentation of redirects properly.
- Instrumentation approach is similar for all versions
- had to fix Vert.x core instrumentation for setTimer context propagation: avoid instrumentation of directly recursive setTimer invocations (as this is a common pattern for endless checks)

- [x] Added to CHANGELOG
- [x] Added to supported technologies page


apm-agent-plugins/apm-vertx/apm-vertx4-plugin/src/main/java/co/elastic/apm/agent/vertx/v4/webclient/WebClientInstrumentation.java
```diff
@@ -0,0 +1,190 @@
...
+                if (null == spanObj) {
+                    spanObj = vertxContext.getLocal(AbstractVertxWebHelper.CONTEXT_TRANSACTION_KEY);
+                    decrementParentReferences = false;
+                }
+
+                if (spanObj instanceof AbstractSpan) {
+                    AbstractSpan<?> parent = (AbstractSpan<?>) spanObj;
+                    webClientHelper.startSpan(parent, httpContext, request);
+                    if (decrementParentReferences) {
+                        parent.decrementReferences();
```
eyalkoren, line 134: Conceptually, reference increment should be coupled with the insertion into `HttpContext` and reference decrement should be coupled with removal (which is actually setting the key to `null` as there doesn't seem to be a `remove` API).
Removing from the map is important - if we decrement references and the span object gets recycled, the `HttpContex` will point to a span with arbitrary contents.
We need to pay attention to both `WEB_CLIENT_PARENT_SPAN_KEY` and `WEB_CLIENT_SPAN_KEY`.

eyalkoren (commented):
I will take over from here, thanks @AlexanderWert!

eyalkoren (approved):
@AlexanderWert it is ready as far as I am concerned.
Please go over the changes I've made to make sure they make sense.

I fixed the `fail()` API instrumentation (decrementing references for the parent span so that it can be reported) and added a test. In addition, I made sure that we at least capture an error for this scenario. Currently, a span is reported with all the details for failed requests only in Vert.x 3. We can add this capability to Vert.x 4 as well, but it will be messy - we will need to rely on internal fields for that (see [details](https://github.com/elastic/apm-agent-java/pull/1824/files#diff-b8b005f46a6aae37ee812d29606909e0f40c8b5cd1bdbda1ae7bbf897e7c6ca5R48-R51)) so I decided an error will do for now 🙂 .

AlexanderWert:
@eyalkoren: LGTM, I'm fine with merging!

PR #1827, opened 2021-05-26
Check minimal JFR size before parsing

eyalkoren:
Context: https://discuss.elastic.co/t/apm-profiler-stopping-with-https-discuss-elastic-co-c-apm-on-latest-adoptopenjdk-openjdk8-openj9-alpine-slim/273428


felixbarny (approved):
I'm fine with the change but also have some mixed feelings.

An argument could be made that the current behavior of throwing an exception is the right one. It seems like it's a situation that can't be recovered from - async-profiler can't profile the current VM. By throwing the exception, the profiler is stopped and it won't attempt to start other profiling sessions that are inevitably going to fail.

eyalkoren:
> An argument could be made that the current behavior of throwing an exception is the right one. It seems like it's a situation that can't be recovered from - async-profiler can't profile the current VM. 

That's what I am missing here - if this cannot be a recoverable state, then certainly throwing an Exception is preferable. Can you think of a scenario where this is a recoverable/temporary/sporadic state?

eyalkoren:
Changes:
- Throwing an Exception and stopping profiler when JFR file is empty
- Throwing an Exception and aborting profiler when profiler is enabled on J9 JVMs
- Documented everywhere 

Since we document that it is not supported, we will have to update anyway if and when J9 is supported, so we can remove the check from code as well.

felixbarny (approved):

PR #1842, opened 2021-06-04
prevent NPE when capturing headers

### LINKED ISSUE #1840: Error in transaction creation based on Kafka record
Kibana version:
7.11.2
Elasticsearch version:
7.11.2
APM Server version:
7.11.2
APM Agent language and version:
java - 1.22.0

Original install method (e.g. download page, yum, deb, from source, etc.) and version:
java client jar

Description of the problem including expected versus actual behavior. Please include screenshots

Our Application logs show

_ERROR co.elastic.apm.agent.kafka.helper.ConsumerRecordsIteratorWrapper - Error in transaction creation based on Kafka record_

Our insights are around a message headers key/value having a null value and apm failing to capture such messages .

Provide logs and/or server output (if relevant):

 ```
 at java.lang.Thread.run(Thread.java:834) [?:?]
    	at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
    	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
    	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.run(KafkaMessageListenerContainer.java:1073) [spring-kafka-2.6.4.jar!/:2.6.4]
    	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.pollAndInvoke(KafkaMessageListenerContainer.java:1176) [spring-kafka-2.6.4.jar!/:2.6.4]
    	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.invokeListener(KafkaMessageListenerContainer.java:1529) [spring-kafka-2.6.4.jar!/:2.6.4]
    	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.invokeRecordListener(KafkaMessageListenerContainer.java:1810) [spring-kafka-2.6.4.jar!/:2.6.4]
    	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.doInvokeWithRecords(KafkaMessageListenerContainer.java:1911) [spring-kafka-2.6.4.jar!/:2.6.4]
    	at co.elastic.apm.agent.kafka.helper.ConsumerRecordsIteratorWrapper.next(ConsumerRecordsIteratorWrapper.java:42) [?:?]
    	at co.elastic.apm.agent.kafka.helper.ConsumerRecordsIteratorWrapper.next(ConsumerRecordsIteratorWrapper.java:100) [?:?]
    	at co.elastic.apm.agent.impl.context.Message.addHeader(Message.java:127) ~[?:?]
    	at co.elastic.apm.agent.impl.context.Headers.add(Headers.java:47) ~[?:?]
    	at co.elastic.apm.agent.util.BinaryHeaderMap.add(BinaryHeaderMap.java:76) ~[?:?]
    	at co.elastic.apm.agent.util.IOUtils.decodeUtf8Bytes(IOUtils.java:130) ~[?:?]
    java.lang.NullPointerException: null

```

### PR DESCRIPTION
**Pull request by SylvainJuge**
## What does this PR do?

Fixes #1840 

## Checklist

- [x] This is a bugfix

## For reviewer:

- [ ] Do we have any usage of `null` header names or keys in practice ? While the underlying maps may allow for `null` values, I *think* that it would not make any sense to try capturing those.

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/util/BinaryHeaderMap.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/util/BinaryHeaderMap.java
@@ -11,9 +11,9 @@
  * the Apache License, Version 2.0 (the "License"); you may
  * not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -52,8 +52,17 @@ public class BinaryHeaderMap implements Recyclable, Iterable<BinaryHeaderMap.Ent
     private static final Logger logger = LoggerFactory.getLogger(BinaryHeaderMap.class);
 
     private CharBuffer valueBuffer;
+    /**
+     * Ordered list of keys
+     */
     private final ArrayList<String> keys;
+
+    /**
+     * Values lengths, key index is provided with {@link #keys}.
+     * Negative length indicate a {@literal null} entry
+     */
     private int[] valueLengths;
+
     private final NoGarbageIterator iterator;
 
     public BinaryHeaderMap() {
@@ -71,7 +80,15 @@ public class BinaryHeaderMap implements Recyclable, Iterable<BinaryHeaderMap.Ent
         return size() == 0;
     }
 
-    public boolean add(String key, byte[] value) {
+    public boolean add(String key, @Nullable byte[] value) {
+        boolean result;
+        if (value == null) {
+            int size = keys.size();
+            keys.add(key);
+            valueLengths[size] = -1;
+            return true;
+        }
+
         int valuesPos = valueBuffer.position();
         CoderResult coderResult = IOUtils.decodeUtf8Bytes(value, valueBuffer);
         while (coderResult.isOverflow()) {
@@ -81,7 +98,7 @@ public class BinaryHeaderMap implements Recyclable, Iterable<BinaryHeaderMap.Ent
             }
             coderResult = IOUtils.decodeUtf8Bytes(value, valueBuffer);
         }
-        boolean result;
+
         if (coderResult.isError()) {
             ((Buffer) valueBuffer).limit(valuesPos);
             result = false;
@@ -94,6 +111,7 @@ public class BinaryHeaderMap implements Recyclable, Iterable<BinaryHeaderMap.Ent
             valueLengths[size] = valueBuffer.position() - valuesPos;
             result = true;
         }
+
         return result;
     }
 
@@ -148,6 +166,7 @@ public class BinaryHeaderMap implements Recyclable, Iterable<BinaryHeaderMap.Ent
         @Nullable
         String key;
         CharBuffer value = CharBuffer.allocate(64);
+        private boolean nullValue;
 
         public String getKey() {
             if (key == null) {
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by eyalkoren** [comment]
> Do we have any usage of null header names or keys in practice ? While the underlying maps may allow for null values, I think that it would not make any sense to try capturing those.

This code path is adding headers to span context by iterating over actual existing record headers, so I assume we should capture them as is in order to reflect metadata about the record. 
 `null` keys are not allowed in the docs, so we can guard from `null` even though such are not expected (because they come from the sent record, so they already passed this screening).

`null` values are allowed, so if the user uses them, they may have a meaning, in which case it worth reporting those. 
I think the easier fix is to add `"null"` to the value buffer in `BinaryHeaderMap#add()` if the value is `null`. 

BTW, in `NoRandomAccessMap#add()` (the text header map counterpart), `null` values are allowed and are added to the map, however they are omitted during serialization. Either way we decide, the behavior or text header and binary headers should be aligned.

**Review by eyalkoren — COMMENTED** [review]
> Do we have any usage of null header names or keys in practice ? While the underlying maps may allow for null values, I think that it would not make any sense to try capturing those.

This code path is adding headers to span context by iterating over actual existing record headers, so I assume we should capture them as is in order to reflect metadata about the record. 
 `null` keys are not allowed in the docs, so we can guard from `null` even though such are not expected (because they come from the sent record, so they already passed this screening).

`null` values are allowed, so if the user uses them, they may have a meaning, in which case it worth reporting those. 
I think the easier fix is to add `"null"` to the value buffer in `BinaryHeaderMap#add()` if the value is `null`. 

BTW, in `NoRandomAccessMap#add()` (the text header map counterpart), `null` values are allowed and are added to the map, however they are omitted during serialization. Either way we decide, the behavior or text header and binary headers should be aligned.

**Comment by SylvainJuge** [comment]
Good point, here is my revised proposal:

- allow for `null` values for both binary and text headers
- serialize them as `null` instead of omitting them, which would be consistent with what we do for HTTP headers

**Comment by eyalkoren** [comment]
The reason I suggested to serialize to `"null"` string is the way the binary headers are implemented. You can also store the bytes of this string and identify it in the iterator and return `null` as the value if you prefer, only make sure that the intake API allows for `null` values for headers.

**Comment by SylvainJuge** [comment]
Both the HTTP and message headers allow for `null` in the protocol schema for headers:

```
            "headers": {
              "description": "Headers received with the message, similar to HTTP request headers.",
              "type": [
                "null",
                "object"
              ],
              "additionalProperties": false,
              "patternProperties": {
                "[.*]*$": {
                  "type": [
                    "null",
                    "array",
                    "string"
                  ],
                  "items": {
                    "type": "string"
                  }
                }
              }
            },
```

File: apm-agent-core/src/main/java/co/elastic/apm/agent/util/BinaryHeaderMap.java [hunk 1]
```diff
@@ -52,8 +52,17 @@
...
 
     private CharBuffer valueBuffer;
+    /**
+     * Ordered list of keys
+     */
     private final ArrayList<String> keys;
+
+    /**
+     * Values lengths, key index is provided with {@link #keys}.
+     * Negative length indicate a {@literal null} entry
```
eyalkoren, line 62: 💡

apm-agent-core/src/main/java/co/elastic/apm/agent/util/BinaryHeaderMap.java
```diff
@@ -71,7 +80,15 @@ public boolean isEmpty() {
         return size() == 0;
     }
 
-    public boolean add(String key, byte[] value) {
+    public boolean add(String key, @Nullable byte[] value) {
+        boolean result;
+        if (value == null) {
+            int size = keys.size();
+            keys.add(key);
+            valueLengths[size] = -1;
```
eyalkoren, line 88: Make sure to `enlargeValueLengths()` if required to avoid `ArrayIndexOutOfBoundsException`.
Better yet, extract this code into a method to be reused with the same done below, something like:
```
int size = keys.size(); 
keys.add(key); 
if (size == valueLengths.length) {
  enlargeValueLengths();
}
valueLengths[size] = length;
```

eyalkoren (changes requested):

apm-agent-core/src/main/java/co/elastic/apm/agent/util/BinaryHeaderMap.java
SylvainJuge: now properly fixed & tested.

eyalkoren (approved):
Thanks! ❤️

Issue #838, opened 2019-09-11
Use default name priority if Spring controller method is unknown


felixbarny:
This lets users leverage the `use_path_as_transaction_name` config.

Context: https://discuss.elastic.co/t/apm-tracing-with-zuul-netflix-oss/198838

https://github.com/elastic/apm-agent-java/blob/aff5c2eb6889b4b5d072fa94da88b872cda1c942/apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/spring/webmvc/SpringTransactionNameInstrumentation.java#L136-L144

antrix190:
Hey @felixbarny I would like to work on this issue.
Could you please explain what needs to be done when methodName is null.

felixbarny:
If the method name is null, set `transaction.withName(className, PRIO_LOW_LEVEL_FRAMEWORK + 1)`

This way, `co.elastic.apm.agent.servlet.ServletTransactionHelper#applyDefaultTransactionName` will override the transaction name if `use_path_as_transaction_name` is set to `true` but it makes sure that it overrides the `DispatcherServlet#doGet` transaction name.

At least that's the theory 🙂

eyalkoren:
Closed by #1906

PR #1843, opened 2021-06-04
Fix use_path_as_transaction_name in Spring controllers

hectorespert:
## What does this PR do?

~~Fixes https://github.com/elastic/apm-agent-java/issues/838~~

Fixes `use_path_as_transaction_name` option when a transaction is generated by a Spring controller.

## Checklist

- [x] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else


eyalkoren:
@hectorespert thanks for this PR!

#838 says:
Currently, we always prefer Spring-MVC-specific transaction name over the generic one (which is taking the `use_path_as_transaction_name` config into consideration). We want to fix that so that if the method name is known, **we still do the same**, but if the method name is unknown, we should prefer the default/generic transaction name setting.
The way to do that is by using different priorities - currently, we always use `PRIO_HIGH_LEVEL_FRAMEWORK` when setting transaction name through the Spring MVC configurations. What I think @felixbarny meant is that we will use this priority only if the method name is known, otherwise we should use `PRIO_LOW_LEVEL_FRAMEWORK + 1`, allowing the default name-setting to take effect, including the `use_path_as_transaction_name` config.

So, while this PR may be considered as a "fix" for the current behaviour, it is not doing what #838 aims for and it defies our current assumption that framework-specific transaction names are always preferable. 
So, while the validity of the current assumption is debatable, we wouldn't want to change it atm.

hectorespert:
@eyalkoren I changed the scope of this PR. It's not a fix for https://github.com/elastic/apm-agent-java/issues/838 by I link that because I started trying to fit it.

eyalkoren:
@hectorespert I see.  As I mentioned above- we give high-level framework naming higher priority comparing to the `use_path_as_transaction_name` config, and we wouldn't want to change it  at them moment. 
So, again, while the validity of this decision is debatable, let's not do it here. If we _do_ decide to change it, it wouldn't be a Spring-MVC-specific change, it should be consistent with all the other frameworks.

Please let me know if and how you want to proceed, or whether you want me to take over.
Thanks!

hectorespert:
> Please let me know if and how you want to proceed, or whether you want me to take over.

@eyalkoren There is no problem if you take over.

apm-agent-plugins/apm-spring-webmvc-plugin/src/test/java/co/elastic/apm/agent/springwebmvc/SpringTransactionNameInstrumentationTest.java
```diff
@@ -79,13 +85,46 @@ void testControllerTransactionName() throws Exception {
         assertThat(reporter.getFirstTransaction().getNameAsString()).isEqualTo("TestController#test");
     }
 
+    @Test
+    void testControllerTransactionNameWhenUsePathAsTransactionNameIsTrue() throws Exception {
+        WebConfiguration webConfiguration = config.getConfig(WebConfiguration.class);
+        when(webConfiguration.isUsePathAsName()).thenReturn(true);
+        mockMvc.perform(get("/test/944207fa-d27b-40f6-be58-6dd6a1b541d4"))
+            .andExpect(content().string("944207fa-d27b-40f6-be58-6dd6a1b541d4"));
+        assertThat(reporter.getFirstTransaction().getNameAsString()).isEqualTo("GET /test/944207fa-d27b-40f6-be58-6dd6a1b541d4");
```
SylvainJuge, line 94: We need to minimize the name cardinality, in this case we should expect to have `"/test/{uuid}"` as transaction name, not the one that contains the full path. The full URL path is still captured in transaction attributes so it's not lost and can be retrieved if needed.

eyalkoren:
Superseded by #1906 as the fix for #838.

While this PR offers additional value, the full implementation for Spring MVC should be similar to what we have for JAX-RS, where we properly read annotation values to infer dynamic path parts and account for annotation inheritance, and this is not something we want to pursue at this point.

Thanks for raising this to awareness @hectorespert 🙏

PR #1850, opened 2021-06-10
updated copyright headers

kananindzya:
added `canUpdateCopyright` param in `license-maven-plugin`
updated headers via : `mvn license:update-file-header`

- [ ] This is something else


eyalkoren:
@kananindzya Thanks for taking the time to apply this update ❤️ 

Since we are about to remove years specification from the copyright notice, I will just close this PR and we will handle it on a different one.

PR #1854, opened 2021-06-10
Added basic support for com.sun.net.httpserver.HttpServer (#1853)

### LINKED ISSUE #1853: Support for com.sun.net.httpserver.HttpServer
We are using (standalone) Eclipse Jersey together with com.sun.net.httpserver.HttpServer for some of our services. Currently no transactions are started when such a service receives a request.

The agent should start a transaction when a http request is handled by com.sun.net.httpserver.HttpServer.

We could use a ContainerRequest/ResponseFilter.

tobiasstadler:
## What does this PR do?
Fixes #1853

## Checklist

- [x] This is a new plugin


tobiasstadler:

tobiasstadler:
@SylvainJuge Maybe you could have a look please?

apm-agent-plugins/apm-jdk-httpserver-plugin/src/main/java/co/elastic/apm/agent/httpserver/HttpHandlerAdvice.java
```diff
@@ -0,0 +1,167 @@
...
+
+            return hostname;
+        }
+        return exchange.getLocalAddress().getAddress().getHostName();
+    }
+
+    @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class, inline = false)
+    public static void onExitHandle(@Advice.Argument(0) HttpExchange exchange, @Advice.Enter @Nullable Object transactionOrNull, @Advice.Thrown @Nullable Throwable t) {
+        ElasticApmTracer tracer = GlobalTracer.getTracerImpl();
+        if (tracer == null || transactionOrNull == null) {
```
SylvainJuge, line 142: No need to check for tracer state as `transactionOrNull` will be `null` when `GlobalTracer.getTracerImpl()` has already returned `null` in `onEnterHandle(...)`

apm-agent-plugins/apm-jdk-httpserver-plugin/src/main/java/co/elastic/apm/agent/httpserver/HttpHandlerInstrumentation.java
```diff
@@ -0,0 +1,58 @@
...
+
+import static net.bytebuddy.matcher.ElementMatchers.hasSuperType;
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.not;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+
+public class HttpHandlerInstrumentation extends ElasticApmInstrumentation {
+
+    @Override
```
SylvainJuge, line 37: Here it's relevant to override `getTypeMatcherPreFilter` to ensure that it's only applied within package `com.sun.net.httpserver`, otherwise that makes `hasSuperType` matcher expensive. This will make it work only with the implementations that are part of the JDK.

-----

Alternatively, if you aim to extend support to implementations that might be provided through `com.sun.net.httpserver.spi.HttpServerProvider`, this would require to apply instrumentation at runtime when instances are returned by `HttpServerProvider`. This could easily be done in a follow-up PR unless required for your use-case.

apm-agent-plugins/apm-jdk-httpserver-plugin/src/main/java/co/elastic/apm/agent/httpserver/HttpHandlerInstrumentation.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public String getAdviceClassName() {
+        return "co.elastic.apm.agent.httpserver.HttpHandlerAdvice";
+    }
+
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Collections.singleton("jdk-httpserver");
```
SylvainJuge, line 56: Should include `experimental` here as you have documented it as experimental in doc (which is definitely a safe choice as a first step).

apm-agent-plugins/apm-jdk-httpserver-plugin/src/main/java/co/elastic/apm/agent/httpserver/HttpHandlerInstrumentation.java
tobiasstadler: Thank you for your review.

I am fine with only instrumenting `com.sun.net.httpserver` `Handler`. But then the `HttpHandlerTest` passes no longer because of `java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory`. The reason for the exception is that `com.sun.net.httpserver.*` is loaded by the platform classloader which cannot load `org/slf4j/LoggerFactory` and neither can the agent class loader, because `org/slf4j` is not listed in `co.elastic.apm.agent.bci.classloading.IndyPluginClassLoaderParent#agentPackages`. This is not an issue in the release build because `org/slf4j`is shaded.

How should I proceed? Remove the logger? Remove the test?

apm-agent-plugins/apm-jdk-httpserver-plugin/src/main/java/co/elastic/apm/agent/httpserver/HttpHandlerInstrumentation.java
tobiasstadler: Done

apm-agent-plugins/apm-jdk-httpserver-plugin/src/main/java/co/elastic/apm/agent/httpserver/HttpHandlerAdvice.java
tobiasstadler: Done

SylvainJuge:
I've modified a bit the implementation to instrument the `HttpHandler` implementation at runtime, which allows to cover the nominal case where the handler class is not part of the `com.sun.net` nor `sun.net` packages. There are two ways to set the `HttpHandler` on the `HttpContext` and both are covered.

I also managed to reproduce the issue with `ClassNotFoundError`, but it does not appear with the current state of the code, thus I'd say that we can get with it for now and investigate further only if required. The underlying issue seemed related to trying to load one of the advice classes directly within the `PlatformClassLoader`, whereas those advice classes should rather be loaded within the agent `PluginClassLoader` which is here a direct child of `PlatformClassLoader` (the instrumentation target classloader) and can also access most agent classes.

SylvainJuge (approved):

tobiasstadler:
Thank You for working on this.

SylvainJuge:
For the record, the `ClassNotFoundError` is triggered when we move the `HttpHandlerHelper#ensureInstrumented` method to `HttpHandlerAdvice` class and then call it directly from within another advice method body. While direct advice-to-advice is not common as we usually use intermediate or helper classes, this might be convenient to share code between advices.

tobiasstadler:
Thank You!

SylvainJuge:
Thank for your contributions! if you haven't subscribed already you might be interesting by the Elastic Contributor program : https://www.elastic.co/community/contributor

PR #1858, opened 2021-06-14
Instrument javax.servlet.Filter the same way as javax.servlet.FilterChain (#1857)

### LINKED ISSUE #1857: Support for javax.servlet.Filter
Standalone Spark (https://sparkjava.com) uses Filters without a FilterChain (see https://github.com/perwendel/spark/blob/54079b0f95f0076dd3c440e1255a7d449d9489f1/src/main/java/spark/embeddedserver/jetty/JettyHandler.java#L50), so no transaction is started by agent.

`javax.servlet.Filter` should be instrumented in the same way as `javax.servlet.FilterChain`

Manually starting the transaction

tobiasstadler:
## What does this PR do?
Fixes #1857

## Checklist


tobiasstadler:

SylvainJuge (approved):

tobiasstadler:
@SylvainJuge I merged the master branch (which disabled the auto-merge), would you mind merging it?

eyalkoren:
@SylvainJuge this may considerably increase the number of times `ServletApiAdvice` can be invoked. Do you think we should add a quick abort when not required? For example - if there is an active transaction when `ServletApiAdvice#onEnterServletService()` starts - abort early; also make sure `ServletApiAdvice#onExitServletService()` aborts early when not required (e.g. when `onEnterServletService` aborted).

tobiasstadler:
@eyalkoren Something like
```
ElasticApmTracer tracer = GlobalTracer.getTracerImpl();
if(tracer == null || tracer.currentTransaction() != null) {
    return null;
}
```

SylvainJuge:
I did run a few wall-clock benchmarks with & without this change and did not found significant impact.

I did test with Tomcat though, which means that the `ApplicationFilterChain` is a parent of any call to `Filter.doFilter` configured within the application, hence making the Advice exit early as the transaction scope is defined by `ApplicationFilterChain` which extends `FilterChain` (but that might be an implementation detail only for Tomcat`.

Alternatively, for the specific case of Spark, the usage of `Servlet` API is quite non-standard, and maybe adding a proper support for Jetty handler to handle the transaction would also work (albeit it's definitely a bit more work).

tobiasstadler:
@SylvainJuge It is the same for Undertow (But `ApplicationFilterChain` is called `FilterChainImpl`)

tobiasstadler:
So how should we proceed?

tobiasstadler:
Thank You!

PR #1859, opened 2021-06-14
Prevent public API related spans from being recycled

eyalkoren:
## What does this PR do?

In order to work around misuse of the public activation API, related spans will not be recycled. 
This way we can ensure there is no improper usage of recycled-and-still-referenced span objects at the cost of creating some garbage.
## Checklist

I decided not to add anything to the documentation about that, as I think it would only cause confusion.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -93,7 +93,8 @@ pipeline {
               stash allowEmpty: true, name: 'build', useDefaultExcludes: false
               archiveArtifacts allowEmptyArchive: true,
                 artifacts: "${BASE_DIR}/elastic-apm-agent/target/elastic-apm-agent-*.jar,${BASE_DIR}/apm-agent-attach/target/apm-agent-attach-*.jar,\
-                      ${BASE_DIR}/apm-agent-attach-cli/target/apm-agent-attach-cli-*.jar,${BASE_DIR}/target/site/aggregate-third-party-report.html",
+                      ${BASE_DIR}/apm-agent-attach-cli/target/apm-agent-attach-cli-*.jar,${BASE_DIR}/apm-agent-api/target/apm-agent-api-*.jar,\
+                      ${BASE_DIR}/target/site/aggregate-third-party-report.html",
                 onlyIfSuccessful: true
             }
           }
--- a/apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/AbstractSpanInstrumentation.java
+++ b/apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/AbstractSpanInstrumentation.java
@@ -132,6 +132,17 @@ public class AbstractSpanInstrumentation extends ApiInstrumentation {
         }
     }
 
+    public static class InitializeInstrumentation extends AbstractSpanInstrumentation {
+        public InitializeInstrumentation() {
+            super(named("initialize").and(takesArgument(0, Object.class)));
+        }
+
+        @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
+        public static void incrementReferences(@Advice.Argument(0) Object span) {
+            ((AbstractSpan<?>) span).incrementReferences();
+        }
+    }
+
     public static class SetStartTimestampInstrumentation extends AbstractSpanInstrumentation {
         public SetStartTimestampInstrumentation() {
             super(named("doSetStartTimestamp").and(takesArguments(long.class)));
--- a/apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/agent/pluginapi/TransactionInstrumentationTest.java
+++ b/apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/agent/pluginapi/TransactionInstrumentationTest.java
@@ -24,7 +24,7 @@
  */
 package co.elastic.apm.agent.pluginapi;
 
-import co.elastic.apm.agent.AbstractInstrumentationTest;
+import co.elastic.apm.AbstractApiTest;
 import co.elastic.apm.agent.impl.TracerInternalApiUtils;
 import co.elastic.apm.api.ElasticApm;
 import co.elastic.apm.api.Outcome;
@@ -38,7 +38,7 @@ import java.security.SecureRandom;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-class TransactionInstrumentationTest extends AbstractInstrumentationTest {
+class TransactionInstrumentationTest extends AbstractApiTest {
 
     private static final SecureRandom random = new SecureRandom();
 
@@ -193,7 +193,7 @@ class TransactionInstrumentationTest extends AbstractInstrumentationTest {
     public void testAgentPaused() {
         // end current transaction first
         endTransaction();
-        reporter.reset();
+        reporter.resetWithoutRecycling();
 
         TracerInternalApiUtils.pauseTracer(tracer);
         int transactionCount = objectPoolFactory.getTransactionPool().getRequestedObjectCount();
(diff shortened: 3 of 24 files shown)


apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/agent/pluginapi/TransactionInstrumentationTest.java
```diff
@@ -193,7 +193,7 @@ public void startSpan() {
     public void testAgentPaused() {
         // end current transaction first
         endTransaction();
-        reporter.reset();
+        reporter.resetWithoutRecycling();
```
SylvainJuge, line 196: [suggestion/question] I think it could be relevant to add another boolean in `MockReporter` like  `checkDestinationAddress` or `checkUnknownOutcomes` to optionally disable the recycling check and always call `reset()`, because here it's not that we don't want to trigger recycling, it's just that we need to not test that everything is recycled.

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-plugin/src/test/java/co/elastic/apm/agent/springwebflux/AbstractServerInstrumentationTest.java
```diff
@@ -72,6 +72,7 @@ static void stopApp() {
     void beforeEach() {
         assertThat(reporter.getTransactions()).isEmpty();
         client = getClient();
+        disableRecyclingValidation();
```
SylvainJuge, line 75: [question] why do we need to do that for webflux (and same for vertx) ? As far as I know those were properly recycling (with the help from GC in the case of webflux).

apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/agent/pluginapi/TransactionInstrumentationTest.java
eyalkoren: It's the other way around - `reset()` does the assertion and then calls `resetWithoutRecycling()`, I just removed the assertion.
Why do I need it to be recycled and not dropped?

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-plugin/src/test/java/co/elastic/apm/agent/springwebflux/AbstractServerInstrumentationTest.java
eyalkoren: How spans are recycled with WebFlux and Vert.x is not affected by this change. What is affected are the tests that use the public API to get transactions, which then are not recycled (this is what this PR does 🙂 ).
But good point- we want to test recycling especially for WebFlux, so let me see how I can accommodate that (maybe switch to using the internal API instead of the public API)

Jenkinsfile
```diff
@@ -93,7 +93,8 @@ pipeline {
               stash allowEmpty: true, name: 'build', useDefaultExcludes: false
               archiveArtifacts allowEmptyArchive: true,
                 artifacts: "${BASE_DIR}/elastic-apm-agent/target/elastic-apm-agent-*.jar,${BASE_DIR}/apm-agent-attach/target/apm-agent-attach-*.jar,\
-                      ${BASE_DIR}/apm-agent-attach-cli/target/apm-agent-attach-cli-*.jar,${BASE_DIR}/target/site/aggregate-third-party-report.html",
+                      ${BASE_DIR}/apm-agent-attach-cli/target/apm-agent-attach-cli-*.jar,${BASE_DIR}/apm-agent-api/target/apm-agent-api-*.jar,\
+                      ${BASE_DIR}/target/site/aggregate-third-party-report.html",
```
SylvainJuge, line 97: While we are at it, maybe adding an extra link in the readme for the snapshot API would be relevant (https://github.com/elastic/apm-agent-java/#snapshots) We only list other artifacts there.

apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/agent/pluginapi/TransactionInstrumentationTest.java
SylvainJuge: discussed it live, my idea was to only have a single `reset` method that always recycle, but only assert for proper recycling like we do for outcome of span type check : assertion enabled by default, but we can opt-out on a per-test basis. We will leave that for later refactor.

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-testapp/pom.xml
```diff
@@ -37,13 +37,6 @@
     </dependencyManagement>
 
     <dependencies>
-
-        <dependency>
-            <groupId>co.elastic.apm</groupId>
-            <artifactId>apm-agent-api</artifactId>
-            <version>${project.version}</version>
-        </dependency>
-
```
SylvainJuge, line 46: The Webflux webapp is stand-alone, thus should not have dependencies on the agent, thus using the API was required for that.

SylvainJuge (approved):
LGTM, I'm fine with having webflux test app rely on internal API for now, we can revisit this later if required to. 99.99% of the time this app will only be used for unit testing.

Jenkinsfile
eyalkoren: Good idea, although we currently don't upload it, so we should handle separately

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-testapp/pom.xml
eyalkoren: We decided that for now we will keep it as is to make sure we continuously test recycling at all cases.
We can get back to this once there is a need to run as a standalone app.

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/AbstractSpanInstrumentation.java
```diff
@@ -132,6 +132,17 @@ public static Object doCreateSpan(@Advice.FieldValue(value = "span", typing = As
...
     }
 
+    public static class InitializeInstrumentation extends AbstractSpanInstrumentation {
+        public InitializeInstrumentation() {
+            super(named("initialize").and(takesArgument(0, Object.class)));
+        }
+
+        @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
+        public static void incrementReferences(@Advice.Argument(0) Object span) {
+            ((AbstractSpan<?>) span).incrementReferences();
```
felixbarny, line 142: Have you considered instrumenting `AbstractSpanImpl` directly and incrementing the reference there? The advantage is that would also work with older versions of the API.

eyalkoren: If you mean the constructor, then - yes, I considered but preferred this.
As far as I see it - there is no problem to any current API user (besides the one that reported this issue) and all future users will use the newer API version.

felixbarny: 👌

eyalkoren:
Thanks for the honest feedback.

We've seen enough misuse that caused a lot of grief to our community, which made us decide that the performance gains of recycling do not worth the pain, only in the narrow context of public API spans/transactions.
The problem with recycled "dirty" spans is that they may be reused while still containing irrelevant data, which can be very difficult to debug, or even track. 

This PR only makes sure that spans/transactions "touched" by the public API pass away in the natural Java way, which is garbage collection, instead of being recycled. So if your recycling validation would have passed, that would be a bug in this PR that requires fixing 🙂 .

> So what should i do now, just disable the RecyclingValidation like this PR do?

Yes, for the time being.
We intend to put more efforts soon to expose all our testing capabilities better for external plugins developers, so if I am to get something constructive from your feedback, that would be to add an abstract test class for external plugins that disables this check. Would that be a good solution in your opinion?

PR #1862, opened 2021-06-15
Writing only in debug when failing to set JMS property

eyalkoren:


eyalkoren:
This PR is done in order to demote the `Failed to set JMS message property` warning message to debug level altogether. However, it was done due to a report from a user that was apparently using an older version, before we already reduce logging level for `MessageNotWriteableException` in #1715.
So, pending on this user's feedback, we can decide whether this is a required change.
Even if not, we may use it to fix an improper logging done in #1730, where the exception is not treated as a `Throwable`.

apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsMessagePropertyAccessor.java
```diff
@@ -82,11 +82,11 @@ public void setHeader(String headerName, String headerValue, Message message) {
...
             message.setStringProperty(headerName, headerValue);
-        } catch (MessageNotWriteableException e) {
-            logger.debug("Failed to set JMS message property {} due to read-only message", headerName, e);
-        } catch (JMSException e) {
-            logger.warn("Failed to set JMS message property {}. Distributed tracing may not work.", headerName);
-            logger.debug("Detailed error: ", e);
+        } catch (Exception e) {
+            if (logger.isDebugEnabled()) {
+                logger.debug(String.format("Failed to set JMS message property %s. This may indicate a real problem that " +
+                    "can prevent from distributed tracing from working, but it may also be a valid scenario.", headerName), e);
```
SylvainJuge, line 88: [suggestion] We did not added a test for this case when doing this previous `try/catch` block with `MessageNotWriteableException`. It might be worth adding it this time
as it's already the 2cnd time we revisit this code.

SylvainJuge (approved):

apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsMessagePropertyAccessor.java
eyalkoren: Currently we have no indication it is still a problem. Apparently, the user that reported this warning message was using an older version, so the 1.23.0 change to `debug` may be sufficient.
So I will close this one now and we'll fix that if indeed required.

PR #1863, opened 2021-06-15
Minor refactors + enable_experimental_instrumentations config

SylvainJuge:
## What does this PR do?

Backport some changes that were made in the opentelemetry-bridge PR (https://github.com/elastic/apm-agent-java/pull/1631)
that were not directly related to it for easier merge later.

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Http.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Http.java
@@ -27,8 +27,6 @@ package co.elastic.apm.agent.impl.context;
 import co.elastic.apm.agent.objectpool.Recyclable;
 
 import javax.annotation.Nullable;
-import java.net.URI;
-import java.net.URISyntaxException;
 
 public class Http implements Recyclable {
 
@@ -51,7 +49,6 @@ public class Http implements Recyclable {
     /**
      * URL used for the outgoing HTTP call
      */
-    @Nullable
     public String getFullUrl() {
         return url.getFull().toString();
     }
@@ -74,30 +71,9 @@ public class Http implements Recyclable {
      */
     public Http withUrl(@Nullable String url) {
         if (url != null) {
-            String sanitized = sanitize(url);
-            if (sanitized != null) {
-                this.url.appendToFull(sanitized);
-            }
+            this.url.withFull(url);
         }
         return this;
-    }
-
-    @Nullable
-    private static String sanitize(String uriString) {
-        if (uriString.indexOf('@') < 0) {
-            return uriString;
-        }
-        final URI uri;
-        try {
-            uri = new URI(uriString);
-            if (uri.getUserInfo() != null) {
-                return new URI(uri.getScheme(), null, uri.getHost(), uri.getPort(), uri.getPath(), uri.getQuery(), uri.getFragment()).toString();
-            } else {
-                return uri.toString();
-            }
-        } catch (URISyntaxException e) {
-            return null;
-        }
     }
 
     public Http withMethod(String method) {
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Http.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Http.java
@@ -50,9 +50,16 @@ public class Http implements Recyclable {
      * URL used for the outgoing HTTP call
      */
     public CharSequence getUrl() {
-        // note: do not expose the underlying Url object, as it does not have
+        // note: do not expose the underlying Url object, as it might does not have
         // all it's properties set due to providing the full URL as-is
         return url.getFull();
+    }
+
+    /**
+     * @return internal {@link Url} instance
+     */
+    public Url getInternalUrl() {
+        return url;
     }
 
     @Nullable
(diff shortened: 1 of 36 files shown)


SylvainJuge:
I thought that having the "log once" logger disable all log levels once it has already been used would have helped reuse it.

However, when trying to use it with `co.elastic.apm.agent.servlet.ServletVersionInstrumentation` that does not work as two log statements are issued, and that might introduce a small concurrency bug as "checking the logger state" and "issuing log statement" are not done in an atomic operation.

felixbarny:
> "checking the logger state" and "issuing log statement" are not done in an atomic operation.

That _is_ done in an atomic operation, see:

```java
if (alreadyLogged.compareAndSet(false, true)) {
    delegate.debug(format, arg);
}
```

felixbarny:
Note that this PR removes the special value `experimental` from `disable_instrumentations`. This should be added to the changelog. More details here: https://github.com/elastic/apm-agent-java/pull/1631#issuecomment-768905415

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Url.java
```diff
@@ -184,6 +185,35 @@ public Url withSearch(@Nullable String search) {
...
+        if (pathname == null) pathname = url.getPath();
+        if (search == null) search = url.getQuery();
+        int port = url.getPort();
+        port = port > 0 ? port : url.getDefaultPort();
+        if (port > 0) {
+            withPort(port);
+        }
+    }
+
+    public void fillFullUrl() {
```
felixbarny, line 200: There's overlap with `updateFull`

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
```diff
@@ -965,7 +965,7 @@ private void serializeHttpContext(final Http http) {
             if (statusCode > 0) {
                 writeField("status_code", http.getStatusCode());
             }
-            writeLastField("url", http.getUrl());
+            writeLastField("url", http.getUrl().getFull());
```
felixbarny, line 968: I wonder if we need to call `fillFullUrl` or `updateFull` to avoid the `url` to be serialized to an empty string. Do we have tests for that case?

SylvainJuge:
I've found a few other related changes that seemed relevant to include:
- remove duplication of the `full` url field writing logic
- embed the URL sanitation directly into `Url` to avoid leaking credentials

----

There are two usages of `Url` object:
- to be used as a string (1): only the value of `getFull()` is used and serialized
- to be serialized as an object with separate properties (2) : all properties are used

For (1), there are two cases:
- set individual properties, then call `Url#updateFull()`, then `getFull()` returns the computed Url string
- use `Url#withFull(...)` with the raw URL value, then `getFull()` returns the same value, optionally cleaned of any user credentials, thus there is no parsing nor allocations most of the time.

For (2), there are two variants:
- set individual properties, then call `Url#updateFull()`
- use `Url#fillFromUri(URI)` or `Url#fillFromUrl(URL)`

With those, all the properties are populated including the `full` property.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
SylvainJuge: We don't have a test for that, and I wonder if we could simply throw an `IllegalStateException` when calling `getFull` when the `full` attribute is empty to indicate those errors:
- individual properties have been properly set, but `updateFull()` has not
- there was no call to `withFull` with a raw string value

SylvainJuge:
Also, while now using `Url` object for when it's meant to be serialized as string (for example in the `Http` class), I think it might be a good idea to not expose the `Url` object directly, but instead only the string representation. Otherwise when serializing it would be easy to assume that because an `Url` object is provided all of it's fields are available, which is just not true when a raw string value is used.

felixbarny:
In `co.elastic.apm.agent.impl.context.Url#getFull` we could also check if the `full` is empty and update `full` if so.

SylvainJuge:
> In `co.elastic.apm.agent.impl.context.Url#getFull` we could also check if the `full` is empty and update `full` if so.

Yes, very good idea !

In that case we can even remove the required calls to `updateFull` and lazily call it internally when `getFull` is called and the `full` property is empty but the `Url` object has some of its properties set.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Http.java
```diff
@@ -52,8 +49,16 @@
     /**
      * URL used for the outgoing HTTP call
      */
-    @Nullable
-    public String getUrl() {
+    public CharSequence getUrl() {
+        // note: do not expose the underlying Url object, as it might does not have
```
felixbarny, line 53:
```suggestion
        // note: do not expose the underlying Url object, as it might not have
```

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Url.java
```diff
@@ -91,36 +98,62 @@ public StringBuilder getFull() {
...
+                // silently ignore URIs that we can't parse
+            }
+            if (uri != null) {
+                fillFrom(uri);
+            }
+        }
         return this;
     }
 
+    private static boolean urlNeedsSanitize(CharSequence sequence) {
```
felixbarny, line 148:
```suggestion
    private static boolean urlNeedsSanitization(CharSequence sequence) {
```

felixbarny (approved):

OrangeDog:
The release notes say

> - Set `disable_instrumentations=true` to enable experimental instrumentations

eyalkoren:
I will fix shortly. Thanks!

PR #1866, opened 2021-06-16
Printing info for NullCheck logger initialization error

eyalkoren:
This is currently only for analysis purposes. We'll see if we use it eventually (with modifications) or close altogether after we get some feedback.


eyalkoren:
@SylvainJuge let's discuss this offline. The `info` log will go for sure (this is the `todo` I left there), but other than that, we'll need to evaluate this print to std err.

SylvainJuge:
I remember that we have a few other places where we write directly to `System.err` and `System.out`, in order to help diagnose cases when it happens, adding a common prefix like `[elastic-apm-agent] {level}` where `{level}` = `ERROR`, `WARN`, `INFO` to those lines could help grep the log files when that happens (and also allows to add this to the troubleshooting documentation).

SylvainJuge (approved):
do we want to keep the `printStackTrace()` call in the final version, or was it just required to debug the related issue ? I'm not sure it's relevant to keep that in production code now that we have an explicit error message when it happens.

eyalkoren:
> do we want to keep the printStackTrace() call in the final version

Since this error logging is for debugging class initialization issue, I say we do. When it happens, we may want to see what caused this class loading as it may be related to the root cause.

PR #1868, opened 2021-06-16
[DOC] Details about the body capture maximum size.

lucabelluccini:
## What does this PR do?

Adds more details about the body capture.

According to https://github.com/elastic/apm-agent-java/blob/a040f5ad380a5bd401f26ccebdf71b61a2630970/apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java#L102 we limit the body capture to 10k characters long.

The unit test shows the body will be truncated: https://github.com/elastic/apm-agent-java/blob/d7de9d2298c71c9dc5bc3ff55001b4a72f40f7fd/apm-agent-plugins/apm-servlet-plugin/src/test/java/co/elastic/apm/agent/servlet/TestRequestBodyCapturing.java#L165

## Checklist

- [x] This is a documentation change
  - [ ] ~I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)~ Not worth a changelog entry

To the reviewer:
- [ ] Please review if this is correct
- [ ] Please backport to applicable versions

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -296,6 +296,9 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
             "\n" +
             "This option is case-insensitive.\n" +
             "\n" +
+            "NOTE: Currently, the body length is limited to 10000 characters and it is not configurable." +
+            "If the body size exceeds the limit, it will be truncated." +
+            "\n" +
             "NOTE: Currently, only UTF-8 encoded plain text HTTP content types are supported.\n" +
             "The option <<config-capture-body-content-types>> determines which content types are captured.\n" +
             "\n" +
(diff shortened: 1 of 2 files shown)


eyalkoren (changes requested):
Useful addition, thanks @lucabelluccini 🙏 
Please note that `configuration.asciidoc` is created automatically though.
So the way to apply add something to configuration docs is to edit the description in the configuration class and then run the `ConfigurationExporterTest` to apply the changes to the docs.
In this case, you would need to add it to the description in https://github.com/elastic/apm-agent-java/blob/master/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java#L287-L307 and run the test.

The build will fail otherwise because there is a diff between the automatically created `configuration.asciidoc` and the existing version.

lucabelluccini:
😮 TIL for me - sorry about that.

lucabelluccini (commented):
Should I amend the previous changes to the `asciidoc` or I can keep them?

eyalkoren:
> Should I amend the previous changes to the asciidoc or I can keep them?

Start by reverting the asciidoc.
Then run `ConfigurationExporterTest` locally. It will fail, but will update the asciidoc.
Then run it again and it should pass.
After you see it pass locally, merge both changes (class and asciidoc) and then the build should pass.

Another option - I can do that if you had enough already 🙂 

Sorry for making it so complicated for you..

PR #1871, opened 2021-06-18
Fixes NPE on Webflux headers

### LINKED ISSUE #1869: NPE when use webflux
apm agent version : 1.24.0
spring cloud gateway version : 2.2.8

## Steps to reproduce

I use apm in spring cloud gateway，every time a request passes through the gateway, a npe is throw。
I found `co.elastic.apm.agent.impl.transaction.TraceContext.FROM_TRACE_CONTEXT_TEXT_HEADERS#asChildOf` has the following code , it need a request header : `tracestate`. 

                if (isValid) {
                    // as per spec, the tracestate header can be multi-valued
                    traceContextHeaderGetter.forEach(TRACESTATE_HEADER_NAME, carrier, child, TRACESTATE_HEADER_CONSUMER);
                }
I try to set the request header myself，it will not report an error。
Do I need to set up this request header myself？
## Debug logs

> java.lang.NullPointerException
	at co.elastic.apm.agent.springwebflux.HeaderGetter.forEach(HeaderGetter.java:44)
	at co.elastic.apm.agent.springwebflux.HeaderGetter.forEach(HeaderGetter.java:33)
	at co.elastic.apm.agent.impl.transaction.TraceContext$4.asChildOf(TraceContext.java:155)
	at co.elastic.apm.agent.impl.transaction.TraceContext$4.asChildOf(TraceContext.java:132)
	at co.elastic.apm.agent.impl.transaction.Transaction.start(Transaction.java:125)
	at co.elastic.apm.agent.impl.ElasticApmTracer.startChildTransaction(ElasticApmTracer.java:191)
	at co.elastic.apm.agent.impl.ElasticApmTracer.startChildTransaction(ElasticApmTracer.java:182)
	at co.elastic.apm.agent.impl.GlobalTracer.startChildTransaction(GlobalTracer.java:94)
	at co.elastic.apm.agent.springwebflux.WebfluxHelper.getOrCreateTransaction(WebfluxHelper.java:59)
	at co.elastic.apm.agent.springwebflux.DispatcherHandlerInstrumentation$HandleAdvice.onEnter(DispatcherHandlerInstrumentation.java:77)
	at org.springframework.web.reactive.DispatcherHandler.handle(DispatcherHandler.java:139)

<details>
<summary>Click to expand</summary>

```
replace this line with your debug logs
```
</details>

SylvainJuge:
## What does this PR do?

Fixes #1869

Adds a few tests to other Text-based header protocols

## Checklist

- [x] This is a bugfix


apm-agent-plugins/apm-apache-httpclient-plugin/src/test/java/co/elastic/apm/agent/httpclient/helper/RequestHeaderAccessorTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.http.message.BasicHttpRequest;
+import org.junit.jupiter.api.Nested;
+
+import java.util.List;
+import java.util.Map;
+
+class RequestHeaderAccessorTest {
+
+    @Nested
+    class GetterTest extends AbstractTextHeaderGetterTest<RequestHeaderAccessor, HttpRequest> {
```
eyalkoren, line 39: Why is this required?

eyalkoren (approved):
Looks like a useful infrastructure 👍 
One minor comment

apm-agent-plugins/apm-apache-httpclient-plugin/src/test/java/co/elastic/apm/agent/httpclient/helper/RequestHeaderAccessorTest.java
SylvainJuge: It's not really needed, but as we only test for the `Getter` part it made more sense to me this way:
- the abstract test works for all `TextHeaderGetter` implemenations, including those that are not `TextHeaderSetter`
- we can only inherit a single class, thus if we were to add an `AbstractTextHeaderSetterTest` that would be required to test for the `Setter` part.

But I agree that's not required with the current state. Do you think it's worth changing that or can we merge as-is ?

eyalkoren: Either way can work, but generally speaking - I prefer not having non-used code only for potential future usage. Maybe the person who adds these setter tests in the future will add them to the same abstract class you created and call it `AbstractTextHeaderAccessorTest`, in which case it will remain irrelevant.

> Do you think it's worth changing that or can we merge as-is ?

Since it's super-easy to change, I'd vote for that, but the PR is approved either way, so no need to.

Issue #1635, opened 2021-01-29
RabbitMQ follow up: support Spring AMQP's SimpleMessageListenerContainer


felixbarny:
See also https://github.com/elastic/apm-agent-java/pull/1328#issuecomment-769186851

kananindzya:
Hi @felixbarny , I created a simple project with `SimpleMessageListenerContainer` and cannot reproduce this case.
Code you can find here (https://github.com/kananindzya/apm-rabbitmq-spring)

![image](https://user-images.githubusercontent.com/48118512/107157130-406dfd80-6937-11eb-8f55-91372a71bb88.png)

eyalkoren:
@kananindzya maybe this reproduces the descried problem... 

@fgabolde can you please provide a more detailed description of what you were missing? Did you not get any `SEND` spans before? Or were there `SEND` spans but there was no distributed tracing? Does the screenshot above look like the incomplete support you got at first?
If you can add a screenshot of how a proper trace should look like in the description of this issue as the **expected result**, that would be great.

fgabolde:
Hi, sure, I could have been clearer in my original comment.

I had two separate Spring Boot applications, one sending and one receiving. The receiver was initially set up with a SimpleMessageListenerContainer, and a service with a decorated method:

```java
  @RabbitListener(queues = "SomeQueueName")
  public void processMessage(SomePojo message) {
    doSomethingThatGeneratesMoreSpans();
  }
```

I'm not sure this is significantly different from your test code. I don't believe it is (but then, there is a lot of magic in Spring).

* Sender app: all traces present, including the `SEND` span
* Receiver app: no traces present *except* for the `RECEIVE` span

The RECEIVE span was related to the appropriate transaction, so in my waterfall I had something like

```
Sender:   incoming HTTP request
Sender:     DB query
Sender:       SEND
Receiver:       RECEIVE
```

After this I should have had more spans in the Receiver app, e.g. calls to other HTTP APIs, S3, etc.

I guessed that the context was lost somewhere by logging the output of `ElasticApm.currentTransaction()` and got "INSTANCE", which looked to me like `NoopTransaction`.

When I switched to the DirectMessageListenerContainer, after reading that the threading model was different, I got all the missing spans back.

Hope this is clearer!

fgabolde:
[...]

PR #1872, opened 2021-06-21
Adding special wrapper for ChannelAwareMessageListener

eyalkoren:
## What does this PR do?

A simple fix attempt for https://github.com/elastic/apm-agent-java/issues/1635#issuecomment-864595814
## Checklist

- [ ] This is a bugfix


eyalkoren:
@kananindzya if it's easy enough, please add your test to this PR - reproduce when disabling this change and verify that this fix works as expected. Otherwise, if you think there is a better way to fix - please suggest.
Lastly, do you think #1716 should be adjusted accordingly as well?

eyalkoren:
@kananindzya thanks for the test!
Can you check why `RabbitListenerTest` fails now?

kananindzya:
> @kananindzya thanks for the test!
> Can you check why `RabbitListenerTest` fails now?

I sent fix for this https://github.com/eyalkoren/apm-agent-java/pull/3

SylvainJuge (approved):

PR #1874, opened 2021-06-21
Remove license header template

eyalkoren:
Apparently, with the former license plugin there is no way to lose the automatically-injected [template](https://www.mojohaus.org/license-maven-plugin/header.html) at the top of the license header, so I had to use a [different plugin](https://github.com/mathieucarbou/license-maven-plugin).
We still use `org.codehaus.mojo:license-maven-plugin` in order to create the third-party license report.


SylvainJuge:
@eyalkoren can you provide the command that you used to update existing headers ? That might be useful for all open branches that add new files, or when dealing with conflicts.

eyalkoren:
@SylvainJuge nothing changed on this regard - like before it's done during the `process-sources` phase, so `mvn compile` should do.

eyalkoren:
Only really interesting things to review here are the [general build changes](https://github.com/elastic/apm-agent-java/blob/1c148cd42b6661163206132135bce687791c12ec/pom.xml#L375-L404), the [gRPC sources build](https://github.com/elastic/apm-agent-java/blob/1c148cd42b6661163206132135bce687791c12ec/apm-agent-plugins/apm-grpc/pom.xml#L98-L109) and the [third-party list creation command](https://github.com/elastic/apm-agent-java/blob/1c148cd42b6661163206132135bce687791c12ec/Jenkinsfile#L91).
The sooner we merge this, the less painful it would be.

SylvainJuge (approved):

PR #1883, opened 2021-06-23
Propagating trace context in HTTP requests without spans

eyalkoren:
## What does this PR do?

Implementing the [context-propagation spec update](https://github.com/elastic/apm/pull/464) - adding the ability to propagate trace context on HTTP requests for which a span is not created.
For example - this enables to set the trace context headers on Elasticsearch REST client queries.

## Checklist

### DIFF
--- a/apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/helper/HttpAsyncRequestProducerWrapper.java
+++ b/apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/helper/HttpAsyncRequestProducerWrapper.java
@@ -19,6 +19,7 @@
 package co.elastic.apm.agent.httpclient.helper;
 
 import co.elastic.apm.agent.http.client.HttpClientHelper;
+import co.elastic.apm.agent.impl.transaction.AbstractSpan;
 import co.elastic.apm.agent.impl.transaction.Span;
 import co.elastic.apm.agent.impl.transaction.TextHeaderSetter;
 import co.elastic.apm.agent.objectpool.Recyclable;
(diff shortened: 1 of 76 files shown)


apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/helper/HttpAsyncRequestProducerWrapper.java
```diff
@@ -58,26 +85,39 @@ public HttpHost getTarget() {
...
+            if (span != null) {
+                RequestLine requestLine = request.getRequestLine();
+                if (requestLine != null) {
+                    String method = requestLine.getMethod();
+                    span.withName(method).appendToName(" ");
+                    span.getContext().getHttp().withMethod(method).withUrl(requestLine.getUri());
+                }
+                span.propagateTraceContext(request, headerSetter);
+            } else if (parent != null) {
+                parent.propagateTraceContext(request, headerSetter);
```
SylvainJuge, line 102: [question] does the case where `span != null` is the one where an HTTP client is used directly and `parent != null` when an high level client on top of HTTP client like ES client is used ? If so, why are mutually exclusive ? 

For example, what happens if the parent span is an internal app span that does not allows to deal with any context propagation headers ? In the context of ES high-level client, both might set HTTP headers, but only the child span should propagate the context.

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientAsyncInstrumentation.java
```diff
@@ -61,27 +61,32 @@ public String getAdviceClassName() {
...
                 OkHttpClientHelper.computeHostName(url.getHost()), url.getPort());
+
+            Request.Builder builder = originalRequest.newBuilder();
             if (span != null) {
                 span.activate();
-                Request.Builder builder = originalRequest.newBuilder();
                 span.propagateTraceContext(builder, OkHttpRequestHeaderSetter.INSTANCE);
                 request = builder.build();
                 callback = CallbackWrapperCreator.INSTANCE.wrap(originalCallback, span);
+            } else {
```
SylvainJuge, line 87: [minor] maybe add a comment here (and in similar code) to explain that here `span == null` means that there is an exit span already active, and thus that we still need to propagate context of the parent.

SylvainJuge (commented):
From what I understand here, when an active exit span is already active (likely created by an higher level HTTP client like ES client), we propagate the context of the parent span in place of the low-level HTTP client.

So, here we assume that the direct parent span that is active when executing the HTTP transaction is the exit span for which we need to propagate context. I think there are cases where it might not be true when > 2 layers are involved: for example having an Spring-something high level abstraction for ES that creates exit spans, which delegates to ES client, which in turn delegates to HTTP client.

In that case, at the HTTP client, the context that should be propagated is the grand-parent, not the direct parent.

eyalkoren:
> So, here we assume that the direct parent span that is active when executing the HTTP transaction is the exit span for which we need to propagate context. I think there are cases where it might not be true when > 2 layers are involved: for example having an Spring-something high level abstraction for ES that creates exit spans, which delegates to ES client, which in turn delegates to HTTP client.
>
> In that case, at the HTTP client, the context that should be propagated is the grand-parent, not the direct parent.

I didn't get that. If there we decided not to create an HTTP span for whatever reason, we propagate the context of the parent and make it non-discardable. This means that the parent span will be reported and the child transaction should be its child. 
Regardless, if I understand your example properly - it should not occur, as we do not exit spans as children of exit spans, but even if we would, I don't understand why it makes more sense to make the child transaction a sibling of the active exit span and not its child.

apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/helper/HttpAsyncRequestProducerWrapper.java
eyalkoren: Sorry, I missed that as well. 
> what happens if the parent span is an internal app span that does not allows to deal with any context propagation headers ?

I don't know what "span that does not allows to deal with any context propagation headers" means.

Let's assume the parent span set context headers based on its trace context and then this HTTP client is invoked. If it decides to create a span - it overrides the context headers with its own (as it should). If it decides not to create a span, it overrides the headers with ones with the same value, which sounds right as well.

Sorry if I didn't get what you had in mind...

apm-agent-plugins/apm-okhttp-plugin/src/main/java/co/elastic/apm/agent/okhttp/OkHttpClientAsyncInstrumentation.java
eyalkoren: It's basically true now, but conceptually it can be something else. If, for whatever reason we decide no to create spans here, as long as we want to support distributed tracing, we need to propagate _some_ trace context headers, and in order for distributed tracing to work, it needs to be the parent's trace context, which we set as non-discardable.
If the parent is not an exit span, it will not be perfect and things like service map will not be populated properly, so we will have to see what we do with that if we find such, but for the purpose of context propagation - I believe this is what we should do.

SylvainJuge (approved):

PR #1889, opened 2021-06-29
Emulate recent server for benchmarks

SylvainJuge:
## What does this PR do?

Fixes a recent increase of memory allocation rate after merging PR #1305 
This was in fact due to integer serialization as a string (for `Url#port`), the extra allocation will only be required for < 7.0.0 servers.


eyalkoren (approved):
Elegant!

PR #1891, opened 2021-06-29
Remove extra allocation for legacy servers

SylvainJuge:
## What does this PR do?

Improvement on https://github.com/elastic/apm-agent-java/pull/1889
Removes extra allocation for legacy servers.

Thanks @felixbarny for the idea.


eyalkoren (approved):
I would actually go with the second part of @felixbarny's suggestion - not optimising for APM server < 7.0...

However, if you tested it manually with a 6.x server, then let's merge.

Issue #712, opened 2019-07-08
Support for serverless functions


eyalkoren:
See the discussion at https://discuss.elastic.co/t/apm-with-aws-lambda-java/185936.

~~Currently implementation should at least include addition of a blocking flush API. Maybe some work around metrics.~~
(It is likely that we would want to work on native support for serverless, rather than offering something that requires specific API for that, but we'll think about implementation when we get to it). 

In addition, add some documentation on how to enable:
- using the programmatic attach API
- provide tools.jar
- a word of warning about overhead and maybe tuning
- caveats (metrics?)

RajatArora08:
Hi @eyalkoren! Is this planned for any release yet?

eyalkoren:
@RajatArora08 thanks for expressing your interest!
It is not planned for any specific release yet, but it is definitely something we are actively looking into, not specifically for the Java agent and not specifically for AWS lambda, but trying to figure out the best way to support server-less architectures across our entire Observability offering. 
Keep tracking this issue for now, so you should get notified when we get to it in the Java agent.

gamefundas:
@eyalkoren it has been a while since this was opened, has there been any progress in this area for serverless observability. I was considering elastic apm for our application and support of AWS lambda is quite important. Came across this and few other threads to learn that currently there is no official support available. Is this true?

eyalkoren:
Not a lot to report at the moment, only [this Python agent PR](https://github.com/elastic/apm-agent-python/pull/716) as a partial POC of one approach.

remmeier:
`a word of warning about overhead and maybe tuning` 

is maybe the main point to address in order to allow serverless functions to start quickly. For example, our non-servlerless services start in about 2-3 seconds. When applying the apm agent, startup times double to 4-5 seconds. A general direction for startup optimizations is to move more things from runtime to build-time. e.g. graal native creates binaries to replace JIT. micronauts/dagger do dependency injection at compile-time. With these tools applications can start in milliseconds. Would be great to allow the integration of elastic with such tech stacks.

One angle could be be the use cases covered by the library:

- publishing of data to apm server
- openapi interface
- integration into the various tech stacks (servlets, jdbc, jpa, etc.)
- apm to inject the various integrations into a running application

[...]

PR #1897, opened 2021-07-05
Add improved hardFlush and softFlush methods

felixbarny:
## What does this PR do?

`softFlush()`
- Waits until all pending events have been processed
- Performs a synced_flush on the gzip buffer and flushes all buffers to the network
- Keeps the HTTP request alive
- Returns immediately if the connection to APM Server is unhealthy (when there's a backoff)
- Allocation free

Intended for serverless instrumentations
- Avoids that every lambda invocation results in a separate HTTP request, repeating the metadata
- Ensures that the data is written out to APM Server before the lambda returns
- As the lambda may stay frozen for some time, non-flushed events may get lost.
- Reduces the blocking time at the end of the lambda compared to a `hardFlush()` as we don't need to wait for the HTTP response from APM Server.

Avoid sending events via a closed connection
- Issue: APM Server may terminate the HTTP request after the request timeout has expired. This may happen when the lambda is frozen for a while.
- Solution:
  - Before serializing events into the HTTP output stream, the agent checks if the request has been open for `apm_request_time`. If so, it ends the request and starts a new one.
  - Issue: When ending a request that APM Server has already closed, socket exceptions will occur.
    - Solution:
      - If the request has been open for `apm_request_time + 1 sec`, and all events have been flushed, the agent ignores socket exceptions when ending the HTTP request and assumes all events were successfully sent to APM Server.

Testing TODOs
- APM Server receives events after calling `softFlush` (so far only verified in unit tests)
- APM Server behavior when HTTP connection gets frozen for a short time
- APM Server behavior when HTTP connection gets frozen for a long time so that the request timeout elapses.
  - Are the events received so far sent to ES?
  - Does the cloud proxy buffer or stream events to APM Server? If it buffers, do the events get lost?
  - How would the logs look like? Can we indicate that that's expected behavior?
- Agent behavior after lambda thaws and the connection has been closed by APM Server in the meantime.
  - Does this trigger a backoff?
  - How would the logs look like? Can we indicate that that's expected behavior?

It seems easier to test these things after merging this PR. As we don't expose any public API, we can remove things if they don't work out.

Relates to https://github.com/elastic/apm-agent-java/issues/712

## Checklist

  - ~[ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)~


apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java
```diff
@@ -155,83 +170,80 @@ public long getReported() {
...
-     * <p>
-     * This method may block for a while until a slot in the ring buffer becomes available.
-     * </p>
-     *
-     * @return A {@link Future} which resolves when the flush has been executed.
-     */
     @Override
     public Future<Void> flush() {
-        final boolean success = disruptor.getRingBuffer().tryPublishEvent(FLUSH_EVENT_TRANSLATOR);
+        CompletableVoidFuture future = new CompletableVoidFuture();
```
eyalkoren, line 175: Since we now have a mechanism to wait for an event to be processed based on the sequence, the only benefit of the returned `Future` is the ability to cancel. If we decide that cancelling is not that important, we can considerably simplify things (no need for the Java-7-compatible `CompletableFuture` and no need to use event translators that get `null` futures.

eyalkoren (commented):
LGTM, as much as possible with a static review for this...
A few fairly minor comments.

Regarding the test of:
> Agent behavior after lambda thaws and the connection has been closed by APM Server in the meantime.

Is it possible to test through the existing integration test by setting a short timeout on the handler and waiting for longer within the test?

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java
felixbarny: We do need the canceling for [`api_request_time`](https://www.elastic.co/guide/en/apm/agent/java/current/config-reporter.html#config-api-request-time).
If we end the request due to `api_request_size` before `api_request_time` elapses, we need to cancel the `timeoutTask` which, in turn, cancels the future.

felixbarny (commented):
> Is it possible to test through the existing integration test by setting a short timeout on the handler and waiting for longer within the test?

Good idea, yes, that should be doable

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
```diff
@@ -21,14 +21,17 @@
 import co.elastic.apm.agent.report.processor.ProcessorEventHandler;
 import co.elastic.apm.agent.report.serialize.PayloadSerializer;
 import co.elastic.apm.agent.premain.ThreadUtils;
+import com.lmax.disruptor.Sequence;
```
eyalkoren, line 24:
```suggestion
```
imports :ocd

felixbarny:
> Is it possible to test through the existing integration test by setting a short timeout on the handler and waiting for longer within the test?

I tried that using `Undertow.builder().setServerOption(UndertowOptions.REQUEST_PARSE_TIMEOUT, 500)` but it didn't seem to have any effect. I suppose it's best to test that with a real APM Server deployed on ESS.

eyalkoren:
> I tried that using Undertow.builder().setServerOption(UndertowOptions.REQUEST_PARSE_TIMEOUT, 500) but it didn't seem to have any effect. I suppose it's best to test that with a real APM Server deployed on ESS.

When you call `endExchange` in the test handler, doesn't it close the connection already? If the test does a `softFlush`, waits a bit (preferably until the reporter confirms that the connection was closed), then reports an additional transaction and invokes another `softFlush` - would that mock what we want?

felixbarny:
I don't think so because it would just send a 200 response back instead of aborting the request.

eyalkoren:
And what about explicitly closing the exchange, or the `ServerConnection` you can get from it?

felixbarny:
Closing the ServerConnection worked!

The issue now is that the next `END_REQUEST` event will trigger a `SocketException` and thus a backoff.
I think what we'll want to do is to have a mode where we can ignore non-HTTP errors on `endRequest()`:

In other words, conditionally execute the `onRequestError(-1, connection.getErrorStream(), e);` path.

https://github.com/elastic/apm-agent-java/blob/d0e00baf62f7443efa533ee9542c7f55a8c8dcb4/apm-agent-core/src/main/java/co/elastic/apm/agent/report/AbstractIntakeApiHandler.java#L155-L161

That is because, on Lambda, it's expected that the request may be ended by APM Server. We'd still catch `SocketException`s and the like and trigger a backoff when failing to start a request.

I'm not yet sure if a config option is a good option. Maybe the Lambda instrumentation should switch that flag on the reporter when it initalizes.

eyalkoren:
Awesome, so that already proved useful.
Not sure it should be a configuration, maybe enough as a reporter flag/state that should certainly be set automatically by the Lambda instrumentation

apm-agent-core/src/main/java/co/elastic/apm/agent/report/AbstractIntakeApiHandler.java
```diff
@@ -156,18 +166,36 @@ public void endRequest() {
...
+                    // we don't want to trigger a backoff or log an error if we have already flushed all data previously
+                    // note: flushing the OutputStream of a closed connection does not necessarily lead to an exception.
+                    // only when closing the connection, the exception occurrs
+                    // use case: on AWS Lambda, we flush after every request
+                    // and the environment may freeze for a long period of time so that the APM Server closes the connection
+                    if (!pendingFlush || clock.nanoTicks() >= significantlyOverApiRequestTime) {
+                        onRequestSuccess();
+                    } else {
+                        onRequestError(-1, connection.getErrorStream(), e);
+                    }
```
eyalkoren, line 182: This contains multiple heuristics:
1. that the `api_request_time` value is close to the `read_timeout` setting in the APM server
2. that one second is "significantly over API request time"
3. that crossing this significantly higher-than-request-duration can be considered as a successful request

While we can certainly live with heuristics if needed, since there is no real use for this functionality at the moment, I suggest we leave it out.

We can still consider leaving the `softFlush` functionality (without assuming freezes), but I don't see a real use case for that at the moment, especially since there are no guarantees that the flushed bytes are also flushed from the local buffer.

eyalkoren (commented):
I think that we should remove the `softFlush` complications if we don't see current use for it.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java
```diff
@@ -155,83 +168,79 @@ public long getReported() {
...
+            try {
+                long sequence = disruptor.getRingBuffer().tryNext();
+                try {
+                    eventTranslator.translateTo(disruptor.get(sequence), sequence, Thread.currentThread());
+                } finally {
+                    disruptor.getRingBuffer().publish(sequence);
                 }
-                if (!isDone()) {
-                    throw new TimeoutException();
+                return waitForEventProcessed(sequence, thresholdNs);
```
eyalkoren, line 205: This creates a race condition- the event has already been published and potentially processed by the disruptor handler thread (including attempting to `unpark`) before this thread gets the chance to `park`.

felixbarny: I don't think so. Note that `waitForEventProcessed` checks if the sequence has already been processed. It's always required to check the condition after park returns as a thread may be unparked for random reasons.

eyalkoren: Right! I missed the check in the for-loop stop condition, only saw the one outside the for-loop 👍

eyalkoren:
@felixbarny what do you say we revive this PR, removing the `softFlush`? The `hardFlush` is a big improvement from the current `flush`, but it should just replace the current implementation without changing the API.

felixbarny:
SGTM

PR #1898, opened 2021-07-05
Auto-inference of destination.service.resource

eyalkoren:
## What does this PR do?

Closing #1848 

## Checklist

  - [x] I added the test JSON spec to the APM repo


eyalkoren:
I started also implementing the addition of the creation of exit spans, as described in the spec, but I think it doesn't make a lot of sense in the Java agent, as we don't expose any API to add custom data. A new API for exit span creation will look something like `startExitSpan(String type, String subtype, String action)`, which means that the `type` or `subtype` will be used for auto-inference of the `service.resource`, however this is essentially a capability we already offer through the API to set the service resource explicitly. @felixbarny WDYT?

felixbarny:
> however this is essentially a capability we already offer through the API to set the service resource explicitly

The difference is that users don't need to remember setting resource fields or worry about which value to set in order for the span to show up in the service map.

apm-agent-core/src/test/resources/json-specs/service_resource_inference.json
```diff
@@ -0,0 +1,183 @@
+[
```
felixbarny, line 1: 😍

eyalkoren:
> The difference is that users don't need to remember setting resource fields or worry about which value to set in order for the span to show up in the service map.

Users will have to remember this new dedicated API and not the "default" one, and I think it's actually better to force them to think what the resource would be, instead of implicitly use `type` or `subtype`.

eyalkoren:
As decided offline with @felixbarny - the exit span creation API is added.

PR #1906, opened 2021-07-08
Use path when high-level framework method is unknown

SylvainJuge:
## What does this PR do?

HTTP Transaction naming can have multiple variants:
- when using `use_path_as_transaction_name` we should use the request path and apply `url_groups` to limit cardinality
- when using a high-level framework, the `ClassName#methodName` is preferred when `use_path_as_transaction_name` is `false`

Currently, the behavior when method name is not known, which happens with some Spring controllers, is to simply use the `ClassName`, which makes many distinct transactions be gathered under a common name.
Thus, in the case where the method name is `null`, using the path is often a better fallback.

## Checklist


tobiasstadler:
@SylvainJuge I think `TransactionNameUtils#setNameFromHttpRequestPath` can also be used in `co.elastic.apm.agent.httpserver.HttpHandlerAdvice`

SylvainJuge:
I've added `ResourceHttpRequestHandler` as a known exception, and thus name is not changed, we could also apply other naming schemes, we'll likely revisit this later if there is a need, a few ideas:
- `GET static resource` for all static resources, which provide a bit more granularity with HTTP verb, could be reused for other frameworks too like `unknown route`.
- `GET *.js` or `GET *.css` : apply extension-based grouping automatically,

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/springwebmvc/SpringTransactionNameInstrumentation.java
```diff
@@ -114,19 +121,23 @@ public static void setTransactionName(@Advice.Argument(2) Object handler) {
...
+                // when method name is not known, using path is a better default as multiple distinct requests
+                // would be handled by a single class. If delegating to a Servlet, then servlet naming will be applied
+                // thus it allows to provide a better fallback in case method name is null and there is no servlet executed.
+                //
+                // ResourceHttpRequestHandler is handling static resources, thus rather high cardinality most of the time
+                // and grouping on a single transaction is preferable.
+                TransactionNameUtils.setNameFromHttpRequestPath(request.getMethod(), request.getServletPath(), request.getPathInfo(), transaction.getAndOverrideName(PRIO_LOW_LEVEL_FRAMEWORK + 1), webConfig.getUrlGroups());
+            } else {
+                TransactionNameUtils.setNameFromClassAndMethod(className, methodName, transaction.getAndOverrideName(PRIO_HIGH_LEVEL_FRAMEWORK));
+            }
```
eyalkoren, line 135: I don't think we need to check `webConfig.isUsePathAsName()`. Instead, the algorithm should be:
```
StringBuilder transactionName;
if (methodName == null) {
  transactionName = getAndOverrideName(PRIO_LOW_LEVEL_FRAMEWORK + 1, false);
} else {
  transactionName = getAndOverrideName(PRIO_HIGH_LEVEL_FRAMEWORK);
}
TransactionNameUtils.setNameFromClassAndMethod(className, methodName, transactionName);
```

By using `getAndOverrideName()` with `false` and the same priority used by `ServletTransactionHelper#applyDefaultTransactionName` to set the path-based name, you separate the logic of setting by path or by class/class-name. The latter WILL override the name because it uses `getAndOverrideName()` with `true`.

eyalkoren (commented):
Great refactoring!
I proposed an alternative approach for the actual fix (transaction naming in Spring MVC) that separates the transaction naming logic of the low-level framework from the high level framework.

eyalkoren:
I applied the new APIs to Javalin and Vert.x and added a fallback option to use `use_path_as_transaction_name` for WebFlux transactions.

Ideally, the `TransactionNameUtils` API can be a bit nicer even:
- not requiring the `url_groups` config, but currently we need it this way for the tests
- the logic of setting `unknown route` unless the `use_path_as_transaction_name` config is set is quite common, so we kind or repeat it
- maybe we can also share some of the logic of which priority to use in each case, but not sure

However, I don't know if it will eventually be better or if it worth the effort, only some points for consideration

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/springwebmvc/SpringTransactionNameInstrumentation.java
SylvainJuge: After testing a bit, we must still check `isUsePathAsName()` and always set the high-level framework name here.
- the naming applied with `ServletTransactionHelper#applyDefaultTransactionName` will only set the low-level name to a path if `isUsePathAsName()` is true
- at spring controller level, the class/method can be either a servlet or another controller, thus we have no guarantee that the default naming will have higher priority.

Of course, we can revisit this later if required.

eyalkoren (approved):

PR #1788, opened 2021-04-26
Issue 1699 extend destination field

kananindzya:
## What does this PR do?
closes #1699 

## Checklist

- [x] This is an enhancement of existing features, or a new feature in existing plugins
  - [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [x] I have added tests that prove my fix is effective or that my feature works
  - [x] Added an API method or config option? Document in which version this will be introduced
  - [x] I have made corresponding changes to the documentation

PR #1909, opened 2021-07-10
Added support for setting the framework using the public api (#1908)

### LINKED ISSUE #1908: Support for setting the framework using the public api
I am developing an external plugin for instrumenting remote EJB calls on WildFly and I would like to set the framework name of the created transactions.

There should be a method named `setFrameworkName` in `co.elastic.apm.api.Transaction`

None

## Additional context
#1815

tobiasstadler:
## What does this PR do?
Fixes #1908

## Checklist


eyalkoren:
Thanks for this PR @tobiasstadler !

Some things are missing:
1. document the added API to `docs/public-api.asciidoc`
2. providing higher precedence to user-set values, both is set before AND after the automatic setting
3. the implementation of 2 should include the ability to set `null` or empty string in order to override the automatically-set value
4. testing 2 and 3 - before/after automatic, with `null`/empty and valid value
5. extend the javadoc and asciidoc to specify the above

You can see examples for all of those in #1788

tobiasstadler:
@eyalkoren I did the requested changes

The javadoc/documentation states
```
Override the name of the framework for the current transaction.
For supported frameworks,
the framework name is determined automatically,
and can be overridden using this function.
```
In my opinion this clearly says that a user set value has a higher precedence that an automatic set value (by a plugin). Do you agree?

eyalkoren:
> In my opinion this clearly says that a user set value has a higher precedence that an automatic set value (by a plugin). Do you agree?

Absolutely agree, only a word that setting `null` or empty value will make the agent omit this field, just as a clarification

eyalkoren (approved):

eyalkoren:
run elasticsearch-ci/docs

tobiasstadler:
Thank You!

eyalkoren:
Thank YOU! 😄

PR #1910, opened 2021-07-11
Ecs reformatting more fields

### LINKED ISSUE #1881: expose service_node_name in the logs generated by log_ecs_reformatting

We are trying to replicate a logging setup that we can do with ecs-logging-java with log_ecs_reformatting. We can't expose certain ECS, ecs-logging-java and custom fields trough the APM Java Agent apm-agent-attach-cli.jar

There is an offical field in the APM agent configuration called service_node_name
https://www.elastic.co/guide/en/apm/agent/java/current/config-core.html#config-service-node-name
it is used in APM data, but it would be useful for us if it would also appear in the reformatted JSON logs.

Every JSON log line should contain a "service.node.name": "foo" similar to the already exposed "service.name"

We considered abusing existing fields like "service.name" and concat all information that we need into this field. Then write some ingest pipelines that split up the field and cleans up the mess. This must then be done for Log and APM ingests. This is shady and not obvious.

## Additional context

We searched for an equivalent of ecs-logging-java "additionalField" functionality and ways to pass in other ecs-logging fields like "includeOrigin" and "eventDataset".
https://www.elastic.co/guide/en/ecs-logging/java/current/setup.html#_use_the_ecs_encoderformatterlayout

We will file a separate github issue about additionalField and includeOrigin. We assume that this is potentially bigger whish then this service_node_name feature request.

### LINKED ISSUE #1882: allow to set Encoder/Layout parameters and additionalFields to logs generated by log_ecs_reformatting

We are trying to replicate a logging setup that we can do with ecs-logging-java with log_ecs_reformatting. We can't expose certain ECS, ecs-logging-java and custom fields trough the APM Java Agent apm-agent-attach-cli.jar

include a nominal or generic way to set the ecs-logging-java Encoder/Layout parameter in apm-agent-attach-cli.jar Konfiguration. (We are using --args-provider)

### nominal

maybe with a prefix like `ecs_logging`?

```
ecs_logging_includeOrigin=true
ecs_logging_additionalFields=key1=value1,key2=value2
```

The `ecs_logging_additionalFields` reuses the JUL nomenclature from https://www.elastic.co/guide/en/ecs-logging/java/current/setup.html#_use_the_ecs_encoderformatterlayout

Obivous disadvantage: Once the ecs-logging-java team adds a new field, this needs to be repeated in APM Java Agent

### generic

some kind of generic approach like anything after `ecs_logging_` will be passed to ecs-logging.

```
ecs_logging_includeOrigin=true
ecs_logging_additionalFields=key1=value1,key2=value2
ecs_logging_futureValue=foo
ecs_logging_misspelledValue=bar
```
Obivous disadvantage: `ecs_logging_misspelledValue`. Where would I see the error message that I misspelled it. Ideally in the APM Java Agent  log file. https://www.elastic.co/guide/en/apm/agent/java/current/config-logging.html#config-log-file

For additionalFields we considered abusing existing fields like "service.name" and concat all information that we need into this field. Then write some ingest pipelines that split up the field and cleans up the mess. This must then be done for Log and APM ingests. This is shady and not obvious.

For includeOrigin we failed to find a solution

## Additional context

We hope/assume that the APM-Java-Agent team reuses the source code from ecs-logging-java as a dependency so our feature request is realisitc.

This is related to https://github.com/elastic/apm-agent-java/issues/1881

eyalkoren:
## What does this PR do?

Closes #1881 
Closes #1882 

Includes the following `log_ecs_reformatting` enhancements:
- Automatically sets the `service.node.name` field if set in the agent configuration
- `log_ecs_reformatting_additional_fields` configuration to support the addition of arbitrary fields to all log events
- Automatically serialize markers as tabs where relevant (log4j2 and Logback)

## Checklist


apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/test/java/co/elastic/apm/agent/log4j2/LegacyLog4j2ShadingTest.java
```diff
@@ -53,6 +53,12 @@ public void testSimpleLogShading() throws Exception {
         super.testSimpleLogShading();
     }
 
+    @Test
```
SylvainJuge, line 56: Why do we need to override method here ? if markers are not supported there is already a conditional on `markersSupported()` in `super.testMarkers()` implementation. If there is only a single (or a few) implementations for which the `testMarkers` is not relevant, then only overriding the method without `@Test` annotation should be enough, which have the benefit of not making like the feature is supported but test is always true.

apm-agent-plugins/apm-log-shader-plugin/apm-log4j1-plugin/src/main/java/co/elastic/apm/agent/log4j1/Log4J1EcsReformattingHelper.java
```diff
@@ -50,10 +51,17 @@ protected String getAppenderName(WriterAppender appender) {
...
+                                        @Nullable Map<String, String> additionalFields, Layout originalFormatter) {
         EcsLayout ecsLayout = new EcsLayout();
         ecsLayout.setServiceName(serviceName);
+        ecsLayout.setServiceNodeName(serviceNodeName);
         ecsLayout.setEventDataset(eventDataset);
+        if (additionalFields != null) {
+            for (Map.Entry<String, String> keyValuePair : additionalFields.entrySet()) {
+                ecsLayout.setAdditionalField(keyValuePair.getKey() + "=" + keyValuePair.getValue());
+            }
+        }
```
SylvainJuge, line 64: [minor] there are multiple variants of this, might be worth having a common code to add extra fields in `AbstractEcsReformattingHelper` to avoid duplication.

SylvainJuge (approved):

apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/test/java/co/elastic/apm/agent/log4j2/LegacyLog4j2ShadingTest.java
eyalkoren: See javadoc for this class. Only tests annotated with the JUnit 4 annotation will run and we do want it for log4j tests, old and new

apm-agent-plugins/apm-log-shader-plugin/apm-log4j1-plugin/src/main/java/co/elastic/apm/agent/log4j1/Log4J1EcsReformattingHelper.java
eyalkoren: There are multiple variants because the way to add additional data is library-specific, in Logback and log4j2 it even requires a library-specific type for the added field.

PR #1914, opened 2021-07-12
Use the resource auto-inference to set context.destination.service.resource for JDBC spans (#1913)

### LINKED ISSUE #1913: resource auto-inference is not working for JDBC spans
## Describe the bug
According to https://github.com/elastic/apm/pull/465 `context.destination.service.resource` should be `${subtype}/${instance}` (or `${type}/${instance}` if subtype is null) for a JDBC Span, e.g. for a postgresql db named test: `postgresql/test`. However, `context.destination.service.resource` is only `${subtype}` (or `${type}`), resp. `postgresql ` in the example above.

The algorithm in `co.elastic.apm.agent.impl.transaction.Span#beforeEnd` is not working because:
* JDBC spans are created as normal spans, not exit spans (see `co.elastic.apm.agent.jdbc.helper.JdbcHelper#createJdbcSpan`)
* JDBC spans usually set `context.destination.service.resource` to `ConnectionMetaData.getDbVendor()` (see `co.elastic.apm.agent.jdbc.helper.JdbcHelper#createJdbcSpan`)

## Expected behavior
For a Postgres db named test `context.destination.service.resource` should be `postgresql/test`

tobiasstadler:
## What does this PR do?
Fixes #1913

## Checklist
- [x] This is a bugfix

### DIFF
--- a/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/JdbcHelper.java
+++ b/apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/JdbcHelper.java
@@ -87,7 +87,7 @@ public class JdbcHelper {
             return null;
         }
 
-        Span span = parent.createSpan().activate();
+        Span span = parent.createExitSpan().activate();
         if (sql.isEmpty()) {
             span.withName("empty query");
         } else if (span.isSampled()) {
@@ -121,7 +121,6 @@ public class JdbcHelper {
                 .withPort(connectionMetaData.getPort());
             destination.getService()
                 .withName(vendor)
-                .withResource(vendor)
                 .withType(DB_SPAN_TYPE);
         }
         span.withSubtype(vendor).withAction(DB_SPAN_ACTION);
--- a/apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/AbstractJdbcInstrumentationTest.java
+++ b/apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/AbstractJdbcInstrumentationTest.java
@@ -59,6 +59,7 @@ public abstract class AbstractJdbcInstrumentationTest extends AbstractInstrument
     private static final long PREPARED_STMT_TIMEOUT = 10000;
 
     private final String expectedDbVendor;
+    private final String expectedDbInstance;
     private Connection connection;
     @Nullable
     private PreparedStatement preparedStatement;
@@ -67,9 +68,10 @@ public abstract class AbstractJdbcInstrumentationTest extends AbstractInstrument
     private final Transaction transaction;
     private final SignatureParser signatureParser;
 
-    AbstractJdbcInstrumentationTest(Connection connection, String expectedDbVendor) throws Exception {
+    AbstractJdbcInstrumentationTest(Connection connection, String expectedDbVendor, String expectedDbInstance) throws Exception {
         this.connection = connection;
         this.expectedDbVendor = expectedDbVendor;
+        this.expectedDbInstance= expectedDbInstance;
         connection.createStatement().execute("CREATE TABLE ELASTIC_APM (FOO INT NOT NULL, BAR VARCHAR(255))");
         connection.createStatement().execute("ALTER TABLE ELASTIC_APM ADD PRIMARY KEY (FOO)");
         transaction = startTestRootTransaction("jdbc-test");
@@ -78,6 +80,8 @@ public abstract class AbstractJdbcInstrumentationTest extends AbstractInstrument
 
     @Before
     public void setUp() throws Exception {
+        reporter.disableCheckDestinationAddress();
+
         preparedStatement = EXECUTOR_SERVICE.submit(new Callable<PreparedStatement>() {
             public PreparedStatement call() throws Exception {
                 return connection.prepareStatement(PREPARED_STATEMENT_SQL);
@@ -394,6 +398,7 @@ public abstract class AbstractJdbcInstrumentationTest extends AbstractInstrument
         Span span = reporter.getFirstSpan();
         StringBuilder processedSql = new StringBuilder();
         signatureParser.querySignature(rawSql, processedSql, preparedStatement);
+        assertThat(span.isExit()).isTrue();
         assertThat(span.getNameAsString()).isEqualTo(processedSql.toString());
         assertThat(span.getType()).isEqualTo(DB_SPAN_TYPE);
         assertThat(span.getSubtype()).isEqualTo(expectedDbVendor);
@@ -419,7 +424,11 @@ public abstract class AbstractJdbcInstrumentationTest extends AbstractInstrument
         }
 
         Destination.Service service = destination.getService();
-        assertThat(service.getResource().toString()).isEqualTo(expectedDbVendor);
+        if (expectedDbInstance != null) {
+            assertThat(service.getResource().toString()).isEqualTo(expectedDbVendor + "/" + expectedDbInstance);
+        } else {
+            assertThat(service.getResource().toString()).isEqualTo(expectedDbVendor);
+        }
 
         assertThat(span.getOutcome())
             .describedAs("span outcome should be explicitly set to either failure or success")
@@ -435,6 +444,7 @@ public abstract class AbstractJdbcInstrumentationTest extends AbstractInstrument
         Span jdbcSpan = reporter.getFirstSpan();
         StringBuilder processedSql = new StringBuilder();
         signatureParser.querySignature(rawSql, processedSql, preparedStatement);
+        assertThat(jdbcSpan.isExit()).isTrue();
         assertThat(jdbcSpan.getNameAsString()).isEqualTo(processedSql.toString());
         assertThat(jdbcSpan.getType()).isEqualTo(DB_SPAN_TYPE);
         assertThat(jdbcSpan.getSubtype()).isEqualTo("unknown");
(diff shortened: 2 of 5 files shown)


tobiasstadler:
@eyalkoren Would you be ok with putting the change behind a configuration option? Of course the current behavior will be the default.

eyalkoren:
What would be the use case of disabling it? Do you have a specific example of something that will change/break by applying it?

The purpose of auto-inference is that all exit spans fill this field.

tobiasstadler:
I do not want to disable auto-inference. I was talking about the value of `context.destination.service.resource` for jdbc spans. Currently they are populated with the db vendor. I want(ed) to change that to vendor + instance (as it is done by the auto-inference). You rejected my change (which I totally understand) because it may break existing setups. So I propose to to add a configuration option which changes the value of `context.destination.service.resource` for jdbc spans to vendor + instance.

eyalkoren:
Ahh, gotcha, sorry for the misunderstanding.

Do your JDBC spans contain the correct value in the `instance` field? I think we first need to improve our `instance` discovery.

This generally sounds like a reasonable request, but, as I said - we are in the process of reevaluating whether to change that by default, so I rather we wait a bit with that.

tobiasstadler:
Yes, in my case (PostgreSQL) `span.db.instance` contains the value I want.

I tried using an ingest pipeline and it works fine for spans, but it doesn't work for the service_destination metricset, because there is no `span.db.instance`.

eyalkoren:
Sorry, I don't follow - the agent sets the `span.context.db.instance` field, but it is not stored as `span.db.instance` in Elasticsearch?

tobiasstadler:
No, it is stored for spans and the ingest pipeline works fine for spans. But the metrics, calculated for the transactions/spans by the apm-server (or kibana?), do not have a `span.db.instance` field, so I cannot change the resource name for them. So after the ingest pipeline the resource name for the span is `postgreSQL/mydb`, but the resource name of the metric is still `postgres`. As these metrics are used to e.g. display the latency, throughput and error rate in the dependencies table, the data is missing their.

Here is an example before the ingest pipeline:
```
{
    "_index": "apm-7.14.0-span-000001",
    "_type": "_doc",
    "_id": "eLvBzHoBOG9QGDzr1Brm",
    "_score": 1,
    "_source": {
      "parent": {
        "id": "a808c835bd49d7fc"
      },
      "agent": {
        "name": "java",
        "ephemeral_id": "84d5f5fd-f453-4298-b48b-640edf56233d",
        "version": "1.24.1-SNAPSHOT.UNKNOWN"
      },
      "destination": {
        "address": "...",
        "port": ...,
        "ip": "..."
      },
      "processor": {
        "name": "transaction",
        "event": "span"
      },
      "observer": {
        "hostname": "0342e6a2b6bb",
        "id": "80ec3261-4080-4fa8-92e7-f68dbd7d807c",
        "ephemeral_id": "34d2e106-e279-4e18-8195-0d3a4f28032d",
        "type": "apm-server",
        "version": "7.14.0",
        "version_major": 7
      },
      "trace": {
        "id": "34d33e78799821071cf2afec9583955b"
      },
      "@timestamp": "2021-07-22T05:48:07.799Z",
      "ecs": {
        "version": "1.10.0"
      },
      "service": {
        "name": "..."
      },
      "event": {
        "outcome": "success"
      },
      "transaction": {
        "id": "a808c835bd49d7fc"
      },
      "span": {
        "duration": {
          "us": 61833
        },
        "stacktrace": [],
        "subtype": "postgresql",
        "destination": {
          "service": {
            "resource": "postgresql"
          }
        },
        "name": "...",
        "action": "query",
        "id": "43a7cf0244756f9f",
        "type": "db",
        "db": {
          "instance": "mydb",
          "statement": "...",
          "type": "sql",
          "user": {
            "name": "..."
          }
        }
      },
      "timestamp": {
        "us": 1626932887799620
      }
    },
    "fields": {
      "span.stacktrace.line.number": [],
      "span.name": [
        "..."
      ],
      "span.action": [
        "query"
      ],
      "destination.port": [
        ...
      ],
      "span.db.user.name": [
        "..."
      ],
      "transaction.id": [
        "a808c835bd49d7fc"
      ],
      "trace.id": [
        "34d33e78799821071cf2afec9583955b"
      ],
      "span.duration.us": [
        61833
      ],
      "span.stacktrace.classname": [],
      "span.db.statement": [
        "..."
      ],
      "processor.event": [
        "span"
      ],
      "agent.name": [
        "java"
      ],
      "destination.address": [
        "..."
      ],
      "event.outcome": [
        "success"
      ],
      "span.stacktrace.exclude_from_grouping": [],
      "service.name": [
        "..."
      ],
      "span.stacktrace.filename": [],
      "processor.name": [
        "transaction"
      ],
      "span.stacktrace.function": [],
      "span.subtype": [
        "postgresql"
      ],
      "span.id": [
        "43a7cf0244756f9f"
      ],
      "observer.version_major": [
        7
      ],
      "span.stacktrace.library_frame": [],
      "span.stacktrace.module": [],
      "destination.ip": [
        "..."
      ],
      "observer.hostname": [
        "0342e6a2b6bb"
      ],
      "span.type": [
        "db"
      ],
      "timestamp.us": [
        1626932887799620
      ],
      "observer.id": [
        "80ec3261-4080-4fa8-92e7-f68dbd7d807c"
      ],
      "@timestamp": [
        "2021-07-22T05:48:07.799Z"
      ],
      "span.db.type": [
        "sql"
      ],
      "observer.ephemeral_id": [
        "34d2e106-e279-4e18-8195-0d3a4f28032d"
      ],
      "observer.version": [
        "7.14.0"
      ],
      "ecs.version": [
        "1.10.0"
      ],
      "observer.type": [
        "apm-server"
      ],
      "agent.ephemeral_id": [
        "84d5f5fd-f453-4298-b48b-640edf56233d"
      ],
      "parent.id": [
        "a808c835bd49d7fc"
      ],
      "agent.version": [
        "1.24.1-SNAPSHOT.UNKNOWN"
      ],
      "span.destination.service.resource": [
        "postgresql"
      ],
      "span.db.instance": [
        "mydb"
      ]
    }
  }
```
```
{
    "_index": "apm-7.14.0-metric-000001",
    "_type": "_doc",
    "_id": "jrvCzHoBOG9QGDzrhBq9",
    "_score": 1,
    "_source": {
      "agent": {
        "name": "java"
      },
      "processor": {
        "name": "metric",
        "event": "metric"
      },
      "metricset.name": "service_destination",
      "observer": {
        "hostname": "0342e6a2b6bb",
        "id": "80ec3261-4080-4fa8-92e7-f68dbd7d807c",
        "ephemeral_id": "34d2e106-e279-4e18-8195-0d3a4f28032d",
        "type": "apm-server",
        "version": "7.14.0",
        "version_major": 7
      },
      "@timestamp": "2021-07-22T05:48:56.913Z",
      "ecs": {
        "version": "1.10.0"
      },
      "service": {
        "name": "..."
      },
      "metricset": {
        "period": 60000
      },
      "event": {
        "ingested": "2021-07-22T05:48:57.916489641Z",
        "outcome": "success"
      },
      "span": {
        "destination": {
          "service": {
            "resource": "postgresql",
            "response_time": {
              "count": 27,
              "sum": {
                "us": 1564268
              }
            }
          }
        }
      }
    },
    "fields": {
      "service.name": [
        "..."
      ],
      "metricset.period": [
        60000
      ],
      "span.destination.service.response_time.count": [
        27
      ],
      "processor.name": [
        "metric"
      ],
      "observer.version_major": [
        7
      ],
      "observer.hostname": [
        "0342e6a2b6bb"
      ],
      "metricset.name": [
        "service_destination"
      ],
      "event.ingested": [
        "2021-07-22T05:48:57.916Z"
      ],
      "observer.id": [
        "80ec3261-4080-4fa8-92e7-f68dbd7d807c"
      ],
      "span.destination.service.response_time.sum.us": [
        1564268
      ],
      "@timestamp": [
        "2021-07-22T05:48:56.913Z"
      ],
      "observer.ephemeral_id": [
        "34d2e106-e279-4e18-8195-0d3a4f28032d"
      ],
      "observer.version": [
        "7.14.0"
      ],
      "ecs.version": [
        "1.10.0"
      ],
      "observer.type": [
        "apm-server"
      ],
      "processor.event": [
        "metric"
      ],
      "agent.name": [
        "java"
      ],
      "span.destination.service.resource": [
        "postgresql"
      ],
      "event.outcome": [
        "success"
      ]
    }
  }
```

eyalkoren:
Yeah, that's right, I was only thinking about getting the nodes in the map, but the DB service metrics won't work.

Since this is not specific to the Java agent, would you mind opening an issue in the https://github.com/elastic/apm repo, describing what you need? I can't find an existing one for this. It would be great to get your input on why you think the instance should be used for DB service node distinction. 
Things to keep in mind for making the discussion efficient:
-  How likely is it that by using the `instance`, we will provide a poor experience OOTB for users that may end up having a very big number of DB nodes? 
   - How many instances do you have in your setup? 
   - What bounds the number of instances? 
   - Are there SQL DBs where the number of instances can typically be very big?
- What is the best way to get the instance for each DB? In the Java agent, we currently use `java.sql.Connection#getCatalog` - are there DB vendors for which this will produce very high cardinality?
- Is the `instance` notion common to all or most DB vendors?
- Are there better options than `instance`, considering all of the above? Could DB name provide better granularity?
- Whether and how should this be configurable? 

I think this is a very important issue and I rather we tackle it properly, not as a patch.
Let me know if I can assist with that.

Thanks for the useful input!

tobiasstadler:
FYI, I created #1928 to enable resource auto-inference via a config option. I think this is a step in the right direction. It enables me (and possibly others) to include `db.instance` in `destination.service.resource` without breaking any existing setup.

In my opinion populating `db.instance` with the "right" value is a separate problem and I will create an issue for that. Personally I am fine with the current behavior (actually I created the initial PR to use `java.sql.Connection#getCatalog`), but it probably won't fit every setup (e.g. for Oracle and DB2 `getCatalog` will return null).

tobiasstadler:
FYI, I created elastic/apm#473

PR #1918, opened 2021-07-13
Add scala Future test case using multiple threads

milanvdm:
Related to https://github.com/elastic/apm-agent-java/pull/1274

The main issue was related to https://github.com/scala/bug/issues/12089 which got introduced in Scala 2.13.0.

From what I can tell, https://github.com/elastic/apm-agent-java/pull/1274 added a new test case that actually deadlocks on 2.13.0-2.13.3. This causes some weird behavior in the new test case.

After upgrading to 2.13.6 and using the ExecutionContext.global instead of a custom ForkJoinPool, the test case passes.

I also added extra test cases with different ExecutionContexts and different approaches on ending transactions.

@henrikno


milanvdm:
Could someone have another look at this or will this wait until there is a decision on how to support community plugins?

milanvdm:
@felixbarny Any progress on this? :)

PR #1920, opened 2021-07-14
Added support for excluding classes from classes_excluded_from_instrumentation_default

### LINKED ISSUE #1919: Support for excluding classes from classes_excluded_from_instrumentation_default
I am developing an external plugin that needs to instrument class which are matched by `classes_excluded_from_instrumentation_default`.

It should be possible for instrumentation's to exclude classes from `classes_excluded_from_instrumentation_default`

manually setting and maintaining `classes_excluded_from_instrumentation_default`

## Additional context
#1815

tobiasstadler:
## What does this PR do?
Fixes #1919

## Checklist

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
@@ -92,6 +92,7 @@ import java.util.concurrent.TimeUnit;
 import static co.elastic.apm.agent.bci.bytebuddy.ClassLoaderNameMatcher.classLoaderWithName;
 import static co.elastic.apm.agent.bci.bytebuddy.ClassLoaderNameMatcher.isReflectionClassLoader;
 import static co.elastic.apm.agent.bci.bytebuddy.CustomElementMatchers.anyMatch;
+import static co.elastic.apm.agent.bci.bytebuddy.CustomElementMatchers.anyMatchExcept;
 import static net.bytebuddy.asm.Advice.ExceptionHandler.Default.PRINTING;
 import static net.bytebuddy.matcher.ElementMatchers.any;
 import static net.bytebuddy.matcher.ElementMatchers.is;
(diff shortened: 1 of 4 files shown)


tobiasstadler:
Could someone please have a look?

tobiasstadler:
@felixbarny Maybe you can spare some time please.

felixbarny:
I had a quick look. At it's current state, it seems a bit hackish tbh. But I also don't have a great alternative. Some questions and thoughts.

- What's the startup time difference when removing `org.jboss.as.*` from `classes_excluded_from_instrumentation_default`?
  - If there's a big difference, where is it coming from? Shouldn't all instrumentations pre-filter on substrings of the class name or check if the class to instrument is even loadable in the current class loader (`CustomElementMatchers::classLoaderCanLoadClass`).
  - If there's no big difference, could we just remove the setting altogether? It can definitely catch you by surprise and causes long and frustrating debugging sessions for any future contributor who wants to instrument something that's excluded. Ideally, I think we should try to get rid of `classes_excluded_from_instrumentation_default`.
- If the above is not an option, instead of adding `getTypeMatcherIncludeDefaultExcludedClasses`, could we add a `getTypeMatcherDefaultExcludedClasses` method that returns the values from `classes_excluded_from_instrumentation_default` by default? In your instrumentation, you could override the method and return `none()` or specifically remove the `org.jboss.as.*` filter.

eyalkoren:
We discussed this offline and decided the best approach would be to introduce a new service that can be used for customization of instrumentation plugins.
It can be a service dedicated only for this purpose, similarly to the new [`PluginClassLoaderRootPackageCustomizer`](https://github.com/elastic/apm-agent-java/blob/f615e69241c9106b5a93b83d65974d8a7e5f0d85/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/PluginClassLoaderRootPackageCustomizer.java) service (only available at the plugin SDK), or a more general interface, similar in concept to OTel's [`IgnoredTypesBuilder`](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/ignore/IgnoredTypesBuilder.java)

PR #1921, opened 2021-07-16
Use Implementation-Title as default service name if it is available

tobiasstadler:
## What does this PR do?
When the MANIFEST.MF of main jar contains the Implementation-Title attribute, it is used as the default service name.

## Checklist
- [x] This is an enhancement of existing features, or a new feature in existing plugins
  - [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [x] I have made corresponding changes to the documentation

PR #1726, opened 2021-03-30
Added support for overwriting the service version per classloader

tobiasstadler:
## What does this PR do?
Fixes #1725

## Checklist
- [x] This is an enhancement of existing features, or a new feature in existing plugins
  - [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [x] I have added tests that prove my fix is effective or that my feature works
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] I have made corresponding changes to the documentation
- [ ] This is a bugfix
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [ ] I have added tests that would fail without this fix
- [ ] This is a new plugin
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [ ] My code follows the [style guidelines of this project](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#java-language-formatting-guidelines)
  - [ ] I have made corresponding changes to the documentation
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/master/CONTRIBUTING.md#testing) pass locally with my changes
  - [ ] I have updated [supported-technologies.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/docs/supported-technologies.asciidoc)
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] Added an instrumentation plugin? Describe how you made sure that old, non-supported versions are not instrumented by accident.
- [ ] This is something else
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)

PR #1922, opened 2021-07-16
Use Implementation-Version as default service version if it is available

tobiasstadler:
## What does this PR do?
When the MANIFEST.MF of main jar contains the Implementation-Version attribute, it is used as the default service version (except for application servers).

## Checklist


tobiasstadler:
Could someone please have a look?

tobiasstadler:
@SylvainJuge Maybe you can spare some time please

tobiasstadler:
@felixbarny Maybe you can spare some time please

tobiasstadler:
Thank You! I am looking forward to it.

tobiasstadler:
@SylvainJuge I guess this won't make it into 1.27.0?

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ServiceVersionUtil.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.jar.Attributes;
+import java.util.jar.JarFile;
+
+public class ServiceVersionUtil {
+
+    public static String getDefaultServiceVersion() {
+        return getDefaultServiceVersion(System.getProperty("sun.java.command"));
+    }
+
+    private static String getDefaultServiceVersion(@Nullable String command) {
```
felixbarny, line 31: There's some duplication from ServiceNameUtil. What about merging the two and returning a single `ServiceInfo` Object:

```java
class ServiceInfo {
    private final String serviceName;
    private final String serviceVersion;
    private final boolean isContainer; // or Boolean with null = unknown?

    public static ServiceInfo create() {
        return create(System.getProperties());
    }
    public static ServiceInfo create(Properties properties) {
        // parsing sun.java.command
        // delegate to create(Manifest manifest) if command is a jar file
        // merge ServiceInfo if required
    }

    public static ServiceInfo create(Manifest manifest) {
        ...
    }

    ServiceInfo merge(ServiceInfo other) {
        // create new immutable copy and merge the two with one having precedence
    }
   ...
}
```

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ServiceVersionUtil.java
```diff
@@ -0,0 +1,61 @@
...
+                }
+
+                break;
+            }
+        }
+
+        return null;
+    }
+
+    private static boolean isContainerCommand(String command) {
```
felixbarny, line 53: Just thinking out loud. Do we even need to exclude the version from containers? When the service name is `tomcat-application`, it might not be a big deal if the service.version shows the tomcat version.
On the other hand, it's also not particularly useful so probably no need to go to lengths to try to get the container version.

I guess what you did is the best compromise: try to not use the service.version for known containers. But if a version slips through, it's not a big deal.

felixbarny (approved):
Good improvement and fits will with the just merged implementation-title detection. Let's refactor the code a bit and I think we're good to go.

felixbarny (changes requested):
Sorry, I mean to request changes instead of approve.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ServiceVersionUtil.java
felixbarny: This overlaps (in a nice way, actually) with the changes in https://github.com/elastic/apm-agent-java/pull/1726/files.

Maybe we don't need the isContainer flag and just not populate the `serviceVersion` instead.

tobiasstadler: @felixbarny Would you please merge #1921? It makes creating `ServiceInfo` easier.

felixbarny: Merged

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ServiceVersionUtil.java
tobiasstadler: I left it as is

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ServiceVersionUtil.java
tobiasstadler: I merged ServiceNameUtil and ServiceVersionUtil. Can you please have a l look at my changes?

tobiasstadler:
/elasticsearch-ci/docs

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ServiceVersionUtil.java
felixbarny: Looks great, much cleaner!

felixbarny (approved):
I took the liberty to push some polishing changes

felixbarny:
run elasticsearch-ci/docs

felixbarny (approved):

tobiasstadler:
@felixbarny Would you please run elasticsearch-ci/docs again?

felixbarny:
run elasticsearch-ci/docs

tobiasstadler:
Thank You!

PR #1925, opened 2021-07-21
Add description about memory pool metrics to docs

eyalkoren:


jackshirazi (commented):
should the `name` comment be the same as the others in the file?
name: The name representing this memory manager (for example G1 Young Generation, G1 Old Generation)

eyalkoren:
> should the name comment be the same as the others in the file?
name: The name representing this memory manager (for example G1 Young Generation, G1 Old Generation)

The other ones are for garbage collectors, not for memory pools.

jackshirazi:
oops, silly me. Do we want to have a similar example? like 
name: The name representing this memory pool (for example Metaspace, Compressed Class Space, G1 Eden Space)

eyalkoren:
> Do we want to have a similar example? like
name: The name representing this memory pool (for example Metaspace, Compressed Class Space, G1 Eden Space)

I don't think this is required - we send those by default, so users can look them up and see what values they get. It may do the opposite and confuse those that use a JVM that uses different names.

jackshirazi (approved):

PR #1928, opened 2021-07-22
Use the resource auto-inference to set context.destination.service.resource for JDBC spans (#1913)

### LINKED ISSUE #1913: resource auto-inference is not working for JDBC spans

tobiasstadler:
## What does this PR do?
Fixes #1913

## Checklist


eyalkoren:
Yes, I understood what you requested, but this is what I referred to as a patch.
A real config will probably not be such, but more something like: `use_instance_for_db_resource`, and it is relevant to all agents, not only Java. We wouldn't want to let make our users think about auto-inference and how it works, but on what this config would change.

So, let us consider that. Maybe we decide to add this as an `internal` config.

tobiasstadler:
Ok, so should I change my implementation and just add the instance to `context.destination.service.resource` in `JdbcHelper` if the internal config option `use_instance_for_db_resource` is true?

tobiasstadler:
@eyalkoren I implemented your suggestion

eyalkoren:
@tobiasstadler sorry for the long pauses between responses, we are extremely busy these days and I was OOO lately. Let me try pushing https://github.com/elastic/apm/issues/473 forward first, and if we can't get that, we will consider a Java-specific configuration. Normally we try to avoid that, as it make future decisions more complicated.

tobiasstadler:
Ok, thanks you @eyalkoren

felixbarny:
@SylvainJuge is working on a concept that plans to use `service.target.name` and `service.target.type` from [ECS](https://www.elastic.co/guide/en/ecs/master/ecs-service.html#_field_reuse_22) instead.

tobiasstadler:
@SylvainJuge / @felixbarny Is there any (rough) timeline (you can share) when your concept will be available?

SylvainJuge:
Hi @tobiasstadler, we aim to start working on this in the next couple of months. As usual any estimate should be taken with a ~~pinch~~ shovel of salt.

Given it involves changing things in multiple projects: all agents, apm-server and apm-ui it's really hard to give you any estimate when it will be available in practice. However, it's something that is an important on a product level, so it is really something that is expected to happen in the short/mid-term time-frame.

tobiasstadler:
Ok, thank You for the information. I guess I have to wait then.

PR #1931, opened 2021-07-22
Semver parsing enhancement

### LINKED ISSUE #1930: Version fails to parse some RedHat built artifacts
## Describe the bug
I had a look in our logs and found the message `Cannot read implementation version based on ProtectionDomain. This should not affect your agent's functionality. Failed with message: For input string: "redhat"`. The message appears because we use some artifacts built by RedHat which often have a ".redhat" after the version, e.g. httpclient-4.5.13.redhat-00001.jar (see https://maven.repository.redhat.com/ga/org/apache/httpcomponents/httpclient/4.5.13.redhat-00001/).

## Expected behavior
RedHat built artifacts should be parsed without errors. I propose the following path:
```
diff --git a/apm-agent-core/src/main/java/co/elastic/apm/agent/util/Version.java b/apm-agent-core/src/main/java/co/elastic/apm/agent/util/Version.java
index 076b8a42..a9761a87 100644
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/util/Version.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/util/Version.java
@@ -36,7 +36,13 @@ public class Version implements Comparable<Version> {
         final String[] parts = version.split("\\-")[0].split("\\.");
         numbers = new int[parts.length];
         for (int i = 0; i < parts.length; i++) {
-            numbers[i] = Integer.valueOf(parts[i]);
+            try {
+                numbers[i] = Integer.valueOf(parts[i]);
+            } catch (NumberFormatException e) {
+                if (!"redhat".equals(parts[i])) {
+                    throw e;
+                }
+            }
         }
     }
 

```

eyalkoren:
## What does this PR do?

Closes #1930 

A quick and dirty approach, that we can improve later if required.
It looks like using a regex may get quite complicated for this.
If I have more time, I would have made it shorter ( _Blaise Pascal_)
## Checklist

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,7 @@ endif::[]
 
 [float]
 ===== Bug fixes
+* Fix failure to parse some forms of the `Implementation-Version` property from jar manifest files - {pull}1931[#1931]
 
 [[release-notes-1.x]]
 === Java Agent version 1.x
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/util/Version.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/util/Version.java
@@ -33,10 +33,30 @@ public class Version implements Comparable<Version> {
     }
 
     private Version(String version) {
-        final String[] parts = version.split("\\-")[0].split("\\.");
-        numbers = new int[parts.length];
-        for (int i = 0; i < parts.length; i++) {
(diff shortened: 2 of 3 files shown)


tobiasstadler:
Thank You!

SylvainJuge (approved):
LGTM for a quick work-around, we can definitely improve that later if needed.

PR #1937, opened 2021-07-28
Make sure trace context headers are added only once

eyalkoren:
## What does this PR do?

Make sure trace context headers are added only once to outgoing calls, even if the protocol supports multiple values per header (like in HTTP).
## Checklist

- [x] This is a bugfix


SylvainJuge (approved):
[minor] For HTTP headers, most (if not all) header accessors allow for ready & write operations, thus maybe we could have a helper method or a variant of `propagateTraceContext` that ensure that header has only a single value.

eyalkoren:
> For HTTP headers, most (if not all) header accessors allow for ready & write operations, thus maybe we could have a helper method or a variant of propagateTraceContext that ensure that header has only a single value.

Yes, that's what I wanted to do at first, but then realized that in some cases it's not efficient to read (allocates iterator for example), so using `set` instead of `add` (i.e. override) is cheaper (even though it's not equivalent).
In Spring `RestTemplate` I used a `containsKey` API on a per-header basis. 
In OkHttp you even need to use a different type for reading and writing.
So I ended up doing it case by case, at least for now.

PR #1940, opened 2021-07-29
Work-around for Jackson & instrumentation issues on Java15+

SylvainJuge:
## What does this PR do?

Work-around for compatibility issue with Jackson JSON serialization on Java 15 and 16.
~~Also, rollback jackson version to 2.12.x branch to ensure Java7 compatibility (2.13.x is built with Java 8) ([see doc](https://github.com/FasterXML/jackson#active-developed-jackson-2x-branches))~~

### DIFF
--- a/apm-agent-core/src/test/java/co/elastic/apm/agent/JsonUtils.java
+++ b/apm-agent-core/src/test/java/co/elastic/apm/agent/JsonUtils.java
@@ -18,12 +18,35 @@
  */
 package co.elastic.apm.agent;
 
+import com.fasterxml.jackson.core.JsonGenerator;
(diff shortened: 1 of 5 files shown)


SylvainJuge:
FYI I realized that changing jackson version is not as trivial as it sounds as we have it in many levels of maven dependencies.
For example, mixing spring-boot versions with the one we already have through core module tends to conflict and create weird classloading issues because jackson dependencies structure changed (thus inclusions/exclusions do not work as expected).

Thus we can consider this still WIP until proper review is required.

SylvainJuge:
Update: should be ready to review/merge, even if the build still does not run without error on Jdk 14-16.

felixbarny (approved):

PR #1948, opened 2021-07-30
Fix Webflux server response status capture

### LINKED ISSUE #1938: WebFlux - transaction has status code 200 where it should be 500
1. spring version 2.2.6 webflux, netty
```
    @Operation(
        summary = "상품 이미지 검색", description = "상품 이미지 검색 api", responses = {
            @ApiResponse(responseCode = "200", description = "상품 이미지 검색 성공", content = @Content(schema = @Schema(implementation = GoodsDTOV1.ImageResponse.class))),
            @ApiResponse(responseCode = "404", description = "상품 이미지 검색 실패", content = @Content(schema = @Schema(implementation = GoodsDTOV1.ImageResponse.class)))
        }
    )
    @GetMapping("/goods")
    @ReactorCacheables(cacheNames = CacheNameConstant.GOODS_IMAGE_LIST, key = "getCacheKey(#method)")
    public Mono<ResponseDTOV2> getGoodsList(@ParameterObject @ModelAttribute @Valid GoodsDTOV1.Search goods) {
        return goodsService.getGoodsList(goods.getUrl(), ImageSearchTypeEnum.getValue(goods.getImageSearchType()))
            .flatMap(goodsImageVO -> {

                final GoodsDTOV1.ImageResponse imageResponse = goodsMapStruct.toDTO(goodsImageVO);

                if (1 == 1) {
                    throw new RuntimeException();
                }

                return Mono.just(
                    ResponseDTOV2.builder().data(imageResponse).build()
                );
            });
    }


Obviously a status 500 internal server error occurred.
In elastic-apm it comes out as 200

```
image

image

```
    @Operation(
        summary = "상품 이미지 검색", description = "상품 이미지 검색 api", responses = {
            @ApiResponse(responseCode = "200", description = "상품 이미지 검색 성공", content = @Content(schema = @Schema(implementation = GoodsDTOV1.ImageResponse.class))),
            @ApiResponse(responseCode = "404", description = "상품 이미지 검색 실패", content = @Content(schema = @Schema(implementation = GoodsDTOV1.ImageResponse.class)))
        }
    )
    @GetMapping("/goods")
    @ReactorCacheables(cacheNames = CacheNameConstant.GOODS_IMAGE_LIST, key = "getCacheKey(#method)")
    public Mono<ResponseDTOV2> getGoodsList(@ParameterObject @ModelAttribute @Valid GoodsDTOV1.Search goods) {

         if (1 == 1) {
                throw new RuntimeException();
         }

        return goodsService.getGoodsList(goods.getUrl(), ImageSearchTypeEnum.getValue(goods.getImageSearchType()))
            .flatMap(goodsImageVO -> {

                final GoodsDTOV1.ImageResponse imageResponse = goodsMapStruct.toDTO(goodsImageVO);



                return Mono.just(
                    ResponseDTOV2.builder().data(imageResponse).build()
                );
            });
    }
```
image

If the exception is outside of mono, it is exposed normally, but if an error occurs inside mono, it does not work properly.

404 is normal
image

SylvainJuge:
## What does this PR do?

Fixes #1938 

We used to instrument two parts of Webflux request processing:
- request dispatch, which wraps request handling
- request handling

This was mostly fine for most cases, but did not worked in the following case:
- using annotated controller that throws an exception during request processing (within mono/flux operator body)
- using an annotated exception handler method

In this case, the transaction was terminated early during the "handling" phase, without having the proper response status code set by the exception handler, which is executed as part of the "dispatch" phase. In practice, it seems that only wrapping the "dispatch" is enough to properly capture the whole HTTP request.

## Checklist

- [x] This is a bugfix


SylvainJuge:
It seems that this PR makes transaction termination less reliable than it previously was, I'm currently investigating what could be the cause here.

SylvainJuge:
It still fails randomly, thus making it back to `draft` status.

SylvainJuge:
Quick status:
- discovered that some random failures were due to cancellation of request processing when client is faster than server and stops listening before the server has actually executed the `onComplete` method that we override, thus some transactions were not properly terminated as neither the `onError` or `onComplete` were being called in this case.
- the fix for that is a bit "ugly" as it simply adds a listener when the subscription is cancelled to properly terminate transaction, and also needs to decrement the reference count when this happens.

I've let the whole test suite running locally for ~45min (about 1800 tests), and did not got a single failure (versus less than 10 executions before getting a failure previously), thus at least we can issue the "works on my machine" certificate.
However, we will likely not include this in the next release (but likely the following) just to let it dry a bit and likely conduct a bit more testing.

eyalkoren:
Why is adding the `CancelTask` "ugly"? Isn't this a proper use of the framework? Asking because it looks good to me, so I wonder if there is a concern I am missing.

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-plugin/src/main/java/co/elastic/apm/agent/springwebflux/TransactionAwareSubscriber.java
```diff
@@ -218,7 +181,26 @@ private void doExit(boolean discard, String method, @Nullable Transaction transa
...
+        debugTrace(true, "cancelTransaction", transaction);
+        try {
+            if (transaction == null) {
+                return;
+            }
+
+            WebfluxHelper.endTransaction(null, transaction, exchange);
+
+            // not 100% sure to understand why it's required, but without it recycling reference count
+            // is always off by one when transaction get cancelled
```
eyalkoren, line 197: This is something we must figure out. Could it be that the transaction was activated but not deactivated in this scenario?

Regardless to this specific (I think), this class stores transaction references in the `transactionMap`, but we do not reflect that in the reference counting. The ref count needs to be incremented when the transaction is added to the map and decremented when removed.

SylvainJuge: After a bit much more investigation following our discussion today, this extra decrement is required because the transaction is not being properly de-activated (which is expected).

- We can't simply limit activation to the wrapped method bodies (`onSubscribe`, `onNext`, `onComplete`, `onError`) like with the reactor plugin because otherwise we have failing tests, for example trying to set a custom transaction name with annotated controllers.
- as it seems to already work better like this, I would suggest to keep it as-is and revisit this later if that becomes an issue.
- there might be other ways to cover the "set custom transaction name" test, for example by wrapping/instrumenting, but that would require some further investigation and investment.
- I think it could have been a bit easier if the "current transaction" was being accessible through Webflux context storage instead of relying on current thread state and activation/deactivation events (but that would be a major change).

As `transactionMap` is an instance of `SpanConcurrentHashMap`, the reference count is already managed by the map implementation.

felixbarny: > there might be other ways to cover the "set custom transaction name" test, for example by wrapping/instrumenting, but that would require some further investigation and investment.

If that allows us to get rid of the unsafe pattern of activating a span in one place and deactivating it somewhere else in favor of activating/deactivating in something resembling a try/finally block, I think that'd be well worth it.

> I think it could have been a bit easier if the "current transaction" was being accessible through Webflux context storage instead of relying on current thread state and activation/deactivation events

We have to explicitly activate so that automatically created spans for HTTP clients connect to the WebFlux transaction, for example. But maybe we can internally rely on the context: For example, before the controllers are called, we can instrument that part of the dispatching, get the transaction from the WebFlux context and activate it.

SylvainJuge:
Update: found a way to propagate "active context" by using reactor context storage, which allows to keep a reference on the active transaction without relying on having it active on current thread.
As a result, we don't have to keep any transaction active between calls on methods `onSubscribe` and `onComplete` or `onError`, which then required an extra reference count decrement due to the lack of proper deactivation of the transaction.

Next (and hopefully final) step: see if we could handle cancellation differently by wrapping the `Subscription` instance, as it would allow to handle cancellation cleanup without having to register a callback.

apm-agent-core/src/test/java/co/elastic/apm/agent/AbstractInstrumentationTest.java
```diff
@@ -150,17 +150,21 @@ protected static void flushGcExpiry(WeakConcurrentMap<?, ?> map, int count) {
...
             long duration = System.currentTimeMillis() - start;
-            System.out.printf("flushGcExpiry - after gc execution %d ms%n", duration);
-            try {
-                Thread.sleep(1);
-            } catch (InterruptedException e) {
-                // silently ignored
+            System.out.printf("flushGcExpiry - after gc execution %d ms (%d/%d) %n", duration, index, count);
+            if (count > 1 && index < count) {
+                try {
+                    Thread.sleep(5);
```
SylvainJuge, line 161: [for reviewer] I'm not sure if the delay increase makes any difference here

apm-agent-plugins/apm-reactor-plugin/src/main/java/co/elastic/apm/agent/reactor/TracedSubscriber.java
```diff
@@ -52,10 +53,20 @@
...
 
+    private final Context context;
+
     TracedSubscriber(CoreSubscriber<? super T> subscriber, Tracer tracer, AbstractSpan<?> context) {
         this.subscriber = subscriber;
         this.tracer = tracer;
         contextMap.put(this, context);
+
+        // store our span/transaction into reactor context for later lookup without relying on active tracer state
+        this.context = subscriber.currentContext().put(AbstractSpan.class, context);
```
SylvainJuge, line 64: [for reviewer] Reactor context is an immutable structure, thus we have to keep a reference. Given the context instance is tied to this class instance, there is no need to increment/decrement reference count.

apm-agent-plugins/apm-reactor-plugin/src/main/java/co/elastic/apm/agent/reactor/TracedSubscriber.java
```diff
@@ -236,8 +247,14 @@ static boolean isHookRegistered() {
...
                 }
 
+                // use active span/transaction if directly active
                 AbstractSpan<?> active = tracer.getActive();
 
+                // fallback to using context-stored span/transaction if not already active
+                if (active == null) {
+                    active = subscriber.currentContext().getOrDefault(AbstractSpan.class, null);
+                }
+
```
SylvainJuge, line 257: [for reviewer] this is what allows Reactor context propagation to retrieve the current context when it's not directly active, but has been stored, which happens when the wrapping occurs outside of any of the wrapped methods.

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-plugin/src/main/java/co/elastic/apm/agent/springwebflux/HandlerAdapterInstrumentation.java
```diff
@@ -98,31 +98,20 @@ public static Object onEnter(@Advice.Argument(0) ServerWebExchange exchange,
...
-                // without the need to use wrapping
+                // in transaction has been made no-op, we must still deactivate it
                 transaction.deactivate();
-                return resultMono;
             }
-
-            // in case an exception it thrown, it's too early to end transaction
-            // otherwise the status code returned isn't the expected one
-
-            return WebfluxHelper.wrapHandlerAdapter(tracer, resultMono, transaction, exchange);
```
SylvainJuge, line 122: [for review] this wraps the request processing, which is not necessary as we instrument the "dispatcher" which includes this.

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-plugin/src/main/java/co/elastic/apm/agent/springwebflux/TransactionAwareSubscriber.java
```diff
@@ -99,17 +60,35 @@
...
+
+        // store transaction into subscriber context it can be looked-up by reactor when the transaction
+        // is not already active in current thread.
+        this.context = subscriber.currentContext().put(AbstractSpan.class, transaction);
+    }
+
+    @Override
+    public Context currentContext() {
+        return context;
+    }
```
SylvainJuge, line 81: [for reviewer] this is exactly like with `TracedSubscriber`

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-plugin/src/main/java/co/elastic/apm/agent/springwebflux/TransactionAwareSubscriber.java
```diff
@@ -99,17 +60,35 @@
...
+
+    @Override
+    public void request(long n) {
+        subscription.request(n);
+    }
+
+    @Override
+    public void cancel() {
+        subscription.cancel();
+        cancelTransaction();
```
SylvainJuge, line 91: [for reviewer] When `onSubscribe` is called, we capture the original subscription and delegate in order to handle cancellation directly without extra callback.

Issue #1911, opened 2021-07-12
[META 352] AWS Lambda support


AlexanderWert:
Add support for AWS lambda as described in the meta issue: https://github.com/elastic/apm/issues/352

PR #1951, opened 2021-08-02
Lambda instrumentation

AlexanderWert:
## What does this PR do?
Closes https://github.com/elastic/apm-agent-java/issues/1911

## Checklist


apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/RequestHandlerInstrumentation.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Override
+    public ElementMatcher<? super NamedElement> getTypeMatcherPreFilter() {
+        return named(GlobalTracer.requireTracerImpl().getConfig(ServerlessConfiguration.class).getAwsLambdaHandler());
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return not(isInterface())
+            .and(hasSuperType(named("com.amazonaws.services.lambda.runtime.RequestHandler")));
```
eyalkoren, line 71: Why is this required? If the `_HANDLER` env variable is always set or we always require `aws_lambda_handler` to be configured, why do we need this? Can we replace this with the implementation of `getTypeMatcherPreFilter`?

(same question for `RequestStreamHandlerInstrumentation`)

eyalkoren: Answering myself: this is required in order to distinguish between `RequestStreamHandler` and `RequestStreamHandler` implementations, which require different instrumentation

apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/helper/AbstractAPIGatewayTransactionHelper.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    @Nullable
+    protected abstract String getHttpMethod(I event);
+
+    protected abstract String getApiGatewayVersion();
+
+    protected void fillHttpRequestData(Transaction transaction, @Nullable String httpMethod, @Nullable Map<String, String> headers, @Nullable String serverName, @Nullable String path, @Nullable String queryString, @Nullable String body) {
+        Request request = transaction.getContext().getRequest();
+        if (null != httpMethod) {
```
eyalkoren, line 60: Why do we give up all data when `httpMethod == null`?

AlexanderWert: You are right we can extract 

```
            fillUrlRelatedFields(request, serverName, path, queryString);
```

from the if block. The rest (i.e. body capturing) depends on the http method.

eyalkoren (approved):
Tested manually:
1. all handler types, with the matching event type and with a non-matching event type
2. both options of handle method specification - `package.Handler` vs `package.Handler::handleRequest`
3. metadata 
4. transaction context and name
5. some special configurations (e.g. `log_level` and `trace_methods`)

Documents were added as well, and an additional release step (not automated atm as it is probably only temporary).

docs/setup-aws-lambda.asciidoc
```diff
@@ -0,0 +1,88 @@
...
+Follow these steps to ensure proper Java Agent attachment to your Lambda function:
+
+1. Download the `elastic-apm-java-aws-lambda-layer-<VERSION>.zip` archive from our
+https://github.com/elastic/apm-agent-java/releases[GitHub Releases page].
+2. Publish the downloaded ZIP archive as an https://docs.aws.amazon.com/lambda/latest/dg/invocation-layers.html?icmpid=docs_lambda_help[AWS Lambda Layer].
+3. Configure your function to use the Java Agent layer.
+4. Within your Function configuration, configure the following environment variables:
+a. `AWS_LAMBDA_EXEC_WRAPPER` (required): set this variable's value to `/opt/elastic-apm-handler`, so that Lambda starts the runtime with this script.
+b. `ELASTIC_APM_LAMBDA_APM_SERVER` (required): this will tell the Elastic APM Extension where to send data to.
+c. `ELASTIC_APM_SECRET_TOKEN` or `ELASTIC_APM_API_KEY` (required): see the relevant https://github.com/elastic/apm-aws-lambda/blob/main/docs/aws-lambda-extension.asciidoc#the-necessary-variables[documentation].
```
AlexanderWert, line 52: Minor thing: I think this cross link here is not required. Just copy the text from there as it is quite short (would improve the experience):

This controls the authentication method that extension uses when sending data to the URL configured via `ELASTIC_APM_LAMBDA_APM_SERVER`.

eyalkoren: Done 👍

PR #1959, opened 2021-08-03
Downloading agent and verifying PGP signature

### LINKED ISSUE #1901: Download agent binaries from attacher
- Add a `--download-agent-version` option to the attacher CLI
- Download the agent binaries from maven central
- Validate the PGP signature and fail if it doesn't match the expected one
- Store them on disk, either in the same folder or in a subfolder
- If the binaries of the requested agent version are already in the folder, it will not re-download them
- Use the downloaded jar when attaching to a JVM

eyalkoren:
## What does this PR do?

Closes #1901 

We use the Bouncy Castle FIPS PGP implementation for the jar signature validation.
Since the related Bouncy Castle jars are doing self-checksum validation based on the jar they are coming from, we package them as is in a dedicated lib within the jar, alongside a minimal jar of ours that contains the Bouncy Castle implementation for the internal `BouncyCastleVerifier#verifyPgpSignature` API.
All these jars are loaded by a dedicated class loader.
## Checklist

- [x] Unit tests:
  - [x] Testing the new argument is parsed properly
  - [x] Testing downloading capabilities
  - [x] Testing PGP signature verification
  - [x] Testing dedicated PGP validation library class loader
- [x] End-to-end tests:
  - [x] Testing downloading and verification
  - [x] Testing downloading after already downloaded
  - [x] Testing wrong version
  - [x] Testing attachment with downloading
  - [x] Testing attachment without downloading (embedded agent or configured agent jar)
  - [x] Testing slim CLI jar - through our integration tests on Jetty, Tomcat and WebLogic with agent 1.25.0
- [x] Add the new CLI option to documentation:
  - [x] in docs
  - [x] in the CLI `--help` output
- [x] Add the Bouncy Castle license to our license notice


apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentDownloader.java
```diff
@@ -0,0 +1,190 @@
...
+            if (Files.isReadable(localAgentJarPath)) {
+                downloadAndVerify = false;
+            } else {
+                // quick and dirty - this is a temporary solution until we download agent from the Fleet package registry
+                logger.info("{} is not readable for the current user, removing and downloading again", localAgentJarPath);
+                Files.delete(localAgentJarPath);
+            }
+        }
+
+        if (downloadAndVerify) {
```
SylvainJuge, line 80: If file exists, we should still check signature, otherwise that's a way to inject untrusted code, skipping download might be still relevant though. Also, if there is any download error and file is corrupted, deleting the file would be a better option otherwise that would require manual cleanup.

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/PgpSignatureVerifierLoader.java
```diff
@@ -0,0 +1,141 @@
...
+    private final String verifierImplementationName;
+
+    static PgpSignatureVerifierLoader getInstance(final String sourceLib, final Path targetLib, String verifierClassName) throws Exception {
+
+        // done lazily so that we create the logger only after proper initialization
+        final Logger logger = LogManager.getLogger(PgpSignatureVerifierLoader.class);
+
+        Path libPath;
+        // first try to see if the lib is available as a resource available to the current class loader
+        URL resourceUrl = PgpSignatureVerifierLoader.class.getResource(sourceLib);
```
SylvainJuge, line 52: [question] why do we have to apply any fallback here ? Is that loading library is different when testing (BC is on the classpath), whereas when packaged we have a different behavior ?

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/Utils.java
```diff
@@ -0,0 +1,72 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class Utils {
+
+    private static final String AGENT_BASE_DIR = "agent";
+
+    private static Path agentDir = null;
+
+    public static Path getTargetAgentDir(String agentVersion) throws IOException {
```
SylvainJuge, line 33: Given that `getTargetAgentDir` is at most called once, there is no need to cache the `agentDir`, and if we call it many times with different versions, it would be confusing:
```
Path path1 = Utils.getTargetAgentDir("1.0.0"); // will return path to version 1.0.0 as expected
Path path2 = Utils.getTargetAgentDir("1.2.3"); // will return path to version 1.0.0
```

Thus, I would suggest to just remove the static field `agentDir` and always re-compute the path from the version.

apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/AgentDownloaderTest.java
```diff
@@ -0,0 +1,99 @@
...
+        String targetFileName = "test.asc";
+        final Path localFilePath = targetDir.resolve(targetFileName);
+        Files.deleteIfExists(localFilePath);
+        agentDownloader.downloadFile(mavenPgpSignatureUrl, localFilePath);
+        assertThat(Files.isReadable(localFilePath)).isTrue();
+        assertThatThrownBy(() -> agentDownloader.downloadFile(mavenPgpSignatureUrl, localFilePath)).isInstanceOf(FileAlreadyExistsException.class);
+    }
+
+    // disabled due to overhead, but very convenient for sanity testing
+    @Disabled
```
SylvainJuge, line 71: I'm not really a fan of having relevant tests disabled by default, do you think it would be possible to only disable them if they introduce test failures due to a dependency on maven central ? Ideally those kind of tests would be enabled but part of a JUnit test group that is run with retries (or for which we allow to have failures, but those aren't blocking the build).

integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/AbstractServletContainerIntegrationTest.java
```diff
@@ -81,11 +81,22 @@
...
+
     static boolean ENABLE_DEBUGGING = false;
     static boolean ENABLE_RUNTIME_ATTACH = true;
+
+    /**
+     * Set to a specific version to test downloading of agent from maven central using the slim cli tool.
+     * Only relevant if {@link #ENABLE_RUNTIME_ATTACH} is set to {@code true}.
+     */
+    @Nullable
+    private static final String AGENT_VERSION_TO_DOWNLOAD_FROM_MAVEN = null;
```
SylvainJuge, line 98: [question] If I understand correctly, this is only set when testing manually, just like the `ENABLE_DEBUGGING` is used to help when debugging issues. If yes, then adding a comment to make that 100% obvious could be relevant.

SylvainJuge (commented):
Another question:
Why do we have binary dependencies in the test classpath ? Are they kept in sync through build ?

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentDownloader.java
eyalkoren: > If file exists, we should still check signature, otherwise that's a way to inject untrusted code

Not really, if attackers have access to the server host, they can do whatever they want, including replacing the cli jar with malicious one. Similarly, we cannot control the verification of the cli jar itself, so if the user doesn't make sure to download it from a trusted place, it may contain malicious contents.

As far as I see it, we are only dealing with man-in-the-middle attacks here - make sure our cli tool downloads only the trusted agent jar.

> Also, if there is any download error and file is corrupted, deleting the file would be a better option otherwise that would require manual cleanup.

You mean - download has an error but ends without exception? Is that a case we need to consider?

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentDownloader.java
```diff
@@ -0,0 +1,190 @@
...
+                    Files.delete(localAgentJarPath);
+                }
+                throw throwable;
+            }
+        }
+        return localAgentJarPath;
+    }
+
+    String computeFileUrl(String mavenArtifactBaseUrl, String artifactFileName) {
+        return String.format(Locale.ROOT, "%s/%s", mavenArtifactBaseUrl, artifactFileName);
```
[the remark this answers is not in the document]
eyalkoren: Right, we should change to use the default

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentDownloader.java
```diff
@@ -0,0 +1,190 @@
...
+        HttpURLConnection urlConnection = openConnection(urlString);
+        urlConnection.setRequestMethod("HEAD");
+        urlConnection.connect();
+        return urlConnection.getResponseCode() == 200;
+    }
+
+    private HttpURLConnection openConnection(String urlString) throws IOException {
+        URL url = new URL(urlString);
+        HttpURLConnection urlConnection = (HttpURLConnection) url.openConnection();
+        urlConnection.addRequestProperty("User-Agent", "elastic-apm-agent-java-attach-cli");
```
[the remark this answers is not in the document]
eyalkoren: 👍

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/PgpSignatureVerifierLoader.java
eyalkoren: Add a comment that we provide both options to enable unit tests

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/Utils.java
eyalkoren: I'll change it so we a Utils instance per version

apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/AgentDownloaderTest.java
eyalkoren: You're right, we download lots of stuff from maven already, this makes sense that it's enabled

integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/AbstractServletContainerIntegrationTest.java
eyalkoren: Yes, you're right

SylvainJuge (approved):
LGTM, discussed few details live.

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentDownloader.java
SylvainJuge: summary of what has been discussed offline:
- when downloading, we should always check the signature, to prevent against "download successful but with wrong content"
- when file is already present, checking the signature of file would only detect if local file has been tampered with when we try to download it, but:
  a) we have no guarantee that attacher with download option will be called
  b) anyone having access to the folder where the attacher is could have replaced both the attacher and agent
  c) it is probably a better responsibility for the Fleet Agent to check for binary integrity (both Java agent and the cli-attacher, which could be implicit for the latter if embedded in Fleet agent).

apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/AgentDownloaderTest.java
eyalkoren: Enabling these tests

Issue #1963, opened 2021-08-06
Support Kotlin Coroutines


bastrich:
Currently Elastic Apm Java Agent doesn't support tracing context inside Kotlin Coroutines. And when you use Elastic Apm in Kotlin project, tracing doesn't work inside coroutines. So it would be quite useful to have Elastic Apm suppurt for Kotlin Coroutines.

I think the best solution is to implement Elastic Apm Java Agent Plugin for supporting Kotlin Coroutines.

Alternatively we can pass tracing context manually to kotlin coroutine context. But that doesn't allow other plugins to work correctly and adds a lot of boilerplate code.

samuelbirocchi:
Any news on this?

jackshirazi:
Hi Kotliners. It's unlikely that Elastic will put resources into this any time soon. Ideally the Kotlin community would add this into the agent. Here's some options
1. We've recently published a [Plugin API with examples](https://www.elastic.co/blog/create-your-own-instrumentation-with-the-java-agent-plugin) which maybe would assist adding this capability
2. Or more effortful would be a PR adding a plugin directly in the agent in this repo
3. Elastic APM accepts OpenTelemetry comms, so if there is an OpenTelemetry Kotlin agent that instruments coroutines, that might be an option

PR #1964, opened 2021-08-06
Implement Kotlin Coroutines Plugin

bastrich:
Implemented #1963 

Here I build CoroutineInterceptor and CoroutineWrapper classes manually with ByteBuddy and in the local scope of the advice method. Otherwise, if I build them in another static context or just write like simple Java/Kotlin classes, I get ClassDefNotFoundError for Kotlin classes when integrating with traced applications. I don't know why, but building classes with ByteBuddy in the local scope of advice function resolved that issue. And I wrote an integration test for such case.

- [x] This is a new plugin

### DIFF
--- a/apm-agent-plugins/apm-kotlin-concurrent-plugin/src/main/java/co/elastic/apm/agent/kotlinconcurrent/CoroutineConstructorAdvice.java
+++ b/apm-agent-plugins/apm-kotlin-concurrent-plugin/src/main/java/co/elastic/apm/agent/kotlinconcurrent/CoroutineConstructorAdvice.java
@@ -1,0 +1,164 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package co.elastic.apm.agent.kotlinconcurrent;
+
+import co.elastic.apm.agent.bci.TracerAwareInstrumentation;
+import co.elastic.apm.agent.impl.transaction.AbstractSpan;
+import kotlin.coroutines.AbstractCoroutineContextElement;
+import kotlin.coroutines.Continuation;
+import kotlin.coroutines.ContinuationInterceptor;
+import kotlin.coroutines.CoroutineContext;
+import net.bytebuddy.ByteBuddy;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.modifier.Visibility;
+import net.bytebuddy.dynamic.loading.ClassLoadingStrategy;
+import net.bytebuddy.implementation.FieldAccessor;
+import net.bytebuddy.implementation.MethodCall;
+import net.bytebuddy.implementation.bytecode.assign.Assigner;
+
+import java.lang.reflect.InvocationTargetException;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public abstract class CoroutineConstructorAdvice {
+    public static final ByteBuddy byteBuddy = new ByteBuddy();
+    public static final Object continuationTracingWrapperClassMonitor = new Object();
+    public static final Object tracingContinuationInterceptorClassMonitor = new Object();
+
+    private CoroutineConstructorAdvice() {
+    }
+
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter(@Advice.Argument(value = 0, readOnly = false) CoroutineContext coroutineContext)
+        throws NoSuchMethodException, InvocationTargetException, InstantiationException, IllegalAccessException {
+
+        final AbstractSpan<?> context = TracerAwareInstrumentation.tracer.getActive();
+        if (context != null) {
+            context.incrementReferences();
+
+            Class<?> continuationTracingWrapperClass;
+            try {
+                continuationTracingWrapperClass = Class.forName("ContinuationTracingWrapper");
+            } catch (ClassNotFoundException e) {
+                synchronized (continuationTracingWrapperClassMonitor) {
+                    try {
+                        continuationTracingWrapperClass = Class.forName("ContinuationTracingWrapper");
+                    } catch (ClassNotFoundException classNotFoundException) {
+                        continuationTracingWrapperClass = byteBuddy
+                            .subclass(Continuation.class)
+                            .modifiers(Visibility.PUBLIC)
+                            .name("ContinuationTracingWrapper")
+                            .defineField("span", AbstractSpan.class, Visibility.PRIVATE)
+                            .defineField("continuation", Continuation.class, Visibility.PRIVATE)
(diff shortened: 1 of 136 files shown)


bastrich:
Do I need to change project.version manually or here is automatic version bump?

bastrich:
@felixbarny should I add you to this PR? or how can it be reviewed?

thecloudgeek:
can we get this one implemented soon?

bastrich:
> can we get this one implemented soon?

not sure, see this thread on the forum - https://discuss.elastic.co/t/support-kotlin-coroutines/281846

patiramyadav:
> > can we get this one implemented soon?
> 
> @thecloudgeek not sure, see this thread on the forum - https://discuss.elastic.co/t/support-kotlin-coroutines/281846
> 
> but generally it works, you can copy the source code and build your own plugin

@bastrich How do I use this plugin in kotlin gradle based project? I have cloned this project and built locally. I have plugin jar file now.

patiramyadav:
```
2021-09-29 22:35:46,770 [reactor-http-nio-3] DEBUG co.elastic.apm.agent.mdc.MdcActivationListener - Class loader jdk.internal.loader.ClassLoaders$AppClassLoader@9e89d68 cannot load log4j API
java.lang.ClassNotFoundException: org.apache.log4j.MDC
	at jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581) ~[?:?]
	at jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178) ~[?:?]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:522) ~[?:?]
	at co.elastic.apm.agent.mdc.MdcActivationListener$2.get(MdcActivationListener.java:74) ~[elastic-apm-agent-1.26.1-SNAPSHOT.jar:?]
	at co.elastic.apm.agent.mdc.MdcActivationListener$2.get(MdcActivationListener.java:68) ~[elastic-apm-agent-1.26.1-SNAPSHOT.jar:?]
	at co.elastic.apm.agent.cache.WeakKeySoftValueLoadingCache$CacheValue.<init>(WeakKeySoftValueLoadingCache.java:119) ~[elastic-apm-agent-1.
```
I have 
```
    // logging deps
    implementation 'io.github.microutils:kotlin-logging:1.7.2'
```

apm-agent-plugins/apm-kotlin-concurrent-plugin/src/main/java/co/elastic/apm/agent/kotlinconcurrent/CoroutineConstructorAdvice.java
```diff
@@ -0,0 +1,166 @@
...
+
+public abstract class CoroutineConstructorAdvice {
+    public static final ByteBuddy byteBuddy = new ByteBuddy();
+    public static final Object continuationTracingWrapperClassMonitor = new Object();
+    public static final Object tracingContinuationInterceptorClassMonitor = new Object();
+
+    private CoroutineConstructorAdvice() {
+    }
+
+    @Advice.OnMethodEnter(suppress = Throwable.class)
```
patiramyadav, line 46: Here I think we also need to include ```inline=false``` like: ```@Advice.OnMethodEnter(suppress = Throwable.class, inline = false)```

jerryjpark:
Is there anything that needs to be configured like JVM configs for this AOP code to work? I'm getting below error when I launch the application from CLI. It is currently working for me only when I run the application in IntelliJ IDEA.

`java.lang.IllegalStateException: Error invoking (accessor)::defineClass
	at co.elastic.apm.agent.shaded.bytebuddy.dynamic.loading.ClassInjector$UsingReflection$Dispatcher$UsingUnsafeInjection.defineClass(ClassInjector.java:1024)
	at co.elastic.apm.agent.shaded.bytebuddy.dynamic.loading.ClassInjector$UsingReflection.injectRaw(ClassInjector.java:276)
	at co.elastic.apm.agent.shaded.bytebuddy.dynamic.loading.ClassInjector$AbstractBase.inject(ClassInjector.java:113)
	at co.elastic.apm.agent.shaded.bytebuddy.dynamic.loading.ClassLoadingStrategy$Default$InjectionDispatcher.load(ClassLoadingStrategy.java:233)
	at co.elastic.apm.agent.shaded.bytebuddy.dynamic.loading.ClassLoadingStrategy$Default.load(ClassLoadingStrategy.java:144)
	at co.elastic.apm.agent.shaded.bytebuddy.dynamic.TypeResolutionStrategy$Passive.initialize(TypeResolutionStrategy.java:100)
	at co.elastic.apm.agent.shaded.bytebuddy.dynamic.DynamicType$Default$Unloaded.load(DynamicType.java:6154)
	at kotlinx.coroutines.AbstractCoroutine.<init>(AbstractCoroutine.kt:35)
	at kotlinx.coroutines.AbstractCoroutine.<init>(AbstractCoroutine.kt:41)
	at kotlinx.coroutines.future.CompletableFutureCoroutine.<init>(Future.kt:51)
	at kotlinx.coroutines.future.FutureKt.future(Future.kt:42)
`

bastrich:
@jerryjpark how do you build and add an elastic apm dependency? which kotlin version?

jerryjpark:
@bastrich thanks for your reply. I don't have access to my computer now and I won't for couple more days.

I should've shared with you more of the stack trace. The underlying reason (which may or may not be apparent in the stack trace I shared above) is because the Coroutine class the complaining classloader was looking for was not available to the classloader. So, yes it could have to do with the particular version of the class binary the claasloader was looking for. So, I matched the (provided scope) Kotlin dependencies for your plugin to the version of Kotlin used in our project (1.5.0) and rebuilt the plugin to rule out version mismatch as an issue. Still, I cannot say for sure that it is not a version mismatch issue until I debug the Gradle package part. 

We are using Gradle and Gradle plugin to package our application with all its dependencies. We are clearly providing the correct version of the Kotlin dependencies that your plugin expects (at least per the Gradle project def).

Once I return to work, I plan to inspect the fat jar file to begin with to make sure the runtime Kotlin dependencies make to where they should be. Then, I will debug the class loaders if there can an issue for the Kotlin dependency to be found by the relevant classloader that throws CNFE.

samuelbirocchi:
Is this dead?

Reizinixc:
Does anyone have a plan or picked this up recently? It seems the PR is out-of-date for a while.

My team would like to use this to implement a new service instead of doing a workaround like the existing one.

Although we're absolute novices, we may consider making some changes to this.

Currently, I successfully converted the code that uses reflection to non-reflection, but it's very limited and still breaks the app for some cases, e.g.,
* Only `-javaagent` works, `apm-agent-attach` fails miserably to find `kotlinx.coroutines.AbstractCoroutine` (not sure yet why)
* Using a `suspend` function without `async` (need to add code to support thread switching). For instance,
   ```kotlin
   coroutineScope {
       val span = ElasticApm.currentSpan().startSpan("db", "mariadb", "query")

       span.setName("SELECT subdivisions").activate().use {
           fun dashIfEmpty(s: String) = if (s == "") "-" else s

           println("1 - ${ElasticApm.currentTransaction().id.let(::dashIfEmpty)}: ${ElasticApm.currentSpan().id.let(::dashIfEmpty)} (${Thread.currentThread().name})")
           delay(300L)
           println("1A - ${ElasticApm.currentTransaction().id.let(::dashIfEmpty)}: ${ElasticApm.currentSpan().id.let(::dashIfEmpty)} (${Thread.currentThread().name})")
           span.end()
       }
   }
   ```
   ```
   1 - c99bfe8ce5f05d85: a57e817c550748c9 (http-nio-8080-exec-1)
   2022-11-14 09:57:06,835 [http-nio-8080-exec-1] WARN  co.elastic.apm.agent.impl.ActiveStack - Deactivating a context ('ExampleController#controller' 00-b7beea814b3b854c500e814a55408e1e-c99bfe8ce5f05d85-01 (595962ca)) which is not the currently active one ('SELECT subdivisions' 00-b7beea814b3b854c500e814a55408e1e-a57e817c550748c9-01 (4a06cfe3)). This can happen when not properly deactivating a previous span or context.
   1A - -: - (kotlinx.coroutines.DefaultExecutor)
   ```

Issue #1555, opened 2020-12-07
Detect when agent is added in application classpath


SylvainJuge:
Agent is required to be loaded in the `bootstrap` classloader to ensure proper visibility of agent classes to the instrumented classes.

In practice, it's quite common that some setup issues trigger `ClassNotFoundException` with agent classes. This happens for example with some OSGi classloaders that do not delegate to `bootstrap` classloader unless they are configured to do so.

Given `ClassNotFoundException` symptom is often associated to `add an extra missing jar to your classpath` in an almost Pavlovian way to every Java developer, it's quite common that users try to add the agent jar to application classpath.

However, as agent jar is not a regular jar, this will cause more trouble than the initial issue, as some agent classes will be loaded in the wrong classloader and will create hard to diagnose issues.

Thus, we should be able to detect such setup errors at runtime and warn end-user of such issues with a proper error message. Optionally, we could even make the agent self-disable.

Heuristic / Implementation proposal(s)
- instrumentation with an empty advice that matches on `ElasticApmAgent` and checks whether the class loader is neither the System CL nor the bootstrap CL.

SylvainJuge:
Another heuristic that we could use to detect when agent is in the application classpath is the following:
- when agent is not packaged (in tests), then don't enforce anything
- when agent is packaged (could be detected by presence of agent `META-INF/MANIFEST.MF` in classpath), then we could easily check if `GlobalTracer` class is being loaded in the expected classloader: either bootstrap CL, or check the agent CL type when (and if) there will be such.

eyalkoren:
I like this proposal, let's do it!

PR #1966, opened 2021-08-08
Warn about using non-initialized tracer

eyalkoren:
## What does this PR do?

Warn when trying to invoke tracing operations before tracer initialization.
## Checklist

  - [x] I have added tests that prove my feature works
  - [x] I have added a description of the issue to the troubleshooting page

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/GlobalTracer.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/GlobalTracer.java
@@ -32,7 +32,7 @@ import java.util.Objects;
 public class GlobalTracer implements Tracer {
 
     private static final GlobalTracer INSTANCE = new GlobalTracer();
-    private volatile Tracer tracer = NoopTracer.INSTANCE;
+    private volatile Tracer tracer = new NoopTracer(TracerState.UNINITIALIZED);
 
     private GlobalTracer() {
     }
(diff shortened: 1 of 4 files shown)


SylvainJuge:
In the related issue that motivated this change, there were at least two setup issues:
- using `-javaagent` + runtime attach at the same time: this PR is not about this, but we should probably do something here
- having agent in the application classpath combined with a custom classloader that triggered having more than one instance of `GlobalTracer` and the second one is not properly initialized.

The change in this PR is about adding a warning when the side effect of the second issue becomes visible, which may be quite specific and uncommon, which also makes it quite hard and complex to reproduce. I think that in this case it would be more relevant to try to detect the root issue which is having the agent jar in application classpath (which is what #1555 is about).

eyalkoren:
Isn't this what we are looking for - when an instrumentation tries to invoke any tracing functionality on an uninitialized tracer?

An example scenario that I think would reproduce this: a webapp that adds the agent as dependency and is installed on Tomcat (that does a child-first lookup).
I assume it would happen as well in our pre-OSGi-fix agent, when added as dependency.

Don't forget that availability of the agent jar through more than one class loader in the CL hierarchy is the norm - the `-javaagent` adds the jar to the System class loader search path and we add it to the boot CL search path. As long as no custom class loading is used, anything in the JVM will always get the class loaded by the boot CL due to delegation.
So, without any custom class loading implementation (e.g. child-first), this CL setup (agent added to application as dependency):

```
Boot CL (can search the agent jar)
  └─ Platform CL
        └─ System CL (can search the agent jar)
              └─ custom CL (can search the agent jar)
```
is equivalent to this CL setup (agent only attached properly):
```
Boot CL (can search the agent jar)
  └─ Platform CL
        └─ System CL (can search the agent jar)
              └─ custom CL 
```

eyalkoren:
Closing in favor of the proposed solution by @SylvainJuge in https://github.com/elastic/apm-agent-java/issues/1555#issuecomment-901740868

PR #1968, opened 2021-08-10
Restoring instrumentation to io.grpc.internal.ManagedChannelImpl

eyalkoren:
## What does this PR do?

Attempting to fix failure to capture gRPC client spans. Context: https://discuss.elastic.co/t/java-grpc-client-not-ending-spans-and-not-reporting-to-apm/280643.

The first stage is just to restore the instrumentation of this `Channel` implementation, but it means client tests for latest version will fail. If we see the issue is resolved when restoring this instrumentation, next step would be to find a way to exclude the extra span.
## Checklist

- [ ] This is a bugfix


eyalkoren:
Closing as this doesn't seem to make a difference for the reported issue

PR #2065, opened 2021-08-16
implement to enable JvmDiscoverer tests on Windows

jackshirazi:
## What does this PR do?
Add implementations for Windows to run JvmDiscoverer functionality, including correctly handling spawned process IO so that a full output buffer from the spawned process doesn't hang the process, ie ensure that the output buffers from both stdout and stderr are fully read before checking if the process terminated

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/JvmDiscoverer.java
+++ b/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/JvmDiscoverer.java
@@ -208,11 +208,14 @@ public interface JvmDiscoverer {
                     //tasklist is always available, so no check for it
                     return JvmInfo.isJ9();
                 } else {
+                    List<String> args = new ArrayList<>();
+                    args.add("ps");
+                    args.add("aux");
                     return 
                             // attachment under hotspot involves executing a kill -3
                             // this would terminate false positive matching processes (ps aux | grep java)
                             JvmInfo.isJ9()
-                            && UserRegistry.User.executeCommand(List.of("ps", "aux")).exitedNormally();
+                            && UserRegistry.User.executeCommand(args).exitedNormally();
                 }
             } catch (Exception e) {
                 logger.error(e.getMessage(), e);
--- a/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/JvmDiscoverer.java
+++ b/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/JvmDiscoverer.java
@@ -161,17 +161,35 @@ public interface JvmDiscoverer {
         @Override
         public Collection<JvmInfo> discoverJvms() throws Exception {
             Collection<JvmInfo> jvms = new ArrayList<>();
-            Process process = new ProcessBuilder("ps", "aux").start();
+            Process process = Platform.isWindows() ?
+                    new ProcessBuilder("tasklist", "/FO", "CSV", "/V").start() :
+                    new ProcessBuilder("ps", "aux").start();
             BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
             for (String line = reader.readLine(); line != null; line = reader.readLine()) {
                 if (line.contains("java")) {
-                    String[] rows = line.split("\\s+");
-                    String pid = rows[1];
-                    String user = rows[0];
+                    String pid;
+                    String user;
+                    if (Platform.isWindows()) {
+                        String[] rows = line.split("\",\"");
+                        pid = rows[1];
+                        user = rows[6];
+                        if (user.contains("\\")) {
+                            //get rid of the domain
+                            user = user.substring(user.lastIndexOf('\\')+1);
+                        }
+                        //false positives might show from "java" being an argument
+                        if (!rows[0].startsWith("\"java")) {
+                            continue;
+                        }
+                    } else {
+                        String[] rows = line.split("\\s+");
+                        pid = rows[1];
+                        user = rows[0];
+                    }
                     try {
                         jvms.add(JvmInfo.of(pid, user, GetAgentProperties.getAgentAndSystemProperties(pid, userRegistry.get(user))));
                     } catch (Exception e) {
-                        logger.debug("Although the ps aux output contains 'java', the process {} does not seem to be a Java process.", pid);
+                        logger.debug("Although the process listing output contains 'java', the process {} does not seem to be a Java process.", pid);
                         logger.debug(line);
                         logger.debug(e.getMessage(), e);
                     }
@@ -184,11 +202,16 @@ public interface JvmDiscoverer {
         @Override
         public boolean isAvailable() {
             try {
-                return !Platform.isWindows()
-                    // attachment under hotspot involves executing a kill -3
-                    // this would terminate false positive matching processes (ps aux | grep java)
-                    && JvmInfo.isJ9()
-                    && new ProcessBuilder("ps", "aux").start().waitFor() == 0;
+                if (Platform.isWindows()) {
+                    //tasklist is always available, so no check for it
+                    return JvmInfo.isJ9();
+                } else {
+                    return 
+                            // attachment under hotspot involves executing a kill -3
+                            // this would terminate false positive matching processes (ps aux | grep java)
+                            JvmInfo.isJ9()
+                            && new ProcessBuilder("ps", "aux").start().waitFor() == 0;
+                }
             } catch (Exception e) {
                 logger.error(e.getMessage(), e);
                 return false;
--- a/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/JvmDiscoverer.java
+++ b/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/JvmDiscoverer.java
@@ -161,9 +161,11 @@ public interface JvmDiscoverer {
         @Override
         public Collection<JvmInfo> discoverJvms() throws Exception {
             Collection<JvmInfo> jvms = new ArrayList<>();
-            Process process = Platform.isWindows() ?
-                    new ProcessBuilder("tasklist", "/FO", "CSV", "/V").start() :
-                    new ProcessBuilder("ps", "aux").start();
+            ProcessBuilder builder = Platform.isWindows() ?
+                    new ProcessBuilder("tasklist", "/FO", "CSV", "/V") :
+                    new ProcessBuilder("ps", "aux");
+            builder.redirectErrorStream(true);
+            Process process = builder.start();
             BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
             for (String line = reader.readLine(); line != null; line = reader.readLine()) {
                 if (line.contains("java")) {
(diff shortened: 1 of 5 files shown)


apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
```diff
@@ -222,5 +249,111 @@ public boolean isCurrentUser() {
...
+                        // if it's not alive but there is still readable input, then continue reading
+                        isAlive = spawnedProcess.isAlive() || in.available() > 0;
+                    }
+                }
+                //would like to call waitFor(TIMEOUT) here, but that is 1.8+
+                //so will pause then have a go at getting the exit value
+                try {
+                    Thread.sleep(50);
+                } catch (InterruptedException e) {
+                    //no action, just means the next loop iteration is earlier
```
eyalkoren, line 291: [minor] which loop?

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
```diff
@@ -222,5 +249,111 @@ public boolean isCurrentUser() {
...
+            Exception exception = null;
+            StringBuilder commandOutput = new StringBuilder();
+            try {
+                spawnedProcess = buildTheProcess.start();
+
+                long timeout = 5 * 1000L; // NOTE 5 second timeout!
+                long start = System.currentTimeMillis();
+                long now = start;
+                boolean isAlive = true;
+                byte[] buffer = new byte[64 * 1000];
```
eyalkoren, line 268: [minor] quite a big buffer for command output, but since this is a standalone JVM it's that big of a deal.

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
```diff
@@ -222,5 +249,111 @@ public boolean isCurrentUser() {
...
+                exception = e1;
+                return new CommandOutput(commandOutput, exitValue, exception);
+            } finally {
+                // Cleanup as well as we can
+                if (spawnedProcess != null && spawnedProcess.isAlive()) {
+                    spawnedProcess.destroy();
+                    try {
+                        Thread.sleep(50);
+                    } catch (InterruptedException e) {
+                        //no action, just means the next loop iteration is earlier
```
eyalkoren, line 310: [minor] which loop?

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
```diff
@@ -222,5 +249,111 @@ public boolean isCurrentUser() {
...
+                    // when no longer need 1.7 compatibility, add these lines
+                    // try{Thread.sleep(50);}catch (InterruptedException e) {}
+                    // if (p.isAlive()) {
+                    // p.destroyForcibly();
+                    // }
+                }
+                try {
+                    exitValue = spawnedProcess.exitValue();
+                } catch (IllegalThreadStateException e2) {
+                    exception = e2;
```
eyalkoren, line 321: These `exitValue` and `exception` are ignored.
I believe you meat to do 
```
return new CommandOutput(commandOutput, exitValue, exception);
```
after the finally.

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
```diff
@@ -222,5 +249,111 @@ public boolean isCurrentUser() {
...
+                try {
+                    exitValue = spawnedProcess.exitValue();
+                } catch (IllegalThreadStateException e2) {
+                    //ignore and let the finally block have the last say
+                }
+
+                return new CommandOutput(commandOutput, exitValue, exception);
+            } catch (IOException e1) {
+                exception = e1;
+                return new CommandOutput(commandOutput, exitValue, exception);
```
eyalkoren, line 302: Wouldn't it be better if the `catch` here catches `Throwable` and then
```
return new CommandOutput(commandOutput, exitValue, exception);
```
would come after `finally`?

eyalkoren (commented):
Nice!
In addition to the minor inlined comments, can we eliminate the duplication of `runAsXXX` and `executeAsXXX` by migrating everything to work with `executeAs` and the `CommandOutput`?

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
jackshirazi:
```suggestion
                        //no action, just means the next loop iteration checking
                        //for timeout or process dead, is earlier
```

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
jackshirazi: Sorry, not sure whether you're saying it matters or doesn't matter. This would be called in the agent not the spawned process, but the buffer would only exist during the call. I can reduce it with no real issue though. What is your recommendation, 8k or 1k?

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
jackshirazi: hmm, I'll add comments to the file. Basically I call exitValue() in the main body just before returning from the body with the CommandOutput and if I don't do that then the exitValue() potentially doesn't get called until after the process exits in the finally. It turns out that somewhere in the JVM the process handling doesn't like it when I wait to the finally block, so sometimes it fails to correctly handle the spawned process.
The final call to exitValue() in the finally block is what finally sets the value in the variable, before the method exits and returns the CommandOutput (from either the main body or the catch clause). It's confusing because the JVM handling of spawned processes is confusing in 1.7 tbh. It gets better in 1.8
Please review after the next commit, if this doesn't make sense, or I got it wrong (entirely possible) then comment that, thx

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
jackshirazi: good call, let me make that change

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
jackshirazi: clarified in next commit

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
jackshirazi: clarified in next commit

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
jackshirazi: actually all of that doesn't matter, as I've found that the isAlive() call needs to be reimplmented for 1.7 compatibility so I have to call exitValue() earlier anyway. Will be in the next commit

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
eyalkoren: Shoot, I missed an important word in this sentence - its _**NOT**_ a big of a deal 🤦‍♂️ 
By standalone JVM I meant the cli tool rather than the user's JVM. I was only referring to the allocation of the buffer and it's really not super important.

Generally speaking, [minor] comments are used to say - "I noticed something, you may want to reconsider, but ignoring is perfectly acceptable". In this case, it was only about the fact that it's more than I'd expect for this purpose.

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
jackshirazi: fixed in commit 3a02a5cd094fd7eb4b033d3167e302463b90bf70

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
jackshirazi: fixed in commit 3a02a5cd094fd7eb4b033d3167e302463b90bf70

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
jackshirazi: fixed the flow in commit 3a02a5cd094fd7eb4b033d3167e302463b90bf70 per your suggestion

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
jackshirazi: fixed the flow in commit 3a02a5cd094fd7eb4b033d3167e302463b90bf70 per your suggestion

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/JvmDiscoverer.java
```diff
@@ -184,11 +202,16 @@ public boolean accept(File dir, String name) {
...
-                    && new ProcessBuilder("ps", "aux").start().waitFor() == 0;
+                if (Platform.isWindows()) {
+                    //tasklist is always available, so no check for it
+                    return JvmInfo.isJ9();
+                } else {
+                    return 
+                            // attachment under hotspot involves executing a kill -3
+                            // this would terminate false positive matching processes (ps aux | grep java)
+                            JvmInfo.isJ9()
+                            && new ProcessBuilder("ps", "aux").start().waitFor() == 0;
```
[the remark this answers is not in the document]
jackshirazi: yes, good catch

jackshirazi:
> In addition to the minor inlined comments, can we eliminate the duplication of `runAsXXX` and `executeAsXXX` by migrating everything to work with `executeAs` and the `CommandOutput`?

checking if this makes things clearer or not

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/UserRegistry.java
jackshirazi: thanks, changed the buffer to 4k in commit b0e8e4cf1c1db67e3cdb2aaedf7e2c26a331d902

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/JvmDiscoverer.java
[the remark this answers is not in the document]
jackshirazi: made the change in commit b0e8e4cf1c1db67e3cdb2aaedf7e2c26a331d902

jackshirazi:
> checking if this makes things clearer or not

It does, thanks for the suggestion, made the changes in commit b0e8e4cf1c1db67e3cdb2aaedf7e2c26a331d902

eyalkoren (approved):
Looks good!

PR #2067, opened 2021-08-18
gRPC instrumentation fix and refactoring

eyalkoren:
## What does this PR do?

This PR started as a fix to an issue reported in [the forum](https://discuss.elastic.co/t/java-grpc-client-not-ending-spans-and-not-reporting-to-apm/280643).

The issue is related to a change we did when tests of gRPC version 1.35 started failing due to duplicated client spans. In order to get rid of the extra span, we excluded `io.grpc.internal.ManagedChannelImpl$RealChannel` from our `ChannelInstrumentation`. However, this broke other scenarios, so this PR restores instrumentation of said channel and adds special handling for the `io.grpc.internal.DelayedClientCall` (added in 1.32) instead. This new client call type is created as a placeholder and starts going through the client call lifecycle, but at some point (different between 1.32 and 1.35) it starts referring to the real client call and delegate invocations to it. The new lifecycle management allows creation of duplicated spans, but when `io.grpc.internal.DelayedClientCall`s are used, their corresponding spans replace the ones created for the "real" calls once they are set as the delegate of the placeholder call, and the spans created for the real calls are discarded.

In addition, we now use `ensureInstrumented` to instrument client calls, server calls, client listeners and server listeners. By doing that, we can both be very efficient in type matching and still support custom types. While customer client/server calls are sort of edge case, custom listeners are probably more expected. Our tests now include custom client call and client/server listeners.

Lastly, I refactored all instrumentations to extract advice methods into separate (inner) classes, so to ensure proper dependencies visibility.

In a separate PR I intend to apply some enhancements to `SpanConcurrentHashMap` and then replace all maps within `GrpcHelper` with such.

## Checklist

  - [x] The reporting user confirmed general functionality and resolution of the reported issue

### DIFF
--- a/apm-agent-plugins/apm-grpc/apm-grpc-plugin/pom.xml
+++ b/apm-agent-plugins/apm-grpc/apm-grpc-plugin/pom.xml
@@ -37,7 +37,12 @@
             <version>${grpc.version}</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-netty-shaded</artifactId>
+            <version>${grpc.version}</version>
+            <scope>test</scope>
+        </dependency>
         <dependency>
             <groupId>io.cucumber</groupId>
             <artifactId>cucumber-java</artifactId>
--- a/apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/GrpcHelper.java
+++ b/apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/GrpcHelper.java
@@ -372,8 +372,8 @@ public class GrpcHelper {
      *     </li>
      *     <li>
      *         a single span was created and either mapped for both the placeholder and the real call (theoretical),
-     *         in which case we need to do nothing, or mapped only to the placeholder, in which to add a mapping for
-     *         the real client call
+     *         in which case we need to do nothing, or mapped only to the placeholder, in which case we need to add a
+     *         span mapping for the real client call
      *     </li>
      * </ul>
      *
--- a/apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/GrpcHelper.java
+++ b/apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/GrpcHelper.java
@@ -18,6 +18,7 @@
  */
 package co.elastic.apm.agent.grpc;
 
+import co.elastic.apm.agent.impl.GlobalTracer;
 import co.elastic.apm.agent.impl.Tracer;
 import co.elastic.apm.agent.impl.context.Destination;
 import co.elastic.apm.agent.impl.transaction.AbstractHeaderGetter;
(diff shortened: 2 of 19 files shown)


apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/GrpcHelper.java
```diff
@@ -312,25 +323,83 @@ public Span startSpan(@Nullable AbstractSpan<?> parent,
...
+            Span spanToMap = spanFromEntry;
+            if (spanToMap == null) {
+                // handling nested newCall() invocations - we still want to map the client call to the same span
+                Span tmp = GlobalTracer.get().getActiveSpan();
+                if (tmp != null && tmp.getSubtype() != null && tmp.getSubtype().equals(GRPC) && tmp.isExit()) {
+                    spanToMap = tmp;
+                }
+            }
+
+            if (spanToMap != null && !spanToMap.isDiscarded()) {
```
SylvainJuge, line 345: [question] in which case does the span would have been discarded ?

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/test/java/co/elastic/apm/agent/grpc/AbstractGrpcClientInstrumentationTest.java
```diff
@@ -156,7 +155,6 @@ void clientStreamingCallShouldBeIgnored() {
     }
 
     @Test
-    @Disabled // disabled for now as it's flaky on CI
```
SylvainJuge, line 159: :tada:

apm-agent-plugins/apm-grpc/apm-grpc-plugin/pom.xml
```diff
@@ -37,7 +37,12 @@
             <version>${grpc.version}</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-netty-shaded</artifactId>
+            <version>${grpc.version}</version>
+            <scope>test</scope>
+        </dependency>
```
SylvainJuge, line 45: [question] why do we need this extra dependency for testing ? is it required only for newer versions ?

SylvainJuge (approved):
LGTM, only a few questions/comments.

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/test/java/co/elastic/apm/agent/grpc/AbstractGrpcClientInstrumentationTest.java
eyalkoren: 🤞

apm-agent-plugins/apm-grpc/apm-grpc-plugin/pom.xml
eyalkoren: I used this for getting sources so I can debug an entire version. Since it's in `test` scope, I think worth keeping it for the same purpose.

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/GrpcHelper.java
eyalkoren: Not sure if it's a real possibility, but since we are discarding spans I think it worth the check - we don't need to continue with mappings for discarded spans.
Theoretical option: two internal gRPC channels delegate `newCall() ` from one to the other, where the nested call already sets the created call as the real call in the corresponding `DelayedClientCall`, causing the active span to be discarded already before the outer call exits.

PR #2076, opened 2021-08-23
classloader check on global tracer init

### LINKED ISSUE #1555: Detect when agent is added in application classpath

SylvainJuge:
## What does this PR do?

Fixes #1555

Ensures that the agent will throw an exception when agent has been added to application dependencies (and thus that the agent is not being loaded from the bootstrap classloader). While exception will likely break the application, it makes it obvious that there is a setup issue that requires immediate action.

## Checklist

- [x] This is something else

## Implementation checklist
- [x] ensure tests run properly with this extra check
- [x] manual testing with sample application with packaged agent as dependency using `Class.forname(...)` that tries to load agent `GlobalTracer`.

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/GlobalTracer.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/GlobalTracer.java
@@ -25,8 +25,10 @@ import co.elastic.apm.agent.impl.transaction.BinaryHeaderGetter;
 import co.elastic.apm.agent.impl.transaction.Span;
 import co.elastic.apm.agent.impl.transaction.TextHeaderGetter;
 import co.elastic.apm.agent.impl.transaction.Transaction;
+import co.elastic.apm.agent.util.VersionUtils;
 
 import javax.annotation.Nullable;
+import java.io.File;
 import java.io.IOException;
 import java.net.URL;
 import java.util.Enumeration;
@@ -79,6 +81,18 @@ public class GlobalTracer implements Tracer {
         String agentLocation = GlobalTracer.class.getProtectionDomain().getCodeSource().getLocation().getFile();
         if (!agentLocation.endsWith(".jar")) {
             // agent is not packaged, thus we assume running tests
+            classloaderCheckOk = true;
+            return;
+        }
+
+        String premainClass = VersionUtils.getManifestEntry(new File(agentLocation), "Premain-Class");
+        if (null == premainClass) {
+            // packaged within a .jar, but not within an agent jar, thus we assume it's still for testing
+            classloaderCheckOk = true;
+            return;
+        }
+
+        if (premainClass.startsWith("co.elastic.apm.agent")) {
             classloaderCheckOk = true;
             return;
         }
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/GlobalTracer.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/GlobalTracer.java
@@ -29,12 +29,7 @@ import co.elastic.apm.agent.util.VersionUtils;
 
 import javax.annotation.Nullable;
 import java.io.File;
-import java.io.IOException;
-import java.net.URL;
-import java.util.Enumeration;
 import java.util.Objects;
-import java.util.jar.Attributes;
-import java.util.jar.Manifest;
 
 public class GlobalTracer implements Tracer {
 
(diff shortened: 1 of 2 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/GlobalTracer.java
```diff
@@ -54,17 +63,52 @@ public static ElasticApmTracer requireTracerImpl() {
...
+                Manifest manifest = new Manifest(resources.nextElement().openStream());
+                Attributes attributes = manifest.getMainAttributes();
+                String preMain = attributes.getValue("Premain-Class");
+                if (preMain != null && preMain.startsWith("co.elastic.apm.agent")) {
+                    return false;
+                }
+            }
+        } catch (IOException e) {
+            throw new IllegalStateException(e);
+        }
```
eyalkoren, line 97: [minor]

Maybe you can reuse https://github.com/elastic/apm-agent-java/blob/f9267ff7fc7df619cb2272b8790fde618055c8f6/apm-agent-core/src/main/java/co/elastic/apm/agent/util/VersionUtils.java#L105

Or even just do something like https://github.com/elastic/apm-agent-java/blob/f9267ff7fc7df619cb2272b8790fde618055c8f6/apm-agent-premain/src/main/java/co/elastic/apm/agent/premain/AgentMain.java#L157 - we only want to see the class is coming from within a jar

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/GlobalTracer.java
```diff
@@ -54,17 +63,52 @@ public static ElasticApmTracer requireTracerImpl() {
...
     }
 
+    private static void checkClassloader() {
+        ClassLoader cl = GlobalTracer.class.getClassLoader();
+
+        if (classloaderCheck(cl)) {
+            return;
+        }
+
+        throw new IllegalStateException("Agent setup error: must be loaded in bootstap classloader, should not be included in application dependencies. Classloader : " + cl);
```
eyalkoren, line 73: [proposal] I think this message should be very focused on how the user can fix it - no need to say anything about class loaders even, only that the agent seems to have been setup improperly, and it is likely referenced by the JVM or application classpath, or used as a dependency.

eyalkoren (approved):
Elegant solution!
Added two proposals for you to consider

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/GlobalTracer.java
SylvainJuge: Very good idea, it's even simpler to check and does not require manifest parsing.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/GlobalTracer.java
SylvainJuge: Definitely agree here, the reason people tend to put this `.jar` on the classpath is because they have heard too much "put things on classpath" :-).

SylvainJuge:
Actually just checking that the class file is packaged in a `.jar` is not enough, as the current state of PR makes tests fail with :

```
java.lang.ExceptionInInitializerError
Caused by: java.lang.IllegalStateException: Agent setup error: agent jar file "/var/lib/jenkins/workspace/_java_apm-agent-java-mbp_PR-2076/src/github.com/elastic/apm-agent-java/apm-agent-core/target/apm-agent-core-1.25.1-SNAPSHOT.jar"  likely referenced in JVM or application classpath
```

Thus I'll revert to the previous version (or add an extra condition to only apply this check when properly packaged as an agent).

eyalkoren:
If the `GlobalTracer` is loaded from within the agent jar, wouldn't the other check also fail do to the presence of manifest?

PR #2081, opened 2021-08-25
Replace WeakConcurrentMaps with SpanConcurrentHashMaps in gRPC plugin

eyalkoren:
Implementing #2092 for the gRPC plugin


apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
```diff
@@ -123,14 +128,16 @@ private static JsonSchema getSchema(String resource) {
     }
 
     /**
-     * Sets all optional checks to their default value (enabled), should be used as a shortcut to reset mock reporter state
-     * after/before using it for a single test execution
+     * Resets checks to be executed based on the default behaviour.
+     * All optional checks are enabled and special requested behaviour is disabled.
```
SylvainJuge, line 132: [minor] we could be more specific here by replacing 'special requested behavior' with `GC execution before recycling checks`.

apm-agent-plugins/apm-grpc/apm-grpc-plugin/src/main/java/co/elastic/apm/agent/grpc/GrpcHelper.java
```diff
@@ -382,16 +384,38 @@ public void onClientCallCreationExit(@Nullable ClientCall<?, ?> clientCall, @Nul
      * @param realClientCall        the client call instance that represents the actual client call
      */
     public void replaceClientCallRegistration(ClientCall<?, ?> placeholderClientCall, ClientCall<?, ?> realClientCall) {
-        Span spanOfPlaceholder = delayedClientCallSpans.remove(placeholderClientCall);
+        // we cannot remove yet, because the span could have been ended already through ClientCall#start(), in which case
+        // it will be recycled ahead of time due to reference decrement when removed from the map
+        Span spanOfPlaceholder = delayedClientCallSpans.get(placeholderClientCall);
         if (spanOfPlaceholder != null) {
```
SylvainJuge, line 390: [minor] we could return early  to avoid one level of `if` nesting to make this slightly easier to read.

SylvainJuge (approved):

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
eyalkoren: It's written generically on purpose, as we keep adding nuances to our tests that are being reflected here. Similarly, we don't elaborate what the optional checks are.

PR #2082, opened 2021-08-25
CI Pipeline cleanup

SylvainJuge:
## What does this PR do?

- remove the `Run_As_Master_Branch` parameter
- remove the unused `push_docker` parameter
- add `integrationtests_ci` boolean parameter to include/exclude Integration Tests
- ensure that integration tests and JDK compatibility tests are not run on documentation updates

## Checklist

- [ ] check for any upstream build pipeline that could be using `push_docker` or `Run_As_Master_Branch`, if they are silently ignored when not present there is no impact.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,11 +32,10 @@ pipeline {
   }
   parameters {
     string(name: 'MAVEN_CONFIG', defaultValue: '-V -B -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn -Dhttps.protocols=TLSv1.2 -Dmaven.wagon.http.retryHandler.count=3 -Dmaven.wagon.httpconnectionManager.ttlSeconds=25', description: 'Additional maven options.')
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -291,7 +289,8 @@ pipeline {
           expression { return env.ONLY_DOCS == "false" }
           anyOf {
             changeRequest()
-            expression { return !params.Run_As_Master_Branch }
+            expression { return params.integrationtests_ci }
+            expression { return env.GITHUB_COMMENT?.contains('integration tests') }
```
v1v, line 293: Can you add this option in: https://github.com/elastic/apm-agent-java/pull/2082/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R31


```
issueCommentTrigger('(?i)(.*(?:jenkins\\W+)?run\\W+(?:the\\W+)?(?:(compatibility|benchmark|integration)\\W+)?tests(?:\\W+please)?.*|^/test(?:\\W+.*)?$)')
```

PR #2094, opened 2021-08-30
Migrate trace_methods to indy plugin

felixbarny:
## What does this PR do?

Relates to #1337 

## Checklist

- [x] This is something else


apm-agent-core/src/main/java/co/elastic/apm/agent/tracemethods/MethodMatcher.java
```diff
@@ -32,6 +33,7 @@
 import static co.elastic.apm.agent.matcher.AnnotationMatcher.annotationMatcher;
 import static co.elastic.apm.agent.matcher.WildcardMatcher.caseSensitiveMatcher;
 
+@GlobalState
```
SylvainJuge, line 36: [question] If I understand it correctly, we only need global state here as it relates to configuration like `MethodMatcherValueConverter`, and thus needs to be loaded in the main agent classloader and not in the plugin classloader, is that correct ? If so, maybe a small comment could make that slightly more obvious.

apm-agent-core/src/main/java/co/elastic/apm/agent/tracemethods/TraceMethodInstrumentation.java
```diff
@@ -153,7 +117,55 @@ public boolean matches(MethodDescription target) {
...
     @Override
-    public boolean indyPlugin() {
-        return false;
+    public String getAdviceClassName() {
+        return getClass().getName() + "$TraceMethodAdvice";
+    }
+
+    public static class TraceMethodAdvice {
+
+        private static final ElasticApmTracer tracer = GlobalTracer.requireTracerImpl();
```
SylvainJuge, line 126: [minor] moving tracer init to the start of the `static` block makes the dependency to retrieve configuration obvious.

SylvainJuge (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/tracemethods/MethodMatcher.java
felixbarny: Correct. Agree on adding an explanatory comment.

PR #2095, opened 2021-08-30
Migrate Kafka to indy plugin

felixbarny:
## What does this PR do?

Relates to #1337

## Checklist

- [x] This is something else


apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/postprocessor/AssignToPostProcessorFactory.java
```diff
@@ -108,33 +112,59 @@ public StackManipulation resolve(TypeDescription typeDescription, MethodDescript
...
+            throw new IllegalStateException("Cannot assign " + adviceMethod.getReturnType() + " to " + target + " in advice method " + adviceMethod.toGenericString());
+        }
+        StackManipulation load = MethodVariableAccess.of(adviceMethod.getReturnType()).loadFrom(exit ? argumentHandler.exit() : argumentHandler.enter());
+        return new StackManipulation.Compound(
+            loadThis ? MethodVariableAccess.loadThis() : StackManipulation.Trivial.INSTANCE,
+            load,
+            assign,
+            store
+        );
+    }
```
SylvainJuge, line 166: :exploding_head: I will trust you on this, congrats on refactoring this to make it more understandable.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/postprocessor/AssignToPostProcessorFactory.java
```diff
@@ -108,33 +112,59 @@ public StackManipulation resolve(TypeDescription typeDescription, MethodDescript
...
-                        throw new IllegalStateException("Advice method has to return Object[] when setting an index");
-                    }
-                    return new StackManipulation.Compound(
-                        MethodVariableAccess.REFERENCE.loadFrom(exit ? argumentHandler.exit() : argumentHandler.enter()),
-                        IntegerConstant.forValue(assignToReturn.index()),
-                        ArrayAccess.REFERENCE.load(),
-                        assigner.assign(TypeDescription.Generic.OBJECT, instrumentedMethod.getReturnType(), Assigner.Typing.DYNAMIC),
-                        MethodVariableAccess.of(instrumentedMethod.getReturnType()).storeAt(argumentHandler.returned())
-                    );
+                    return objectArrayStackManipulation(adviceMethod, assigner, false, argumentHandler, exit, assignToReturn.index(), instrumentedMethod.getReturnType(), MethodVariableAccess.of(instrumentedMethod.getReturnType()).storeAt(argumentHandler.returned()));
```
SylvainJuge, line 116: [minor] maybe reformat to have one parameter per line for methods calls that have plenty of parmeters/builders.

apm-agent-plugins/apm-kafka-plugin/apm-kafka-headers-plugin/src/main/java/co/elastic/apm/agent/kafka/ConsumerRecordsIteratorInstrumentation.java
```diff
@@ -58,24 +52,19 @@ public String getAdviceClassName() {
         return "co.elastic.apm.agent.kafka.ConsumerRecordsIteratorInstrumentation$ConsumerRecordsAdvice";
     }
 
-    @SuppressWarnings("rawtypes")
     public static class ConsumerRecordsAdvice {
 
+        public static final KafkaInstrumentationHeadersHelper helper = KafkaInstrumentationHeadersHelper.get();
```
SylvainJuge, line 57: [question/minor] do we have to make it `public` ?

apm-agent-plugins/apm-kafka-plugin/apm-kafka-base-plugin/src/main/java/co/elastic/apm/agent/kafka/KafkaProducerInstrumentation.java
```diff
@@ -66,36 +61,29 @@ public String getAdviceClassName() {
         return "co.elastic.apm.agent.kafka.KafkaProducerInstrumentation$KafkaProducerAdvice";
```
SylvainJuge, line 61: [minor] use `getClass().getName()` to make it consistent with others, doing that with all `getAdviceClassName` implementations could be a good idea.

apm-agent-plugins/apm-kafka-plugin/apm-kafka-headers-plugin/src/main/java/co/elastic/apm/agent/kafka/KafkaProducerHeadersInstrumentation.java
```diff
@@ -66,26 +55,23 @@ public KafkaProducerHeadersInstrumentation(ElasticApmTracer tracer) {
 
     @Override
     public String getAdviceClassName() {
-        return "co.elastic.apm.agent.kafka.KafkaProducerHeadersInstrumentation$KafkaProducerHeadersAdvice";
+        return getClass().getName() + "$KafkaProducerHeadersAdvice";
     }
 
     @SuppressWarnings("rawtypes")
```
SylvainJuge, line 61: [minor] can't we remove the `@SuppressWarnings("rawtypes")`here while we are at it ?

apm-agent-plugins/apm-kafka-plugin/apm-kafka-headers-plugin/src/main/java/co/elastic/apm/agent/kafka/KafkaProducerHeadersInstrumentation.java
```diff
@@ -94,56 +80,45 @@ public static Span beforeSend(@Advice.FieldValue("apiVersions") final ApiVersion
...
+                ProducerRecord copy = new ProducerRecord<>(record.topic(), record.partition(), record.timestamp(),
+                    record.key(), record.value(), record.headers());
+                headersHelper.removeTraceContextHeader(copy);
+                headersSupported = false;
+                thiz.send(copy, callback);
+                overrideThrowable = new Object[]{null};
             }
-            return false;
+            helper.onSendEnd(span, record, thiz, throwable);
+            return overrideThrowable;
```
SylvainJuge, line 120: [question] Here, could we just return `Throwable` to override the exception that we might have triggered ? Do we need to return an `Object[]` here, even if just with a single value ?

If I understand correctly, the behavior from the returned value here is
- `null` : don't change the thrown exception if there is any
- `Object[]{null}` : remove the exception
- `Object[]{ new Exception() }` : override/set the exception to throw

To me, returning an array would only be required in the unlikely case where we try to set the return value or a class attribute, in which case we would have to always return the full array with the original exception we got from `@Advice.Thrown` argument.

SylvainJuge (approved):
Great job on migrating this one !

apm-agent-plugins/apm-kafka-plugin/apm-kafka-headers-plugin/src/main/java/co/elastic/apm/agent/kafka/KafkaProducerHeadersInstrumentation.java
felixbarny: That doesn't work because we can only remove the exception with `Object[]{null}`.
When changing the return type to `Throwable` and returning `null`, it would still not change the exception if there's any.
That is to be consistent with the `@AssignTo.*` annotations that don't allow assigning something to `null`. The reason is that if an advice method throws an exception (which we suppress), the return value will always be null. We want to avoid assigning that null value as that will have unintended consequences.

PR #2101, opened 2021-09-01
Require separate advice class

felixbarny:
## What does this PR do?

Avoids class loading issues when the advice refers to library classes.

This allows to validate the advice class when the instrumentations get initialized
as opposed to on indy bootstrap.
That caches more errors and fails faster.

relates to #1337


SylvainJuge (approved):
LGTM, good idea to use `AdviceClass` as a generic default to avoid having to properly find dozen of relevant class names.

felixbarny:
Blocked by https://github.com/elastic/apm-agent-java/pull/2107/

PR #2102, opened 2021-09-03
LRU TypePoolCache

felixbarny:
## What does this PR do?
Reduces memory overhead by a smarter type pool caching strategy.

The type pool cache improves the startup times by speeding up type matching (determining whether a class that's about to be loaded should be instrumented). Generally, the more types that are cached, the faster the startup.

The old strategy did not impose a limit to the cache but cleared it after it hasn't been accessed in a while. However, load test have discovered that the cache may never be cleared and leave a permanent overhead of 23mb. The actual size of the cache highly depends on the application and loosely correlates with the number of loaded classes.

The new caching strategy targets to allocate 1% of the committed heap, at least 0.5mb and max 10mb. If a particular entry hasn't been accessed within 20s, it will be removed from the cache. 

 The results based on load testing are very positive:
* Equivalent startup times (within the margins of error of the previous strategy)
* Equivalent allocation rate (within the margins of error of the previous strategy)
* Reduced avg heap utilization from 10%/15mb (previous strategy) to within margins of error without the agent
* Reduced GC time due to the additional headroom that the application can utilize.
* Based on heap dump analysis, after warmup, the cache size is now around 59kb (down from 23mb with the previous strategy).

## Checklist

- [x] This is something else


felixbarny:
The integration tests for Tomcat and Jetty are failing.
There seem to be fewer method matches for the servlet instrumentation.
Compared to master, the logs match the following line only once instead of 4 times:
```
Method match for instrumentation ServletInstrumentation
```

jackshirazi (approved):
nice work!

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/LruTypePoolCache.java
```diff
@@ -0,0 +1,279 @@
...
+        cacheProviders.expungeStaleEntries();
+    }
+
+    private ConcurrentMap<String, ResolutionsByClassLoader> createCache() {
+        return new ConcurrentLinkedHashMap.Builder<String, ResolutionsByClassLoader>()
+            .maximumWeightedCapacity(maxCacheSize)
+            .weigher(new EntryWeigher<String, ResolutionsByClassLoader>() {
+                @Override
+                public int weightOf(String key, ResolutionsByClassLoader value) {
+                    return Math.max(1, value.size());
```
SylvainJuge, line 128: [question] what does weight is being used for here ? If I understand correctly that makes the entries added last to have an higher weight.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/LruTypePoolCache.java
```diff
@@ -0,0 +1,279 @@
...
+     * <p>
+     * By default, {@link ClassLoader#getSystemClassLoader()} is used as such a key as any resource location for the
+     * bootstrap class loader is performed via the system class loader within Byte Buddy as {@code null} cannot be queried
+     * for resources via method calls such that this does not make a difference.
+     * </p>
+     *
+     * @return A class loader to represent the bootstrap class loader.
+     */
+    private ClassLoader getBootstrapMarkerLoader() {
+        return ClassLoader.getSystemClassLoader();
```
SylvainJuge, line 166: [question] If I understand it correctly, it's only safe/correct to use the system CL here because we are just using it to load resources, and not to load classes themselves. Thus, it's a bit more than just a "marker" for map key.

SylvainJuge (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/LruTypePoolCache.java
felixbarny: I'd like to stick with the terminology used within Byte Buddy itself, which is where I copied this method from.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/LruTypePoolCache.java
felixbarny: clarified in comment

```java
// the key of the ConcurrentLinkedHashMap is the class name
// the value holds the resolutions by class loader
// if a given class name is loaded by more than one class loader,
// we need to adjust the weight of that entry accordingly
```

PR #2118, opened 2021-09-13
Fixing GlobalThreadLocal to use per-thread default values

### LINKED ISSUE #2117: `GlobalThreadLocal` uses constant default value
Instead, the constructor should be changed so that it receives a `Callable<T>` or a custom type (as we can't use `Supplier<T>`, `Function<Thread, T>` due to Java 7 compatibility).

Currently, there is an issue in the `co.elastic.apm.agent.okhttp.OkHttpClientHelper` which uses the same `StringBuilder` instance for every thread.

eyalkoren:
## What does this PR do?

Closes #2117 
## Checklist

- [x] This is a bugfix

### DIFF
--- a/apm-agent-plugin-sdk/src/test/java/co/elastic/apm/agent/sdk/state/GlobalThreadLocalTest.java
+++ b/apm-agent-plugin-sdk/src/test/java/co/elastic/apm/agent/sdk/state/GlobalThreadLocalTest.java
@@ -94,8 +94,7 @@ public class GlobalThreadLocalTest {
         future = executor.submit(threadLocal::getAndRemove);
         // we can rely on that because we use a single thread executor
         assertThat(future.get()).isEqualTo("worker");
-        future = executor.submit(threadLocal::getAndRemove);
-        // we can rely on that because we use a single thread executor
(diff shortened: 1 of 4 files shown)


apm-agent-plugin-sdk/src/test/java/co/elastic/apm/agent/sdk/state/GlobalThreadLocalTest.java
```diff
@@ -20,12 +20,103 @@
...
-        threadLocal.set(null);
+        GlobalThreadLocal.get(GlobalThreadLocalTest.class, "setNullValueShouldNotThrow").set(null);
+    }
+
+    @Test
+    void testNonConstantDefaultValue() throws ExecutionException, InterruptedException {
+        final GlobalThreadLocal<Object> threadLocal = GlobalThreadLocal.get(
+            GlobalThreadLocalTest.class,
+            "testNonConstantDefaultValue",
+            new ObjectValueSupplier()
```
felixbarny, line 46: [minor] may be replaced with lambda
```suggestion
            () -> new Object()
```

felixbarny (approved):

PR #2134, opened 2021-09-16
[release] use the vault secret/role credentials explicitly

v1v:
## What does this PR do?

Credentials where the release run do not match what we use in other CI instances, so let's override the default ones with the ones which are defined.

Fixes

image

Requires https://github.com/elastic/apm-pipeline-library/pull/1282


SylvainJuge (approved):
LGTM, I'll trust you on the content of the vault credentials :-)

v1v:
This PR is not validated by the CI, since it's the release pipeline. So I'll merge this PR now to proceed with the release

PR #2138, opened 2021-09-20
[release] post-release cleanup & improvements

SylvainJuge:
- [x] restore manually disabled steps
- [x] attempt to fix the call to `githubReleasePublish`

### DIFF
--- a/.ci/release/Jenkinsfile
+++ b/.ci/release/Jenkinsfile
@@ -266,8 +266,10 @@ pipeline {
                     return ret == 0
                 }
               }
-              githubEnv()
-              githubReleasePublish(id: "${env.RELEASE_ID}", name: "Release ${TAG_VER}")
+              script {
+                githubEnv()
+                githubReleasePublish(id: "${env.RELEASE_ID}", name: "Release ${TAG_VER}")
+              }
             }
           }
         }
--- a/.ci/release/Jenkinsfile
+++ b/.ci/release/Jenkinsfile
@@ -72,9 +72,6 @@ pipeline {
       options { skipDefaultCheckout () }
       stages{
         stage('Check oss.sonatype.org') {
-          when {
-            expression { return false }
-          }
           steps {
             // If this fails, an exception should be thrown and execution will halt
             dir("${BASE_DIR}"){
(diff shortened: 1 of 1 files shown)


v1v (approved):

SylvainJuge:
@v1v do you have any ideas how to better handle the items listed for clarification here ?

v1v:
> @v1v do you have any ideas how to better handle the items listed for clarification here ?

I'll comment in reverse order

>  Is there a way to have a git experience on CI that is closer to what we are used to locally ? For example, why do we have to always use git fetch ... commands in the pipeline ?

The CI, by default, uses a detached workspace, therefore it's required to fetch all the commits when doing a release. That particular automation, IIRC, it was delegated to the [withGitRelease](https://github.com/elastic/apm-pipeline-library/blob/master/vars/withGitRelease.groovy) and that particular step calls [setup-git-release.sh](https://github.com/elastic/apm-pipeline-library/blob/master/resources/scripts/setup-git-release.sh).

Other APM agents use it for their releases:

- https://github.com/elastic/apm-agent-ruby/blob/de937105d696b98720ad2152b4d6cfde50dcad2a/.ci/Jenkinsfile#L416-L419
- https://github.com/elastic/apm-agent-rum-js/blob/52f9118e587bef1831a743eee3a3166f9eba141a/Jenkinsfile#L547-L551

I don't know the reason those are needed to be explicit in the release pipeline though :/

>  the TAG_BARE and TAG_VER are set at the start of the pipeline and reused afterwards, but when the version is changed (for example when jumping from maven 1.25.1-SNAPSHOT to 1.26.0-SNAPSHOT), the git tag is not set to v1.26.0 until the maven release is executed, thus in this case the variables will have the values from the previous release !

Do you mean to reset those env variables afterwards? If so, `stage('Create GitHub release draft')` and  `stage('Publish release on GitHub')` are the ones that need those new values, is that right?

SylvainJuge:
Yes, using the `withGitRelease` seems probably a better option here instead of directly tinkering with shell & git commands.
If you agree with that, let's add another checkbox to the issue description so we can improve that.

----

Yes, the `TAG_BARE` and `TAG_VER` are implicitly changed by the `Set release version` step, and used afterwards for the stages `Create GitHub release draft` and `Publish release on GitHub`. Previously, they used to be referenced in shell scripts.

If we can have global variables (or even write a small wrapper function to replace them), that should be better.
Let me know what is better and I'll add another item to the issue to change this.

PR #2139, opened 2021-09-20
Improved capturing of logged exceptions when using Log4j2

tobiasstadler:
## What does this PR do?
There are multiple overloaded versions of Logger.error(message, cause) for Log4j 2, so relaxe the method matcher a bit

## Checklist

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -25,6 +25,7 @@ endif::[]
 
 [float]
 ===== Features
+* Improved capturing of logged exceptions when using Log4j2 {pull}2139[#2139]
 
 [float]
 ===== Bug fixes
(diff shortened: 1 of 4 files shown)


SylvainJuge:
Thanks for this PR @tobiasstadler !

Just did a very quick look at this PR, I have a couple of questions:
- should we relax the constraint only for Log4j2, or do other logging frameworks also have this ?
- there are a few cases where the `Throwable` parameter is not always in the second position ([see javadoc](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/index.html?org/apache/logging/log4j/message/AbstractMessageFactory.html)), are those indirectly covered by delegation ?

tobiasstadler:
> should we relax the constraint only for Log4j2, or do other logging frameworks also have this ? 

At the moment only SLF4J and Log4J2 are supported and afaik only Log4J2 supports non String messages

> there are a few cases where the `Throwable` parameter is not always in the second position (see javadoc), are those indirectly covered by delegation

It depends on the actual Logger implementation, but most implementations delegate to a private method. Thus they are not covered, neither before the PR nor with the PR.

SylvainJuge (approved):

SylvainJuge:
jenkins run the tests please

SylvainJuge:

tobiasstadler:
Thank You!

PR #2158, opened 2021-09-28
JDK 15+16 test fixes

SylvainJuge:
- jedis test fails due to known jdk 15 bug
- fix test SSL context init work-around
- tried to investigate a bit why log shading fails on Java 15 without success, thus disabling those tests on this JDK.


SylvainJuge:
Marked as blocked by https://github.com/elastic/apm-agent-java/pull/2184
Those JDK 15 and 16 test failures are too noisy, thus fixing JDK 17 first has higher priority.

SylvainJuge:
Closing this for now as JDK 16 support is currently OK and JDK 15 is documented as having issues in #2327

PR #2160, opened 2021-09-29
add jakarta.jms* JMS plugin support (EE9 renamed the packages)

jackshirazi:
## What does this PR do?
For EE9 Jakarta renamed all javax.* packages under their control to jakarta.* (because of a copyright conflict). We want to support JVMs using the new jakarta.* package names for plugins that we already have that support the older javax.* names. This PR specifically implements support for the JMS package name changes, javax.jms* to jakarta.jms*. We need our plugins to support both sets of package names independently and without conflict - a JVM could import both sets of names.

The strategy is to take our existing JMS plugin, and refactor it to support both package name sets.

Note we have some ActiveMQ tests. Artemis is the new version of ActiveMQ, and it is now dependent on jakarta.jms instead of javax.jms, but at the time of this PR is not entirely independent of javax.jms, leading to some incompatibilities, so some tests are commented out for now. An issue has been opened with ActiveMQ related to this.

The changes in this PR:
- add the dependency for jakarta.jms* (from jakarta-jms-api) - means that both javax.jms* and jakarta.jms* are available
- push the other dependencies from the common pom down into the project poms to avoid clashes of the same dependency jars with different versions (some versions depend on javax.jms, some on jakarta.jms)
- refactor the plugin so that either javax.jms or jakarta.jms is plugged in separately (and can do both in the same JVM) by the following steps:
- rename all files that "import javax.jms*" to Javax<Prevclassname>
- breakout into JavaxBaseJmsInstrumentation the javax.jmx specific part of BaseJmsInstrumentation
- add an equivalent JakartaBaseJmsInstrumentation
- copy all the Javax* classes to Jakarta* classes, replacing javax with jakarta in the new classes
- edit the package info to reference jakarta.jms as well as javax.jms
- add the new Jakarta class names to the resources/META-INF/services/co.elastic.apm.agent.sdk.ElasticApmInstrumentation file
- split the tests into separate sub-projects for Javax* and Jakartax* (but not the spring plugin tests for the reason below)
- leave the spring plugin tests as is only testing javax.jms as spring is not yet setup to handle jakarta.jms 

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else


jackshirazi:
now redone in https://github.com/elastic/apm-agent-java/pull/2282/

PR #2164, opened 2021-10-01
Simplify DynamicTransformer API

felixbarny:
## What does this PR do?

Instead of `DynamicTransformer.Accessor.get().ensureInstrumented`,users can just call `DynamicTransformer.ensureInstrumented`

Makes the API similar in style to the recent changes in `WeakConcurrent`.

## Checklist

- [ ] This is something else

### DIFF
--- a/apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/DynamicTransformer.java
+++ b/apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/DynamicTransformer.java
@@ -25,7 +25,7 @@ import java.util.Collection;
 import java.util.ServiceLoader;
 
 public class DynamicTransformer {
-    private static final DynamicTransformerSupplier transformer;
+    private static final DynamicTransformerProvider transformer;
 
     static {
         ClassLoader classLoader = DynamicTransformer.class.getClassLoader();
@@ -33,7 +33,7 @@ public class DynamicTransformer {
             classLoader = ClassLoader.getSystemClassLoader();
         }
         // loads the implementation provided by the core module without depending on the class or class name
-        transformer = ServiceLoader.load(DynamicTransformerSupplier.class, classLoader).iterator().next();
+        transformer = ServiceLoader.load(DynamicTransformerProvider.class, classLoader).iterator().next();
     }
 
     /**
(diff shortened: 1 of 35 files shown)


apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/DynamicTransformer.java
```diff
@@ -38,23 +48,13 @@
...
-        }
-
-        public static DynamicTransformer get() {
-            return transformer;
-        }
+    public static void ensureInstrumented(Class<?> classToInstrument, Collection<Class<? extends ElasticApmInstrumentation>> instrumentationClasses) {
+        transformer.ensureInstrumented(classToInstrument, instrumentationClasses);
     }
 
+    public interface DynamicTransformerSupplier {
```
eyalkoren, line 55: [minor] I don't think `DynamicTransformerSupplier` is a good name for this interface. I would call it `DynamicTransformer` and find a better name for the declaring class, even `StaticDynamicTransformer`, but preferably something better 🙂

eyalkoren (approved):
LGTM, only a minor naming suggestion

apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/DynamicTransformer.java
felixbarny: I agree that the name of the interface isn't ideal. But I wanted to optimize for the actual user facing class which is `DynamicTransformer`. While `DynamicTransformerSupplier` needs to be public, users never really get to see it. Also, it's in line with `WeakConcurrentSupplier`.

eyalkoren: It's just that `DynamicTransformer` doesn't supply anything (as opposed to the weak map supplier).
What about `RuntimeInstrumentor` for the interface?

And an additional suggestion - I just noticed that `DynamicTransformerImpl` seems redundant, why not make `ElasticApmAgent` the service provider implementing the interface (whatever it's called 🙂  ).

Either way, feel free to ignore both suggestions and merge if you disagree. It's not that important.

PR #2165, opened 2021-10-03
Introduce async profiler upgrade utility and upgrade to 1.8.7

eyalkoren:
## What does this PR do?

Introduces a JUnit utility for upgrading async profiler. The test class is disabled by default, and should be invoked manually when an upgrade is required.

In addition, async profiler is upgraded to version 1.8.7, and uses the newly introduced `safemode` setting option through system property.
## Checklist

- [x] async profiler is upgraded to 1.8.7
- [x] the agent sets the `AsyncProfiler.safemode` system property


apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/asyncprofiler/AsyncProfiler.java
```diff
@@ -71,6 +73,8 @@ public static AsyncProfiler getInstance(String profilerLibDirectory) {
                         "profiling_inferred_spans_enabled to false");
                 }
                 try {
+                    // set the AsyncProfiler.safemode with the configured safemode, so that optimizations can be applied already at load time
+                    System.setProperty(SAFEMODE_SYSTEM_PROPERTY_NAME, String.valueOf(safemode));
```
felixbarny, line 77: Can the safemode still be set to a different value at runtime?
Also, which optimizations?

SylvainJuge:
I did not looked at the added unit test details, but what are the benefits over using a test disabled by default over the `./apm-agent-plugins/apm-profiling-plugin/update-binaries.sh` shell script ? Having two ways to update the agent would likely be confusing.

eyalkoren:
> what are the benefits over using a test disabled by default over the ./apm-agent-plugins/apm-profiling-plugin/update-binaries.sh shell script ?

I actually didn't remember we had such 🤦‍♂️ 
If I did, I wouldn't bother. Now that it is implemented, benefits are:
- Readability
- Version verification after upgrade

Since I already spent this time, does it make sense to keep?

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/asyncprofiler/AsyncProfiler.java
eyalkoren: I will add comments.
The new System property is going to be removed in 2.x, where `CompiledMethodLoad` are not going to be enabled at all (which also answers the "optimization" question.

SylvainJuge:
> > what are the benefits over using a test disabled by default over the ./apm-agent-plugins/apm-profiling-plugin/update-binaries.sh shell script ?
> 
> I actually didn't remember we had such man_facepalming If I did, I wouldn't bother. Now that it is implemented, benefits are:
> 
> * Readability
> * Version verification after upgrade
> 
> Since I already spent this time, does it make sense to keep?

Let's keep the test version then, maybe instead of having a disabled test just a executable entry point so it's not reported as disabled/skipped in reports and it does not require any modification to execute.

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/asyncprofiler/AsyncProfiler.java
eyalkoren: https://github.com/elastic/apm-agent-java/pull/2165/commits/3f35ae1f7b69ebccabba0bc7083f8ba36552a648

CHANGELOG.asciidoc
```diff
@@ -37,6 +37,7 @@ exclude known static files in all applications might be changed to `/app/static/
 [float]
 ===== Features
 * Improved capturing of logged exceptions when using Log4j2 - {pull}2139[#2139]
+* Update to async-profiler 1.8.7 and set configured `safemode` already at load time though a new system property - {pull}2165[#2165]
```
SylvainJuge, line 40:
```suggestion
* Update to async-profiler 1.8.7 and set configured `safemode` at load time though a new system property - {pull}2165[#2165]
```

apm-agent-plugins/apm-profiling-plugin/pom.xml
```diff
@@ -15,5 +15,25 @@
...
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-compress</artifactId>
+            <version>1.21</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>tools.profiler</groupId>
+            <artifactId>async-profiler</artifactId>
+            <version>1.8.3</version>
```
SylvainJuge, line 34: [minor] might be worth having a build property to define the version, then reuse this in `AsyncProfilerUpgrader` to avoid any confusion. Or, maybe simpler if `AsyncProfilerUpgrader` would be able to extract the version from this dependency at runtimeand remove the `AsyncProfilerUpgrader#TARGET_VERSION`

SylvainJuge (approved):

apm-agent-plugins/apm-profiling-plugin/pom.xml
eyalkoren: This is the Java API dependency, not the lib itself - you see its version is 1.8.3. We don't need to maintain this unless the API breaks

Issue #1953, opened 2021-08-02
Update CI, tests & doc for Java 17 when GA


SylvainJuge:
Java 17 (GA expected in September 2021) is an LTS release.

- [x] ensure infra repository has binaries for JDK 17
- [x] update CI integration tests to add JDK 17 to it https://github.com/elastic/apm-agent-java/pull/2184
- [x] review any test that is disabled on those JDKs, see usages of `org.junit.jupiter.api.condition.DisabledOnJre` (or the Junit4 equivalent).
- [x] update documentation for supported versions

SylvainJuge:
Relates to https://github.com/elastic/apm-agent-java/issues/1944 for instrumentation errors.

SylvainJuge:
Proposal for running tests in CI:
- keep the build on JDK 11 for now, thus JDK 11 (LTS) is tested by default by the build pipeline
- keep the `compatibility_ci` toggle (still off by default) that enables running unit tests with few other JDKs
- remove all the non-LTS versions from the JDKs tested with `compatibility_ci`, only keep JDK 17 (LTS) and 18 (non-LTS but under development)
- further compatibility testing should be done through other means, either with load-testing or another form of integration test to test all the JDK versions that we might support (for example our unit tests do not run in Java7, but the agent supports it).

Proposal for documentation:
- link to Oracle official support page to clarify what LTS and non-LTS means: https://www.oracle.com/java/technologies/java-se-support-roadmap.html
- officially support LTS releases (at least for Oracle/OpenJDK): 7, 8, 11 and 17
- mark that non-LTS releases are usually compatible but are not officially supported, some of them (like JDK15) have known bugs that might make them not suitable to be used in production, with or without our agent.
- keep other vendors (IBM, HP-UX, SAP) as-is for now

jackshirazi:
- How long do we support/test non-LTS versions? My suggestion is 1 year after they have EOLed (they live for 6 months then are EOLed, but we don't start testing immediately anyway, so I'm suggesting we refer to a period from EOL rather than from GA)
- I think the HP, SAP are hotspot versions, so no need to test. But the IBM one is a completely different architecture from the ground up, so we should test, I'd say more important than doing both Oracle and OpenJDK since these are both just hotspot versions with very few differences
- the LTS cadence is moving to 2 years, we're already supporting 4 LTS versions, we need to also discuss removing official support for old LTS versions

SylvainJuge:
[...]

Issue #1942, opened 2021-07-29
system lambda jdk 16 compatibility issue


SylvainJuge:
We use [system-lambda](https://github.com/stefanbirkner/system-lambda) for testing custom environment variables.
There is a known compatibility issue with [JDK16](https://github.com/stefanbirkner/system-lambda/issues/23) that makes some of our tests to fail.

There is no reliable work-around for now, the only option would be to modify our test implementation (and possibly use another way to mock dependencies on `System` class, at least to set environment variables.

Known affected test(s)
- `co.elastic.apm.agent.impl.payload.ContainerInfoTest.testKubernetesDownwardApi`

How to reproduce:
- simply run tests with JDK 16

Work-around (not usable outside of testing)
- add `--add-opens=java.base/java.util=ALL-UNNAMED` into JVM arguments

SylvainJuge:
Alternatively, we could also create a very small abstraction over the calls to `System` class when retrieving environment variables.
When testing, we can use an alternate implementation that allows setting arbitrary values.

PR #2184, opened 2021-10-12
Compatibility test JDK 17

SylvainJuge:
## What does this PR do?

Part of #1953 

- disable jdk 16 and jdk 15 compatibility tests as they produce noise for now
- add jdk 17 to the list of tested JDKs for compatibility tests
- upgrade maven wrapper + version to 3.8.3 + upgrade a few plugins

## Checklist

TODO : add here one item in checklist per failing test that require investigation
- [x] System Lambda in `ContainerInfoTest` is not compatible with Java 16+ (see #1942 as well)
- [x] Log4j1.x reached EOL and MDC relies on Java version parsing
- [x] Change how we ensure that we don't initialize default SSL context before the application.


felixbarny:
Seems like Java 17 is not available on CI currently

```
Error: JAVA_HOME is not defined correctly.
 We cannot execute /var/lib/jenkins/.java/openjdk17/bin/java
```

SylvainJuge:
run compatibility tests

jackshirazi (approved):

PR #2195, opened 2021-10-18
Adding utility to mock environment variables

eyalkoren:
Adding a utility to test scenarios with custom env variables.
Originally implemented for removal of system-lambda (as part of elastic/apm-agent-java#2184), but I want to use it in a current PR I am doing now.


eyalkoren:
closes #1942

PR #2219, opened 2021-10-30
Support quartz v1.x

### LINKED ISSUE #2212: Add support for Quartz 1.x
Currently, we only support Quartz 2.x. 
However, it seems to be quite easy to also support Quartz 1.x users.

The main issue is that `org.quartz.JobExecutionContext` changed from being a class in 1.x to an interface in 2.x. Therefore, we have to have two dedicated modules and compile against 1.x and 2.x. That's because compiling against an interface yields in an `invokeinterface` rather than an `invokevirtual` byte code instruction.

The simplest solution would be to just copy/paste the instrumentation.

Alternatively, we could create a quartz-common module that contains the instrumentation, the advice, and a test suite that works with 1.x and 2.x. Then create quartz-1 and quartz-2 modules that implement an interface that's provided by the quartz-common module. The interface takes a generic argument, representing either the interface-based or class-based `org.quartz.JobExecutionContext` and returns a `String` representing the job key (`context.getJobDetail().getKey().toString()`). The quartz-common module implements the interface, too, accepting an `Object` as the argument. It then checks whether the object represents an interface or a class and calls the corresponding implementation, accordingly. The dispatcher needs to know the exact class names from the quartz-1/2 sub-modules and loads them in the constructor and stores them in two dedicated fields.

kananindzya:
## What does this PR do?

closes #2212

## Checklist


eyalkoren (approved):
Looks awesome! 👏 
I pushed some minor changes so to reduce some back and forth, see if you disagree with anything.
Thanks! ❤️

eyalkoren:
run elasticsearch-ci/docs

Issue #2176, opened 2021-10-07
Off-by-one errors in Java APM Integration Tests


cachedout:
Over the past year, we have seen an ongoing pattern of cases wherein we see [APM Integration tests](https://github.com/elastic/apm-integration-testing) fail because a single event which we believe to have sent was not received and an assertion fails.

### Error
AssertionError: queried for [('processor.event', 'transaction'), ('service.name', ['springapp'])], expected 1000, got 999

### Examples

https://apm-ci.elastic.co/job/apm-integration-tests/job/7.x/751/testReport/junit/tests.agent/test_java/Integration_Tests___All___test_concurrent_req_java_spring/

https://apm-ci.elastic.co/job/apm-integration-tests/job/7.x/738/testReport/junit/tests.agent/test_java/Integration_Tests___All___test_concurrent_req_java_spring/

### Discussion

I would like to discuss either finding a fix for the underlying issue or perhaps adjusting the assertion to allow for either 999 or 1,000 as acceptable values if we agree that this condition does not represent an actual software defect.

If we could do this, it would go a long way toward increasing the overall stability of the APM Integration Test suite which is used by many teams to verify the functionality of the APM suite of products.

Thanks!

cc: @elastic/observablt-robots

eyalkoren:
This is the case for years, thanks for taking the initiative to address that!!

Is it specific only to Java? If so, should we still assume the event is sent and not received?

In order to understand whether all events are sent, if we set agent with `log_level=debug` and ingest the test logs somewhere, we may be able to count log events. 

Any serialization and send attempt should log (expecting 1000):
```
Receiving TRANSACTION event (sequence <ring-buffer-slot-number>)
```
Any serialization/write failure should log:
```
Failed to handle event of type TRANSACTION with this error: ...
```
Failure to create a connection with the APM Server should log:
```
Failed to get APM server connection, dropping event: ...
```
Failure to send should log:
```
Error sending data to APM server: ...
```

In addition, at shutdown we log:
```
Reported events: <num-events>
Dropped events: <num-events>
```
where the dropped events should contain failures to send.

Once we get to a conclusion whether the right number was sent, we know whether to investigate upstream (agent tracing logic) or downstream (APM Server).

jackshirazi:
I believe this is no longer an issue. Leaving open for a couple of days, then will close if no concerns raised

PR #2224, opened 2021-11-01
Delaying reporter start

eyalkoren:
Currently, this is just a test to POC the approach of reporting events before starting the disruptor without losing events.

@felixbarny this is what we discussed - it can be used in `ApmServerReporter#start` that would asynchronously invoke a blocking `ReportingEventHandler#init` and start the disruptor only after it returns. It can then block on metadata discovery as well as APM Server client and health check, so that we can start accumulating events in the ring buffer before we are ready to send them.

One drawback of this approach is that it won't cover for intermittent connection failures with APM server, so maybe the best approach is to simply resubmit events to the buffer when send fails.

I currently open this as draft so that we can further look into it later on. 
Applying better coverage to these scenarios may also resolve the inconsistent integration tests failures reported in #2176

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java
@@ -113,8 +113,8 @@ public class ApmServerReporter implements Reporter {
 
     @Override
     public void start() {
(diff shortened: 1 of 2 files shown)


eyalkoren:
Won't merge this way. We may consider this or other approach in the future.

PR #2228, opened 2021-11-02
Improve flush

felixbarny:
## What does this PR do?

Like #1897, but only the hard flush improvements

Benefits
- More efficient flushing (less wait time and garbage-free)
- Flushing does not block when connection to APM Server is unhealthy
- Agent does not delay application shutdown when connection to APM Server is unhealthy

## Checklist

  - ~[ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)~


apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
```diff
@@ -132,77 +167,43 @@ private void writeEvent(ReportingEvent event) {
...
-
-        @Override
-        public boolean cancel() {
-            final boolean cancel = super.cancel();
-            final Future<Void> flush = this.flush;
-            if (flush != null) {
-                flush.cancel(false);
-            }
-            return cancel;
+            reporter.scheduleWakeupEvent();
```
eyalkoren, line 206: Still makes sense to wrap within try-catch, so that the thread doesn't die from any reason

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
```diff
@@ -37,22 +39,28 @@
...
-    private final Timer timeoutTimer;
+    private final ScheduledExecutorService timeoutTimer;
     @Nullable
-    private ApmServerReporter reporter;
-    @Nullable
-    private TimerTask timeoutTask;
+    private Runnable timeoutTask;
+    private final AtomicLong processed = new AtomicLong();
+
+    public IntakeV2ReportingEventHandler(ReporterConfiguration reporterConfiguration, ProcessorEventHandler processorEventHandler,
```
eyalkoren, line 47:
```suggestion
   IntakeV2ReportingEventHandler(ReporterConfiguration reporterConfiguration, ProcessorEventHandler processorEventHandler,
```

Maybe also javadoc that this is intended for tests

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
```diff
@@ -132,77 +167,43 @@ private void writeEvent(ReportingEvent event) {
...
-    }
-
     @Override
     @Nullable
     protected HttpURLConnection startRequest(String endpoint) throws Exception {
         HttpURLConnection connection = super.startRequest(endpoint);
         if (connection != null) {
-            if (reporter != null) {
-                timeoutTask = new FlushOnTimeoutTimerTask(reporter);
+            if (timeoutTask != null) {
```
eyalkoren, line 175: Can this ever be null?

apm-agent-core/src/test/java/co/elastic/apm/agent/report/ApmServerReporterIntegrationTest.java
```diff
@@ -136,17 +164,53 @@ void testSecretToken() throws ExecutionException, InterruptedException {
...
-        reporter.flush().get();
+        reporter.flush();
+        assertThat(reporter.getDropped()).isEqualTo(0);
+        assertThat(receivedIntakeApiCalls.get()).isEqualTo(1);
+    }
+
+    @Test
+    void testTimeout() {
+        doReturn(TimeDuration.of("1ms")).when(reporterConfiguration).getApiRequestTime();
+        doAnswer(invocation -> System.nanoTime()).when(clock).nanoTicks();
```
eyalkoren, line 183: Why is this required? Isn't `System.nanoTime()` the default implementation anyway?

apm-agent-core/src/test/java/co/elastic/apm/agent/report/ApmServerReporterIntegrationTest.java
```diff
@@ -136,17 +164,53 @@ void testSecretToken() throws ExecutionException, InterruptedException {
...
+        reporter.report(new Transaction(tracer));
+        reporter.flush(1, TimeUnit.SECONDS);
+        reporter.report(new Transaction(tracer));
+        reporter.flush(1, TimeUnit.SECONDS);
+        reporter.report(new Transaction(tracer));
+        reporter.flush(1, TimeUnit.SECONDS);
+        reporter.report(new Transaction(tracer));
+
+        assertThat(v2handler.isHealthy()).isFalse();
+        assertThat(reporter.flush(1, TimeUnit.SECONDS)).isFalse();
```
eyalkoren, line 213:
```suggestion
        assertThat(reporter.flush(0, TimeUnit.SECONDS)).isFalse();
```
Should be shortcut by the fact that `isHealthy() == false`, right?

apm-agent-core/src/test/java/co/elastic/apm/agent/report/ApmServerReporterIntegrationTest.java
```diff
@@ -136,17 +164,53 @@ void testSecretToken() throws ExecutionException, InterruptedException {
...
+    void testFailingApmServer() {
+        statusCode = HttpStatus.SERVICE_UNAVAILABLE_503;
+        // try to report a few events to trigger backoff
+        reporter.report(new Transaction(tracer));
+        reporter.flush(1, TimeUnit.SECONDS);
+        reporter.report(new Transaction(tracer));
+        reporter.flush(1, TimeUnit.SECONDS);
+        reporter.report(new Transaction(tracer));
+        reporter.flush(1, TimeUnit.SECONDS);
+        reporter.report(new Transaction(tracer));
```
eyalkoren, line 210: Why not test with much lower timeouts? We want to see that the default case is very quick

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
felixbarny: Not in practice. But `timeoutTask` is `@Nullable` as it's initialized in the `init` method.

apm-agent-core/src/test/java/co/elastic/apm/agent/report/ApmServerReporterIntegrationTest.java
felixbarny: I removed the whole clock abstraction as it was only really needed for the soft flush tests.

apm-agent-core/src/test/java/co/elastic/apm/agent/report/ApmServerReporterIntegrationTest.java
felixbarny: Yes, the default should be quick. But I want to ensure that the backoff is triggered. Waiting for longer than needed ensures this test is less flaky. If the timeout is too low, we're not waiting for the mock APM Server to respond and thus, would not trigger the backoff. I'm flushing multiple times here as the fist backoff period is 0 seconds, meaning it retries immediately.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
felixbarny: I removed the Clock abstraction (see below).

eyalkoren (approved):

PR #2235, opened 2021-11-04
Add the (current) transaction name to the error object

tobiasstadler:
## What does this PR do?
Since elastic/apm-server#6539 it is possible to add the transaction name to an error object. I added support for it to the java agent.

## Checklist
- [x] This is something else

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/error/ErrorCapture.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/error/ErrorCapture.java
@@ -239,6 +239,11 @@ public class ErrorCapture implements Recyclable {
          */
         private boolean isSampled;
         /**
+         * The related TransactionInfo name
+         */
+        @Nullable
+        private String name;
+        /**
          * The related TransactionInfo type
          */
         @Nullable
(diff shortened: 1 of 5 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/error/ErrorCapture.java
```diff
@@ -238,6 +238,11 @@ public ErrorCapture deactivate() {
          * A hint for UI to be able to show whether a recorded trace for the corresponding transaction is expected
          */
         private boolean isSampled;
+        /**
+         * The related TransactionInfo name
+         */
+        @Nullable
+        private String name;
```
felixbarny, line 245: What do you think about changing this to a `StringBuilder` so that it can be recycled?

felixbarny:
Implements https://github.com/elastic/apm/pull/541

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/error/ErrorCapture.java
tobiasstadler: So copying the content on `AbstractSpan.getNameForSerialization` to `TransactionInfo.name`?

felixbarny: yes, exactly.

tobiasstadler: Done

felixbarny (approved):

jackshirazi (approved):

felixbarny:
run elasticsearch-ci/docs

tobiasstadler:
Thank You!

PR #2238, opened 2021-11-04
No Content-Encoding when disabling compression for localhost

felixbarny:
## What does this PR do?
Implements the latest changes in the spec (https://github.com/elastic/apm/pull/518) that disallows the `NO_COMPRESSION` level.


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/stacktrace/StacktraceConfiguration.java
```diff
@@ -89,6 +89,9 @@ public int getStackTraceLimit() {
     }
 
     public long getSpanFramesMinDurationMs() {
+        if (spanFramesMinDurationMs.isDefault()) {
+
+        }
```
eyalkoren, line 94: Leftover?

eyalkoren (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/stacktrace/StacktraceConfiguration.java
felixbarny, line 94: Yep
```suggestion
```

PR #2240, opened 2021-11-05
Refactor ResourceExtractionUtil to use NIO

felixbarny:


apm-agent-common/src/main/java/co/elastic/apm/agent/common/util/ResourceExtractionUtil.java
```diff
@@ -18,98 +18,127 @@
...
+                    // wait until other JVM instances have fully written the file
+                    // multiple JVMs can read the file at the same time
+                    try (FileLock readLock = channel.lock(0, Long.MAX_VALUE, true)) {
+                        if (!hash(Files.newInputStream(tempFile)).equals(resourceHash)) {
+                            throw new IllegalStateException("Invalid checksum of " + tempFile + ". Please delete this file.");
+                        } else if (!Files.getOwner(tempFile).equals(currentUserPrincipal)) {
+                            throw new IllegalStateException("File " + tempFile + " is not owned by '" + currentUserPrincipal.getName() + "'. Please delete this file.");
                         }
                     }
                 }
```
eyalkoren, line 112: [proposal] This is a `synchronized` method, I think a clearer path is to start with checking `java.nio.file.Files#exists()` and if so do all that for the internal JVM synchronization. 
You can extract this to a different method and do the same check and handling after getting the `writeLock` for the cross-JVMs synchronization.
I think it's clearer than relying on getting a `FileAlreadyExistsException`.

eyalkoren (approved):

apm-agent-common/src/main/java/co/elastic/apm/agent/common/util/ResourceExtractionUtil.java
felixbarny: It's actually important to rely on the exception as `Files#exists` is vulnerable to check-then-act race conditions.

`FileChannel.open`, can atomically create a new file with a specific set of file attributes. It's guaranteed that if it returns without throwing an exception, no other user has created a file with the same name.

eyalkoren: OK

PR #2246, opened 2021-11-09
Fix error transaction name serialization check and add tests

eyalkoren:


apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
```diff
@@ -339,9 +339,7 @@ private void serializeError(ErrorCapture errorCapture) {
     private void serializeErrorTransactionInfo(ErrorCapture.TransactionInfo errorTransactionInfo) {
         writeFieldName("transaction");
         jw.writeByte(JsonWriter.OBJECT_START);
-        if (errorTransactionInfo.getName() != null) {
```
tobiasstadler, line 342: Sorry, I missed that when I converted String to StringBuilder

eyalkoren: No worries! Effectively that was equivalent as it was always `true`. 
We are thankful for your great contributions!

PR #2252, opened 2021-11-10
Ignoring a deliberately old version of Byte Buddy in dependabot.yml

eyalkoren:
## What does this PR do?
Context: #2251 
We do want automatic PRs for Byte Buddy version upgrades, but we have a specific place where we want to keep an older version that can remain constant.
Based on docs, the [`ignore`](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#ignore) option with specified `versions` should work.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,3 +31,7 @@ updates:
       - dependency-name: "com.squareup.okhttp3:okhttp"
       - dependency-name: "com.datastax.cassandra:cassandra-driver-core"
       - dependency-name: "com.datastax.oss:java-driver-core"
+    ignore:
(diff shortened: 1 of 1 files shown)


cachedout (approved):
[The docs](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#ignore) are a bit unclear about versions but it _seems_ like this is the correct way to pin a version. If not, we can always adjust it again. :)

SylvainJuge (approved):

PR #2255, opened 2021-11-11
Register and schedule Micrometer MeterRegistry when Meters are registered

eyalkoren:
## What does this PR do?

As reported in https://github.com/spring-projects/spring-boot/issues/26630, premature initialization of Micrometer's `MeterRegistry` in Spring applications may cause metrics loss.
This PR attempts to work around this issue assuming that `Meter`s are only registered to fully initialized meter registries.
This assumption is yet to be verified.
## Checklist

- [x] This is a bugfix
  - [x] I verified the described issue is resolved using this fix where it is reproducible without this fix
  - [x] I upgraded Micrometer dependency to latest and updated [`dependabot.yml`](https://github.com/elastic/apm-agent-java/blob/master/.github/dependabot.yml) so that we continuously test against the latest version
  - [x] I added a `MicrometerInstrumentationVersionsIT` to test with version 1.0.1


apm-agent-plugins/apm-micrometer-plugin/src/main/java/co/elastic/apm/agent/micrometer/MicrometerInstrumentation.java
```diff
@@ -22,38 +22,27 @@
...
     @Override
     public ElementMatcher<? super TypeDescription> getTypeMatcher() {
-        return hasSuperType(named("io.micrometer.core.instrument.MeterRegistry"));
+        return named("io.micrometer.core.instrument.MeterRegistry");
     }
 
     @Override
     public ElementMatcher<? super MethodDescription> getMethodMatcher() {
-        return isPublic().and(not(isStatic()));
+        return named("registerMeterIfNecessary");
```
felixbarny, line 45: As this is a private method, we should add micrometer to the allow list in [`dependabot.yml`](https://github.com/elastic/apm-agent-java/blob/master/.github/dependabot.yml) and always test against the latest version. To ensure compatibility with older versions, let's add a `VersionsIT`.

eyalkoren: Agreed, a `VersionsIT` is required for the instrumentation test. 
I already know that upgrading the default version to the latest will require adaptation of the reporter test, so we'll first verify this solves the problem.

eyalkoren: Added

apm-agent-plugins/apm-micrometer-plugin/src/test/java/co/elastic/apm/agent/micrometer/MicrometerMetricsReporterTest.java
```diff
@@ -368,7 +368,6 @@ void testExclusionOfFailedGauge_singleGauge() {
             .isEmpty();
 
         getMetricSets();
-        assertThat(metricsReporter.getFailedMeters().iterator().next().getId().getName()).isEqualTo("gauge1");
```
SylvainJuge, line 371: [minor] Was this change required due to upgrading micrometer that have different behavior between 1.0.1 and 1.8.0 or due to the change of instrumented method ?

SylvainJuge (approved):
LGTM, should be good as it has been validated in the field.

apm-agent-plugins/apm-micrometer-plugin/src/test/java/co/elastic/apm/agent/micrometer/MicrometerMetricsReporterTest.java
eyalkoren: It's a different behavior - in the new version the abstract meters capture exceptions quietly, so we don't get to see that they failed

PR #2264, opened 2021-11-16
[release] split git commands in sh steps with some labels

v1v:
## What does this PR do?

Add the `label` to help with the UI visualisation and the semantic
Add `--force` in case the branch already exists when running the checkout

```
       -f, --force
           When switching branches, proceed even if the index or the working tree differs from HEAD. This is used to throw away local changes.

           When checking out paths from the index, do not fail upon unmerged entries; instead, unmerged entries are ignored.
```


eyalkoren (approved):
LGTM, but shouldn't it start with a clean work tree?

PR #2265, opened 2021-11-16
[release] enable stage temporarily for bypassing input parameter

v1v:
## What does this PR do?

Disable the original stage and create a new one to avoid using maven but the input parameter only


eyalkoren (commented):
If there's a way to manually set the relevant env variables for the specific job, I think it's better. Is there a way?

eyalkoren (approved):
Let's use it, but make sure not to leave it in :)

PR #2268, opened 2021-11-16
[release] branch and checkout 

v1v:
## What does this PR do?

Replace checkout --track with  branch and checkout


eyalkoren (commented):
You need to `checkout` before doing this

eyalkoren (approved):
Yes, let's try this one 🙂

PR #2269, opened 2021-11-16
[release] use git checkout and reset

v1v:
## What does this PR do?

Another attempt to solve the issues when restarting the workspace to its previous state


eyalkoren (commented):
You cannot do `-D` if it is the current branch

eyalkoren (commented):
Let's also `fetch origin/${BRANCH_SPECIFIER}`

eyalkoren (commented):
The `'Update Cloudfoundry'` stage should always update master, regardless of which branch was used for the release

PR #2283, opened 2021-11-24
Improve runtime attach configuration

SylvainJuge:
This is an improvement of previous changes https://github.com/elastic/apm-agent-java/pull/1469 in this area

### Issue description

If `config_file` is used in runtime attach API, the configuration is merged with runtime attach parameters:
- the log shows `Attachment parameters` for the configuration source, whereas the configuration comes from a configuration properties file, which might be confusing for the user trying to debug configuration.
- because the configuration is copied to attach parameters, the external configuration is loaded with low priority that do not match [documentation](https://www.elastic.co/guide/en/apm/agent/java/current/configuration.html) for `config_file`

### Implemented changes
- for runtime attach load `config_file` in the exact same way as with an environment variable or JVM property.
- Provide a reliable hint in agent logs on the actual configuration origin: path to a configuration file, classpath resource, runtime attach parameters.
- test CLI & self-attach configuration with integration tests (which rely on some agent logs parsing).

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -777,8 +777,8 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
      * Makes sure to not initialize ConfigurationOption, which would initialize the logger
      */
     @Nullable
-    public static String getConfigFileLocation(List<ConfigurationSource> configurationSources, boolean runtimeAttach) {
-        String configFileLocation = runtimeAttach ? null : DEFAULT_CONFIG_FILE;
+    public static String getConfigFileLocation(List<ConfigurationSource> configurationSources, boolean premain) {
+        String configFileLocation = premain ? DEFAULT_CONFIG_FILE : null;
         for (ConfigurationSource source : configurationSources) {
             String value = source.getValue(CONFIG_FILE);
             if (value != null) {
@@ -787,7 +787,7 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
             }
         }
 
-        if (configFileLocation == null || runtimeAttach || !configFileLocation.contains(AGENT_HOME_PLACEHOLDER)) {
+        if (configFileLocation == null || !premain || !configFileLocation.contains(AGENT_HOME_PLACEHOLDER)) {
             return configFileLocation;
         }
 
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -25,10 +25,10 @@ import co.elastic.apm.agent.configuration.converter.TimeDuration;
 import co.elastic.apm.agent.configuration.converter.TimeDurationValueConverter;
 import co.elastic.apm.agent.configuration.validation.RegexValidator;
 import co.elastic.apm.agent.impl.transaction.Span;
-import co.elastic.apm.agent.matcher.WildcardMatcher;
-import co.elastic.apm.agent.matcher.WildcardMatcherValueConverter;
 import co.elastic.apm.agent.matcher.MethodMatcher;
 import co.elastic.apm.agent.matcher.MethodMatcherValueConverter;
+import co.elastic.apm.agent.matcher.WildcardMatcher;
+import co.elastic.apm.agent.matcher.WildcardMatcherValueConverter;
 import org.stagemonitor.configuration.ConfigurationOption;
 import org.stagemonitor.configuration.ConfigurationOptionProvider;
 import org.stagemonitor.configuration.converter.MapValueConverter;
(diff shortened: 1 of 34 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -777,25 +777,26 @@ public TimeDuration getSpanMinDuration() {
...
-            if (agentHome != null) {
-                return configFileLocation.replace(AGENT_HOME_PLACEHOLDER, agentHome);
-            } else {
-                return null;
-            }
-        } else {
+
+        if (configFileLocation == null || !premain || !configFileLocation.contains(AGENT_HOME_PLACEHOLDER)) {
             return configFileLocation;
         }
```
eyalkoren, line 792: It's confusing and I don't think necessary.
At this point, if `configFileLocation != null` always return `configFileLocation.replace(AGENT_HOME_PLACEHOLDER, agentHome)`, otherwise return `null`

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/source/PropertyFileConfigurationSource.java
```diff
@@ -36,30 +39,59 @@
...
-        Properties props = getFromClasspath(location, ClassLoader.getSystemClassLoader());
-        if (props == null) {
-            props = getFromFileSystem(location);
+        SimpleSource source = new SimpleSource(name);
+        for (String key : properties.stringPropertyNames()) {
+            source.add(key, properties.getProperty(key));
+        }
+        return source;
+    }
+
```
eyalkoren, line 85: I prefer we don't have this here - those are factory methods in a type of a config source that create a different type of a config source

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
SylvainJuge: Should be simplified now.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/source/PropertyFileConfigurationSource.java
SylvainJuge: Created `ConfigSources` class just for that purpose.

integration-tests/runtime-attach/runtime-attach-app/src/main/java/co/elastic/apm/testapp/AppMain.java
```diff
@@ -50,18 +54,36 @@ public static void main(String[] args) {
...
+                System.out.format("Using self attach with classpath configuration\n");
+                ElasticApmAttacher.attach();
+            } else {
+                System.out.format("Using self attach with external configuration %s\n", agentConfig.getAbsolutePath());
+                Map<String, String> config = new HashMap<>();
+                config.put("service_name", "self-attach-external-config");
+                config.put("config_file", agentConfig.getAbsolutePath());
+                ElasticApmAttacher.attach(config);
+            }
             ElasticApmAttacher.attach();
```
jackshirazi, line 83: shouldn't this be removed (the attach happens in the `then` and `else` clauses?

integration-tests/runtime-attach/runtime-attach-test/src/test/java/co/elastic/apm/testapp/RuntimeAttachTest.java
```diff
@@ -88,41 +103,137 @@ void after() {
...
-            "disable_send=true",
-            "application_packages=co.elastic.apm.testapp",
-            "cloud_provider=NONE");
+        waitForInstrumentedWork();
+    }
+
+    @Test
+    void testExternalConfigurationCli(@TempDir File tmp) throws IOException {
+        // only testing with cli-attach for simplicity as it's easier to provide custom parameters
+        // should behave in a similar way with
```
jackshirazi, line 151: should behave in a similar way with what?

SylvainJuge: similar was as using the Attacher API.

integration-tests/runtime-attach/runtime-attach-app/src/main/java/co/elastic/apm/testapp/AppMain.java
SylvainJuge: Good catch, trying to attach twice will be ignored in practice if already attached.

apm-agent-core/src/test/java/co/elastic/apm/agent/configuration/source/ConfigSourcesTest.java
```diff
@@ -0,0 +1,108 @@
...
+    void loadFromClasspath() throws IOException {
+        ConfigurationSource source = ConfigSources.fromClasspath("test.elasticapm.properties", ClassLoader.getSystemClassLoader());
+        assertThat(source).isNotNull();
+
+        assertThat(source.getName()).isEqualTo("classpath:test.elasticapm.properties");
+        assertThat(source.getValue("application_packages")).isEqualTo("co.elastic.apm");
+        assertThat(source.getValue("log_level")).isEqualTo("DEBUG");
+
+        // should be no-op
+        source.reload();
```
jackshirazi, line 45: this doesn't show if it's not a no-op but doesn't throw an exception. Maybe change a value before the reload, then check that the reload didn't re-set the value back?

SylvainJuge: Maybe it might be better to just check the implementation type (here it should be a `SimpleSource`), and not a subclass of it.

jackshirazi (approved):

PR #2286, opened 2021-11-24
Only remove domain when FQDN is expected

### LINKED ISSUE #2284: About hostname
My server name is service type+IP, which was correct in kibana before version 1.26. The new version 1.27 omits the part
The name is  like `business-10.0.0.1`,now the kibana display `business-10`
so ,How can the apm to display the full name? or I need to set property apm.hostname?

eyalkoren:
Fixes #2284


jackshirazi (approved):

SylvainJuge:
I've just found that there was another minor issue with using external commands, the output is not trimmed, thus locally I got (on linux) `myhostname\n` and it's sent as-is to apm-server :facepalm:.

PR #2293, opened 2021-11-26
Support for excluding JDBC spans from being captured

tobiasstadler:
## What does this PR do?
Adds support to exclude SQLs from being captured.

## Checklist


apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/JDBCConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+import co.elastic.apm.agent.configuration.converter.ListValueConverter;
+import co.elastic.apm.agent.matcher.WildcardMatcher;
+import co.elastic.apm.agent.matcher.WildcardMatcherValueConverter;
+import org.stagemonitor.configuration.ConfigurationOption;
+import org.stagemonitor.configuration.ConfigurationOptionProvider;
+
+import java.util.Collections;
+import java.util.List;
+
+public class JDBCConfiguration extends ConfigurationOptionProvider {
```
tobiasstadler, line 30: Maybe it is better to rename it to `DbConfiguration` and move it to apm-agent-core

felixbarny:
Have you considered dropping these spans in an ingest node processor instead? See also https://www.elastic.co/guide/en/elasticsearch/reference/master/drop-processor.html

The tradeoff is that this doesn't decrease the overhead of capturing, serializing, and sending the span. But if it's not an overwhelming amount of spans that need to be dropped, it seems reasonable and works across all agents. 

Could you describe your use case in a little more detail? Do you want to drop the spans for connection validation queries? In that case, it seems like an ingest node processor would be a good option.

If that's not an option, it might make sense to make the feature more generic so that it works for all spans with a `db.statement`. The tradeoff is that the spans would still need to be collected and would then be dropped after the fact (for example by calling `span.requestDiscarding()` after the span has ended.

tobiasstadler:
Currently it is only about validation queries, which account for about 30% of our db queries. We may want to exclude out queries as well, but we haven't decided yet.

I didn't think of using an ingest pipeline. Thanks for the tip. We will implement that for the time being.

Yes, a more general solution is probably the better approach. But since we are only using JDBC for DB access, I went for a JDBC only solution.

Dropping the spans after they ended might impact all spans, as for every span it has to be checked if `db.statement` exists.

tobiasstadler:
While the ingest pipeline works fine, I still think excluding statements in the agent is valuable. In our case we can exclude many SQL statements that currently need to be send to the APM-Server and Elasticsearch only to be dropped there.  Maybe we can avoid that occasionally the queue of our APM-Server is full (without tuning the server/increasing its resources).

I am fine if this is an experimental/internal option until a more general solution is available (which might need an issue/pr in the apm repo).

tobiasstadler:
@felixbarny I created elastic/apm#590 to move this forward

tobiasstadler:
@felixbarny Would `sqls_excluded_from_jdbc_instrumentation` being an internal configuration option be an option?

PR #2295, opened 2021-11-29
Always include detected hostname

SylvainJuge:
## What does this PR do?

With `1.27.0` we do not send the `detected_hostname` field when `configured_hostname` is being sent.
That avoids running an external command, however it triggers the following [bug](https://github.com/elastic/kibana/issues/119737) in Kibana.

This PR changes the current implementation that was introduced in `1.27.0` with the following behavior for `metadata`:
- provide the `system.hostname` for APM server < 7.4
- provide `system.detected_hostname` for APM server >= 7.4
- provide `system.configured_hostname` for APM server >= 7.4 and when agent `hostname` configuration is set.

## Checklist

- [x] This is a bugfix

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,6 +28,7 @@ endif::[]
 * Fixing missing Micrometer metrics in Spring boot due to premature initialization - {pull}2255[#2255]
 * Fixing hostname trimming of FQDN too aggressive - {pull}2286[#2286]
 * Fixing agent `unknown` version - {pull}2289[#2289]
+* Always capture agent hostname - {pull}2295[#2295]
 
 
 [[release-notes-1.x]]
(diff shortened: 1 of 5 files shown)


jackshirazi (approved):
looks good

eyalkoren:
The current [spec](https://github.com/elastic/apm/blob/master/specs/agents/metadata.md) reflects the fact that `detected_hostname` is not mandatory. Maybe it should be updated accordingly, or we should inform other agent teams otherwise.

SylvainJuge:
So, after a discussion with @eyalkoren, merging this PR does not seem the right thing to do for the following reasons:
- the `host.hostname` field that was used by Kibana is deprecated and should not be relied on
- the impact on users is quite minimal, it only shows `N/A` for hostname in one place, and it does not breaks any product feature
- changing this behavior at agents level implies to modify the spec + ensure that all agents comply to it (thus quite some overhead)
- apm server implementation relies on `detected_hostname` to be optional, there is no usage of sending both `configured_hostname` and `detected_hostname` as the detected one is ignored when one is explicitly configured.
- `detected_hostname` relies on external commands execution, which make them depend on OS and platform, we have some heuristics to avoid FQDN, both of which are likely to be unreliable and require a simple work-around if they fail. With current implementation setting `configured_hostname` allows to provide this work-around in case hostname detection does not behave as expected.

PR #2302, opened 2021-12-01
[release] fix+refactor major branch update

SylvainJuge:
## What does this PR do?

- rewrite major branch update steps in bash
- move cloudfoundry update script to `.ci/release` folder

The most annoying part was dealing with "shallow clone" used in Jenkins which is easily reproduced using similar checkout command:
```
git clone git@github.com:SylvainJuge/git-playground.git --depth 1 --no-tags
```

With "shallow clone", we have to explicitly enable fetching branches afterwards with the following command:
```
git remote set-branches ${REMOTE_NAME} '*'
```

## Checklist
- [x] test with another git repository as apm-agent-java
- [x] ensure script works with shallow clone (which is used in Jenkins jobs)
- [x] ensure script works with "regular" clone
- [x] test with invalid input version: should fail with explicit message
- [x] test when remote tag is missing: should fail with explicit message
- [x] test with existing major branch: should force-update existing
- [x] test with non-existing major branch: should create branch
- [x] test with existing local major branch (which happens quite often outside of CI): the local branch should be deleted


v1v (commented):
Can you also delete -> https://github.com/elastic/apm-agent-java/blob/c013264202922422ba1285f9d65d2488c9d34583/.ci/release/Jenkinsfile#L140-L158 ?

It's another leftover :/

v1v (approved):

PR #2310, opened 2021-12-05
Attacher CLI docs and logging changes

### LINKED ISSUE #2196: Testing Agent attach with Fleet

eyalkoren:
Closes #2196 

Currently `--include/exclude-main` and `--include/exclude-vmargs` use regex to match, whereas `--include/exclude-user` and `--include-pid` use exact match.
I think that's the right way to go.


apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentAttacher.java
```diff
@@ -503,15 +503,15 @@ static void printHelp(PrintStream out) {
             out.println("        A list of PIDs to include.");
             out.println();
             out.println("    --include-main/--exclude-main <pattern>...");
-            out.println("        A list of regular expressions of fully qualified main class names or paths to JARs of applications the java agent should be attached to.");
+            out.println("        A regular expression of fully qualified main class names or paths to JARs of applications the java agent should be attached to.");
             out.println("        Performs a partial match so that `foo` matches `/bin/foo.jar`.");
             out.println();
-            out.println("    --include-vmarg/--exclude-vmarg <pattern>...");
-            out.println("        A list of regular expressions matched against the arguments passed to the JVM, such as system properties.");
+            out.println("    --include-vmargs/--exclude-vmargs <pattern>...");
```
felixbarny, line 509: The UI and the server are using `vmarg` now. Maybe we should just change that in the attacher to match the docs, too. Or document `vmarg` and accept both `vmarg` and `vmargs`.

eyalkoren: I was not aware of that, I'll make sure our documentation is consistent with only `vmarg` and change the attacher to allow both `vmarg` and `vmargs` (and treat them the same) so to not make it a breaking change

stuartnelson3: The apm-server config has been modified to only accept `vmargs` right now, since that is correct for the version bundled with the apm-server for 8.0. Can you open an issue in the apm-server when this change happens, so we can update the code there?

felixbarny (approved):

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentAttacher.java
```diff
@@ -503,15 +507,15 @@ static void printHelp(PrintStream out) {
             out.println("        A list of PIDs to include.");
             out.println();
             out.println("    --include-main/--exclude-main <pattern>...");
```
SylvainJuge, line 509:
```suggestion
            out.println("    --include-main/--exclude-main <pattern>");
```

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentAttacher.java
```diff
@@ -503,15 +507,15 @@ static void printHelp(PrintStream out) {
...
-            out.println("        A list of regular expressions of fully qualified main class names or paths to JARs of applications the java agent should be attached to.");
+            out.println("        A regular expression of fully qualified main class names or paths to JARs of applications the java agent should be attached to.");
             out.println("        Performs a partial match so that `foo` matches `/bin/foo.jar`.");
             out.println();
             out.println("    --include-vmarg/--exclude-vmarg <pattern>...");
```
SylvainJuge, line 513:
```suggestion
            out.println("    --include-vmarg/--exclude-vmarg <pattern>");
```

apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentAttacher.java
```diff
@@ -503,15 +507,15 @@ static void printHelp(PrintStream out) {
...
-            out.println("        A list of regular expressions matched against the arguments passed to the JVM, such as system properties.");
+            out.println("        A regular expression that is matched against the arguments passed to the JVM, such as system properties.");
             out.println("        Performs a partial match so that `attach=true` matches the system property `-Dattach=true`.");
             out.println();
             out.println("    --include-user/--exclude-user <user>...");
```
SylvainJuge, line 517:
```suggestion
            out.println("    --include-user/--exclude-user <user>");
```

SylvainJuge (approved):

PR #2313, opened 2021-12-06
Fix APM Server version parsing

eyalkoren:
## What does this PR do?

Adjusting APM Server version parsing to match 7.16+
## Checklist

  - [x] Tested manually with
    - [x] 6.8.20
    - [x] 7.0.0
    - [x] 7.11.2
    - [x] 7.16.0
    - [x] 8.0.0-beta1

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/util/Version.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/util/Version.java
@@ -72,4 +72,14 @@ public class Version implements Comparable<Version> {
         }
         return 0;
     }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < numbers.length - 1; i++) {
+            sb.append(numbers[i]).append('.');
+        }
+        sb.append(numbers[numbers.length - 1]);
+        return sb.toString();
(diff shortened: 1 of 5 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/util/Version.java
```diff
@@ -72,4 +72,14 @@ public int compareTo(Version another) {
...
         return 0;
     }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < numbers.length - 1; i++) {
+            sb.append(numbers[i]).append('.');
+        }
+        sb.append(numbers[numbers.length - 1]);
```
jackshirazi, line 82: this breaks if numbers is an empty array

jackshirazi (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/util/Version.java
eyalkoren: Yeah, I didn't care because I only use it for the APM Server, but you're right, if it's here it needs to be safe always

jackshirazi (approved):

Issue #2092, opened 2021-08-29
Replace WeakConcurrentMaps with SpanConcurrentHashMaps


eyalkoren:
## What is this about?
Switching to `SpanConcurrentHashMap` wherever applicable has two major benefits:
1. reference counting is enforced and maintained centrally, ensuring spans are not recycled ahead of time and still being referenced, therefore potentially used
2. in combination with our `MockReporter` recycling assertions, this enables us locating places where spans/transactions are not actively removed from `WeakConcurrentMap`s, thus relying on GC to clear them away. By reducing those to a minimum, we can dramatically decrease the number of `WeakReference`s that are "active" at any given moment, thus considerably reduce GC-related overhead. 

## Plugin Checklist

- [x] gRPC - #2081
- [x] AsyncHttpClient - #2093 
- [x] Dubbo - #2093 
- [x] Generic concurrent plugin - #2173 
- [x] Lettuce - #2093 
- [x] External process tracing - #2093 
- [x] HttpUrlConnection - #2300
- [x] Vert.x 3 - #2322
- [x] Scala - #2323

SylvainJuge:
Added this to the next milestone so we can prioritize this migration.

PR #2322, opened 2021-12-08
Switch to using Span concurrent weak map in Vert.x 3

eyalkoren:
Implementing #2092 for Vert.x 3

Switching to the dedicated span concurrent map helped to discover a small transaction leak (not causing memory leak because GC still cleans) due to a change in internal API.

### DIFF
--- a/apm-agent-plugins/apm-vertx/apm-vertx3-plugin/src/main/java/co/elastic/apm/agent/vertx/v3/web/http1/HttpServerRequestImplEndInstrumentation.java
+++ b/apm-agent-plugins/apm-vertx/apm-vertx3-plugin/src/main/java/co/elastic/apm/agent/vertx/v3/web/http1/HttpServerRequestImplEndInstrumentation.java
@@ -42,7 +42,7 @@ public class HttpServerRequestImplEndInstrumentation extends WebInstrumentation 
 
     @Override
     public ElementMatcher<? super MethodDescription> getMethodMatcher() {
-        return named("doEnd").and(takesNoArguments());
+        return named("doEnd").or(named("onEnd")).and(takesNoArguments());
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


apm-agent-plugins/apm-vertx/apm-vertx3-plugin/src/main/java/co/elastic/apm/agent/vertx/v3/web/http1/HttpServerRequestImplEndInstrumentation.java
```diff
@@ -42,7 +42,7 @@
 
     @Override
     public ElementMatcher<? super MethodDescription> getMethodMatcher() {
-        return named("doEnd").and(takesNoArguments());
+        return named("doEnd").or(named("onEnd")).and(takesNoArguments());
```
AlexanderWert, line 45: did we miss this initially? `onEnd` is available in 3.9 but not in 3.6.
How did you notice this?
Maybe we need a unit test for making sure context gets cleaned up properly, WDYT?

AlexanderWert (approved):
LGTM.

Minor question about unit test for context clean-up

apm-agent-plugins/apm-vertx/apm-vertx3-plugin/src/main/java/co/elastic/apm/agent/vertx/v3/web/http1/HttpServerRequestImplEndInstrumentation.java
eyalkoren: This is exactly what this map migration does- it triggers our existing global recycling test assertions, thus uncovering such cases.
What we had before is a map that holds references to spans without counting these references, so the Vert.x API could change, causing spans to remain in the map and still get recycled. Now that we count these map references, if you don't properly remove them from the map, the recycling assertion causes the v3-latest tests to fail, which is how I found that.

PR #2323, opened 2021-12-08
Switching to use Span concurrent map in Scala plugin

eyalkoren:
Implements #2092 for Scala concurrency plugin


jackshirazi (approved):
nice simplification! Maybe worth raising an issue for adding span lifecycle tests in scala?

eyalkoren:
> Maybe worth raising an issue for adding span lifecycle tests in scala?

Good idea. We can first review and see what's missing (e.g. trying to propagate and activate a span that is already ended asynchronously). But assuming that this specific PR is not affecting that, I'll merge and keep that separately.

PR #2332, opened 2021-12-10
Disable log4j lookups

felixbarny:
## What does this PR do?
Disables log4j2 lookups

### DIFF
--- a/apm-agent-attach-cli/src/test/resources/log4j2.xml
+++ b/apm-agent-attach-cli/src/test/resources/log4j2.xml
@@ -2,7 +2,7 @@
 <Configuration status="WARN">
     <Appenders>
         <Console name="Console" target="SYSTEM_OUT">
-            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
+            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg{nolookups}%n"/>
         </Console>
     </Appenders>
     <Loggers>
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4j2ConfigurationFactory.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4j2ConfigurationFactory.java
@@ -174,7 +174,7 @@ public class Log4j2ConfigurationFactory extends ConfigurationFactory {
         if (logFormat == LogFormat.PLAIN_TEXT) {
             return builder
                     .newLayout("PatternLayout")
-                    .addAttribute("pattern", "%d [%thread] %-5level %logger{36} - %msg%n");
+                    .addAttribute("pattern", "%d [%thread] %-5level %logger{36} - %msg{nolookups}%n");
         } else {
             String serviceName = getValue(CoreConfiguration.SERVICE_NAME, sources, ServiceNameUtil.getDefaultServiceName());
             return builder.newLayout("EcsLayout")
--- a/apm-agent-core/src/test/resources/log4j2-test.xml
+++ b/apm-agent-core/src/test/resources/log4j2-test.xml
@@ -2,7 +2,7 @@
 <Configuration status="WARN">
     <Appenders>
         <Console name="Console" target="SYSTEM_OUT">
-            <PatternLayout pattern="%d [%thread] %-5level %logger{36} - %msg%n"/>
+            <PatternLayout pattern="%d [%thread] %-5level %logger{36} - %msg{nolookups}%n"/>
         </Console>
     </Appenders>
     <Loggers>
--- a/apm-agent/pom.xml
+++ b/apm-agent/pom.xml
@@ -421,12 +421,13 @@
                                 </filter>
                                 <filter>
                                     <artifact>org.apache.logging.log4j:log4j-core</artifact>
-                                    <!--
-                                    Eliminating exposure to the log4j2 vulnerability related to the SMTP appender -
-                                    https://nvd.nist.gov/vuln/detail/CVE-2020-9488#vulnCurrentDescriptionTitle
-                                    -->
                                     <excludes>
+                                        <!--
+                                        Eliminating exposure to the log4j2 vulnerability related to the SMTP appender -
+                                        https://nvd.nist.gov/vuln/detail/CVE-2020-9488#vulnCurrentDescriptionTitle
+                                        -->
                                         <exclude>org/apache/logging/log4j/core/appender/SmtpAppender.class</exclude>
+                                        <exclude>org/apache/logging/log4j/core/lookup/JndiLookup.class</exclude>
                                     </excludes>
                                 </filter>
                                 <filter>


felixbarny:
I tested manually that lookups don't work anymore. Log4j gracefully handles the absence of `JndiLookup`.

cconboy (approved):

felixbarny:
This will be fixed in versions 1.28.1+
To disable lookups in versions 1.17.0-1.28.0, set system property `-Dlog4j2.formatMsgNoLookups=true`. This also works in 1.17.0-1.26.0, even though log4j2 has been relocated in these versions. Versions prior to 1.17.0 did not use log4j2.

Based on some investigation, I could only trigger a lookup when `log_level=trace`. In other log levels, there was no sign of user-supplied information being logged.

felixbarny:
Update: 1.28.1 is available on [maven central](https://repo1.maven.org/maven2/co/elastic/apm/elastic-apm-agent/1.28.1/) now.

debu99:
that causes false alert triggered by several scanning system(trivy, aws inspector, tenable) ....

felixbarny:
I understand. But there's not much we can do about it. Good news is that it's a false positive alert.

debu99:
Is there any plan to use the new log4j version in next release?

jwedel:
@felixbarny 
> Based on some investigation, was only able to only trigger a lookup when log_level=trace. In other log levels, there was no sign of user-supplied information being logged.

On a scale from 1-10, how sure are you about this? :)

Joking aside, I could image that this is pretty hard to find as all log statements can potentially be exploited. That includes both direct variable or string interpolation, but (maybe) also exceptions messages / stack traces logged from java core classes.

felixbarny:
> Is there any plan to use the new log4j version in next release?

New log4j versions are not compatible with Java 7 but the agent provides Java 7 compatibility. So we'd either have to drop Java 7 support or use a completely different logging framework.

> On a scale from 1-10, how sure are you about this? :)
> 
> Joking aside, I could image that this is pretty hard to find as all log statements can potentially be exploited. That includes both direct variable or string interpolation, but (maybe) also exceptions messages / stack traces logged from java core classes.

The agent tries to avoid emitting log events that occur per-request as that would increase the impact it has on the application. I did not verify all possible code paths to all log statements, though. We can't guarantee that log levels other than `trace` aren't affected. At this point, we only know that log_level=trace is affected.

The only reliable way is to upgrade the agent to 1.28.1 or to set the system property as explained in https://github.com/elastic/apm-agent-java/pull/2332#issuecomment-991001330. Doing either of the two is strongly recommend.

debu99:
could we have a version without java7 support, otherwise we need to either disable security scanning or compiling ourself

albonney:
> could we have a version without java7 support, otherwise we need to either disable security scanning or compiling ourself

Seconding this.  Elastic libs have had a Low vulnerability that has shown up in security scanners (Twistlock, etc...) for over a year an a half.  Many orgs have a 180 day compliance policy to fix these issues.  Now there is a CRITICAL vulnerability that will show up and blow up many scan results.  Often orgs have 15-30 days to fix these critical/high vulnerabilities.

felixbarny:
> Elastic libs have had a Low vulnerability that has shown up in security scanners (Twistlock, etc...) for over a year an a half

Is that about CVE-2020-9488? This has been mitigated similar to CVE-2021-44228, by removing the vulnerable classes form the distribution:

https://github.com/elastic/apm-agent-java/blob/b85d718c36704bc5b4261658538c8d5482cf45d5/apm-agent/pom.xml#L425-L437

> there is a CRITICAL vulnerability that will show up and blow up many scan results

I want to re-emphasize that these vulnerabilities have been fixed but I agree that it's not ideal that they still show up as false-positives in vulnerability scanners. Is in an option to mark these as false-positive alerts on your scanners?

Thanks for all the feedback, much appreciated.

albonney:
> I want to re-emphasize that these vulnerabilities have been fixed but I agree that it's not ideal that they still show up as false-positives in vulnerability scanners. Is in an option to mark these as false-positive alerts on your scanners?

That is possible, but doing manual workarounds does not seem like a viable long-term solution, instead of just keeping up-to-date with the latest log4j vulnerabilities.  Speaking from a purely selfish standpoint, we operate in the US Federal space, and marking a Critical as as false-positive creates mountains of extra paperwork and additional risk-acceptance process.

felixbarny:
Thanks for the additional context that helps!

felixbarny:
The fine folks from log4j have back ported the recent security fixes to the [2.12](https://github.com/apache/logging-log4j2/tree/log4j-2.12) branch. Once 2.12.2 is released, we'll update to that version.

Huge shout out to the log4j team ❤️. If you can, show them your love and support and consider donating.

amarsharma222:
We have used the apm agent 1.28.1 for addressing the vulnerability. However , as part of the recent update in https://www.lunasec.io/docs/blog/log4j-zero-day-update-on-cve-2021-45046/ , it is recommended to use log4j version of 2.16. 

Please confirm if we are good with 1.28.1 ?

felixbarny:
Yes, we're good. See https://github.com/elastic/apm-agent-java/issues/2345#issuecomment-994460356

codefactor:
This issue gave a lot of context on the issue. In my org our security scanners have been constantly creating new bugs for me to resolve in 30 days all because of elastic-apm, log4j keeps coming up over and over again.

> > could we have a version without java7 support, otherwise we need to either disable security scanning or compiling ourself
> 
> Seconding this. Elastic libs have had a Low vulnerability that has shown up in security scanners (Twistlock, etc...) for over a year an a half. Many orgs have a 180 day compliance policy to fix these issues. Now there is a CRITICAL vulnerability that will show up and blow up many scan results. Often orgs have 15-30 days to fix these critical/high vulnerabilities.

+1

I don't think it's wise to continue a course where you KNOW that so many organizations are going to get false-positive results and come back to you (like I'm doing right now) by doing hacks on hacks just because for some reason Java 7 is important to support.

Here's a list of things of your options:

1. Get rid of log4j (preferred) replace it with slf4j/logback?
2. Stop supporting Java 7, for gods sake upgrade log4j
3. Do nothing, and continue to add work for everyone else to work with their security scanning vendors, and continue to have `elastic-apm-agent` to be on the minds of thousands of organizations who need to emergency patch their stuff and wish to get rid of log4j but can't because a transitive of a transitive of a transitive dependency chose to support Java 7

felixbarny:
Hi @codefactor,

I'm sorry for the frustration this has caused.

Which version of the Elastic APM Java Agent are you using? The latest version ships with log4j 2.12.4. As described in the [log4j security page](https://logging.apache.org/log4j/2.x/security.html), 2.12.4 contains backports for all the security issues.

> Each time the version changed, it started at log4j 2.11.1, then new bugs created for 2.11.2, 2.11.3, and now 2.11.4

Note that this was also true for the latest versions that support Java 8+ (2.16.0, 2.17.0, 2.17.1).

For the latest version of the Elastic APM Java Agent that contains log4j 2.12.4, do you still get vulnerability warnings from your security scanners? If so, which scanner are you using?

codefactor:
@felixbarny ,
Here is an example of the `./gradlew dependencies` output
```
co.elastic.apm:apm-agent-api:1.28.4
```
As I mentioned this is a transitive dependency so I don't even directly use it, and since log4j is embedded somewhere in my binary I have no way to check what version of log4j is built in there, I'm not even sure precisely which JAR log4j class files are located.

This version 1.28.4 seems to be the highest available version [here](https://mvnrepository.com/artifact/co.elastic.apm/apm-agent) but yet the security scanner thinks that I have log4j 2.11.4?

What JAR should I look for inside my spring boot JAR which might have log4j in it, and where exactly in that JAR would I be able to see the version of the log4j library?

felixbarny:
> `co.elastic.apm:apm-agent-api:1.28.4`

The API of the agent doesn't have any dependencies. Only the agent itself (`elastic-apm-agent.jar`) uses log4j. But it never used 2.11.x. So log4j must be sneaking in via another dependency.

> What JAR should I look for inside my spring boot JAR which might have log4j in it

Does `jour-app.jar!/BOOT-INF/lib` contain a `log4j-core*.jar`? This is typically where Spring Boot adds the dependencies.

Other than that, transitive dependencies should also show up in gradle's dependency report. See also https://docs.gradle.org/current/userguide/viewing_debugging_dependencies.html. If all of that doesn't help, you could try to extract all jars inside `BOOT-INF/lib` and check if any jar contains `org/apache/logging/log4j/core/Logger.class`

felixbarny:
I just checked - there is no version 2.11.4 of log4j. The 2.11.x line only goes to 2.11.2. See https://search.maven.org/artifact/org.apache.logging.log4j/log4j-core and https://repo1.maven.org/maven2/org/apache/logging/log4j/log4j-core.

codefactor:
So sorry, I misread the tickets somehow before -- I said 2.11.X before, but I really meant was 2.12.1 through 2.12.4 - and so now the most recently opened ticket (which the scanner is still complaining about) is talking about 2.12.4 having a critical severity issue CVE-2020-9488.

My project has 3 different elastic dependencies (all 3 levels deep in the dep tree) `co.elastic.apm:apm-opentracing`, `co.elastic.apm:apm-agent-api`, `co.elastic.apm:apm-agent-attach`. They are all at the same version `1.28.4`

I tried searching for `Logger.class` in any of my distribution JAR files inside my Spring Boot distribution JAR, I ran a command to first unzip my distribution JAR, then a command like this: `for i in `ls verp/BOOT-INF/lib` ; do echo $i && jar -tvf my-application/BOOT-INF/lib/$i | grep log4j ; done` and there was nothing much there, a few related files but then I noticed that inside `apm-agent-attach` there was another jar called `elastic-apm-agent` tricky tricky... this one does contain a bunch of `log4j2` things inside, but who knows what are `.esclazz` extensions... Anyway, I can feel pretty confident here that the scanner is giving me a false positive about log4j because of this `elastic-agent-attach.jar`

felixbarny:
That's correct, `apm-agent-attach-1.28.4.jar` contains `elastic-apm-agent.jar`, which contains Log4j 2.12.4. CVE-2020-9488 has been fixed in 2.12.3, which is documented in https://logging.apache.org/log4j/2.x/security.html. Could you tell me which vulnerability scanner you are using? If you don't want to disclose that publicly, could you please send an email to security@elastic.co?

codefactor:
@felixbarny ,
Does the following scanner sound familiar: `protecode-dockerimage` ?
Sorry I'm not a security engineer, I think this is the name of the scanner that looks at the binaries.

felixbarny:
Sorry for the delay in response. I've been in touch with our infosec team. They're working on getting the CVE updated so that it doesn't falsely flag Log4j 2.12.4 as vulnerable against CVE-2020-9488.

cc @douglasday

felixbarny:
We got in touch with the security team from the Apache foundation. The CVE has been updated so that it is correctly flagged as fixed in 2.12.3.
See also
* https://nvd.nist.gov/vuln/detail/CVE-2020-9488
* https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-9488

@codefactor are you still getting warnings from protecode? If so, maybe an update of the policies is required.

felixbarny:
Quoting from the CVEs
[CVE-2020-9488](https://nvd.nist.gov/vuln/detail/CVE-2020-9488)
> Fixed in Apache Log4j 2.12.3 and 2.13.1

[CVE-2021-44228](https://nvd.nist.gov/vuln/detail/CVE-2021-44228)
> Apache Log4j2 2.0-beta9 through 2.15.0 (excluding security releases 2.12.2, 2.12.3, and 2.3.1)

The Java agent is using log4j 2.12.4, which is greater than the mentioned fixed version 2.12.3.
Therefore, this is a false positive. 

@douglasday is there a way for us to reach out to Protecode to make them aware of the false-positive? I suppose they either are not aware of the amendment of the CVE that flags 2.12.3 as fixed or the scanner is confused as it thinks only 2.12.3 is exempted and doesn't realize 2.12.4 also includes the fix.

felixbarny:
I've talked to our application security team about this. They said that they're seeing lots of false-positives from a variety of different tools. While they're happy to help correct CVEs, they're a small team and don't have the capacity to notify every security scanner vendor if they produce false-positives.

So your options are either to close the ticket as a proven false positive or to contact your security scan vendor so that they can update their policies to avoid these false positives.

Thank you for your patience and sorry that we can't help you with contacting your vendors.
Also, thanks to our application security team for their help and for getting the CVE updated!

PR #2343, opened 2021-12-14
Fixing log4j configuration initialization

eyalkoren:
## What does this PR do?
Fixes an issue introduced in 1.28.0 - reported at https://discuss.elastic.co/t/agent-1-28-1-error-no-log4j-2-configuration-file-found/291668.
NOT RELATED TO [CVE-2021-44228](https://nvd.nist.gov/vuln/detail/CVE-2021-44228)

## Checklist

- [x] This is a bugfix
  - [x] I have reproduced and verified the fix manually


apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
```diff
@@ -298,34 +297,7 @@ public void assertValid(Boolean value) {
...
-        // The initialization of log4j may produce errors if the traced application uses log4j settings (for
-        // example - through file in the classpath or System properties) that configures specific properties for
-        // loading classes by name. Since we shade our usage of log4j, such non-shaded classes may not (and should not)
-        // be found on the classpath.
-        // All handled Exceptions should not prevent us from using log4j further, as the system falls back to a default
-        // which we expect anyway. We take a calculated risk of ignoring such errors only through initialization time,
-        // assuming that errors that will make the logging system non-usable won't be handled.
-        String initialListenersLevel = System.setProperty(INITIAL_LISTENERS_LEVEL, "OFF");
-        String initialStatusLoggerLevel = System.setProperty(INITIAL_STATUS_LOGGER_LEVEL, "OFF");
-        String defaultListenerLevel = System.setProperty(DEFAULT_LISTENER_LEVEL, "OFF");
```
felixbarny, line 310: I think we can also remove the corresponding constants

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
```diff
@@ -298,34 +297,7 @@ public void assertValid(Boolean value) {
...
-        }
-    }
-
-    private static void restoreSystemProperty(String key, @Nullable String originalValue) {
-        if (originalValue != null) {
-            System.setProperty(key, originalValue);
-        } else {
-            System.clearProperty(key);
-        }
+        ConfigurationFactory.setConfigurationFactory(new Log4j2ConfigurationFactory(sources, ephemeralId));
```
felixbarny, line 300: What's the difference between `ConfigurationFactory.setConfigurationFactory` and `Configurator.initialize`? What triggered the error and what exactly did it resolve?

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
```diff
@@ -298,34 +297,7 @@ public void assertValid(Boolean value) {
         .buildWithDefault(LogFormat.PLAIN_TEXT);
 
     public static void init(List<ConfigurationSource> sources, String ephemeralId) {
-        // The initialization of log4j may produce errors if the traced application uses log4j settings (for
-        // example - through file in the classpath or System properties) that configures specific properties for
-        // loading classes by name. Since we shade our usage of log4j, such non-shaded classes may not (and should not)
```
felixbarny, line 303: We're not shading anymore. Does that lead to problems when users configure log4j with system properties as it also affects the agent's logging setup?

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
eyalkoren: `ConfigurationFactory` is a singleton that allows overriding its instance through `ConfigurationFactory.setConfigurationFactory` (not considered thread safe, but since we do it so early on when there is only the main thread it should be OK). Once you did that, any logger created thereafter will be configured through it.
`Configurator.initialize` only initializes a `LoggerContext`, so that all loggers within this context will implicitly get the same configuration. However, once you try to create a logger in a different context, it will attempt to get a configuration and will look it up from what it can find (e.g. properties/xml file).
The default context locator is based on class loader and now (> 1.26.0), when we have many class loaders that load classes that request loggers, only loggers within the same context as the initialized one (which is the agent class loader) will be properly configured and all the other contexts (eg plugin class loaders) won't.
There are multiple ways to handle that, like making sure we initialize all contexts or use a different context selector (will probably still require us to make sure we initialize multiple contexts, for example - threads), but I think the proper way is to utilize the fact that we have isolated logging system and so we can just make sure everything within this system is configured the same.

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
eyalkoren: With the approach I used here, we simply override any other way to configure loggers in our isolated system. The only way to configure is through our custom `ConfigurationFactory`, so AFAIU, all other options are ignored.
It doesn't mean that there aren't system properties that may affect us, but not ones that I know of related to the general configuration.

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
```diff
@@ -298,34 +294,11 @@ public void assertValid(Boolean value) {
...
-    private static void restoreSystemProperty(String key, @Nullable String originalValue) {
-        if (originalValue != null) {
-            System.setProperty(key, originalValue);
-        } else {
-            System.clearProperty(key);
-        }
+        // org.apache.logging.log4j.core.config.ConfigurationFactory is a singleton that allows overriding its instance
+        // through setConfigurationFactory. This API is not considered thread safe, but since we do it so early on when
+        // there is only the main thread it should be OK. Once we override the default factory instance, any logger
+        // created thereafter will be configured through our custom factory.
```
felixbarny, line 300:
```suggestion
        // created thereafter will be configured through our custom factory.
        // This avoids that loggers in plugin class loaders get their own context based on the thread context class loader,
        // which would initialize the loggers based on the properties/xml file that the target application ships with
```

felixbarny (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
felixbarny: Thanks, much clearer now!

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
eyalkoren: I'll add something in this sprit, thanks.
The default context selector is using the caller class loader, not the thread context CL and the problem is either when there is existing log4j config in the classpath or there isn't (then we get an error of non-existing config)

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
eyalkoren: Having said that, maybe I should restore this false errors that do not really affect us in case some configs come from system properties, WDYT?

CHANGELOG.asciidoc
```diff
@@ -46,6 +46,7 @@ endif::[]
 [float]
 ===== Bug fixes
 * Preferring controller names for Spring MVC transactions, `use_path_as_transaction_name` only as a fallback - {pull}2320[#2320]
+* Fixing log4j configuration error - {pull}2343[#2343]
```
felixbarny, line 49: This needs to be in the 1.28.2 release notes

felixbarny (changes requested):

CHANGELOG.asciidoc
eyalkoren: Ohh, of course 🤦 
Thanks

felixbarny (approved):

PR #2360, opened 2021-12-21
Upgrading Byte Buddy version to 1.12.6

eyalkoren:


v1v:
Thanks Eyal, I understand now the whole problem! Then I guess the <= should be the way to go as you said

v1v (approved):

tobiasstadler:
Maybe the problem is that ignore does not check against the version in the pom but the version in "Maven Central"? So an update from < 1.9.16 to version 1.9.16 would be ignored, but not an update from 1.9.16 to something > 1.9.16?

tobiasstadler:
@eyalkoren yes, I also read that. But for me it is not really clear what the ignored version refers to. https://github.community/t/configure-dependabot-to-only-create-prs-for-minor-or-patch-updates/117656 suggests it is the new version not the version in the pom.

tobiasstadler:
Maybe 
```
 - package-ecosystem: "maven"
    directory: "/"
   ....
-  package-ecosystem: "maven"
    directory: "/integration-tests/runtime-attach/runtime-attach-app/"
    ...
    ignore:
      - dependency-name: "net.bytebuddy:byte-buddy-agent"
```
does work?

PR #2362, opened 2021-12-21
Fix keystore for APM-server communication

SylvainJuge:
## What does this PR do?

For the communication with APM server, we use a separate SSL context as the default that is used for the whole JVM.
That allows to implement TLS fallback and disable server certificate validation through configuration when needed.

One side effect of this was that the standard `keyStore` configuration parameters that are provided through JVM parameters would not be taken in account by this separate SSL context.

In short, it now means that the following parameters are now also effective for the communication with APM server, hence allowing to have client-side certificate validation.

- `javax.net.ssl.keyStore`
- `javax.net.ssl.keyStorePassword`
- `javax.net.ssl.keyStoreType`
- `javax.net.ssl.keyStoreProvider`

## Checklist

- [x] This is a bugfix
  - [x] Manual testing with certificates provided by https://badssl.com/

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ssl/SslUtils.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ssl/SslUtils.java
@@ -122,12 +122,12 @@ public class SslUtils {
             sslContext = SSLContext.getInstance("TLS");
         }
 
-        KeyManager[] keyManagers;
+        KeyManager[] keyManagers = null;
         try {
             keyManagers = getKeyManagers();
         } catch (IOException | GeneralSecurityException e) {
-            logger.error("unable to setup key managers", e);
-            throw e;
+            logger.warn("unable to setup key managers, keystore options will be ignored");
+            logger.debug("key manager creation error stack trace: ", e);
         }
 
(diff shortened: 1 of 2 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/report/ssl/SslUtils.java
```diff
@@ -108,19 +113,28 @@ public static SSLSocketFactory getSSLSocketFactory(boolean validateCertificates)
...
             sslContext = SSLContext.getInstance("TLS");
         }
-        sslContext.init(null, trustAllCerts, new java.security.SecureRandom());
+
+        KeyManager[] keyManagers;
+        try {
+            keyManagers = getKeyManagers();
+        } catch (IOException | GeneralSecurityException e) {
+            logger.error("unable to setup key managers", e);
+            throw e;
```
jackshirazi, line 130: do we want to error here, or just revert to previous behaviour with null keymanagers since most of our customers don't need this?

SylvainJuge: That's a very good point, here issuing a warning and then silently use the previous behavior seems appropriate.

jackshirazi (approved):
nice work!

PR #2365, opened 2021-12-22
Upgrading ECS logging to 1.3.0

eyalkoren:


eyalkoren:
Merging with Payara ITs off, testing on master

PR #2368, opened 2021-12-22
Adding instrumentation warmup

eyalkoren:
## What does this PR do?

Tries to invoke as much as possible instrumentation-related code paths, including the instrumentation logic and the `invokedynamic` runtime linkage path, on the agent-attaching thread before registering the`ClassFileTransformer` to the JVM-global `java.lang.instrument.Instrumentation`.
This ensures that most Byte Buddy and agent types related to these processes are loaded and initialized on a single thread, thus reducing the chances for deadlock.
We will start with having warmup on by default, but with a hidden config flag that enables disabling it.
## Checklist

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/InstallationListenerImpl.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/InstallationListenerImpl.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package co.elastic.apm.agent.bci.bytebuddy;
+
+import co.elastic.apm.agent.bci.ElasticApmAgent;
+import co.elastic.apm.agent.bci.IndyBootstrap;
+import net.bytebuddy.agent.builder.AgentBuilder;
+import net.bytebuddy.agent.builder.ResettableClassFileTransformer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.invoke.ConstantCallSite;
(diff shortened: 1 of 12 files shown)


felixbarny (approved):

eyalkoren:
I upgraded to Byte Buddy 1.12.7 which allows to try out the instrumented bytecode created during warmup

apm-agent-core/src/main/java/co/elastic/apm/agent/bbwarmup/WarmupInstrumentation.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return not(isStatic()).and(named("isInstrumented")).and(takesNoArguments()).and(returns(boolean.class));
+    }
+
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Collections.emptyList();
+    }
+
```
felixbarny, line 54:
```suggestion

    @Override
    public boolean includeWhenInstrumentationIsDisabled() {
        return true;
    }

```

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/InstallationListenerImpl.java
```diff
@@ -0,0 +1,79 @@
...
+                }
+
+                String instrumentedClassName = Instrumented.class.getName();
+                if (transformedInstrumentedClass != null) {
+                    logger.debug("Bytecode transformation {} succeeded", instrumentedClassName);
+
+                    // warm up the complete invokedynamic linkage route
+                    HashMap<String, byte[]> typeDefinitions = new HashMap<>();
+                    typeDefinitions.put(instrumentedClassName, transformedInstrumentedClass);
+                    ByteArrayClassLoader.ChildFirst childFirstCL = new ByteArrayClassLoader.ChildFirst(null, typeDefinitions);
```
felixbarny, line 57: Why can't we just call `Instrumented.isInstrumented()`? Why do we need to create a dedicated class loader?

apm-agent-core/src/main/java/co/elastic/apm/agent/bbwarmup/WarmupInstrumentation.java
eyalkoren: Good point, thanks!

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/InstallationListenerImpl.java
eyalkoren: Warmup doesn't retransform the class, it only applies the bytecode transformation. 
In order to retransform classes, it would have to register the BB class file transformer, thus kicking the instrumentation race when invoking `retransformClasses`, which misses the point of avoiding deadlocks.
This allows me to warmup the entire `invokedynamic` linkage path on the attaching thread before installing the BB agent.

felixbarny: Ah, understood. A small comment about that in the code would be nice.

PR #2376, opened 2021-12-29
Fix slf4j LinkageErrors for external plugins

### LINKED ISSUE #2350: slf4j MDC with java.lang.LinkageError， within the  transaction created by external plugin and log correlation enabled

eyalkoren:
## What does this PR do?

Fixes #2350 

This PR touches several places related to external plugins:
- Class loading (@felixbarny please review this part)
     - The CL hierarchy for external plugin is currently not what we intended for the API classes. If we load the API classes by the IndyPluginCL, then the API plugin classes are loaded by another IndyPluginCL, one of which parents is the agent CL and the other is the external plugin IndyPluginCL. So I made sure those are loaded by the `ExternalPluginClassLoader`. 
     - If slf4j classes are packaged into the plugin, I made sure to exclude them as well from the external plugin IndyPluginCL, because it is child-first, which means it will cause clashes with the slf4j that is available in the agent CL. It means that the `ExternalPluginClassLoader` will still have visibility to slf4j, but those cannot effectively be loaded, so the plugin slf4j usages will use the agent slf4j.
     - Our API plugin advice uses the instrumented class's (the API class) CL as the "initiating CL" (a term with a different meaning in the Java spec BTW 🙂 ). When the API is added to an application as a dependency, this makes perfect sense. But when it is loaded by an external plugin CL, this doesn't make sense. I changed it currently so that `TraceContext` will just ignore attempts to set any type of agent CL as the initiating CL. This is enough to cause the usage of the context CL as a fallback, which will probably be sufficient in many cases. #2369 contains a discussion with multiple proposals of how to improve that. Let's discuss this offline.
- External plugin test apps (@jackshirazi please review this part)
     - Added ability to create an external-plugin-based transaction (was only span creation before)
     - Added log correlation by default
     - Added slf4j to the external plugin. This with the two above reproduced #2350 
     - Added slf4j to the external plugin target class (just to make sure we don't break stuff this way)
     - Added Jakarta version for the external plugin test app
     - Fixed and added tests for automatic outcome setting for API spans/transactions when capturing Exceptions

## Checklist

- [x] This is a bugfix


eyalkoren:
@jackshirazi just proposed that we get rid of both log4j and slf4j by taking only the bare minimum from whichever logging framework we choose and repackage it. It sounds like it could save us ton of trouble

felixbarny:
> @jackshirazi just proposed that we get rid of both log4j and slf4j by taking only the bare minimum from whichever logging framework we choose and repackage it. It sounds like it could save us ton of trouble

This may make it harder to use ecs-logging-java. If feasible, I'd like to keep us using the standard slf4j and log4j libraries. However, we can definitely think about stripping unused classes if possible. For example, `org.apache.logging.log4j.core.appender.<db|mom|nosql>`.

> If slf4j classes are packaged into the plugin

We may also just fail to load plugins that include classes that shouldn't be in there (slf4j, Byte Buddy, the Plugin SDK) by throwing an exception. This seems a bit safer as no loading them as there may be binary incompatibilities. But probably not a big issue.

> Our API plugin advice uses the instrumented class's (the API class) CL as the "initiating CL" 

Could we detect if the initiating class loader is the indy plugin CL and programmatically access the target CL?

eyalkoren:
> We may also just fail to load plugins that include classes that shouldn't be in there (slf4j, Byte Buddy, the Plugin SDK) by throwing an exception. This seems a bit safer as no loading them as there may be binary incompatibilities. But probably not a big issue.

What if we just remove slf4j and use our own facade over log4j? If we add that through the SDK, external plugins can log into agent logs as well

> Could we detect if the initiating class loader is the indy plugin CL and programmatically access the target CL?

Yes, I have this working. The problem is that for external plugin, the target CL is not necessarily any better, it could be just the instrumented library CL that probably doesn't have access to the logging library.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
```diff
@@ -362,10 +363,22 @@ public static ConstantCallSite bootstrap(MethodHandles.Lookup lookup,
...
             if (instrumentationClassLoader instanceof ExternalPluginClassLoader) {
-                pluginClasses.addAll(((ExternalPluginClassLoader) instrumentationClassLoader).getClassNames());
+                List<String> externalPluginClasses = ((ExternalPluginClassLoader) instrumentationClassLoader).getClassNames();
+                for (String externalPluginClass : externalPluginClasses) {
+                    if (// API classes have no dependencies and don't need to be loaded by an IndyPluginCL
+                        !(externalPluginClass.startsWith("co.elastic.apm.api")) &&
+                        !(externalPluginClass.startsWith("co.elastic.apm.opentracing")) &&
+                        // slf4j classes, if used by the plugin, should be loaded by the agent Class Loader and not the
+                        // IndyPluginCL, which is child-first, as this would cause LinkageErrors due to clash with the agent's slf4j
+                        !(externalPluginClass.startsWith("org.slf4j"))
```
eyalkoren, line 377: We don't need that if we merge #2390

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
```diff
@@ -712,7 +713,7 @@ public int hashCode() {
     }
 
     void setApplicationClassLoader(@Nullable ClassLoader classLoader) {
-        if (classLoader != null) {
+        if (classLoader != null && !ClassLoaderUtils.isAgentClassLoader(classLoader)) {
```
jackshirazi, line 716: `classLoader != null` test unnecessary now

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -390,14 +390,17 @@ public boolean isExit() {
         return isExit;
     }
 
-
-    public void captureException(long epochMicros, Throwable t) {
-        tracer.captureAndReportException(epochMicros, t, this);
+    @Nullable
+    public String captureException(long epochMicros, @Nullable Throwable t) {
```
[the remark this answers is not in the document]
felixbarny: It should either return `T` or be named `captureExceptionAndGetErrorId`.

integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/AbstractTomcatIT.java
```diff
@@ -33,7 +33,7 @@ public AbstractTomcatIT(final String tomcatVersion) {
 
     @Override
     protected void enableDebugging(GenericContainer<?> servletContainer) {
-        servletContainer.withEnv("CATALINA_OPTS", "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005");
+        servletContainer.withEnv("CATALINA_OPTS", "-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005");
```
jackshirazi, line 36: is `suspend=y` right?

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
```diff
@@ -362,10 +363,22 @@ public static ConstantCallSite bootstrap(MethodHandles.Lookup lookup,
             MethodHandle instrumentedMethod = args.length >= 5 ? (MethodHandle) args[4] : null;
 
             ClassLoader instrumentationClassLoader = ElasticApmAgent.getInstrumentationClassLoader(adviceClassName);
+            ClassLoader targetClassLoader = lookup.lookupClass().getClassLoader();
             ClassFileLocator classFileLocator;
             List<String> pluginClasses = new ArrayList<>();
             if (instrumentationClassLoader instanceof ExternalPluginClassLoader) {
-                pluginClasses.addAll(((ExternalPluginClassLoader) instrumentationClassLoader).getClassNames());
+                List<String> externalPluginClasses = ((ExternalPluginClassLoader) instrumentationClassLoader).getClassNames();
```
felixbarny, line 370: Possibly outside the scope of this PR but I think we should add more validation rules in the constructor of `ExternalPluginClassLoader` to make sure external plugins don't contain slf4j and Byte Buddy, similar to how we validate that the Plugin SDK isn't packaged in the plugin.

https://github.com/elastic/apm-agent-java/blob/e756114a054faf650e92e62b1793b0a1aa996e5a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/classloading/ExternalPluginClassLoader.java#L46-L48

While looking at it, the `%s` placeholder is unresolved.

felixbarny (approved):
We should probably merge #2390 first.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
eyalkoren: It _is_ necessary - if `classLoader == null`, then `!ClassLoaderUtils.isAgentClassLoader(classLoader)` would evaluate to `true` and enter the `if` block, which is not what we want.

integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/AbstractTomcatIT.java
eyalkoren: Since debugging is disabled by default (through `co.elastic.apm.servlet.AbstractServletContainerIntegrationTest#ENABLE_DEBUGGING`), this is not super crucial. I think that I use `suspend=y` more often than not, but let's keep `n` as the default.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
felixbarny: [minor] I think the expression would be easier to grasp if inverted and returning

```java
if (classLoader == null || ClassLoaderUtils.isAgentClassLoader(classLoader)) {
    return;
}
``` 

Or even


```java
if (ClassLoaderUtils.isBootstrapClassLoader(classLoader) || ClassLoaderUtils.isAgentClassLoader(classLoader)) {
    return;
}
```

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
eyalkoren: Yes, let's handle separately (except from the Exception message String format).
With #2390 would be allowed, but all dependencies we package within the agent (not only Byte Buddy) may indeed cause `LinkageError` if used within the external plugin AND available to the CL that loads the instrumented classes.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
eyalkoren: OK

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
eyalkoren: I was primarily thinking about `IndyPluginClassLoader` created for external plugins' instrumentation, which may cause this specific kind of `LinkageError` - "loader constraint violation" (because of the child-first nature of this CL). 
However, since the agent CL is a direct parent of the `ExternalPluginClassLoader`, packaging classes that are available within the agent may either pass quietly, or cause other types of `LinkageError` (e.g. when using mismatching versions).
So I take this back - it is within the context of this PR to disallow external plugins to package libraries that are packaged in the agent.

eyalkoren:
I added a verification mechanism to ensure that external plugins that contain agent dependencies are rejected with a proper log message.
I also added a test to ensure that the dependencies list is kept up to date

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/classloading/ExternalPluginClassLoader.java
```diff
@@ -55,7 +58,27 @@ public ExternalPluginClassLoader(File pluginJar, ClassLoader agentClassLoader) t
             while (entries.hasMoreElements()) {
                 JarEntry jarEntry = entries.nextElement();
                 if (jarEntry.getName().endsWith(".class")) {
-                    tempClassNames.add(jarEntry.getName().replace('/', '.').substring(0, jarEntry.getName().length() - 6));
+                    String fqcn = jarEntry.getName().replace('/', '.').substring(0, jarEntry.getName().length() - 6);
+                    if (fqcn.startsWith("org.slf4j") || fqcn.startsWith("org.apache.logging")) {
```
felixbarny, line 62: These are also in the list of agent dependencies so this case is not strictly needed. But I suppose you still opted for it because of the better error message, right?

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/classloading/ExternalPluginClassLoader.java
```diff
@@ -55,7 +58,27 @@ public ExternalPluginClassLoader(File pluginJar, ClassLoader agentClassLoader) t
...
+                        throw new IllegalStateException(String.format("Package \"%s\" is used within plugin %s. This is not allowed " +
+                                "because it is already used by the agent. For logging purposes, use the agent SDK logging facade - %s",
+                            fqcn.substring(0, fqcn.lastIndexOf('.')),
+                            pluginJar.getName(),
+                            LoggerFactory.class.getName()
+                        ));
+                    }
+                    for (String agentDependencyPackage : AgentInfo.getAgentDependencyPackages()) {
```
felixbarny, line 70: Could we make the `ExternalPluginClassLoader` child-first and allow plugins to have a dependency on `HdrHistrogram`, for example?

felixbarny (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/classloading/ExternalPluginClassLoader.java
eyalkoren: Right

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/classloading/ExternalPluginClassLoader.java
eyalkoren: It will create further complications...
We create an `IndyPluginClassLoader` for each external plugin, which can load most of the plugin classes. If you leave it as is (i.e. include these libraries within the IndyPlugin CL classpath), then you may get `LinkageError` if the `ExternalPluginClassLoader ` also loads these classes (e.g. because the instrumentation class uses it), because it is in the plugin CL delegation hierarchy. If you exclude these classes from the ones loaded by the Indy Plugin CL, then by default it would delegate loading requests of them to the agent CL (because this is how the `DiscriminatingMultiParentClassLoader` works), which may lead to other `LinkageError`

PR #2384, opened 2022-01-05
Fix dynamic log_level configuration

### LINKED ISSUE #2366: Log level configuration is not fully dynamic
## Describe the bug
Logging level configuration is not fully dynamic after #2343 .
The [code](https://github.com/elastic/apm-agent-java/blob/master/apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java#L342-L352) that sets the logging level is still applying the config for the agent CL context only, instead of applying to all agent CLs.
## Steps to reproduce
1. Start the agent with some logging level
2. Invoke some requests to load some instrumentation plugins
3. Change to a different level and wait until applied
4. Logging level of plugins does not change
## Expected behavior

Changing `log_level` in properties file should affect all agent loggers.

eyalkoren:
## What does this PR do?

Fixes #2366 
## Checklist

- [x] This is a bugfix
  - [x] I manually tested that plugin loggers' level changes when configuration changes

### DIFF
--- a/apm-agent-core/src/test/java/co/elastic/apm/agent/logging/LoggingConfigurationTest.java
+++ b/apm-agent-core/src/test/java/co/elastic/apm/agent/logging/LoggingConfigurationTest.java
@@ -1,0 +1,206 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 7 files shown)


apm-agent-core/src/test/java/co/elastic/apm/agent/logging/LoggingConfigurationTest.java
```diff
@@ -0,0 +1,206 @@
...
+            return null;
+        }
+    }
+
+    private static final class TestLog4jContextFactory extends Log4jContextFactory {
+
+        ContextSelector contextSelector = new TestContextSelector();
+
+        @Nullable
+        LoggerContext getContext(Logger slf4jLogger) {
```
jackshirazi, line 123: why is this argument named `slf4jLogger` ?

eyalkoren: Because it's an `org.slf4j.Logger`, as opposed to the `org.apache.logging.log4j.core.Logger` used two lines below. It's a utility for the test that only knows slf4j loggers.

jackshirazi (approved):
nice work

PR #2389, opened 2022-01-11
[jjbb] filter only for the default branch, PRs and tags

v1v:
## What does this PR do?

Filter the main pipeline with:
- the default branch -> master
- PRs
- tags with the format `vd.d.d.*` -> therefore v1.0.0 and v1.0.1-RC1 are supported 

Therefore `1.x` and `1.26` won't be in the list of branches to be built in the CI

## Test

*The below screenshots are obsoleted since the tags are not anymore matching the regex*

image

image

image


kuisathaverat (approved):

SylvainJuge:
Thanks for this.

Is there any value to keep building existing tags ?
- they are (or should be) immutable, thus there is no need to rebuild them once the release is done
- when doing the release, it means an extra execution will be triggered when the tag is created, which just duplicates the work done by the release job that is creating the release artifact.

Of course, it would be different if our "main build pipeline" would automatically create release artifacts when executed on a tag, but it's not the case currently.

v1v:
> Is there any value to keep building existing tags ?

No strong opinion, this PR was intended to fulfil the initial requirement. So I'd say the one regarding whether to listen from tags in the main pipeline versus those tags were already verified in the release pipeline could be done in a follow up. 

What do you think?

.ci/jobs/apm-agent-java-mbp.yml
```diff
@@ -13,7 +13,7 @@
             discover-pr-forks-trust: permission
             discover-pr-origin: merge-current
             discover-tags: true
-            head-filter-regex: '^(?!stable).*$'
+            head-filter-regex: '(master|PR-.*|v\d+\.\d+\.\d+.*)'
```
SylvainJuge, line 16:
```suggestion
            head-filter-regex: '(master|PR-.*)'
```

SylvainJuge (changes requested):
We really don't have any use for re-building tags here, and the change seems quite minimal, thus let's remove them.

SylvainJuge (approved):

PR #2393, opened 2022-01-17
Split integration tests

felixbarny:


SylvainJuge:
As per comments in the shared doc on test naming, I would suggest instead to use the "test scope" to remove ambiguities:
- "Agent Integration Tests", for what we call "integration tests" in the agent, we could then later split that into "Plugins Integration Tests" and "Core Integration Tests" if we need to.
- "APM Integration Tests" could also be used as an alternative to "End-to-End Integration Tests", but I'm fine to keep this one as you suggested.

For simplicity, I think it could be relevant to have a 1:1 mapping between maven profiles and the pipeline stages, so each stage maps to a maven profile, and "test partitioning" is written once for all at the maven profiles level instead of enabling/disabling profiles in a shell script.

SylvainJuge:
One advantage of having maven profiles for the pipeline stages is that it would avoid repeating the `maven.javadoc.skip` to skip javadoc generation at maven invocation time.

Also prefixing those profiles with `ci_` could help to understand that they are mostly used for the build and aren't usually meant to be used when building the project locally.

SylvainJuge (approved):

PR #2400, opened 2022-01-18
Implement optional build steps

SylvainJuge:
## What does this PR do?

- make end-to-end GH check name consistent
- keep benchmarks only on master for now
- disable agent integration tests by default
- selectively enable parts of the pipeline using PR labels:
  - `ci:agent-integration` : for agent integration tests (both App server and non-App-server)
  - `ci:end-to-end` : for APM end-to-end tests, even if they are run asynchronously they aren't required to execute until the PR is somehow ready to be merged
  - `ci:jdk-compatibility` for the JDK compatibility tests

## Checklist

- [x] check that changing PR draft status toggles optional parts of the pipeline.
- [x] check that adding/removing labels on this PR allows to skip parts of the pipeline
- [ ] follow-up tasks:
  - [ ] modify branch protection rules to add agent integration tests & end-to-end tests as required
  - [x] ~~optional: automate adding `ci:agent-integration` and `ci:end-to-end` label when the PR is marked as ready for internal PRs, and when there is at least one approval for community PRs (not sure if it's the right trigger)~~
  - [x] trigger a build when the PR goes from draft to ready in order to properly cover the steps that might have been disabled when the PR was a draft. handled through #2404

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,15 +30,35 @@ pipeline {
     quietPeriod(10)
   }
   triggers {
-    issueCommentTrigger("(${obltGitHubComments()}|^run (compatibility|benchmark|integration) tests)")
+    issueCommentTrigger("(${obltGitHubComments()}|^run (jdk compatibility|benchmark|integration|end-to-end) tests)")
   }
   parameters {
     string(name: 'MAVEN_CONFIG', defaultValue: '-V -B -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn -Dhttps.protocols=TLSv1.2 -Dmaven.wagon.http.retryHandler.count=3 -Dmaven.wagon.httpconnectionManager.ttlSeconds=25', description: 'Additional maven options.')
+
+    // Note about GH checks and optional steps
+    //
+    // All the steps executed by default are included in GH checks
+    // All the mandatory steps not included by default need to be added in GH branch protection rules.
+
+    // enabled by default, required for merge through default GH check
     booleanParam(name: 'test_ci', defaultValue: true, description: 'Enable Unit tests')
-    booleanParam(name: 'agent_integration_tests_ci', defaultValue: true, description: 'Enable Agent Integration tests')
-    booleanParam(name: 'endtoend_tests_ci', defaultValue: true, description: 'Enable APM End-to-End Integration tests')
-    booleanParam(name: 'bench_ci', defaultValue: true, description: 'Enable benchmarks')
-    booleanParam(name: 'compatibility_ci', defaultValue: false, description: 'Enable JDK compatibility tests')
+
+    // disabled by default, but required for merge, there are two GH checks:
+    // - Non-Application Server integration tests
+    // - Application Server integration tests
+    // opt-in with 'ci:agent-integration'
+    booleanParam(name: 'agent_integration_tests_ci', defaultValue: false, description: 'Enable Agent Integration tests')
+
+    // disabled by default, but required for merge, GH check name is ${GITHUB_CHECK_ITS_NAME}
+    // opt-in with 'ci:end-to-end' tag on PR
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -30,15 +30,35 @@ pipeline {
...
     booleanParam(name: 'test_ci', defaultValue: true, description: 'Enable Unit tests')
-    booleanParam(name: 'agent_integration_tests_ci', defaultValue: true, description: 'Enable Agent Integration tests')
-    booleanParam(name: 'endtoend_tests_ci', defaultValue: true, description: 'Enable APM End-to-End Integration tests')
-    booleanParam(name: 'bench_ci', defaultValue: true, description: 'Enable benchmarks')
-    booleanParam(name: 'compatibility_ci', defaultValue: false, description: 'Enable JDK compatibility tests')
+
+    // disabled by default, but required for merge, there are two GH checks:
+    // - Non-Application Server integration tests
+    // - Application Server integration tests
+    // opt-in with 'ci:agent-integration'
```
felixbarny, line 49: Didn't we agree to only disable this for draft PRs?
But if we do so, does marking a PR as ready for review re-trigger a build?

SylvainJuge: > Didn't we agree to only disable this for draft PRs?
Yes, but currently we can't filter steps execution on draft/ready status of the PR, thus only on labels for now.

> But if we do so, does marking a PR as ready for review re-trigger a build?
Currently it does not, but we could probably implement this through a GH automation. As written in the checklist above maybe having a distinction between internal/community PRs might be required.

felixbarny: Not sure if it is enabled but there's https://github.com/jenkinsci/pipeline-github-plugin

Looks like we could do something like

```
when {
    env.CHANGE_ID && pullRequest.draft
}
```

Having required checks that aren't executed by default seems to slow down the feedback process.
And I'm not sure if we gain a lot other than saving some compute time.

v1v:
```
when {
    env.CHANGE_ID && pullRequest.draft
}
```

I'm not sure if that's possible, declarative pipeline with `name.method` does not work unless the `scripts` closure is added, at least that's how it works in the `steps` section.

Jenkinsfile
```diff
@@ -30,15 +30,35 @@ pipeline {
...
+    booleanParam(name: 'agent_integration_tests_ci', defaultValue: false, description: 'Enable Agent Integration tests')
+
+    // disabled by default, but required for merge, GH check name is ${GITHUB_CHECK_ITS_NAME}
+    // opt-in with 'ci:end-to-end' tag on PR
+    booleanParam(name: 'endtoend_tests_ci', defaultValue: false, description: 'Enable APM End-to-End tests')
```
v1v, line 54: To be aligned with the label and stage name:

```suggestion
    booleanParam(name: 'end_to_end_tests_ci', defaultValue: false, description: 'Enable APM End-to-End tests')
```

Jenkinsfile
SylvainJuge: Well, it seems that using this in an expression block seems to work!

- Here is a build where integration tests are skipped because the PR was in draft: [PR 2400 # build 5](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/PR-2400/5/pipeline)
- Here is one that I just launched after marking the PR as ready: [PR 2400 # build 6](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/PR-2400/6/pipeline)

If the latter includes both agent integration tests (even if any of them fail), it means filtering on PR draft state works as expected.

SylvainJuge: Update: it definitely works as expected :-).

SylvainJuge:
The PR is fine to merge (but build fails due to an unrelated unit test)

@felixbarny and @v1v , can we proceed and merge this ? (PR approval is a good answer).

v1v (approved):

felixbarny (approved):

PR #2414, opened 2022-01-24
jjbb: enable tags for the opbeans automation

v1v:
## What does this PR do?

https://github.com/elastic/apm-agent-java/pull/2389 removed the tags, but they are required to let the automation for the opbeans to work similar what it has been done in the past and in the other APM agents.

See https://github.com/elastic/apm-agent-java/blob/9448cc7185747c4e2b25895936a162ae4beaa87e/Jenkinsfile#L412-L438

### DIFF
--- a/.ci/jobs/apm-agent-java-mbp.yml
+++ b/.ci/jobs/apm-agent-java-mbp.yml
@@ -13,7 +13,7 @@
             discover-pr-forks-trust: permission
             discover-pr-origin: merge-current
             discover-tags: true
-            head-filter-regex: '(master|PR-.*)'
+            head-filter-regex: '(master|PR-.*|v\d+\.\d+\.\d+.*)'
             notification-context: 'apm-ci'
             repo: apm-agent-java
             repo-owner: elastic


kuisathaverat (approved):

SylvainJuge:
I think a better option here would be to move this step in the release pipeline.
- it's skipped 99% of the builds in the main pipeline
- it's only relevant once after a release
- it's quite simple, thus unless the related credentials are not available in the internal CI moving it should be quite simple.

v1v:
The release pipeline runs in a different CI controller, so I'd rather prefer to use the `apm-ci` since it's what we have done with the rest of the APM agents, so the more homogeneous the easy to debug/troubleshoot.

If the major concern is regarding the previous stages that should not be triggered, then we can add a when condition in each stage, to warranty only the after-release should happen

What do you think?

SylvainJuge:
I like the idea to have everything in the same pipeline, what is just a bit bothering me is that we have two parts that seem mutually exclusive.

Adding a boolean condition to ignore most steps in the pipeline when on a tag, just to execute a single one feels like we should split them to two different pipelines and have two separate jobs with different triggers, but that might come with other complications as well.

What I think could be  good compromise though, is to have two main stages in the pipeline that are mutually exclusive, in a similar was as described in https://issues.jenkins.io/browse/JENKINS-41187
- one that is executed on tags, and just updates obpeans app if required
- one that is executed on regular branches

The difference here would be that instead of complicating again our `when` conditions for each step of the pipeline, we would rather nest them through nested stages, the condition being defined and maintained only in the top-level stage.

v1v:
+1, I like the idea.

I'd say to be done in a different PR to keep the aim of this PR as it is, put back the process we had in place, then the other PR will be the one that implements the changes regarding the meta-stages to improve the CI workflows

SylvainJuge (approved):

PR #2416, opened 2022-01-25
ci: three meta-stages (checkout, build, release)

v1v:
## What does this PR do?

Refactor the main pipeline in three meta-stages:

1) `Checkout` (it runs by default)
2) `Build` (it runs on branches and PRs) It's composed by all the stages but the opbeans or stable ones
3) `Release` (it runs for the `main` branch or tags). It's composed by the `opbeans` and `stable` stages.

I also removed the duplicated entries in the environment sections, when running for Windows then we might need to revisit this!

## Why

See https://github.com/elastic/apm-agent-java/pull/2414#issuecomment-1020952186

## Test

Default behaviour with a draft PR

image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -61,8 +61,35 @@ pipeline {
     booleanParam(name: 'jdk_compatibility_ci', defaultValue: false, description: 'Enable JDK compatibility tests')
   }
   stages {
-    stage('Initializing'){
+    stage('Checkout') {
       options { skipDefaultCheckout() }
+      steps {
+        pipelineManager([ cancelPreviousRunningBuilds: [ when: 'PR' ] ])
+        deleteDir()
+        gitCheckout(basedir: "${BASE_DIR}", githubNotifyFirstTimeContributor: true, shallow: false,
+                    reference: '/var/lib/jenkins/.git-references/apm-agent-java.git')
+        stash allowEmpty: true, name: 'source', useDefaultExcludes: false
+        script {
+          dir("${BASE_DIR}"){
+            // Skip all the stages except docs for PR's with asciidoc and md changes only
+            env.ONLY_DOCS = isGitRegionMatch(patterns: [ '.*\\.(asciidoc|md)' ], shouldMatchAll: true)
+            // Prepare the env variables for the benchmark results
+            env.COMMIT_ISO_8601 = sh(script: 'git log -1 -s --format=%cI', returnStdout: true).trim()
+            env.NOW_ISO_8601 = sh(script: 'date -u "+%Y-%m-%dT%H%M%SZ"', returnStdout: true).trim()
+            env.RESULT_FILE = "apm-agent-benchmark-results-${env.COMMIT_ISO_8601}.json"
+            env.BULK_UPLOAD_FILE = "apm-agent-bulk-${env.NOW_ISO_8601}.json"
+          }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -61,38 +61,42 @@ pipeline {
     booleanParam(name: 'jdk_compatibility_ci', defaultValue: false, description: 'Enable JDK compatibility tests')
   }
   stages {
-    stage('Initializing'){
```
v1v, line 64: Initialising is not anymore the meta-stage for the checkout and build

Jenkinsfile
```diff
@@ -61,38 +61,42 @@ pipeline {
...
+    stage('Checkout') {
       options { skipDefaultCheckout() }
+      steps {
+        pipelineManager([ cancelPreviousRunningBuilds: [ when: 'PR' ] ])
+        deleteDir()
+        gitCheckout(basedir: "${BASE_DIR}", githubNotifyFirstTimeContributor: true, shallow: false,
```
v1v, line 69: Meta-Stage for the checkout only, it runs always

Jenkinsfile
```diff
@@ -61,38 +61,42 @@ pipeline {
...
+            // Prepare the env variables for the benchmark results
+            env.COMMIT_ISO_8601 = sh(script: 'git log -1 -s --format=%cI', returnStdout: true).trim()
+            env.NOW_ISO_8601 = sh(script: 'date -u "+%Y-%m-%dT%H%M%SZ"', returnStdout: true).trim()
+            env.RESULT_FILE = "apm-agent-benchmark-results-${env.COMMIT_ISO_8601}.json"
+            env.BULK_UPLOAD_FILE = "apm-agent-bulk-${env.NOW_ISO_8601}.json"
+          }
+        }
+      }
+    }
+    stage('Builds'){
```
v1v, line 85: Meta-stage for building/testing on PR basis and merge to branches

Jenkinsfile
```diff
@@ -127,310 +131,288 @@ pipeline {
...
-                reportTestResults()
               }
             }
           }
         }
       }
     }
-    stage('Stable') {
-      options { skipDefaultCheckout() }
+    stage('Releases'){
```
v1v, line 365: Meta-stage for the release specific process, it runs on merge to master or tags. Each child-stage defines the when

SylvainJuge:
One potential downside is that Jenkins does not seem to properly display the steps as nested: https://issues.jenkins.io/browse/JENKINS-60651

So here the meta-stages and the nested stages are displayed sequentially.
I wonder if having the meta-stages implemented as a `parallel` block, even if they only have one item would help to somehow make it display correctly.

v1v:
Meta stages are presented as green, but I don't really see the issue but a minor cosmetic UX. Bear in mind, the Blue Ocean is not maintained (https://groups.google.com/g/jenkinsci-users/c/xngZrSsXIjc/m/d606K7lHBgAJ)

SylvainJuge:
I agree, let's not complicate the pipeline for the sake of making it look nice in Jenkins UI.

mdelapenya (approved):

SylvainJuge (approved):
LGTM for now, we might improve naming of meta steps to make them clearer or even try hacking UI with fake `parallel` stages later.

PR #2422, opened 2022-01-26
ci: enable optional stages in the branch

v1v:
## What does this PR do?

Enable optional stages on a PR basis to be mandatory when merging into `main`:

- `'Non-Application Server integration tests'`
- `'Application Server integration tests'`
- `'End-To-End Integration Tests'`

Simplify the when for:
- `'Benchmarks'`
- `Javadoc`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -177,6 +177,7 @@ pipeline {
                   expression { return env.GITHUB_COMMENT?.contains('integration tests') }
                   expression { matchesPrLabel(label: 'ci:agent-integration') }
                   expression { return env.CHANGE_ID != null && !pullRequest.draft }
+                  not { changeRequest() }
                 }
               }
               steps {
@@ -206,6 +207,7 @@ pipeline {
                   expression { return env.GITHUB_COMMENT?.contains('integration tests') }
                   expression { matchesPrLabel(label: 'ci:agent-integration') }
                   expression { return env.CHANGE_ID != null && !pullRequest.draft }
+                  not { changeRequest() }
                 }
               }
               steps {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -177,6 +177,7 @@ pipeline {
                   expression { return env.GITHUB_COMMENT?.contains('integration tests') }
                   expression { matchesPrLabel(label: 'ci:agent-integration') }
                   expression { return env.CHANGE_ID != null && !pullRequest.draft }
+                  not { changeRequest() }
```
v1v, line 180: image

Jenkinsfile
```diff
@@ -237,11 +239,9 @@ pipeline {
               }
               when {
                 beforeAgent true
-                allOf {
```
v1v, line 240: Otherwise a github comment will never force the `benchmark` to get triggered since `params.bench_ci`= false by default

Jenkinsfile
```diff
@@ -271,10 +271,6 @@ pipeline {
             stage('Javadoc') {
               agent { label 'linux && immutable' }
               options { skipDefaultCheckout() }
-              when {
```
v1v, line 274: This is now delegated to the `meta-stage`

Jenkinsfile
felixbarny, line 180: Wouldn't that be equivalent to what you proposed?
```suggestion
                  expression { return env.CHANGE_ID == null || !pullRequest.draft }
```

v1v: 'Non-Application Server integration tests', 'Application Server integration tests', 'End-To-End Integration Tests' and `Bencharmks` are skipped in any branch (so far only `main`)

The idea with this PR is to keep the existing behaviour for PRs but enable those stages when merging into `main`.

Since I was not sure about the branch strategy, I played safe with `not { changeRequest() }` therefore there is no need to change the CI pipeline if a new branch is created for whatever reasons.

I'd prefer to use `not { changeRequest() }` , even though it's a negative condition

v1v: I think so, although it's not purely declarative based, but simplifies two conditions in one. No strong opinion about this

SylvainJuge: I'm fine with keeping it as-is, even if it's negative (covid learned us that negative things aren't that bad after all :smile: ).

SylvainJuge: I would rather keep those into two distinct conditions for clarity (if you don't mind).

SylvainJuge (approved):

Issue #2346, opened 2021-12-15
Fix Lambda Instrumentation for non existent status code


AlexanderWert:
With API Gateway trigger v1 on the lambda instrumentation the following happens, if the instrumented code does not explicitly set the status code on the response object:

- The `statusCode` variable is an `Integer` object on the [APIGatewayProxyResponseEvent](https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayProxyResponseEvent.java)(used with API Gateway v1 triggers)
- this `statusCode` is not initialized properly with the constructor of APIGatewayProxyResponseEvent
- thus, if the instrumented code does not set the `statusCode` explicitly, then the value is `null`
- the lambda instrumentation then fails with a NPE at this [stage](https://github.com/elastic/apm-agent-java/blob/c30b5a0f380e3d9854863ff5dfb1d3a46abb3cbd/apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/helper/APIGatewayProxyV1TransactionHelper.java#L89)

Issue #2394, opened 2022-01-17
[META 584] Implement changes on AWS lambda context capturing


AlexanderWert:
[![issue details](https://img.shields.io/endpoint?label=meta-issue:&url=https%3A%2F%2Fgiss.app.elstc.co%2Fapi%2Fstatus%2Felastic%2Fapm%2F584)](https://github.com/elastic/apm/issues/584)

Spec PR: https://github.com/elastic/apm/pull/579

## Changes
- **Change `transaction.name` for API Gateway-triggered Lambda functions**: To be consistent with other HTTP-based instrumentations include the low-cardinality resourcePath into the name. 
- **Respect setting for `use_path_as_transaction_name`**: If `use_path_as_transaction_name` is enabled use the *full* HTTP path for the `transaction.name` in API Gateway-triggered Lambda functions.
- **Respect `service_name` and `service_version` agent config options**: If `service_name` and/or `service_version` are explicitly configured use the configured values instead of the derived ones.
- **Use the full domain name of the API Gateway as `service.origin.name`**
- **Replaced `service.id` metadata field with new transaction-level field `faas.id`**: To being able to capture different Lambda function ARN values (semantically correct) under one APM service umbrella (for the scenario of consolidating multiple lambda functions in one APM service).

Issue #2305, opened 2021-12-01
[META 569] Add query param flushed=true to final intake request at the end of lambda function


AlexanderWert:
[![issue details](https://img.shields.io/endpoint?label=meta-issue:&url=https%3A%2F%2Fgiss.app.elstc.co%2Fapi%2Fstatus%2Felastic%2Fapm%2F569)](https://github.com/elastic/apm/issues/569)

[![issue details](https://img.shields.io/endpoint?label=spec-issue:&url=https%3A%2F%2Fgiss.app.elstc.co%2Fapi%2Fstatus%2Felastic%2Fapm%2F570)](https://github.com/elastic/apm/issues/570)

The AWS lambda extension flushes the APM data it has collected to the APM server and considers the function invocation complete when it receives an agent intake request with the query param `flushed=true`. This will most likely optimize the overall execution duration because the extension won't have to wait for the Logs API signal that the function completed.

Ensure that the final intake request to the extension from the agent contains the query param `flushed=true`.

eyalkoren:
The Java agent opens a connection and streams events to it, so a query param is not a good option for us.
I think that a typical Lambda function execution will create a single HTTP request. Either way, only when it ends will the agent know whether it's the final or not.

eyalkoren:
As discussed offline with @AlexanderWert, we will just do an additional empty request after sending the transaction event to the same endpoint with the required query param. Since it is localhost, overhead is not a concern.
@estolfo Can you please verify that the extension can receive empty requests?

AlexanderWert:
> @estolfo Can you please verify that the extension can receive empty requests?

This test covers this: https://github.com/elastic/apm-aws-lambda/pull/97

eyalkoren:
Implemented in #2424

PR #2424, opened 2022-01-27
Implemented AWS lambda spec changes

AlexanderWert:
- [x] Fixes https://github.com/elastic/apm-agent-java/issues/2346
- [x] Implements https://github.com/elastic/apm-agent-java/issues/2394
   - **Change `transaction.name` for API Gateway-triggered Lambda functions**: To be consistent with other HTTP-based instrumentations include the low-cardinality resourcePath into the name. 
   - **Respect setting for `use_path_as_transaction_name`**: If `use_path_as_transaction_name` is enabled use the *full* HTTP path for the `transaction.name` in API Gateway-triggered Lambda functions.
   - **Respect `service_name` and `service_version` agent config options**: If `service_name` and/or `service_version` are explicitly configured use the configured values instead of the derived ones.
   - **Use the full domain name of the API Gateway as `service.origin.name`**
   - **Replaced `service.id` metadata field with new transaction-level field `faas.id`**: To being able to capture different Lambda function ARN values (semantically correct) under one APM service umbrella (for the scenario of consolidating multiple lambda functions in one APM service).
- [x] Send empty request with `flushed=true` query param on Lambda data flush (https://github.com/elastic/apm-agent-java/issues/2305)
- [x] Update docs:
   - [x] Remove the `experimental` flag from the Lambda feature
   - [x] Add something about potential long start and possible configurations - `enable_instrumentations`, increasing timeout, increasing heap size


apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/helper/APIGatewayProxyV1TransactionHelper.java
```diff
@@ -48,20 +50,33 @@ public static APIGatewayProxyV1TransactionHelper getInstance() {
...
+
+        if (null != transaction) {
+            fillHttpRequestData(transaction, getHttpMethod(apiGatewayEvent), apiGatewayEvent.getHeaders(), host,
+                apiGatewayEvent.getRequestContext().getPath(), getQueryString(apiGatewayEvent), apiGatewayEvent.getBody());
+        }
+
+        return transaction;
+    }
+
+    @Nonnull
```
eyalkoren, line 63:
```suggestion
```

apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/helper/APIGatewayProxyV1TransactionHelper.java
```diff
@@ -86,27 +101,57 @@ private String getQueryString(APIGatewayProxyRequestEvent apiGatewayEvent) {
...
+        if (null != transactionName && requiredDataForTransactionNameAvailable(event)) {
+            transactionName.append(getHttpMethod(event)).append(" ");
+            if (webConfiguration.isUsePathAsName()) {
+                transactionName.append(event.getRequestContext().getPath());
+            } else {
+                setResourcePathBasedName(event, transactionName);
+            }
+        } else {
+            super.setTransactionName(transaction, event, lambdaContext);
+        }
```
eyalkoren, line 139: Some clarification questions:
1. if `webConfiguration.isUsePathAsName()` - shouldn't it be used anyway?
2. wouldn't a best-effort transaction naming approach be better than the all-or-nothing approach through `requiredDataForTransactionNameAvailable`?

Also, I see that you extracted this from `AbstractAPIGatewayTransactionHelper` into the concrete helpers, but there seem to still be a lot of shared logic.

eyalkoren (commented):
Looks good!
Missing:
- Serialization of `faas` fields (`id`, `name`, `version`)
- Tests for `faas.name` and `faas.version` in the Lambda tests

Let's sort out our transaction naming approach offline and I will make adjustments if required and implement #2305

apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/helper/APIGatewayProxyV1TransactionHelper.java
```diff
@@ -86,27 +101,57 @@ private String getQueryString(APIGatewayProxyRequestEvent apiGatewayEvent) {
...
     }
 
     @Override
-    protected String getApiGatewayVersion() {
-        return "1.0";
+    protected void setTransactionName(Transaction transaction, APIGatewayProxyRequestEvent event, Context lambdaContext) {
+        StringBuilder transactionName = transaction.getAndOverrideName(AbstractSpan.PRIO_HIGH_LEVEL_FRAMEWORK);
+        if (null != transactionName && requiredDataForTransactionNameAvailable(event)) {
+            transactionName.append(getHttpMethod(event)).append(" ");
+            if (webConfiguration.isUsePathAsName()) {
```
eyalkoren, line 132: We want to look at this one first and always use the path if configured so

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/metadata/MetaDataTest.java
```diff
@@ -81,13 +81,9 @@ void testCloudProvider_NONE_and_configured_hostname() throws InterruptedExceptio
 
     @Test
     void testCloudProvider_ForAWSLambda_fromEnvVariables() throws Exception {
-        when(coreConfiguration.getServiceName()).thenReturn("");
         MetaData awsLambdaMetaData = createAwsLambdaMetaData();
 
         Service service = awsLambdaMetaData.getService();
-        assertThat(service.getName()).isEqualTo("function-name");
-        assertThat(service.getId()).isEqualTo("service-id");
-        assertThat(service.getVersion()).isEqualTo("function-version");
```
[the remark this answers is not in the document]
eyalkoren: Tested through [`ServiceInfoTest`](https://github.com/elastic/apm-agent-java/pull/2424/files#diff-8135da994b60d778923776dee7fc16a2d246fe4a34e680dfde5dcf2de646310dR79-R96), I'll just simplify this one a bit

apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/RequestHandlerInstrumentation.java
```diff
@@ -68,10 +68,11 @@ public String getAdviceClassName() {
         @Nullable
         @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
         public static Object handlerEnter(@Advice.Argument(value = 0) Object input, @Advice.Argument(value = 1) Context lambdaContext) {
-            if (input instanceof APIGatewayV2HTTPEvent) {
+            if (input instanceof APIGatewayV2HTTPEvent && ((APIGatewayV2HTTPEvent) input).getRequestContext() != null
+                && ((APIGatewayV2HTTPEvent) input).getRequestContext().getHttp() != null) {
```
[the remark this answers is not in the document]
eyalkoren: Not using a specific helper would still create a transaction and fallback to the default implementation that uses the function name

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
```diff
@@ -80,11 +81,15 @@ public boolean isProcessed(long sequence) {
...
         if (event.getType() == null) {
             return;
         }
         switch (event.getType()) {
+            case MAKE_FLUSH_REQUEST:
+                endRequest();
+                connection = startRequest(INTAKE_V2_FLUSH_URL);
+                // continuing to behave as END_REQUEST
             case END_REQUEST:
                 endRequest();
```
felixbarny, line 94:
```suggestion
            case MAKE_FLUSH_REQUEST:
                endRequest();
                connection = startRequest(INTAKE_V2_FLUSH_URL);
                endRequest();
```

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerReporter.java
```diff
@@ -166,7 +173,10 @@ public void scheduleWakeupEvent() {
     }
 
     @Override
-    public boolean flush(long timeout, TimeUnit unit) {
+    public boolean flush(long timeout, TimeUnit unit, boolean followupWithFlushRequest) {
+        if (followupWithFlushRequest) {
+            return publishAndWaitForEvent(timeout, unit, MAKE_FLUSH_REQUEST_EVENT_TRANSLATOR);
+        }
         return publishAndWaitForEvent(timeout, unit, END_REQUEST_EVENT_TRANSLATOR);
```
felixbarny, line 180: There are a few potential issues when you have two events instead of one:
- The total blocked time may be higher: you may have to wait for the events to be flushed that were enqueued while waiting for the flush request event.
- In situations where the queue is almost full, one of the two events may make it into the queue.
- A related issue is increased wait time due to more re-tries when the queue is full. Essentially, the effective timeout is twice the requested timeout.

Therefore, I think it makes sense to just have one flushing event. We could even keep it simple and always both end the request and execute a flush request. As the only user for the flush event is the Lambda instrumentation, I don't think we need the `followupWithFlushRequest` argument. All production usages would provide `true`.

If a use case comes up where we do differ between end request and executing a flush request, we could add a boolean flag to the `ReportingEvent`. But for now, I'd keep it simple and always end the request, start a flush request and end the flush request.

eyalkoren: Not sure I understand what you mean, it's a single event - either end or end and flush.
The comment about continuing to the `END_REQUEST` case is only regarding the `switch`-`case` execution flow

eyalkoren: That's how I guaranteed atomicity

felixbarny: But maybe we just need the flush request event and not the end request event. But I leave that up to you.

eyalkoren: Let's leave it as is for now

apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/helper/AbstractLambdaTransactionHelper.java
```diff
@@ -80,6 +88,18 @@ public Transaction startTransaction(I input, Context lambdaContext) {
         return null;
     }
 
+    private String getFaasId(Context lambdaContext) {
+        if (functionArn == null) {
+            functionArn = lambdaContext.getInvokedFunctionArn();
+            String[] arnSegments = functionArn.split(":");
+            if (arnSegments.length > 6) {
```
AlexanderWert, line 95:
```suggestion
            if (arnSegments.length > 7) {
```

This is a bug, needs to be 7, as the standard Lambda function ARN without version segment contains 7 segments.

AlexanderWert (commented):
I reviewed the changes.
Except a small bug, that I found in my own code, everything looks good.

Tested the interplay with the extension: The extension receives the `flush=true` signal as expected. So all good on that side as well.

apm-agent-core/src/test/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandlerTest.java
```diff
@@ -147,7 +147,7 @@ void testUrls() throws MalformedURLException {
     }
 
     @Test
-    void testReport() {
+    void testReport() throws Exception {
```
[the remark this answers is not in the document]
eyalkoren: These tests use `reportTransaction`, which if you go deep enough you'll see using `IntakeV2ReportingEventHandler#dispatchEvent()` that now started using `IntakeV2ReportingEventHandler#startRequest()`, which `throws Exception`. I could catch it somewhere along the way, but I prefer it would be handled by the disruptor, like any other event handling error would.
You can see that `com.lmax.disruptor.EventHandler#onEvent`, which invokes `dispatchEvent()`, also `throws Exception`, so it is expected by the disruptor API.

apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/helper/AbstractLambdaTransactionHelper.java
eyalkoren: I wonder how stable this is. Is there a documented specification regarding the ARN structure?

apm-agent-core/src/main/java/co/elastic/apm/agent/report/IntakeV2ReportingEventHandler.java
```diff
@@ -56,7 +57,7 @@ public void init(ApmServerReporter reporter) {
     }
 
     @Override
-    public void onEvent(ReportingEvent event, long sequence, boolean endOfBatch) {
+    public void onEvent(ReportingEvent event, long sequence, boolean endOfBatch) throws Exception {
```
felixbarny, line 60: What happens when we propagate the exception to the disruptor? Which exception handler do we have configured? Is there any chance of this preventing this event handler to process further events?

eyalkoren: It doesn't seem we configured one. We should, otherwise the default handler throws an exception as well.
However, there is nothing special here, any NPE or other `RuntimeException` caused by one of the many things we try to do within this loop could fail and cause event processing halt.

felixbarny: The usage of JUL is fine as we replace that with our JulBridgeLogger.

This is a snippet of the decompiled class that we ship:

```java
import co.elastic.apm.agent.logging.JulBridgeLogger;
import java.util.logging.Level;

public final class IgnoreExceptionHandler implements ExceptionHandler<Object> {
    private static final JulBridgeLogger LOGGER = JulBridgeLogger.getLogger(IgnoreExceptionHandler.class.getName());
    private final JulBridgeLogger logger;
```

eyalkoren: Right! I completely forgot about `JulBridgeLogger` even though I just recently refactored it 🤦 
We're all good then.

PR #2428, opened 2022-01-27
Reimplementing log correlation

### LINKED ISSUE #2387: Fix JUL ECS logging no longer converting plain text logs to JSON
## Describe the bug
Setting the config in tomcat for java.util.logging to use a org.apache.juli.AsyncFileHandler as a handler and in turn setting the formatter for that to co.elastic.logging.jul.EcsFormatter should result in JSON format logs and did in 1.18.1 but stopped working in 1.19.0. From 1.19.0+ the log lines no longer convert to JSON format (and so remain in the plain text format).

## Steps to reproduce
1. setup tomcat to be able to run
2. add or modify a setenv.sh in the tomcat bin directory to add a ecs-logging-core and jul-ecs-formatter to the classpath, eg
`CLASSPATH=/SOMEPATH0/jul-ecs-formatter-1.3.2.jar:/SOMEPATH0/ecs-logging-core-1.3.2.jar`
and add the agent startup flag, together with specifying the apm config file, eg
`export CATALINA_OPTS="-javaagent:/SOMEPATH0/elastic-apm-agent-1.19.0.jar -Delastic.apm.config_file=/SOMEPATH1/FNAME1.properties $CATALINA_OPTS"`
Note the config file can be empty and doesn't need to specify pointing at a valid apm server, though setting the agent log file is a good idea with the `log_file` property, eg
`log_file=/SOMEPATH2/FNAME2.log`
Note you need the 3 jars listed to be present and in the path you specify, SOMEPATH0 is just an example
3. In the tomcat logging properties file (usually `logging.properties` in the tomcat `conf` directory set the handler to AsyncFileHandler and set it's formatter to `co.elastic.logging.jul.EcsFormatter`, eg like the example below
4. Start and stop tomcat, and check the format of the entries in the log files specified in the `logging.properties` file (no need to execute any requests, the startup logging is sufficient). The log entries should be plain text
5. Change the agent to the prior version 1.18.1 and repeat, the log entries should be in JSON format

Example logging properties entries:
```
handlers = 1catalina.org.apache.juli.AsyncFileHandler, 2localhost.org.apache.juli.AsyncFileHandler, 3manager.org.apache.juli.AsyncFileHandler

.handlers = 1catalina.org.apache.juli.AsyncFileHandler

############################################################
# Handler specific properties.
# Describes specific configuration info for Handlers.
############################################################

1catalina.org.apache.juli.AsyncFileHandler.level = FINE
1catalina.org.apache.juli.AsyncFileHandler.directory = /var/log/tomcat
1catalina.org.apache.juli.AsyncFileHandler.maxDays = 15
1catalina.org.apache.juli.AsyncFileHandler.prefix = catalina.
1catalina.org.apache.juli.AsyncFileHandler.suffix = .json
1catalina.org.apache.juli.AsyncFileHandler.encoding = UTF-8
1catalina.org.apache.juli.AsyncFileHandler.formatter = co.elastic.logging.jul.EcsFormatter

2localhost.org.apache.juli.AsyncFileHandler.level = FINE
2localhost.org.apache.juli.AsyncFileHandler.directory = /var/log/tomcat
2localhost.org.apache.juli.AsyncFileHandler.maxDays = 15
2localhost.org.apache.juli.AsyncFileHandler.prefix = localhost.
2localhost.org.apache.juli.AsyncFileHandler.suffix = .json
2localhost.org.apache.juli.AsyncFileHandler.encoding = UTF-8
2localhost.org.apache.juli.AsyncFileHandler.formatter = co.elastic.logging.jul.EcsFormatter

3manager.org.apache.juli.AsyncFileHandler.level = FINE
3manager.org.apache.juli.AsyncFileHandler.directory = /var/log/tomcat
3manager.org.apache.juli.AsyncFileHandler.maxDays = 15
3manager.org.apache.juli.AsyncFileHandler.prefix = manager.
3manager.org.apache.juli.AsyncFileHandler.suffix = .json
3manager.org.apache.juli.AsyncFileHandler.encoding = UTF-8
3manager.org.apache.juli.AsyncFileHandler.formatter = co.elastic.logging.jul.EcsFormatter

#java.util.logging.ConsoleHandler.level = FINE
#java.util.logging.ConsoleHandler.formatter = co.elastic.logging.jul.EcsFormatter
co.elastic.logging.jul.EcsFormatter.serviceName=tomcat-default
#java.util.logging.ConsoleHandler.encoding = UTF-8


############################################################
# Facility specific properties.
# Provides extra control for each logger.
############################################################

org.apache.catalina.core.ContainerBase.[Catalina].[localhost].level = INFO
org.apache.catalina.core.ContainerBase.[Catalina].[localhost].handlers = 2localhost.org.apache.juli.AsyncFileHandler

org.apache.catalina.core.ContainerBase.[Catalina].[localhost].[/manager].level = INFO
org.apache.catalina.core.ContainerBase.[Catalina].[localhost].[/manager].handlers = 3manager.org.apache.juli.AsyncFileHandler
```

## Expected behavior
The logs specified in the `logging.properties` should hold JSON format log lines
eg 
`{"@timestamp":"2022-01-06T21:42:34.247Z", "log.level": "INFO", "message":"ContextListener: contextInitialized()", "ecs.version": "1.2.0","service.name":"tomcat-default","event.dataset":"tomcat-default","process.thread.id":1,"log.logger":"org.apache.catalina.core.ContainerBase.[Catalina].[localhost].[/examples]"}`
and not plain text, eg 
`06-Jan-2022 20:07:59.725 INFO [main] org.apache.catalina.core.ApplicationContext.log ContextListener: contextInitialized()`

### LINKED ISSUE #2412: Find more generic and robust log-correlation mechanism
The current log correlation mechanism relies on updates to the MDC mechanism (or `ThreadContext` in log4j2) that are invoked through span activation/deactivation listeners. The problem is that these event listeners don't necessarily have access to the logging library through the available class loaders. We have a mechanism that overcomes this in some cases by setting a class loader that loads the instrumented code, but there is no guarantee that the instrumented code has access to these classes as well. 
Whenever we don't have a valid class loader to access the MDC/ThreadContext, log correlation would not work.

We want to come up with a more robust mechanism. 
Ideally, the log event context will be enriched with the active transaction/trace/error IDs just when the event is being created or handled within the logging framework. One way to achieve that is by instrumenting the supported logging frameworks. Another way is to register logging-framework-specific listeners.

eyalkoren:
## What does this PR do?
Closes #2412 
Closes #2374 
Closes #2369 
Closes #2387 

Implementing a new approach for log correlation (which is about adding active `trace.id`, `transaction.id` and `error.id` to log events automatically). The current mechanism makes MDC updates when trace events are being activated or deactivated. This is efficient, but there are cases where the logging library is not available within the context of event activation, or there is access to the wrong one.
Instead, the new approach is to do JIT-MDC-updates, within the logging library, based on the active tracing events, through instrumentation. 

### General concept
Conceptually, the most robust way is to add the tracing IDs to the MDC right before the logging event is created and remove it right after for the following reasons:
- current state is guaranteed to be correct
- the relevant MDC is guaranteed to be available at this context
- even if the event is logged with delay on a different thread, the logging framework will make sure to capture the proper status when creating it

Related disadvantages are:
- logging events are not part of the logging API, so their creation may be done in multiple ways and using unstable internal APIs
- there may be a performance hit in comparison to adding once and removing once per tracing event activation/deactivation

By instrumenting the public logging API themselves, we can benefit from all listed advantages and avoid the first disadvantage.
As for performance- since we use the actual logging framework best practice to decorate logging events with thread-contextual data, we can rely on optimizations done by the framework, so if there is a performance hit, it should be minimal. However, below there are some framework-specific pointers in case we want to optimize.

### Prior research (documenting for later reference)
The current approach is to instrument public logging APIs. If this proves to be too impactful in terms of performance, following are some library-specific options for optimization.

#### log4j2
Trying to capture the `LoggingEvent` creation is not trivial because of API instability, multiple creation options (constructor, static factory method and builder at least) and because events may be reusable.
One option is to try and instrument at least two `org.apache.logging.log4j.core.impl.ContextDataFactory#createContextData()` overloads, but it can get complicated as the returned `StringMap`s may be either mutable or immutable (see `org.apache.logging.log4j.util.StringMap#freeze()`).
It is difficult to find a single method that is also stable across versions through which all logging APIs go through, and not sure this would improve anything.

#### logback
The Logback MDC implementation tracks map updates and reads, and creates a copy of the map every time there is an update after a read (see `ch.qos.logback.classic.util.LogbackMDCAdapter`). This means that our approach will incur more map copy actions.
The logging event's immutable context map gets populated either when the event is logged, or when it is decided to defer its logging. `ch.qos.logback.classic.spi.LoggingEvent#getMDCPropertyMap()` may be a good instrumentation target as it seems stable across versions I believe it returns a mutable collection.
By doing so we can avoid updates of the MDC-underlying map.

## To-do list
- [x] Remove `apm-log-correlation-plugin`
- [x] Merge `apm-error-logging-plugin` into one plugin with the shading plugin
- [x] Rename the shading plugin to something general (eg logging-integration)
- [x] Rearrange packages within the unified plugin, so that ECS-reformatting, correlation and error-capturing is separated
- [x] Get rid of `co.elastic.apm.agent.impl.transaction.TraceContext#applicationClassLoader`
- [x] Make nesting check global so that if two logging frameworks are used at the same time, we only update the MDC once
- [x] Make sure we don't do MDC updates when the the logging API is called with an insufficient level (meaning - the event won't be logged)
- [x] Test Error creation and correlation for log4j2's `fatal` and `catching` APIs, and log4j1's `fatal` API
- [x] Check whether #2387 is related and resolved by this PR
- [x] Add trace metadata to agent log if `log_format_file=JSON` - will be handled through #2495 
- [x] Make log correlation on by default
- [x] Remove the `enable_log_correlation` config

## Checklist

  - [x] I have added tests that prove my fix is effective on log4j2 (oldest and newest supported versions)
  - [x] I have added tests that prove my fix is effective on logback (oldest and newest supported versions)
  - [x] I have added tests that prove my fix is effective on log4j1 
  - [x]  I have added tests for JBoss logging framework
  - [x] I have added tests for slf4j
  - [ ] ~~I added tests for log correlation within agent logs~~ - will be handled through #2495
  - [x] I have made corresponding changes to the documentation - error event capturing for log4j1 was, no direct support for slf4j log correlation anymore


apm-agent-plugins/apm-log-shader-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/Log4j2TraceCorrelationInstrumentation.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public ElementMatcher<? super NamedElement> getTypeMatcherPreFilter() {
+        return nameContains("Logger");
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
+        return hasSuperType(named("org.apache.logging.log4j.Logger"));
```
felixbarny, line 55: Might be worth checking if we could instrument `org.apache.logging.log4j.core.impl.LogEventFactory#createEvent` instead.

felixbarny:
> The current mechanism makes MDC updates when trace events are being activated or deactivated. This is efficient […]
> there may be a performance hit in comparison to adding once and removing once per tracing event activation/deactivation

The current mechanism is not more efficient in all cases. In fact, it’s often less efficient. Generally, the more the application logs, the more efficient the current strategy with the MdcActivationListener is. The less an application logs, the more efficient it is to only add the correlation ids on demand. An extreme case would be a bad n+1 query that does 10k calls to Redis. With the current mechanism, the hook would be executed for every span activation/deactivation, even if there are no log events in the scope of these spans.

Due to that, I think we can even enable log correlation by default now and maybe also add the `span.id` to the MDC. That’s because with the new mechanism, the Strings for the correlation ids only get allocated on demand - if and when they’re actually needed.

Also, I think adding to `ThreadContex`t before each log is not significantly more or less efficient than using `ContextDataInjector` or `ContextDataProvider` as they also require allocating a new `Map`/`StringMap` for each enabled log event.

---

However, I’d like to challenge the approach to instrument all logger methods, such as `trace`, `debug`, `info`, etc.
The biggest concern I have is around performance and the ability for these methods to get inlined. Specifically the methods with vararg parameters (such as `org.slf4j.Logger#info(java.lang.String, java.lang.Object…)`) rely on being inlined. Also, we’ll have to check for re-entrancy and whether the log level is enabled. Maybe not a huge cost but it may add up, especially for `trace` logs. Matching and instrumenting a lot of methods is also more expensive.

I think there are good alternative methods we can instrument that are executed after checking whether the log level is enabled. It looks like these methods are always executed and are within a few frames of the logger method calls:

- Logback: `ch.qos.logback.classic.Logger#callAppenders`
- Log4j2: 
    - `org.apache.logging.log4j.core.impl.LogEventFactory#createEvent`
    - `org.apache.logging.log4j.core.impl.LocationAwareLogEventFactory#createEvent`
- Log4j1: `org.apache.log4j.Category::callAppenders`
- JUL: `java.util.logging.Logger#log(java.util.logging.LogRecord)`

tobiasstadler:
I just wanted to mention that log correlation is (still) working for me with this changes.

eyalkoren:
> Also, I think adding to ThreadContext before each log is not significantly more or less efficient than using ContextDataInjector or ContextDataProvider as they also require allocating a new Map/StringMap for each enabled log event.

That's absolutely right, but it's different when it's the user's choice to add this overhead. Either way, I don't think this aspect of overhead would be an issue, it's more points to consider and possibly optimize.

> However, I’d like to challenge the approach to instrument all logger methods, such as trace, debug, info, etc.

No need to challenge, the description of this PR explains why event creation is ideal in terms of JIT-timing. I instrumented the API in order to test the general approach and overcome the concerns described above. It's still in draft state because we are still exploring our options.
There is considerable value in simplicity, and the fact that very simple instrumentation was sufficient to demonstrate support for these three logging libraries has its value. That is not to say that simplicity outweighs performance, the opposite is true, but we should give up the simplest solution only if there is indeed an overhead concern, and I am absolutely sure this would be the case eventually.

One big advantage of instrumenting based on the APIs, is the instrumentation of slf4j.

eyalkoren:
I just came back to this for a brief review, adding comments for discussion: I do like the simplicity and robustness (e.g. stability across versions and independency of actual implementation) of the API-based instrumentation, but I agree overhead is a major factor, so let's review the separate concerns:

> The biggest concern I have is around performance and the ability for these methods to get inlined.

That's a valid concern indeed. Since our advices are not inlined anymore, we add less bytecode today. It's hard to say what the effect would be, but we can do some rough estimation. I just compared `ch.qos.logback.classic.Logger#info(java.lang.String, java.lang.Object...)` before and after and the increase is from 9 bytecode instructions to 33. From what I could find, the default value for `-XX:FreqInlineSize` for  Linux 64-bit is 325, so I think we should be OK. As far as I can see in log4j2 and Logback, these API methods I instrumented are not nested, so this bytecode is added once within each code path, even when including slf4j.

> Also, we’ll have to check for re-entrancy ...

Reentrance guard is already implemented in the proposed `AbstractLogCorrelationHelper`.

> ...  and whether the log level is enabled.

Whether log level is enabled is nice to test, although adding to thread-local maps is quite cheap, so not sure it worth the effort. If we decide it is, then we can use the related APIs for that.

> Matching and instrumenting a lot of methods is also more expensive.

True, but we apply method matching only for a limited number of classes. From what I could see, even if it's 10X comparing to single-method matching, it's still a matter of few milliseconds for the entire library.

eyalkoren:
Summarizing offline discussion:
- We decided not to provide a generic support for slf4j for now, and instead support the underlining frameworks only. This allows us to do a more performant and framework-specific instrumentation, like @felixbarny suggested above
- We will add support for JULI in a different PR
- We want to add ECS-logging support and then reformatting and log correlation to JBoss logging - always used as a facade for the basic logging implementation, for MDC it either delegates to log4j2 (`ThreadContext`), or implements internally (when used with JULI). This means an issue in ECS-logging (maybe only testing) and an issue for reformatting and log correlation within the agent.

eyalkoren:
run integration tests

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogCorrelationHelper.java
```diff
@@ -0,0 +1,93 @@
...
+            return added;
+        }
+
+        @Override
+        protected void removeFromMdc() {
+            removeFromMdc(TRACE_ID_MDC_KEY);
+            removeFromMdc(TRANSACTION_ID_MDC_KEY);
+        }
+
+        protected abstract boolean addToMdc(String key, String value);
```
felixbarny, line 89: All implementations seem to always return `true. Am I missing something?

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogCorrelationHelper.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    protected abstract boolean addToMdc();
+
+    protected abstract void removeFromMdc();
+
+    /**
+     * Default abstract implementation for {@link AbstractLogCorrelationHelper}, which retrieves the currently active context and
+     * adds metadata key-by-key.
+     */
+    public static abstract class DefaultLogCorrelationHelper extends AbstractLogCorrelationHelper {
```
felixbarny, line 68: Much more elegant than what I did 👏

apm-agent-plugins/apm-log-shader-plugin/apm-log-shader-plugin-common/src/main/java/co/elastic/apm/agent/log/shader/AbstractLogCorrelationHelper.java
eyalkoren: No no, you're right. I thought the API makes more sense this way because the abstract implementation doesn't really know anything about the concrete helper (for example - `MDC` implementation is noop, or rejects the key for some reason or even the helper decides not to add to MDC because it know it's an agent log and the format is not JSON).
But it's all theoretical as it is now, so I can change the API accordingly.

eyalkoren:
run integration tests

eyalkoren:
run integration tests

eyalkoren:
@tobiasstadler as this PR changed a lot since you tried it last time, if you get a chance, please verify that the [latest snapshot](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/PR-2428/33/artifacts) still works as expected.
Thanks!

apm-agent-plugins/apm-logging-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/correlation/Log4j2LogCorrelationInstrumentation.java
```diff
@@ -0,0 +1,118 @@
...
+    @Override
+    public Collection<String> getInstrumentationGroupNames() {
+        return Collections.singleton("log4j2-correlation");
+    }
+
+    @Override
+    public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
+        return not(isBootstrapClassLoader())
+            .and(not(CustomElementMatchers.isAgentClassLoader()))
+            .and(classLoaderCanLoadClass("org.apache.logging.log4j.core.impl.LogEventFactor"));
```
tobiasstadler, line 55:
```suggestion
            .and(classLoaderCanLoadClass("org.apache.logging.log4j.core.impl.LogEventFactory"));
```

tobiasstadler:
@eyalkoren Log correlation for Log4J works for me. But log correlation for jboss-logmanger (https://github.com/jboss-logging/jboss-logmanager) does not work (anymore). I can have a look/take care of it, if you like.

As far as I can see `enable_log_correlation` is not used any more. Was that by intention.

apm-agent-plugins/apm-logging-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/correlation/Log4j2LogCorrelationInstrumentation.java
eyalkoren: Thanks, what a miss! 🤦‍♂️ 
Tests passing with this mean that `Log4j2_17_1InstrumentationTestRunner` and `Log4j2_6SInstrumentationTestRunner` are not being ran on CI. 
I am not sure if this is due to a mix between JUnit 4 and JUnit 5 on the same directory or due to test class naming. I'll look into that

tobiasstadler: By default Maven Surefire does not consider `*TestRunner` a test, see https://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html

eyalkoren:
@tobiasstadler like always, thanks a lot for your valuable input!

> @eyalkoren Log correlation for Log4J works for me

Thanks for the validation.

> But log correlation for jboss-logmanger (https://github.com/jboss-logging/jboss-logmanager) does not work (anymore). I can have a look/take care of it, if you like.

Yes, if you can say what's missing, that would be awesome! I removed the generic JBoss logging support (always adding to JBoss MDC on span activation and removing on span deactivation), assuming that the other log correlation handling will cover for that with the new mechanism, which adds to MDC active span/transaction/error metadata just before logging an event and removing just after. Since JBoss logging can be used with other frameworks, we figured that all is missing is JUL log correlation and what I did for that is instrument only these [specific loggers](https://github.com/elastic/apm-agent-java/pull/2428/files#diff-1e60adf4201bda8a34d45d470622da8a144f6c1f53010ad9b92d779f244b0c22). One of my followup plans was to add JBoss integration test (through #2494 ), so if you can provide input on what's not working, that would be a great help. My intention was to see if we can cover that through [`JBossIT`](https://github.com/elastic/apm-agent-java/blob/main/integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/JBossIT.java) somehow.

> As far as I can see enable_log_correlation is not used any more. Was that by intention.

The intention is to "promote" log correlation now to be on by default, so it's not really required with the new mechanism. If it creates issues, users will be able to shut it down by disabling the relevant instrumentation.
However, we may need to support that for backward compatibility... @felixbarny do you think we need to continue supporting this flag? I think the only non-backwards-compatible scenario here is if someone uses an agent with `enable_log_correlation==false` and a logging pattern that includes all MDC keys - for such users, logs will start showing our tracing IDs, as opposed to before. My opinion is that this doesn't justify maintaining this flag.

tobiasstadler:
@eyalkoren I will take care of the jboss-logmanager once this PR is merged (I already have a working implementation).

eyalkoren:
> @eyalkoren I will take care of the jboss-logmanager once this PR is merged (I already have a working implementation).

I rather merge this PR with JBoss integration working instead of having a middle state where it gets broken. 
Can you make a PR on this branch?
Can you explain what is not working?
Thanks!

tobiasstadler:
> Can you make a PR on this branch?

See eyalkoren/apm-agent-java#5

tobiasstadler:
> Can you explain what is not working?
JBoss LogManger is a logging implementation bases on JUL. Since there is no support for JUL at the moment, JBoss LogManager needs its one implementation for log correlation.

eyalkoren:
> JBoss LogManger is a logging implementation bases on JUL. Since there is no support for JUL at the moment, JBoss LogManager needs its one implementation for log correlation.

Hmm, I assumed a logger is obtained through `Logger.getLogger()`, in which case the LogManager is covered by the instrumentation of [`org.jboss.logging.JBossLogManagerLogger`](https://github.com/elastic/apm-agent-java/pull/2428/files#diff-1e60adf4201bda8a34d45d470622da8a144f6c1f53010ad9b92d779f244b0c22R56), but this doesn't cover `LogManager.getLogManager().getLogger()`...

apm-agent-plugins/apm-logging-plugin/apm-log4j2-plugin/src/main/java/co/elastic/apm/agent/log4j2/correlation/CorrelationIdMapAdapter.java
```diff
@@ -0,0 +1,179 @@
...
+        }),
+        new LazyEntry(TRANSACTION_ID_MDC_KEY, new Callable<String>() {
+            @Override
+            @Nullable
+            public String call() {
+                AbstractSpan<?> activeSpan = tracer.getActive();
+                if (activeSpan == null) {
+                    return null;
+                }
+                return activeSpan.getTraceContext().getId().toString();
```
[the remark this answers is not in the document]
eyalkoren: > This now sets the transaction.id to the active span's ID. I don't think that's what we want.

True!! I need to see again why the test did not cover that!
The intention was to do `activeSpan.getTraceContext().getTransactionId().toString()`, is getting the transaction object from the span preferable for any reason?

> Alternatively, we could change the implementation of currentTransaction() to return getActive().getTransaction() 

Makes sense, but I prefer doing it separately just in case there are implications we don't think about now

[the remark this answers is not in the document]
eyalkoren: > I need to see again why the test did not cover that!

Well, that's because the active span was the active transaction 🙂 
I'll add a test that creates and activates a span as well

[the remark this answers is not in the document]
felixbarny: > The intention was to do activeSpan.getTraceContext().getTransactionId().toString(), is getting the transaction object from the span preferable for any reason?

I forgot that we had the transaction id directly within `TraceContext`.

> Makes sense, but I prefer doing it separately just in case there are implications we don't think about now

👌

apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/test/java/co/elastic/apm/agent/loginstr/LoggingInstrumentationTest.java
```diff
@@ -82,71 +95,70 @@ public void setup() throws Exception {
 
         // IMPORTANT: keep this last, so that it doesn't interfere with Mockito settings above
         serviceName = Objects.requireNonNull(tracer.getMetaDataFuture().get(2000, TimeUnit.MILLISECONDS).getService().getName());
+
+        transaction = Objects.requireNonNull(tracer.startRootTransaction(null)).activate();
```
SylvainJuge, line 99: [minor] could be replaced with a call to `startRootTransaction()`

SylvainJuge (commented):
I haven't finished the review yet, I only have a few comments/questions so far.
That's definitely an impressive refactor and the end result is definitely worth the effort as everything now looks better and easier to extend and maintain.

Some tests are disabled with `@Disabled` then executed only through Junit4 integration tests, I wonder if we could have a better option for this in the future as 1) it may result in having lots of tests that appear to be disabled in CI but are not in practice as they are also integration tests, and 2) it creates a bit of duplication so having a generic Junit4-wrapper would be nice if such thing exists.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/error/ErrorCapture.java
```diff
@@ -44,6 +44,8 @@
 
     private static final Logger logger = LoggerFactory.getLogger(ErrorCapture.class);
 
+    private static final ThreadLocal<ErrorCapture> activeError = new ThreadLocal<>();
```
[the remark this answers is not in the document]
eyalkoren: You reminded me that this one made me think I need to take another look into our thread local usages in general. 
In this specific case, I don't think it is required, as it is set and removed atomically within a single method execution. Once a method bytecode starts to execute, I don't think any of its dependencies may change before it ends.
But it may be something we want to set general rules for and make sure to follow all across.

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
```diff
@@ -149,36 +149,14 @@ static LogLevel mapLogLevel(LogLevel original) {
...
     private final ConfigurationOption<LogEcsReformatting> logEcsReformatting = ConfigurationOption.enumOption(LogEcsReformatting.class)
         .key("log_ecs_reformatting")
         .configurationCategory(LOGGING_CATEGORY)
         .tags("added[1.22.0]", "experimental")
         .description("Specifying whether and how the agent should automatically reformat application logs \n" +
             "into {ecs-logging-ref}/index.html[ECS-compatible JSON], suitable for ingestion into Elasticsearch for \n" +
             "further Log analysis. This functionality is available for log4j1, log4j2 and Logback. \n" +
-            "Once this option is enabled with any valid option, log correlation will be activated as well, " +
-            "regardless of the <<config-enable-log-correlation,`enable_log_correlation`>> configuration. \n" +
+            "Once this option is enabled with any valid option, log correlation will be activated as well. \n" +
```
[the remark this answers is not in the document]
eyalkoren: I do want to say something about correlation being part of it, but you're right, the existing description is misleading, I'll change it 👍

apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/main/java/co/elastic/apm/agent/loginstr/error/AbstractLoggerErrorCapturingInstrumentation.java
```diff
@@ -89,9 +88,4 @@ public static void logExit(@Advice.Enter @Nullable Object errorCaptureObj) {
         ret.add("logging");
         return ret;
     }
-
-    @Override
-    public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
-        return not(ofType(nameStartsWith("co.elastic.apm.")));
-    }
```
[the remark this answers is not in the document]
eyalkoren: It was required before and now as well for log4j2, because we load it within the agent. However, while it didn't interfere with tests of other frameworks before, it does now, so I removed it from the abstract and implemented it specifically for log4j2, where we test with `TestClassWithDependencyRunner`, which uses `co.elastic.test.ChildFirstURLClassLoader` especially to escape these matchers.

apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/main/java/co/elastic/apm/agent/loginstr/correlation/AbstractLogCorrelationHelper.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Default abstract implementation for {@link AbstractLogCorrelationHelper}, which retrieves the currently active context and
+     * adds metadata key-by-key.
+     */
+    public static abstract class DefaultLogCorrelationHelper extends AbstractLogCorrelationHelper {
+
+        private final Tracer tracer = GlobalTracer.get();
+
+        @Override
+        protected boolean addToMdc() {
```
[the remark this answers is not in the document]
eyalkoren: It implements a `protected abstract` method that needs to have `protected` access at least.

apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/test/java/co/elastic/apm/agent/loginstr/LoggingInstrumentationTest.java
eyalkoren: I changed to using `startTestRootTransaction()` instead, assuming that this is what you meant. Let me know it was something else.

eyalkoren:
> Some tests are disabled with @Disabled then executed only through Junit4 integration tests, I wonder if we could have a better option for this in the future as 1) it may result in having lots of tests that appear to be disabled in CI but are not in practice as they are also integration tests, and 2) it creates a bit of duplication so having a generic Junit4-wrapper would be nice if such thing exists.

Yes, we have to improve on this front, but I think it's out of the scope of this PR

apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/test/java/co/elastic/apm/agent/loginstr/reformatting/UtilsTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Nullable
+    private final String logEcsFormattingDestinationDir = config.getConfig(LoggingConfiguration.class).getLogEcsFormattingDestinationDir();
+
+    private String computeReformattedLogFilePathWithConfiguredDir(String logFilePath) {
+        return Utils.computeLogReformattingFilePath(logFilePath, logEcsFormattingDestinationDir);
+    }
+
+    @Test
+    void testShadePathComputation() {
```
SylvainJuge, line 43: [minor] seems a good candidate for changing the `computeReformattedLogFilePathWithConfiguredDir` entry point and test to use `Path` instead of a raw file path in a `String` that requires to rewrite path.

SylvainJuge (approved):
LGTM, this is really nice and I only have a few minor comments.

Another minor thing that might be nice to add if you have time is to make sure to properly document n javadoc which methods are instrumented in `*Instrumentation` classes to make it easier to navigate to the instrumented classes.

apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/test/java/co/elastic/apm/agent/loginstr/LoggingInstrumentationTest.java
```diff
@@ -161,43 +173,43 @@ protected boolean markersSupported() {
...
 
     @Test
-    public void testLazyShadeFileCreation() throws Exception {
-        initializeShadeDir("delayed");
+    public void testLazyEcsFileCreation() throws Exception {
+        initializeReformattingDir("delayed");
         logger.trace(TRACE_MESSAGE);
         logger.debug(DEBUG_MESSAGE);
-        assertThat(Files.exists(Paths.get(getShadeLogFilePath()))).isFalse();
+        assertThat(Files.exists(Paths.get(getLogReformattingFilePath()))).isFalse();
```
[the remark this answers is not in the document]
eyalkoren: > we can rely on assertJ assertions for files.

❤️ TIL

> As a further improvement, making the getLogReformattingFilePath() method return a Path instead of a String might be relevant too.

Some usages require the path as a String, since it's only within the test I will skip this one

efdknittlfrank:
Awesome, looking forward to the next release of the agent! Thank you for all your incredible work :heart:

alejavigo:
Maybe this change in 1.30.0 has to do with my problem here? :
https://github.com/elastic/apm-agent-java/issues/3563

PR #2430, opened 2022-01-28
Fixed Java LDAP tests in JDK 17

tobiasstadler:
## What does this PR do?
Fixed test on JDK 17

## Checklist
- [x] This is something else

### DIFF
--- a/apm-agent-plugins/apm-java-ldap-plugin/pom.xml
+++ b/apm-agent-plugins/apm-java-ldap-plugin/pom.xml
@@ -28,4 +28,15 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <argLine>@{argLine} --add-exports java.naming/com.sun.jndi.ldap=ALL-UNNAMED</argLine>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 </project>


apm-agent-plugins/apm-java-ldap-plugin/pom.xml
```diff
@@ -28,4 +28,15 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <argLine>@{argLine} --add-exports java.naming/com.sun.jndi.ldap=ALL-UNNAMED</argLine>
```
felixbarny, line 37: What happens if someone uses Java 17 to run their app and does add this export? Will the agent break their app or throw errors?

tobiasstadler: As exceptions, that are thrown from plugins, are caught/handled by ByteBuddy/the agent, I assume the plugin won't work, but the rest of the app work.

felixbarny: That's true but these exceptions still affect the application - more overhead due to creation and logging and the logs are going to be spammed, too.
Just disabling the module on Java 17 is probably also not the best option as even Java 11 can be configured to enforce the module boundaries instead of warning about them.
I think we'll probably have to remove the LDAP plugin until we have found a solution for that.

tobiasstadler: Ok, I will create an external plugin in the mean time.

tobiasstadler: Are you taking care of the removal or should I?

felixbarny: I took care of it: #2431

tobiasstadler: For the record, the plugin can be found here: https://github.com/tobiasstadler/apm-java-ldap-plugin

felixbarny: What we could do to have the plugin built-in is try to load a class from `com.sun.jndi.ldap` and skip loading the plugin if the class can't be loaded.

tobiasstadler: some sort of static block inside `LdapClientInstrumentation`?

felixbarny: Either that or within a matcher. I wonder if also resource loading would fail so that we could use `CustomElemetMatchers.classLoaderCanLoadClass`.

tobiasstadler: ` classLoaderCanLoadClass("com.sun.jndi.ldap.Connection")` seems to work.

felixbarny: Seems to work in the sense that it returns `false` on Java 17 without `--add-exports java.naming/com.sun.jndi.ldap=ALL-UNNAMED` but `true` with that flag?

felixbarny: The downside of that is that this eagerly links `com.sun.jndi.ldap.LdapResult` and initializes a bunch of classes that otherwise would not be initialized at this point. We usually try not to impact the order in which classes are loaded as this can negatively impact the application. For example, if java.util.logging is used somewhere in the prematurely initialized classes, app server can't set up their custom log handlers.

tobiasstadler: yes, this is true.

PR #2355, opened 2021-12-18
Support for the Java LDAP client

tobiasstadler:
## What does this PR do?
I added support for instrumenting the Java LDAP client

## Checklist
- [x] This is an enhancement of existing features, or a new feature in existing plugins
  - [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/master/CHANGELOG.asciidoc)
  - [x] I have added tests that prove my fix is effective or that my feature works
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] I have made corresponding changes to the documentation

PR #2431, opened 2022-01-28
Remove ldap plugin

felixbarny:
Removes the ldap plugin for the time being, until we have resolved the issues mentioned in https://github.com/elastic/apm-agent-java/pull/2355#pullrequestreview-866104634 and https://github.com/elastic/apm-agent-java/pull/2430/files#r794536232


felixbarny:
I don't like this PR 😿

PR #2440, opened 2022-02-02
Service name auto discovery docs and polishing

felixbarny:
## What does this PR do?

- Documents service name auto discovery mechanisms
- Documents how to set `Implementation-Title` and `Implementation-Version` manifest entries with maven
- Avoid standalone spring applications to have two different service names, one based on the jar name, the other based on `spring.application.name`.

## Checklist

### DIFF
--- a/apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/springwebmvc/SpringServiceNameInstrumentation.java
+++ b/apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/springwebmvc/SpringServiceNameInstrumentation.java
@@ -20,8 +20,8 @@ package co.elastic.apm.agent.springwebmvc;
 
 import co.elastic.apm.agent.bci.TracerAwareInstrumentation;
 import co.elastic.apm.agent.configuration.ServiceInfo;
-import co.elastic.apm.agent.sdk.logging.Logger;
-import co.elastic.apm.agent.sdk.logging.LoggerFactory;
+import co.elastic.apm.agent.servlet.ServletServiceNameHelper;
+import co.elastic.apm.agent.servlet.adapter.JavaxServletApiAdapter;
 import net.bytebuddy.asm.Advice;
 import net.bytebuddy.description.NamedElement;
 import net.bytebuddy.description.method.MethodDescription;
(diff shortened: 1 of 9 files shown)


apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/springwebmvc/SpringServiceNameInstrumentation.java
```diff
@@ -67,50 +67,47 @@ public String getAdviceClassName() {
 
     public static class SpringServiceNameAdvice {
 
-        private static final Logger logger = LoggerFactory.getLogger(SpringServiceNameAdvice.class);
-
         @Advice.OnMethodExit(suppress = Throwable.class, inline = false)
         public static void afterInitPropertySources(@Advice.This WebApplicationContext applicationContext) {
+            // avoid having two service names for a standalone jar
+            // one based on Implementation-Title and one based on spring.application.name
+            if (!ServiceInfo.autoDetected().isMultiServiceContainer()) {
```
SylvainJuge, line 74: I think we could still honor the spring application name here in case the manifest entry is not set.
- If the manifest entry is available, we make `isMultiServiceContainer()` return `false` and then `ServiceInfo.autoDetected()` will match what was read in the manifest
- If the manifest entry is not available, we should still set it, so in the case of our test application it should still work as expected.

integration-tests/spring-boot-2/spring-boot-2-base/src/test/java/co/elastic/apm/spring/boot/AbstractSpringBootTest.java
```diff
@@ -102,7 +102,9 @@ public void greetingShouldReturnDefaultMessage() throws Exception {
         assertThat(transaction.getContext().getUser().getId()).isEqualTo("id");
         assertThat(transaction.getContext().getUser().getEmail()).isEqualTo("email");
         assertThat(transaction.getContext().getUser().getUsername()).isEqualTo("username");
-        assertThat(transaction.getTraceContext().getServiceName()).isEqualTo("spring-boot-test");
+        // as this test runs in a standalone application and not in a servlet container,
+        // the service.name will not be overwritten for the webapp class loader based on spring.application.name
+        assertThat(transaction.getTraceContext().getServiceName()).isNull();
```
SylvainJuge, line 107: We should be able to still support this in the case where the auto-detected value is not read from manifest (see previous comment).

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/springwebmvc/SpringServiceNameInstrumentation.java
felixbarny, line 74: Some background:
The risk of that is that if we don't properly detect that we're running on an application server/servlet container. In that case, we wouldn't be able to determine the service name based on `spring.application.name`. That's why I didn't add the check in `Tracer.overrideServiceInfoForClassLoader` but made it specific to this instrumentation. This way, we at least can still override based on the servlet context path and the display-name. It's quite unlikely that a standalone application that's using embedded Tomcat, for example, sets a display-name or even a servlet context path. Therefore, if any of these are set, that's a pretty good indication that we're running in a container and not in a standalone jar.

SylvainJuge (approved):

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/springwebmvc/SpringServiceNameInstrumentation.java
SylvainJuge: Clarified this through an offline discussion: we have another fallback here for the `ServiceInfo.autoDetected()` that relies on the `.jar` name. Unfortunately when running in the IDE the test app is not packaged and thus not available.

PR #2443, opened 2022-02-03
Added Instrumentation for `ServletContextListener#contextInitialized`

tobiasstadler:
## What does this PR do?
Improves service name and version auto detection

## Checklist


felixbarny (approved):

felixbarny:
run elasticsearch-ci/docs

tobiasstadler:
Thank You!

felixbarny:
Thank you too!

Do you still have a need for Tomcat, JBoss/WildFly specific instrumentations to init earlier and to make it work for JMS? I think now that we have more fundamentals in place, such as`ServletServiceNameHelper` as well as generic instrumentation via the Servlet API, I'm also more comfortable with adding container-specific instrumentations. That's because if they don't work in a new version of the container, we at least have a fallback that's good enough for most cases.

To make these container-specific instrumentations even more lightweight, I'd just add them in the apm-servlet-plugin. For example under `co.elastic.apm.agent.servlet.init` or `co.elastic.apm.agent.servlet.servicename`.

tobiasstadler:
Yes, at the moment I still prefer having the "correct" service name/version early in the process.

PR #2444, opened 2022-02-03
Support for setting the application class loader for a transaction via the public api

tobiasstadler:
## What does this PR do?
Implements the solution suggested in #2369

## Checklist

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -43,6 +43,7 @@ endif::[]
 * Added support for instrumenting Struts 2 static resource requests - {pull}1949[#1949]
 * Added support for Java/Jakarta WebSocket ServerEndpoint - {pull}2281[#2281]
 * Added support for setting the service name on Log4j2's EcsLayout - {pull}2296[#2296]
+* Added support for setting the application class loader for a transaction via the public api - {pull}2444[#2444]
 
 [float]
 ===== Bug fixes
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/NoopTransaction.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/NoopTransaction.java
@@ -226,4 +226,10 @@ enum NoopTransaction implements Transaction {
     public Span setDestinationService(@Nullable String resource) {
         return this;
     }
+
+    @Nonnull
+    @Override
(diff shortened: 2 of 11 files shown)


felixbarny (approved):
I've added some changes, let me know if this works for you.
LGTM, but I'd like to wait for @eyalkoren's feedback before merging

felixbarny:
run elasticsearch-ci/docs

felixbarny (approved):

felixbarny:
run elasticsearch-ci/docs

tobiasstadler:
> I've added some changes, let me know if this works for you.

Much better that what I did! Thank you for your support.

Yes, service name/version and MDC work for me.

felixbarny:
run elasticsearch-ci/docs

eyalkoren:
I actually intend to get rid of `TraceContext#applicationClassLoader` as part of #2428 ...
We don't need this reference anymore and it's always a good idea to store one less weak reference (and any CL reference).
Since you are actually only interested in the implicit side-effect of `TraceContext#setApplicationClassLoader`, which is to override the service info for the transaction, the API seems inappropriate to me. Calling it `overrideServiceInfoFor(ClassLoader)` would be better (the javadoc would how the provided `ClassLoader` is used).

Generally speaking, I would prefer a cleaner and more generic API such as `overrideServiceInfo(String serviceName, String serviceVersion)`. If extracting the service version and name are simple enough- just implement that instead of relying on the automatic discovery. If relying on the automatic discovery is important, I assume this is mostly relevant for external plugins (not for other usages of the API), so we can expose `ElasticApmTracer#getServiceInfo(ClassLoader)` through the sdk by making it part of the `co.elastic.apm.agent.impl.Tracer` interface. WDYT?

tobiasstadler:
Hm, I find it quite inconvenient to (always) have to to call `getServiceInfo` before I can call `overrideServiceInfo`. What about something like `Transaction#withServiceInfoForClassCloader(ClassLoader)` (or maybe use instead of with or another better name) or #2369. What do you think.

@eyalkoren I will implement your suggestion in the next days

tobiasstadler:
@eyalkoren Please see #2451 for my implementation of your suggestion

tobiasstadler:
#2451 is the preferred api

PR #2445, opened 2022-02-03
fix windows hostname test

### LINKED ISSUE #2436: SystemInfoTest.testHostnameDiscoveryFallbackThroughInetAddress fails on Windows
Output on CI build

```
expected: "apm-ci-immutable-windows-2019-docker-1643640951187090662"
 but was: "APM-CI-IMMUTABL"
```

The reference to Docker here is only due to CI, the test fails in a similar way on a Windows machine.

## Steps to reproduce
Just execute `testHostnameDiscoveryFallbackThroughInetAddress` on Windows.

SylvainJuge:
## What does this PR do?

Fixes #2436 

The test failed on Windows, but not on *nix because of the combination of a few things:
- our test instrumentation did not allowed to emulate a non-existing env variable
- by default the tested env variables `HOST` and `HOSTNAME` are not set on *nix, which is the case also on CI, so the test that tried to cover this case behaved as expected by accident.
- on Windows the `COMPUTERNAME` env variable seems to always be present (including in CI), thus not being able to emulate a missing value did not worked as expected and as a result the test failed.

Morale of the story: testing on Windows helped us to find a bug in or test instrumentation, but it was not due to an underlying bug in the agent, only in testing code.

## Checklist

- [x] This is a bugfix
  - [ ] ~~I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)~~ not required here
  
  
## Other failing tests

There are other tests that are currently failing on Windows and will need further investigation, fixing them could be done either in this PR or separately. Fixing them is expected to trigger new failing tests that weren't previously known to fail

- [ ] co.elastic.apm.agent.log4j2.Log4j2ShadingTest#testLazyShadeFileCreation
- [ ] co.elastic.apm.agent.log4j2.Log4j2ShadingTest#testLogOverride

### DIFF
--- a/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentDownloadUtils.java
+++ b/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentDownloadUtils.java
@@ -48,7 +48,13 @@ public class AgentDownloadUtils {
         }
 
         URL jarLocation = AgentDownloadUtils.class.getProtectionDomain().getCodeSource().getLocation();
-        Path attacherCliDir = Paths.get(jarLocation.getPath()).getParent();
+        String jarLocationPath = jarLocation.getPath();
+        if (jarLocationPath.matches("/\\w\\:.*")){
+            // looks something like /C:/path... - Paths.get() doesn't like that
+            // so strip the leading /
+            jarLocationPath = jarLocationPath.substring(1);
+        }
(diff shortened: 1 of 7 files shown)


SylvainJuge:
As discussed today we will merge this PR even if tests still fail as it already helps fixing a few.
Further follow-up PRs are expected in the future to fix the remaining flaky/broken tests.

PR #2447, opened 2022-02-04
Fix dubbo flaky tests

SylvainJuge:
Dubbo tests randomly fail on CI and locally complaining about a network port being already busy.

- add retry logic when registering client & server as it seems to be randomly taking some time (given there are 2 clients and 2 servers, their startup ordering isn't guaranteed)
- only start the whole test setup once per JVM as it seems that otherwise a global registry is used and the test still randomly fails after a few iterations.

Tested locally: goes from ~20 tests until a failure happens to ~500+ without a single failure.


apm-agent-plugins/apm-dubbo-plugin/src/test/java/co/elastic/apm/agent/dubbo/AbstractDubboInstrumentationTest.java
```diff
@@ -38,54 +39,82 @@
...
-    void startRootTransaction() {
-        when(coreConfig.getCaptureBody()).thenReturn(CoreConfiguration.EventType.OFF);
+    void beforeEach() {
+
+        if (null == dubboTestApi) {
+            // only start test dubbo once, but with delegation to subclass for creating it
+            // thus we can't do that in @BeforeAll
 
-        // using context classloader is required here
-        ClassLoader cl = Thread.currentThread().getContextClassLoader();
```
SylvainJuge, line 65: In practice it does not seem to make any difference here, I don't know why (and if) it used to be required previously.

eyalkoren (approved):

Issue #2349, opened 2021-12-16
Investigate/fix increasing startup time with agent


SylvainJuge:
The total application startup time increases with more recent versions of the agent:

With Opbeans application, we get the following metrics
```
1.17.0: Started OpbeansApplication in 9.419 seconds (JVM running for 12.495)
1.18.0: Started OpbeansApplication in 9.573 seconds (JVM running for 14.468)
1.23.0: Started OpbeansApplication in 9.492 seconds (JVM running for 14.298)
1.24.0: Started OpbeansApplication in 9.226 seconds (JVM running for 14.415)
1.25.0: Started OpbeansApplication in 9.79 seconds (JVM running for 16.123)
1.26.0: Started OpbeansApplication in 9.936 seconds (JVM running for 20.367)
1.27.1: Started OpbeansApplication in 9.335 seconds (JVM running for 19.06)
1.28.1: Started OpbeansApplication in 9.883 seconds (JVM running for 19.801)
```

In order to help figure this out, using async profiler with [wall-clock profiling](https://github.com/jvm-profiling-tools/async-profiler#wall-clock-profiling) seems relevant.

Also, computing a diff flame graph from samples can be easily done with https://github.com/brendangregg/FlameGraph. Recording them with async-profiler requires to use the `collapsed` format with option `-o collapsed`.

`./FlameGraph/difffolded.pl ./{before,after}.txt | ./FlameGraph/flamegraph.pl > diff.svg`

jackshirazi:
I've validated the benchmark on an unloaded 4vcpu isolated VM with nothing else running, across 10 runs for each version, the cumulative total times (with individual run times) for the 10 runs are
1.17.0 154.929 : 15.415 16.914 15.686 15.322 15.94 15.418 15.045 14.814 15.047 15.328
1.18.0 164.699 : 17.557 16.786 16.875 16.105 16.388 16.278 16.128 15.985 15.934 16.663
1.23.0 166.765 : 17.972 16.758 17.096 16.896 16.429 15.935 15.806 16.335 16.442 17.096
1.24.0 170.021 : 18.192 16.929 17.089 17.906 17.146 16.596 16.07 16.538 16.634 16.921
1.25.0 171.276 : 18.128 17.375 16.863 17.611 17.908 16.805 16.499 16.186 16.692 17.209
1.26.0 177.194 : 18.063 17.89 17.743 18.176 18.144 17.323 16.978 17.25 16.979 18.648
[...]

PR #2448, opened 2022-02-04
Adds support for logging the used instrumentation groups

tobiasstadler:
## What does this PR do?
Adds support for logging the used instrumentation groups.

## Checklist
- [x] This is something else

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
@@ -45,7 +45,7 @@ import co.elastic.apm.agent.sdk.weakconcurrent.WeakMap;
 import co.elastic.apm.agent.tracemethods.TraceMethodInstrumentation;
 import co.elastic.apm.agent.util.DependencyInjectingServiceLoader;
 import co.elastic.apm.agent.util.ExecutorUtils;
-import co.elastic.apm.agent.util.InstrumentationUsageUtil;
+import co.elastic.apm.agent.util.InstrumentationUsageReporter;
 import net.bytebuddy.ByteBuddy;
 import net.bytebuddy.agent.builder.AgentBuilder;
 import net.bytebuddy.agent.builder.AgentBuilder.RedefinitionStrategy;
@@ -330,7 +330,7 @@ public class ElasticApmAgent {
         int numberOfAdvices = 0;
         for (final ElasticApmInstrumentation advice : instrumentations) {
             if (isIncluded(advice, coreConfiguration)) {
-                InstrumentationUsageUtil.addInstrumentation(advice);
+                InstrumentationUsageReporter.addInstrumentation(advice);
                 try {
                     agentBuilder = applyAdvice(tracer, agentBuilder, advice, advice.getTypeMatcher());
                     numberOfAdvices++;
@@ -460,7 +460,7 @@ public class ElasticApmAgent {
                         if (matches) {
                             logger.debug("Method match for instrumentation {}: {} matches {}",
                                 instrumentation.getClass().getSimpleName(), methodMatcher, target);
-                            InstrumentationUsageUtil.addUsedInstrumentation(instrumentation);
+                            InstrumentationUsageReporter.addUsedInstrumentation(instrumentation);
                         }
                         return matches;
                     } finally {
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
@@ -259,9 +259,7 @@ public class ElasticApmAgent {
         Runtime.getRuntime().addShutdownHook(new Thread(ThreadUtils.addElasticApmThreadPrefix("init-instrumentation-shutdown-hook")) {
             @Override
             public void run() {
-                if (tracer.getConfig(CoreConfiguration.class).logUsedInstrumentationGroupsOnExit()) {
-                    logger.info("Used instrumentation groups: {}", InstrumentationUsageUtil.getUsedInstrumentationGroups());
-                }
+                logger.info("Used instrumentation groups: {}", InstrumentationUsageUtil.getUsedInstrumentationGroups());
                 tracer.stop();
                 matcherTimers.clear();
                 LoggingConfiguration.shutdown();
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
@@ -36,7 +36,6 @@ import co.elastic.apm.agent.configuration.CoreConfiguration;
 import co.elastic.apm.agent.impl.ElasticApmTracer;
 import co.elastic.apm.agent.impl.ElasticApmTracerBuilder;
 import co.elastic.apm.agent.impl.GlobalTracer;
-import co.elastic.apm.agent.logging.LoggingConfiguration;
 import co.elastic.apm.agent.matcher.MethodMatcher;
 import co.elastic.apm.agent.sdk.ElasticApmInstrumentation;
 import co.elastic.apm.agent.sdk.logging.Logger;
@@ -259,10 +258,8 @@ public class ElasticApmAgent {
         Runtime.getRuntime().addShutdownHook(new Thread(ThreadUtils.addElasticApmThreadPrefix("init-instrumentation-shutdown-hook")) {
             @Override
             public void run() {
-                logger.info("Used instrumentation groups: {}", InstrumentationUsageUtil.getUsedInstrumentationGroups());
                 tracer.stop();
                 matcherTimers.clear();
-                LoggingConfiguration.shutdown();
             }
         });
         matcherTimers.clear();
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
@@ -46,6 +46,7 @@ import co.elastic.apm.agent.sdk.weakconcurrent.WeakMap;
 import co.elastic.apm.agent.tracemethods.TraceMethodInstrumentation;
 import co.elastic.apm.agent.util.DependencyInjectingServiceLoader;
 import co.elastic.apm.agent.util.ExecutorUtils;
+import co.elastic.apm.agent.util.InstrumentationUsageUtil;
 import net.bytebuddy.ByteBuddy;
 import net.bytebuddy.agent.builder.AgentBuilder;
 import net.bytebuddy.agent.builder.AgentBuilder.RedefinitionStrategy;
@@ -83,7 +84,6 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Future;
@@ -127,8 +127,6 @@ public class ElasticApmAgent {
      */
     private static final ConcurrentMap<String, ClassLoader> adviceClassName2instrumentationClassLoader = new ConcurrentHashMap<>();
     private static final ConcurrentMap<String, Collection<String>> pluginPackages2pluginClassLoaderRootPackages = new ConcurrentHashMap<>();
-
-    private static final ConcurrentMap<ElasticApmInstrumentation, Boolean> usedInstrumentations = new ConcurrentHashMap<>();
 
     /**
      * Called reflectively by {@code co.elastic.apm.agent.premain.AgentMain} to initialize the agent
@@ -221,7 +219,7 @@ public class ElasticApmAgent {
     }
 
     private static synchronized void initInstrumentation(final ElasticApmTracer tracer, Instrumentation instrumentation,
-                                                         final Iterable<ElasticApmInstrumentation> instrumentations, boolean premain) {
+                                                         Iterable<ElasticApmInstrumentation> instrumentations, boolean premain) {
         CoreConfiguration coreConfig = tracer.getConfig(CoreConfiguration.class);
         if (!coreConfig.isEnabled()) {
             return;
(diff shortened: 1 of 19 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4j2ConfigurationFactory.java
```diff
@@ -121,7 +121,8 @@ public Configuration getConfiguration(LoggerContext loggerContext, Configuration
     Configuration getConfiguration() {
         ConfigurationBuilder<BuiltConfiguration> builder = newConfigurationBuilder();
         builder.setStatusLevel(Level.ERROR)
-            .setConfigurationName("ElasticAPM");
+            .setConfigurationName("ElasticAPM")
+            .setShutdownHook("disable");
```
tobiasstadler, line 125: I had to change this so logging is still initialized during the shutdown hook.

@eyalkoren You did a lot in this area recently. Would you please have a look.

tobiasstadler:
@jackshirazi I just saw that you mentioned something similar in #2349.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -85,6 +85,13 @@
         .tags("added[1.0.0,Changing this value at runtime is possible since version 1.15.0]")
         .buildWithDefault(true);
 
+    private final ConfigurationOption<Boolean> logUsedInstrumentationGroupsOnExit = ConfigurationOption.booleanOption()
```
tobiasstadler, line 88: Maybe `LoggingConfiguration` is a more suitable place?

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -257,8 +259,19 @@ private static synchronized void initInstrumentation(final ElasticApmTracer trac
         Runtime.getRuntime().addShutdownHook(new Thread(ThreadUtils.addElasticApmThreadPrefix("init-instrumentation-shutdown-hook")) {
             @Override
             public void run() {
+                if (tracer.getConfig(CoreConfiguration.class).logUsedInstrumentationGroupsOnExit()) {
```
felixbarny, line 262: Instead of a new config option, maybe just log on debug?

apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/ElasticApmInstrumentation.java
```diff
@@ -89,6 +89,17 @@
  *
  */
 public abstract class ElasticApmInstrumentation {
+
+    private boolean used;
```
felixbarny, line 93: Instead of making the instrumentations aware of whether they have been used, I'd encapsulate this state in a dedicated class that holds a collection of used/unused instrumentations or just two sets of Strings for used/unused.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -257,8 +259,19 @@ private static synchronized void initInstrumentation(final ElasticApmTracer trac
...
+                    Set<String> usedInstrumentationGroups = new TreeSet<>();
+                    for (ElasticApmInstrumentation instrumentation : instrumentations) {
+                        if (!instrumentation.isUsed()) {
+                            continue;
+                        }
+                        usedInstrumentationGroups.addAll(instrumentation.getInstrumentationGroupNames());
+                    }
+                    logger.info("Used instrumentation groups: {}", usedInstrumentationGroups);
```
felixbarny, line 270: This set doesn't contain the minimum set of instrumentation groups that are enabled. The reason is that often an instrumentation has a generic group, such as `http-client`, and a more specific one, such as `okhttp`. If only `okhttp` had a match, but `apache-httpclient` did not, we would not want to include `http-client`, as it would enable both, even though `apache-httpclient` never matched.

Therefore, after compiling the list of used instrumentation groups, remove all the groups that did not match.
For example `[http-client, okhttp] - [http-client, apache-httpclient] = [okhttp]`.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
tobiasstadler: My motivation is to enable this function without having to enable debug logs. Enabling debug logs produces many too many logs and negatively impacts performance.

What do you think about removing the config option and keeping the log level at info? As it is only logged when the application exists, it shouldn't negatively impact the performance.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
tobiasstadler: This sounds very useful.

What if `CaptureSpanInstrumentation` is used but `CaptureTransactionInstrumentation` is not? Both have "public-api" and "annotations" as instrumentation groups. What should be logged in this case? Both groups?

apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/ElasticApmInstrumentation.java
tobiasstadler: Done

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
felixbarny: Excellent question. Yes, both groups should be logged.

You could implement this by maintaining two `Set<Collection<String>>`, one for matched and one for non-matched instrumentation groups. As both `CaptureSpanInstrumentation` and `CaptureTransactionInstrumentation` have the same collection of group names, they're collapsed into one entry of `["public-api", "annotations"]`. So when one instrumentation matches, the instrumentation groups are moved out of the set of non-matched instrumentation groups and get added to the set of matched groups.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
tobiasstadler: Done

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
tobiasstadler: Done

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -126,6 +128,8 @@
     private static final ConcurrentMap<String, ClassLoader> adviceClassName2instrumentationClassLoader = new ConcurrentHashMap<>();
     private static final ConcurrentMap<String, Collection<String>> pluginPackages2pluginClassLoaderRootPackages = new ConcurrentHashMap<>();
 
+    private static final ConcurrentMap<ElasticApmInstrumentation, Boolean> usedInstrumentations = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
tobiasstadler: Done

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4j2ConfigurationFactory.java
eyalkoren: Another option is to use `loggerContext.addShutdownListener()` within the other (API) `getConfiguration` overrides, but maybe using our own shutdown hook is preferable. Let me get back to you on that

eyalkoren: OK, let's disable the log4j shutdown hook and replace it with the existing mechanism we already have within the agent for this purpose: make `Log4jLoggerFactoryBridge` implement `co.elastic.apm.agent.context.LifecycleListener` and use the `stop` API to do both the logging and the shutdown

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -257,8 +259,10 @@ private static synchronized void initInstrumentation(final ElasticApmTracer trac
         Runtime.getRuntime().addShutdownHook(new Thread(ThreadUtils.addElasticApmThreadPrefix("init-instrumentation-shutdown-hook")) {
             @Override
             public void run() {
+                logger.info("Used instrumentation groups: {}", InstrumentationUsageUtil.getUsedInstrumentationGroups());
                 tracer.stop();
                 matcherTimers.clear();
+                LoggingConfiguration.shutdown();
```
eyalkoren, line 265: Do this through `co.elastic.apm.agent.context.LifecycleListener#stop`

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4jLoggerFactoryBridge.java
```diff
@@ -33,6 +35,13 @@
     private static final String FQCN = Log4jLoggerFactoryBridge.class.getName();
     private static final String PACKAGE = "co.elastic.apm.agent.sdk.logging";
 
+    public static void shutdown() {
```
eyalkoren, line 38: Make this as a `co.elastic.apm.agent.context.LifecycleListener#stop` implementation

eyalkoren (changes requested):

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
tobiasstadler: Done

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4jLoggerFactoryBridge.java
tobiasstadler: Done

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4j2ConfigurationFactory.java
tobiasstadler: A downside of the `LifecycleListener#stop` approach is that `Tracer switched to STOPPED state` won't get logged anymore.

tobiasstadler: Aso `Tracer stop stack trace: ` won't be logged anymore

eyalkoren: Good point and even if we fix that by lifting these loggings a bit higher, there would be a race with other listeners, like `MatcherTimerLifecycleListener`.
So let's revert that and call the shutdown from the end of `co.elastic.apm.agent.impl.ElasticApmTracer#stop` instead. Sorry about that...

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -257,6 +259,7 @@ private static synchronized void initInstrumentation(final ElasticApmTracer trac
         Runtime.getRuntime().addShutdownHook(new Thread(ThreadUtils.addElasticApmThreadPrefix("init-instrumentation-shutdown-hook")) {
             @Override
             public void run() {
+                logger.info("Used instrumentation groups: {}", InstrumentationUsageUtil.getUsedInstrumentationGroups());
```
eyalkoren, line 262: Please move this part to `MatcherTimerLifecycleListener#stop` to combine these two related functionalities

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4j2ConfigurationFactory.java
tobiasstadler: Done

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
tobiasstadler: Done

apm-agent-core/src/main/java/co/elastic/apm/agent/util/InstrumentationUsageUtil.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+public final class InstrumentationUsageUtil {
+
+    private static final Set<ElasticApmInstrumentation> allInstrumentations = new HashSet<>();
+
+    private static final ConcurrentMap<ElasticApmInstrumentation, Boolean> usedInstrumentations = new ConcurrentHashMap<>();
```
felixbarny, line 34: If there's no reason why we can't, I think we should make all of the fields and methods non-static and rename to something like `InstrumentationUsageReporter/Logger`.
I think static state should be confined to the minimum set of classes that really need to be static. While we can't really get rid of static state in `ElasticApmAgent`, I think we can do here. `ElasticApmAgent` would then have a static field of this class.

apm-agent-core/src/main/java/co/elastic/apm/agent/util/InstrumentationUsageUtil.java
```diff
@@ -0,0 +1,70 @@
...
+    public static void addUsedInstrumentation(ElasticApmInstrumentation instrumentation) {
+        usedInstrumentations.put(instrumentation, Boolean.TRUE);
+    }
+
+    public static Collection<String> getUsedInstrumentationGroups() {
+        Set<String> usedInstrumentationGroups = new TreeSet<>();
+        for (ElasticApmInstrumentation instrumentation : allInstrumentations) {
+            if (!usedInstrumentations.containsKey(instrumentation)) {
+                continue;
+            }
```
felixbarny, line 54: Isn't this the same as iterating over `usedInstrumentations`?

tobiasstadler: Yes, you are right. I will change that

apm-agent-core/src/main/java/co/elastic/apm/agent/util/InstrumentationUsageUtil.java
felixbarny: Yeah, a lifecycle listener makes sense. I guess the static state would be the easiest then. But instead of including it in `MatcherTimerLifecycleListener`, I'd make `InstrumentationUsageUtil` implement `LifecycleListener`. In `stop`, it would then log the static state.
Ideally though, we would introduce more lifecycle events, such as `onInstrumentationInitialized` and `onInstrumentrationMatched`. But that might be over-engineered for that one use-case. If we see the need for these lifecycle methods more often, we can still do it then.

felixbarny: Ha, race condition with c518c74 and b773b4b. That works for me as well.

tobiasstadler: Should is still move the logging? Is that ok for you @eyalkoren?

tobiasstadler: Looks [b729244](https://github.com/elastic/apm-agent-java/pull/2448/commits/b72924488637248a21396ca4d61300782314d505) good?

eyalkoren (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4jLoggerFactoryBridge.java
```diff
@@ -33,6 +35,13 @@
...
+        LoggerContextFactory factory = LogManager.getFactory();
+        for (LoggerContext context : ((Log4jContextFactory) factory).getSelector().getLoggerContexts()) {
```
eyalkoren, line 40: Tests fail on this cast, please take a look

eyalkoren (changes requested):

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4jLoggerFactoryBridge.java
tobiasstadler: fixed

felixbarny (approved):

tobiasstadler:
run elasticsearch-ci/docs

eyalkoren:
run elasticsearch-ci/docs

tobiasstadler:
Thank You!

PR #2450, opened 2022-02-07
Make integration tests run on Testcontainers Cloud

felixbarny:
## What does this PR do?
Makes the test suite fit for Testcontainers Cloud (TCC)
- Enables running all tests on M1 processors
- Biggest change is that `withFileSystemBind` has to be replaced with `withCopyFileToContainer`
- I've tested remote debugging to a Tomcat that runs on TCC. The performance seems to be good enough.

## What's not yet working
- When running the full test suite, the TCC workers run out of disk space (TCC is working on increasing disk space)

## Follow ups
- Currently, this is mainly for local development. But we may want to consider TCC for CI, too.
  - As Docker-based tests are run on TCC, we could switch to faster and cheaper ARM-based workers for the rest of the build.
  - Instead of splitting up the integration tests into several parallel steps, we could just execute them in one and increase the JUnit fork count. This will create one TCC worker per fork. Note that this is not implemented yet on TCC.
  - Rely on TCC's image cache. Currently we don't cache any Docker images on our CI(!)

### DIFF
--- a/apm-agent-core/src/test/java/co/elastic/apm/agent/impl/context/UrlTest.java
+++ b/apm-agent-core/src/test/java/co/elastic/apm/agent/impl/context/UrlTest.java
@@ -130,7 +130,7 @@ class UrlTest {
         // only the 'full' property is always computed
         // other attributes are not parsed unless sanitization is applied (which is an implementation detail)
         if(input.contains("@")){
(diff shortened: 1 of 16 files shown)


eyalkoren:
run end-to-end tests

PR #2451, opened 2022-02-07
Added support for setting service name and version for a transaction via the public api

tobiasstadler:
## What does this PR do?
Adds support for setting service name and version for a transaction via the public api

## Checklist


tobiasstadler:
I will add unit tests once you decide this is the way to go (Manuel testing looks good).

tobiasstadler:
@felixbarny @eyalkoren Are you both comfortable with this approach?

tobiasstadler:
I added tests in [9be1a8a](https://github.com/elastic/apm-agent-java/pull/2451/commits/9be1a8a532ecbebc8318c9b9451dc8f5b7dd9cf7)

tobiasstadler:
I just saw you are already prepared the changeling for 1.29.0. So I guess this wont't make it into the release?

tobiasstadler:
Ok, hopefully 1.29.1 or 1.30 then

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
```diff
@@ -232,4 +232,32 @@ public static void captureException(@Nullable Throwable e) {
...
     }
 
+    /**
+     * Gets the service name and version that is associated with a class loader
+     *
+     * @param classLoader the class loader for which the service name and version should be looked up
+     * @return the service name and version for the given class loader null
+     */
+    @Nullable
+    public static ServiceInfo getServiceInfoForClassLoader(@Nullable ClassLoader classLoader) {
```
eyalkoren, line 242: Get the service name and version through separate APIs that return a `String`, or just remove (I prefer the latter).

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
```diff
@@ -232,4 +232,32 @@ public static void captureException(@Nullable Throwable e) {
...
+     * </p>
+     * <p>
+     * NOTE: If the class loader already is associated with a service name and version,
+     * the existing information will not be overwritten.
+     * </p>
+     *
+     * @param classLoader the class loader which should be associated with the given service name and version
+     * @param serviceInfo the service name and version
+     */
+    public static void setServiceInfoForClassLoader(@Nullable ClassLoader classLoader, @Nullable ServiceInfo serviceInfo) {
```
eyalkoren, line 260: Use service name and version separately as `String` (through this singe API method)

apm-agent-api/src/main/java/co/elastic/apm/api/ServiceInfo.java
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package co.elastic.apm.api;
+
+public final class ServiceInfo {
```
eyalkoren, line 21: Remove

apm-agent-api/src/main/java/co/elastic/apm/api/Transaction.java
```diff
@@ -67,6 +67,18 @@
...
+     * Sets the service name and version for this transaction and its child spans.
+     * <p>
+     * NOTE: If this method is called after child spans are already created,
+     * they may have the wrong service name and version.
+     * </p>
+     *
+     * @param serviceInfo the service name and version
+     */
+    @Nonnull
+    Transaction setServiceInfo(ServiceInfo serviceInfo);
```
eyalkoren, line 80: Set the service name and method as `String` args (through the same API method)

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -505,7 +505,8 @@ public static void validateAdvice(ElasticApmInstrumentation instrumentation) {
         for (MethodDescription.InDefinedShape exitAdvice : typeDescription.getDeclaredMethods().filter(isStatic().and(isAnnotatedWith(Advice.OnMethodExit.class)))) {
             validateAdviceReturnAndParameterTypes(exitAdvice, adviceClassName);
             validateLegacyAssignToIsNotUsed(exitAdvice);
-            if (exitAdvice.getReturnType().asRawType().getTypeName().startsWith("co.elastic.apm")) {
+            String returnType = exitAdvice.getReturnType().asRawType().getTypeName();
+            if (returnType.startsWith("co.elastic.apm") && !returnType.startsWith("co.elastic.apm.api")) {
```
eyalkoren, line 509: Let's continue disallowing usage of API types within advices

apm-agent-plugins/apm-api-plugin/pom.xml
```diff
@@ -20,7 +20,7 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>apm-agent-api</artifactId>
             <version>${project.version}</version>
-            <scope>test</scope>
+            <scope>provided</scope>
```
eyalkoren, line 23: Let's add the new functionality without adding this dependency

eyalkoren (changes requested):
Thanks for the PR @tobiasstadler !

The API types are only known within the API itself and not elsewhere within the agent, including the API plugin. Let's keep it this way. We don't want version assumptions that may make the agent backwards incompatible. Even though it is protected within your implementation by the fact that the `getServiceInfoForClassLoader` and `setServiceInfoForClassLoader` methods won't be matched, allowing this dependency may bite us in the future.

There are ways to do that within the API (like we do with multiple APIs through usage of internal methods that use `Object` in their signature), but I think the simplest would be to lose the `ServiceInfo` type. 

One option is to get the service name and version through two APIs. It is not as neat, but it avoids the object allocation and it would be required within the internal API anyway if we want to keep the complete separation between API types and agent types.
Another option is to get rid of `getServiceInfoForClassLoader` altogether.
The fact that you added `setServiceInfoForClassLoader` is enough if the `co.elastic.apm.api.Transaction` interface has a `useServiceInfoOf(ClassLoader)` (instead of what you did in #2444) . You can reference from the javadoc of each method 
to the other to give users hints how they can be used. A `co.elastic.apm.api.Transaction#setServiceInfo(String serviceName, String serviceVersion)` is still useful as well.

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/TransactionInstrumentation.java
```diff
@@ -147,4 +148,41 @@ public static void addCustomContext(@Advice.FieldValue(value = "span", typing =
...
+        public SetServiceInfoInstrumentation() {
+            super(named("setServiceInfo"));
+        }
+
+        public static class AdviceClass {
+            @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
+            public static void setServiceInfo(@Advice.FieldValue(value = "span", typing = Assigner.Typing.DYNAMIC) Object transaction,
+                                              @Advice.Argument(0) String serviceName, @Advice.Argument(1) String serviceVersion) {
+                if (transaction instanceof Transaction) {
+                    ((Transaction) transaction).getTraceContext().setServiceName(serviceName);
```
tobiasstadler, line 162: I wonder if we should check for `serviceName` being `null`

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/ElasticApmApiInstrumentation.java
```diff
@@ -147,4 +148,16 @@ public static void captureException(@Advice.Origin Class<?> clazz, @Advice.Argum
...
 
+    public static class SetServiceInfoForClassLoaderInstrumentation extends ElasticApmApiInstrumentation {
+        public SetServiceInfoForClassLoaderInstrumentation() {
+            super(named("setServiceInfoForClassLoader"));
+        }
+
+        public static class AdviceClass {
+            @Advice.OnMethodExit(suppress = Throwable.class, inline = false)
+            public static void setServiceInfoForClassLoader(@Advice.Argument(0) @Nullable ClassLoader classLoader, @Advice.Argument(1) String serviceName, @Advice.Argument(2) @Nullable String serviceVersion) {
+                tracer.setServiceInfoForClassLoader(classLoader, ServiceInfo.of(serviceName, serviceVersion));
```
tobiasstadler, line 159: I wonder if we should check for `serviceName` being `null`

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
tobiasstadler: Done

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -561,7 +562,7 @@ private static void checkNotAgentType(TypeDescription.Generic type, String descr
         // We have to use 'raw' type as framework classes are not accessible to the boostrap classloader, and
         // trying to resolve them will create exceptions.
         String name = type.asRawType().getTypeName();
-        if (name.startsWith("co.elastic.apm")) {
+        if (name.startsWith("co.elastic.apm") && !name.startsWith("co.elastic.apm.api")) {
```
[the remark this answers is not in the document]
tobiasstadler: Done

apm-agent-plugins/apm-api-plugin/pom.xml
tobiasstadler: Done

apm-agent-api/src/main/java/co/elastic/apm/api/Transaction.java
tobiasstadler: Done

apm-agent-api/src/main/java/co/elastic/apm/api/ServiceInfo.java
tobiasstadler: Done

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
tobiasstadler: Done

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
tobiasstadler: Done

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/ElasticApmApiInstrumentation.java
eyalkoren: A `null` service name will have no effect here - it won't be mapped and not even cause allocation of a `ServiceInfo` object. 
However, I wonder if we should allow `null` service version going forward, now that #1726 is merged.
I think providing a service name without a version should be considered a bug, because it means we will use the global service version for a manually set service name, and there's no real reason why those would be related.
@felixbarny WDYT? Should we only allow overriding these two values together?

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/TransactionInstrumentation.java
eyalkoren: Not sure about this one...
If we allow `null` here, then what happens? If the `service_name` config was manually set, then it would have no effect, right? So this is only a way to use the somewhat useless fallback auto service name?
If so - we can check and reject null values here

docs/public-api.asciidoc
```diff
@@ -192,6 +192,19 @@ public Response onIncomingRequest(Request request) throws Exception {
 
 NOTE: If the protocol does not support multi-value headers, use <<api-start-transaction-with-remote-parent-header>>
 
+[float]
+[[api-set-service-info-for-class-loader]]
+==== `void setServiceInfoForClassLoader(ClassLoader, String, String)` added[1.29.0]
```
eyalkoren, line 197:
```suggestion
==== `void setServiceInfoForClassLoader(ClassLoader, String, String)` added[1.30.0]
```

docs/public-api.asciidoc
```diff
@@ -329,6 +342,27 @@ transaction.setFrameworkName("My Framework");
 
 * `frameworkName`: The name of the framework
 
+[float]
+[[api-transaction-set-service-info]]
+==== `Transaction setServiceInfo(String serviceName, String serviceVersion)` added[1.29.0]
```
eyalkoren, line 347:
```suggestion
==== `Transaction setServiceInfo(String serviceName, String serviceVersion)` added[1.30.0]
```

docs/public-api.asciidoc
```diff
@@ -329,6 +342,27 @@ transaction.setFrameworkName("My Framework");
...
+Sets the service name and version for this transaction and its child spans.
+
+NOTE: If this method is called after child spans are already created, they may have the wrong service name and version.
+
+* `serviceName`: the service name
+* `serviceVersion`: the service version
+
+[float]
+[[api-transaction-use-service-info-for-class-loader]]
+==== `Transaction useServiceInfoForClassLoader(ClassLoader classLoader)` added[1.29.0]
```
eyalkoren, line 357:
```suggestion
==== `Transaction useServiceInfoForClassLoader(ClassLoader classLoader)` added[1.30.0]
```

eyalkoren (commented):
Looks good, thanks!
Once we set our minds about how we handle `null` values, please update the tests accordingly to verify that.

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/ElasticApmApiInstrumentation.java
tobiasstadler: I think its is fine to set a service name without a version. The global service version won't be used in such a case. But there service version will be serialized as null and dropped when ingested by the apm-server

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/TransactionInstrumentation.java
tobiasstadler: if `serviceName` is null, but `serviceVersion` is not, `service.name` won't be serialized, but `service.version` will be. I am not sure what the apm-server does in such a case. But I suspect it will drop  transaction specific service configuration and use the global one.

tobiasstadler: I personally think both methods should return early (do no nothing) when `serviceName` is `null`.

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/ElasticApmApiInstrumentation.java
eyalkoren: You mean - currently APM Server ignores the global version when we use a service name in transaction/metric context?

tobiasstadler: Sorry, my bad. This logic is only implemented for metrics. I created elastic/apm-server#7281 to fix this for transactions also.

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/TransactionInstrumentation.java
felixbarny: We could do that centrally by creating a `setServiceInfo(String serviceName, String serviceVersion)` method in `TraceContext`.
Similar to the refactoring in https://github.com/elastic/apm-agent-java/pull/2444, `TraceContext#setServiceInfo` and `TraceContext#setApplicationClassLoader` would be the only methods that set the service name and version.

felixbarny (approved):
One optional suggestion to bring in some of the improvements from #2444. Otherwise, LGTM.

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/TransactionInstrumentation.java
tobiasstadler: Done!

tobiasstadler:
Thank You!

eyalkoren:
Thank you for your great collaboration ❤️

PR #2460, opened 2022-02-08
Setting agent CL as context CL for agent threads

eyalkoren:
## What does this PR do?
We never saw a need for that, but recently we got a report of a weird JULI bug in Tomcat 9, where it doesn't expect the context CL to be `null`:
```
[elastic-apm-metadata-1] WARN  co.elastic.apm.agent.impl.metadata.CloudMetadataProvider - Unexpected error during automatic discovery process for cloud provider
java.lang.NoClassDefFoundError: Could not initialize class sun.net.www.protocol.http.HttpURLConnection
        at sun.net.www.protocol.http.Handler.openConnection(Handler.java:62) ~[?:?]
        at ...
        at co.elastic.apm.agent.impl.metadata.CloudMetadataProvider.executeRequest(CloudMetadataProvider.java:418) ~
        ...
[elastic-apm-cloud-metadata-1] ERROR co.elastic.apm.agent.util.ExecutorUtils - Could not initialize class sun.net.www.protocol.http.HttpURLConnection
java.lang.NoClassDefFoundError: Could not initialize class sun.net.www.protocol.http.HttpURLConnection
        at sun.net.www.protocol.http.Handler.openConnection(Handler.java:62) ~[?:?]
        at ...
[elastic-apm-cloud-metadata-0] ERROR co.elastic.apm.agent.util.ExecutorUtils - null
java.lang.ExceptionInInitializerError: null
        at sun.net.www.protocol.http.Handler.openConnection(Handler.java:62) ~[?:?]
        at ...
Caused by: java.lang.NullPointerException
        at org.apache.juli.ClassLoaderLogManager.findProperty(ClassLoaderLogManager.java:311) ~[?:?]
        at org.apache.juli.ClassLoaderLogManager.getProperty(ClassLoaderLogManager.java:291) ~[?:?]
        at ...
        at sun.net.www.protocol.http.HttpURLConnection.<clinit>(HttpURLConnection.java:432) ~[?:?]
```

## Checklist

- [x] This is a bugfix

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/util/ExecutorUtils.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/util/ExecutorUtils.java
@@ -84,7 +84,7 @@ public final class ExecutorUtils {
             thread.setDaemon(true);
             thread.setName(threadName);
             ClassLoader originalContextCL = thread.getContextClassLoader();
-            thread.setContextClassLoader(null);
+            thread.setContextClassLoader(ExecutorUtils.class.getClassLoader());
             logThreadCreation(originalContextCL, threadName);
             return thread;
         }
@@ -116,7 +116,7 @@ public final class ExecutorUtils {
             String threadName = ThreadUtils.addElasticApmThreadPrefix(threadPurpose) + "-" + threadCounter.getAndIncrement();
             thread.setName(threadName);
             ClassLoader originalContextCL = thread.getContextClassLoader();
-            thread.setContextClassLoader(null);
+            thread.setContextClassLoader(ExecutorUtils.class.getClassLoader());
             logThreadCreation(originalContextCL, threadName);
             return thread;
         }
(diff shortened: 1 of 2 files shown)


jackshirazi:
wasn't this (null setting) to avoid warnings from tomcat about "possible memory leak"? I have an open PR https://github.com/apache/tomcat/pull/464 to fix on their side, but seems to be very low priority

SylvainJuge (approved):
LGTM, simple and straightforward

SylvainJuge:
It sounded familiar, but I forgot to search for a past issue: it seems that https://github.com/elastic/apm-agent-java/pull/1459 was exactly doing the opposite.

eyalkoren:
Based on https://github.com/apache/tomcat/blame/main/java/org/apache/catalina/loader/WebappClassLoaderBase.java#L1802, the warning and automatic thread stopping is only being executed for threads of which context CL is the `WebappClassLoader` being stopped, so setting the context CL of our threads to agent CL may work as well as setting it to `null`. WDYT?

eyalkoren:
I tested on Tomcat 9.0.58 and everything seems to be working as expected, including when stopping and restarting a webapp multiple times.
I was not able to reproduce these Tomcat warnings when not touching the default agent threads' context CL though as well. The reason for that is that I didn't encounter any case where a thread is started lazily through a web application CL, so the default context CL for all threads I saw being created were the system CL.
I think we'll go with that indeed.

PR #2467, opened 2022-02-14
Updated AWS lambda docs to use ARN for layer installation

AlexanderWert:


docs/setup-aws-lambda.asciidoc
```diff
@@ -49,41 +39,42 @@ explicitly.
...
+
+1. Pick the right ARN from the https://github.com/elastic/apm-agent-java/releases[ARN table for the APM Java Agent Lambda Layer]. Note: The APM Java Agent Layer needs to be in the _same region_ as your Lambda function.
+2. Navigate to your function in the AWS Console
+3. Scroll to the Layers section and click the _Add a layer_ button
+4. Choose the _Specify an ARN_ radio button
+5. Enter the Version ARN of the APM Java Agent Layer in the _Specify an ARN_ text input
+6. Click the _Add_ button
+
+Once the APM Java Agent Layer is in place, you'll need to configure some environment variables _in addition_ to the environment variables you already configured for the APM Lambda extension. 
+Configure the following environment variables for the APM Java Agent Layer:
```
eyalkoren, line 73: I don't think it is better than what we had before:
```
NOTE: Not all environment variables specified within The Necessary Variables
section of the extension documentation need to be configured. The Java agent 
utilizes a wrapper script to set some of these up, so follow the next section to 
see which are required.
```

The AWS Lambda extension [docs](https://www.elastic.co/guide/en/apm/guide/current/aws-lambda-extension.html#aws-lambda-variables) include env variables that are not required for Java. 
This section even starts with a note:
```
The necessary environment variables depend on the APM agent being used. Follow the 
Node.js agent setup guide, Python agent setup guide or Java agent setup guide, r
espectively, for specific instructions on setting the environment variables.
```
But now we don't list the minimal required env variables

eyalkoren: I just saw https://github.com/elastic/apm-aws-lambda/pull/122 (the online docs don't include it yet).
So I guess we moved this entire section about - not all agents require the same variables?
What is telling our users that the layer script sets some of the env variables but not others?

AlexanderWert: Yes, the extension docs only contain the extension-specific configs now. And all of the agents are expected to set meaningful default for most of the config options, so no need to document it there. 

With that we can focus in the Java docs on the Java-specific env vars.

eyalkoren: OK, I reviewed in context of the updated extension docs, makes sense. Thanks

Issue #2084, opened 2021-08-25
[META 497] Collecting statistics about dropped spans


AlexanderWert:
As described in [this spec](https://github.com/elastic/apm/blob/master/specs/agents/handling-huge-traces/tracing-spans-dropped-stats.md), agents should collect statistics on dropped spans.

Meta issue: [![issue details](https://img.shields.io/endpoint?url=https%3A%2F%2Fgiss.app.elstc.co%2Fapi%2Fstatus%2Felastic%2Fapm%2F497)](https://github.com/elastic/apm/issues/497)

Issue #2083, opened 2021-08-25
[META 496] Dropping fast exit spans


AlexanderWert:
As described in the [Dropping fast exit spans](https://github.com/elastic/apm/blob/master/specs/agents/handling-huge-traces/tracing-spans-drop-fast-exit.md) spec, agents should introduce the config option `exit_span_min_duration` that defines the minimal duration for exit spans to not being dropped.

The implementation should be aligned with the overall approach described for [Handling huge traces](https://github.com/elastic/apm/tree/master/specs/agents/handling-huge-traces).

Meta issue: [![issue details](https://img.shields.io/endpoint?url=https%3A%2F%2Fgiss.app.elstc.co%2Fapi%2Fstatus%2Felastic%2Fapm%2F496)](https://github.com/elastic/apm/issues/496)

PR #2477, opened 2022-02-18
Added (initial) support for compressing spans

### LINKED ISSUE #1847: [META 432] Implement compressed spans algorithm
Implement the compressed spans algorithm in the agent as described in https://github.com/elastic/apm/issues/432.

## Status: 

[issue details](https://github.com/elastic/apm-agent-java/issues/1847)

tobiasstadler:
## What does this PR do?
Fixes #1847 

## Checklist

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -714,6 +714,36 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
         .dynamic(false)
         .buildWithDefault(false);
 
+    private final ConfigurationOption<Boolean> spanCompressionEnabled = ConfigurationOption.booleanOption()
+        .key("span_compression_enabled")
+        .configurationCategory(CORE_CATEGORY)
+        .tags("added[1.30.0]", "internal")
+        .description("Setting this option to true will enable span compression feature.\n" +
+            "Span compression reduces the collection, processing, and storage overhead, and removes clutter from the UI. " +
+            "The tradeoff is that some information such as DB statements of all the compressed spans will not be collected.")
+        .dynamic(true)
+        .buildWithDefault(false);
+
+    private final ConfigurationOption<TimeDuration> spanCompressionExactMatchMaxDuration = TimeDurationValueConverter.durationOption("ms")
+        .key("span_compression_exact_match_max_duration")
+        .configurationCategory(CORE_CATEGORY)
+        .tags("added[1.30.0]", "internal")
+        .description("Consecutive spans that are exact match and that are under this threshold will be compressed into a single composite span. " +
+            "This option does not apply to composite spans. This reduces the collection, processing, and storage overhead, and removes clutter from the UI. " +
+            "The tradeoff is that the DB statements of all the compressed spans will not be collected.")
+        .dynamic(true)
+        .buildWithDefault(TimeDuration.of("50ms"));
+
+    private final ConfigurationOption<TimeDuration> spanCompressionSameKindMaxDuration = TimeDurationValueConverter.durationOption("ms")
+        .key("span_compression_same_kind_max_duration")
+        .configurationCategory(CORE_CATEGORY)
+        .tags("added[1.30.0]", "internal")
+        .description("Consecutive spans to the same destination that are under this threshold will be compressed into a single composite span. " +
+            "This option does not apply to composite spans. This reduces the collection, processing, and storage overhead, and removes clutter from the UI. " +
+            "The tradeoff is that the DB statements of all the compressed spans will not be collected.")
+        .dynamic(true)
+        .buildWithDefault(TimeDuration.of("5ms"));
+
     public boolean isEnabled() {
         return enabled.get();
     }
@@ -941,6 +971,18 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
 
     public boolean isEnablePublicApiAnnotationInheritance() {
         return enablePublicApiAnnotationInheritance.get();
+    }
+
+    public boolean isSpanCompressionEnabled() {
+        return spanCompressionEnabled.get();
+    }
+
+    public TimeDuration getSpanCompressionExactMatchMaxDuration() {
+        return spanCompressionExactMatchMaxDuration.get();
+    }
+
+    public TimeDuration getSpanCompressionSameKindMaxDuration() {
+        return spanCompressionSameKindMaxDuration.get();
     }
 
     public enum EventType {
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -714,36 +714,6 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
         .dynamic(false)
         .buildWithDefault(false);
 
-    private final ConfigurationOption<Boolean> spanCompressionEnabled = ConfigurationOption.booleanOption()
-        .key("span_compression_enabled")
-        .configurationCategory(CORE_CATEGORY)
-        .tags("added[1.30.0]", "internal")
-        .description("Setting this option to true will enable span compression feature.\n" +
-            "Span compression reduces the collection, processing, and storage overhead, and removes clutter from the UI. " +
-            "The tradeoff is that some information such as DB statements of all the compressed spans will not be collected.")
-        .dynamic(true)
-        .buildWithDefault(false);
-
-    private final ConfigurationOption<TimeDuration> spanCompressionExactMatchMaxDuration = TimeDurationValueConverter.durationOption("ms")
-        .key("span_compression_exact_match_max_duration")
-        .configurationCategory(CORE_CATEGORY)
-        .tags("added[1.30.0]", "internal")
-        .description("Consecutive spans that are exact match and that are under this threshold will be compressed into a single composite span. " +
-            "This option does not apply to composite spans. This reduces the collection, processing, and storage overhead, and removes clutter from the UI. " +
-            "The tradeoff is that the DB statements of all the compressed spans will not be collected.")
-        .dynamic(true)
-        .buildWithDefault(TimeDuration.of("50ms"));
-
-    private final ConfigurationOption<TimeDuration> spanCompressionSameKindMaxDuration = TimeDurationValueConverter.durationOption("ms")
-        .key("span_compression_same_kind_max_duration")
-        .configurationCategory(CORE_CATEGORY)
-        .tags("added[1.30.0]", "internal")
-        .description("Consecutive spans to the same destination that are under this threshold will be compressed into a single composite span. " +
-            "This option does not apply to composite spans. This reduces the collection, processing, and storage overhead, and removes clutter from the UI. " +
-            "The tradeoff is that the DB statements of all the compressed spans will not be collected.")
-        .dynamic(true)
-        .buildWithDefault(TimeDuration.of("5ms"));
-
     public boolean isEnabled() {
         return enabled.get();
     }
@@ -971,18 +941,6 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
 
     public boolean isEnablePublicApiAnnotationInheritance() {
         return enablePublicApiAnnotationInheritance.get();
-    }
-
-    public boolean isSpanCompressionEnabled() {
-        return spanCompressionEnabled.get();
-    }
-
-    public TimeDuration getSpanCompressionExactMatchMaxDuration() {
-        return spanCompressionExactMatchMaxDuration.get();
-    }
-
-    public TimeDuration getSpanCompressionSameKindMaxDuration() {
-        return spanCompressionSameKindMaxDuration.get();
     }
 
     public enum EventType {
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package co.elastic.apm.agent.configuration;
+
+import co.elastic.apm.agent.configuration.converter.TimeDuration;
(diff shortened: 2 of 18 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -714,6 +714,36 @@ public String toString(Collection<String> value) {
         .dynamic(false)
         .buildWithDefault(false);
 
+    private final ConfigurationOption<Boolean> spanCompressionEnabled = ConfigurationOption.booleanOption()
+        .key("span_compression_enabled")
+        .configurationCategory(CORE_CATEGORY)
+        .tags("added[1.30.0]", "internal")
```
tobiasstadler, line 720: Until #2083 and #2084 are done

felixbarny:
Thanks for your PR! Note that @jackshirazi is already working on this, so please coordinate before proceeding.

@AlexanderWert has now implemented a status indicator in the corresponding issue (#1847) so that it's easier to spot if something is in progress already.

tobiasstadler:
@jackshirazi You pinged the correct person

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+import co.elastic.apm.agent.configuration.converter.TimeDuration;
+import co.elastic.apm.agent.configuration.converter.TimeDurationValueConverter;
+import org.stagemonitor.configuration.ConfigurationOption;
+import org.stagemonitor.configuration.ConfigurationOptionProvider;
+
+public class SpanConfiguration extends ConfigurationOptionProvider {
+
+    public static final String HUGE_TRACES_CATEGORY = "Huge Traces";
+
+    private final ConfigurationOption<Boolean> spanCompressionEnabled = ConfigurationOption.booleanOption()
```
tobiasstadler, line 30: Maybe this should be cached in the transaction to provide a consistent compression

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+        .description("Setting this option to true will enable span compression feature.\n" +
+            "Span compression reduces the collection, processing, and storage overhead, and removes clutter from the UI. " +
+            "The tradeoff is that some information such as DB statements of all the compressed spans will not be collected.")
+        .dynamic(true)
+        .buildWithDefault(false);
+
+    private final ConfigurationOption<TimeDuration> spanCompressionExactMatchMaxDuration = TimeDurationValueConverter.durationOption("ms")
+        .key("span_compression_exact_match_max_duration")
+        .configurationCategory(HUGE_TRACES_CATEGORY)
+        .tags("added[1.30.0]", "internal")
```
tobiasstadler, line 43: Maybe this should be cached in the transaction to provide a consistent compression

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+        .description("Consecutive spans that are exact match and that are under this threshold will be compressed into a single composite span. " +
+            "This option does not apply to composite spans. This reduces the collection, processing, and storage overhead, and removes clutter from the UI. " +
+            "The tradeoff is that the DB statements of all the compressed spans will not be collected.")
+        .dynamic(true)
+        .buildWithDefault(TimeDuration.of("50ms"));
+
+    private final ConfigurationOption<TimeDuration> spanCompressionSameKindMaxDuration = TimeDurationValueConverter.durationOption("ms")
```
tobiasstadler, line 50: Maybe this should be cached in the transaction to provide a consistent compression

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -292,13 +304,123 @@ public void beforeEnd(long epochMicros) {
...
+                    if (parent.bufferedSpan.compareAndSet(buffered, null)) {
+                        this.tracer.endSpan(buffered);
+                    }
+                }
+                this.tracer.endSpan(this);
+                return;
+            }
+            if (buffered == null) {
+                if (!parent.bufferedSpan.compareAndSet(null, this)) {
+                    this.tracer.endSpan(this);
```
tobiasstadler, line 320: We might miss an opportunity to compress, but in my opinion this special case is not with the effort.

tobiasstadler, line 320: We might miss an opportunity to compress, but in my opinion this special case is not with the effort.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -292,13 +304,123 @@ public void beforeEnd(long epochMicros) {
...
+                }
+                return;
+            }
+            if (!buffered.tryToCompress(this)) {
+                if (!parent.bufferedSpan.compareAndSet(buffered, this)) {
+                    this.tracer.endSpan(buffered);
+                    this.tracer.endSpan(this);
```
tobiasstadler, line 327: We might miss an opportunity to compress, but in my opinion this special case is not with the effort.

jackshirazi: I agree, but worth pointing out it's a deliberate choice with a comment like "the failed update would ideally lead to a compression attempt with the new buffer, but we're dropping the compression attempt to keep this simple"

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
tobiasstadler: Ok, I added a (your) comment

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -292,13 +303,133 @@ public void beforeEnd(long epochMicros) {
...
+                    // but we're dropping the compression attempt to keep things simple
+                    this.tracer.endSpan(this);
+                }
+                return;
+            }
+            if (!buffered.tryToCompress(this)) {
+                if (!parent.bufferedSpan.compareAndSet(buffered, this)) {
+                    // the failed update would ideally lead to a compression attempt with the new buffer,
+                    // but we're dropping the compression attempt to keep things simple
+                    this.tracer.endSpan(buffered);
```
jackshirazi, line 329: the buffer is still set to `buffered` here, but it's being reported, I'm think that leads to multiple reports of the same span?

tobiasstadler: I think you are (partly) right here. If we fail to update the buffer, the buffer is already set to something different than `buffered`. So I don't think `buffered` will be reported multiple times (in the future). But as the buffer is set to something different than `buffered`, `buffered` should already have been reported. So  I think `buffered` will be reported for the second time here.

@jackshirazi Do you agree/disagree?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -292,13 +303,133 @@ public void beforeEnd(long epochMicros) {
...
+            if (buffered == null) {
+                if (!parent.bufferedSpan.compareAndSet(null, this)) {
+                    // the failed update would ideally lead to a compression attempt with the new buffer,
+                    // but we're dropping the compression attempt to keep things simple
+                    this.tracer.endSpan(this);
+                }
+                return;
+            }
+            if (!buffered.tryToCompress(this)) {
+                if (!parent.bufferedSpan.compareAndSet(buffered, this)) {
```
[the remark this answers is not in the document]
tobiasstadler: Will do that

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -292,13 +303,132 @@ public void beforeEnd(long epochMicros) {
...
+
+        return true;
+    }
+
+    private boolean tryToCompressRegular(Span sibling) {
+        if (!isSameKind(sibling)) {
+            return false;
+        }
+
+        long currentDuration = duration.get();
```
jackshirazi, line 377: move down to after the immediately following clause as it's not needed until then

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/transaction/SpanCompressionIT.java
```diff
@@ -0,0 +1,135 @@
...
+    void resetReporter() {
+        reporter.reset();
+    }
+
+    @AfterAll
+    static void shutdownExecutor() {
+        executor.shutdown();
+    }
+
+    @RepeatedTest(100)
```
tobiasstadler, line 74: I run 100000 iterations on my machine without any failures

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/transaction/SpanCompressionIT.java
```diff
@@ -0,0 +1,135 @@
...
+    void testParallelExitSpanCreation() {
+        runInTransactionScope((transaction, i) -> {
+            return () -> createExitSpan(transaction, i, 1000L + i);
+        });
+
+        assertReportedSpans(reporter.getSpans());
+    }
+
+    @RepeatedTest(100)
```
tobiasstadler, line 83: I run 100000 iterations on my machine without any failures

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/transaction/SameKindCompressionStrategyTest.java
```diff
@@ -0,0 +1,38 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package co.elastic.apm.agent.impl.transaction;
+
+class SameKindCompressionStrategyTest extends AbstractCompressionStrategyTest {
+
+    private int spansCreated = 0;
```
jackshirazi, line 23: should this be static?

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/transaction/SpanCompressionIT.java
jackshirazi: nice

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/transaction/SpanCompressionIT.java
jackshirazi: nice

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/transaction/SameKindCompressionStrategyTest.java
tobiasstadler: My intention was that each test start with has the same sequence of span names. But since the span name doesn't really matter (only that they are different), I can make it static if you like.

jackshirazi: leave it, the scenario I'm thinking of - maintenance error with reused instance - should be caught by review

jackshirazi (approved):
Great work, thank you!

jackshirazi:

tobiasstadler:
Thank You!

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -292,13 +303,146 @@ public void beforeEnd(long epochMicros) {
 
     @Override
     protected void afterEnd() {
-        this.tracer.endSpan(this);
+        if (transaction != null && transaction.isSpanCompressionEnabled() && parent != null) {
+            Span buffered = parent.bufferedSpan.get();
```
felixbarny, line 307: Seems like this may lead to multiple threads concurrently having access to the buffered span instance.
In my initial POC for span compression, I tried to avoid that by atomically getting and removing the buffered span or setting it to the current span if there's no buffered span.

https://github.com/felixbarny/apm-agent-java/blob/d15716baab62eaa8ff7b677da704d9c8d780d285/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java#L190-L219

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
felixbarny: While I agree that this is an edge case, it's also handled in my POC implementation linked above.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -292,13 +303,146 @@ public void beforeEnd(long epochMicros) {
...
+        } else {
+            this.tracer.endSpan(this);
+        }
+    }
+
+    private boolean isCompressionEligible() {
+        return isExit() && isDiscardable() && (outcomeNotSet() || getOutcome() == Outcome.SUCCESS);
+    }
+
+    private boolean tryToCompress(Span sibling) {
```
felixbarny, line 349: Seems like this method could also be simplified when we can guarantee that we have exclusive access to both `this` span and the sibling. (see above linked POC)

jackshirazi: @felixbarny I agree your solution was the more elegant one. I decided that in the high contention case (which is really when these alternatives matter) we want to avoid the backoff and retry loop, ie that under conflict it's better to drop the compression and allow the thread to proceed asap rather than maximally try to compress. So felt this solution was acceptable

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
jackshirazi: Multiple threads can concurrently have access, that's okay here. It is resolved at the update or report stage which you can fully reason about here (nice when it's concurrent). There are 3 possible atomic concurrent updates, and in each case either it succeeds (true) or fails (false). The cases are complete

```
buffer -> null
true: report buffer
false: do nothing

null -> this
true: do nothing
false: report this

buffer -> this
true: report buffer
false: report this
```

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
jackshirazi: note the loops in the try are getting maxes after compression has succeeded, so should not really cause contention, but we may need to review

felixbarny: Thanks for explaining, makes sense. I agree that the approach taken in this PR is probably the one that causes less contention. I find it a little harder to reason about and verify correctness as there's more that can happen concurrently. But from what I can tell, it seems like all the cases are handled properly.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
tobiasstadler: Yes, concurrent access is fine here and also covered by tests.

PR #2478, opened 2022-02-20
Restricting Logback ECS-reformatting instrumentation to version 1.0.0

eyalkoren:
In order to implement that, I added a fallback to look at the jar manifest's `Bundle-Version` if the `Implementation-Version` is unavailable


eyalkoren:
Not required

PR #2491, opened 2022-03-03
Added support for dropping fast exit spans

### LINKED ISSUE #2083: [META 496] Dropping fast exit spans

tobiasstadler:
## What does this PR do?
Fixes #2083 

## Checklist

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -32,6 +32,7 @@ endif::[]
 * Added support for setting service name and version for a transaction via the public api - {pull}2451[#2451]
 * Added support for en-/disabling each public annotation on each own - {pull}2472[#2472]
 * Added support for compressing spans - {pull}2477[#2477]
+* Added microsecond durations with `us` as unit - {pull}2496[#2496]
 * Added support for dropping fast exit spans - {pull}2491[#2491]
 
 [float]
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -32,6 +32,7 @@ endif::[]
 * Added support for setting service name and version for a transaction via the public api - {pull}2451[#2451]
 * Added support for en-/disabling each public annotation on each own - {pull}2472[#2472]
 * Added support for compressing spans - {pull}2477[#2477]
+* Added support for dropping fast exit spans - {pull}2491[#2491]
 
 [float]
 ===== Performance improvements
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
@@ -65,7 +65,7 @@ public class SpanConfiguration extends ConfigurationOptionProvider {
             "\n" +
             "NOTE: If a span propagates distributed tracing ids, it will not be ignored, even if it is shorter than the configured threshold. This is to ensure that no broken traces are recorded.")
         .dynamic(true)
-        .buildWithDefault(TimeDuration.of("1ms"));
+        .buildWithDefault(TimeDuration.ofFine("1ms"));
 
     public boolean isSpanCompressionEnabled() {
         return spanCompressionEnabled.get();
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
@@ -57,7 +57,7 @@ public class SpanConfiguration extends ConfigurationOptionProvider {
         .dynamic(true)
         .buildWithDefault(TimeDuration.of("5ms"));
 
-    private final ConfigurationOption<TimeDuration> exitSpanMinDuration = TimeDurationValueConverter.durationOption("ms")
+    private final ConfigurationOption<TimeDuration> exitSpanMinDuration = TimeDurationValueConverter.fineDurationOption("ms")
         .key("exit_span_min_duration")
         .tags("added[1.30.0]")
         .configurationCategory(HUGE_TRACES_CATEGORY)
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
@@ -57,7 +57,7 @@ public class SpanConfiguration extends ConfigurationOptionProvider {
         .dynamic(true)
         .buildWithDefault(TimeDuration.of("5ms"));
 
-    private final ConfigurationOption<TimeDuration> exitSpanMinDuration = TimeDurationValueConverter.fineDurationOption("ms")
+    private final ConfigurationOption<TimeDuration> exitSpanMinDuration = TimeDurationValueConverter.fineDurationOption()
         .key("exit_span_min_duration")
         .tags("added[1.30.0]")
         .configurationCategory(HUGE_TRACES_CATEGORY)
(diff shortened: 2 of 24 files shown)


CHANGELOG.asciidoc
```diff
@@ -32,6 +32,7 @@ endif::[]
 * Added support for setting service name and version for a transaction via the public api - {pull}2451[#2451]
 * Added support for en-/disabling each public annotation on each own - {pull}2472[#2472]
 * Added support for compressing spans - {pull}2477[#2477]
+* Added support for dropping fast exit spans - {pull}2491[#2491]
```
tobiasstadler, line 35: Maybe this should be listed in `Potentially breaking changes`?

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
```diff
@@ -57,6 +57,16 @@
         .dynamic(true)
         .buildWithDefault(TimeDuration.of("5ms"));
 
+    private final ConfigurationOption<TimeDuration> exitSpanMinDuration = TimeDurationValueConverter.durationOption("ms")
```
jackshirazi, line 60:
```suggestion
    private final ConfigurationOption<TimeDuration> exitSpanMinDuration = TimeDurationValueConverter.fineDurationOption("ms")
```

tobiasstadler: Hm, I thought I had done that already

tobiasstadler: Ah, I only did the `buildWithDefault`. Sorry!

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
```diff
@@ -57,6 +57,16 @@
         .dynamic(true)
         .buildWithDefault(TimeDuration.of("5ms"));
 
+    private final ConfigurationOption<TimeDuration> exitSpanMinDuration = TimeDurationValueConverter.fineDurationOption("ms")
```
felixbarny, line 60: We're trying to get rid of default units. The reason they still exit is backward compatibility. For new options, please don't specify a default duration suffix.
Unfortunately, this is not really reflected in `TimeDurationValueConverter`. I think we should change `withDefaultFineDuration` and `fineDurationOption` so that they don't accept a default duration and provide an empty string to the `TimeDurationValueConverter` constructor. `TimeDurationValueConverter#withDefaultDuration(String)` and `durationOption(String) should be deprecated and we should create overloads that don't accept a default duration.

tobiasstadler: Ok, should I do this in this PR or should it be a separate PR?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -383,20 +384,27 @@ public void endSpan(Span span) {
             span.decrementReferences();
             return;
         }
-        if (!span.isComposite()) {
+        SpanConfiguration spanConfiguration = getConfig(SpanConfiguration.class);
+        if (span.isExit()) {
+            if (span.getDuration() < spanConfiguration.getExitSpanMinDuration().getMicros()) {
```
jackshirazi, line 389: Does this handle the composite span case where the span is less than the config but the composite span is larger?

tobiasstadler: yes, because `duration = start - end` and start/end is the min/max value of all spans that compose the composite span

jackshirazi: excellent, thanks

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -383,20 +384,27 @@ public void endSpan(Span span) {
             span.decrementReferences();
             return;
         }
-        if (!span.isComposite()) {
+        SpanConfiguration spanConfiguration = getConfig(SpanConfiguration.class);
```
jackshirazi, line 387: why does this have wider scope than the following if()?

tobiasstadler: No specific reason. I will change it.

CHANGELOG.asciidoc
jackshirazi: We've decided to reverse the default on/off, so default should be 0ms, which means it's not a breaking change ... and sorry but you have a few changes needed to flip that for the config and the tests. We'll change the spec soon

tobiasstadler: Ok, can I do the changes now or should I wait until the spec is officially changed?

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/SpanConfiguration.java
tobiasstadler: @felixbarny #2503

CHANGELOG.asciidoc
tobiasstadler: Done

jackshirazi (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -383,20 +384,27 @@ public void endSpan(Span span) {
             span.decrementReferences();
             return;
         }
-        if (!span.isComposite()) {
+        if (span.isExit()) {
+            SpanConfiguration spanConfiguration = getConfig(SpanConfiguration.class);
```
felixbarny, line 388: Store spanConfiguration in a field so that we don't have to look up the config on each span end.

tobiasstadler: Should I do the same for `coreConfiguration.getSpanMinDuration()`?

felixbarny: It's not strictly necessary but a field lookup is cheaper than a map lookup and it does not increase the complexity in any way. This was not meant as a blocker to this PR but a suggestion for a minor improvement 🙂 

The `coreConfiguration` is already a field, so no changes needed there, I think.

jackshirazi: what looks a little more worrying is that the map is a hashmap rather than a concurrent map. It looks like the registration is all done first before accessing, so this should be fine but I'm not sure it's guaranteed

jackshirazi: @tobiasstadler may as well make the 2 configs consistent, ie both fields, before we merge this

tobiasstadler: Done

felixbarny (approved):

tobiasstadler:
Thank You!

PR #2496, opened 2022-03-04
Add micros time duration

jackshirazi:
## What does this PR do?
https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-drop-fast-exit.md requires us to support microsecond time durations with `us` as the unit, so that's added here

## Checklist
- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/converter/TimeDuration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/converter/TimeDuration.java
@@ -23,14 +23,14 @@ import java.util.regex.Pattern;
 
 public class TimeDuration implements Comparable<TimeDuration> {
 
-    public static final Pattern DURATION_PATTERN = Pattern.compile("^(-)?(\\d+)(ms|s|m)$");
+    public static final Pattern DURATION_PATTERN = Pattern.compile("^(-)?(\\d+)(us|ms|s|m)$");
     private final String durationString;
 
-    private final long durationMs;
+    private final long durationMicros;
 
-    private TimeDuration(String durationString, long durationMs) {
+    private TimeDuration(String durationString, long durationMicros) {
         this.durationString = durationString;
-        this.durationMs = durationMs;
+        this.durationMicros = durationMicros;
     }
 
     public static TimeDuration of(String durationString) {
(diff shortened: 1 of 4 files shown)


felixbarny:
There's one potential compatibility issue with adding `us` support to all the existing config options: When setting an option through central config with a `us` suffix that applies to both new and old agents, some may not recognize the `us` unit and fall back to the default value.

That's why there's a dedicated [`GranularDuration`](https://github.com/elastic/apm/blob/main/specs/agents/configuration.md#configuration-value-types) unit that only applies to `exit_span_min_duration`. The other options also don't seem like they'd benefit from being able to specify durations <1ms.

SylvainJuge (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/converter/TimeDuration.java
```diff
@@ -48,23 +48,25 @@ public static TimeDuration of(String durationString) {
...
-                return 1000 * 60;
+                return 1000 * 1000 * 60;
             default:
                 throw new IllegalStateException("Duration unit '" + unit + "' is unknown");
         }
     }
 
     public long getMillis() {
-        return durationMs;
+        return durationMicros/1000;
```
tobiasstadler, line 65: `TimeDuration.of("999us").getMillis()` will be `0`. I don't think this is the expected behavior.

tobiasstadler: I also stumbled upon this problem when implementing #2491. Unfortunately, I didn't find a good solution so far.

jackshirazi: personally I think that's correct in this situation as these are for configurations rather than statistics

jackshirazi: We're aiming to treat these differently with https://github.com/elastic/apm-agent-java/pull/2497 so you can only use microseconds if those are explicitly supported by the config - which for now will only be `exit_span_min_duration`

tobiasstadler: Ok, understood. I will adapt my PR once https://github.com/elastic/apm-agent-java/pull/2497 is merged.

PR #2505, opened 2022-03-09
Added support for collecting statistics about dropped exit spans

### LINKED ISSUE #2084: [META 497] Collecting statistics about dropped spans

tobiasstadler:
## What does this PR do?
Fixes #2084

## Checklist

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/DroppedSpanStats.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/DroppedSpanStats.java
@@ -18,7 +18,11 @@
  */
 package co.elastic.apm.agent.impl.transaction;
 
+import co.elastic.apm.agent.objectpool.Allocator;
+import co.elastic.apm.agent.objectpool.ObjectPool;
 import co.elastic.apm.agent.objectpool.Recyclable;
+import co.elastic.apm.agent.objectpool.impl.QueueBasedObjectPool;
+import org.jctools.queues.atomic.MpmcAtomicArrayQueue;
 
 import java.util.Iterator;
 import java.util.Map;
@@ -28,15 +32,20 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
-public class DroppedSpanStats implements Iterable<Map.Entry<DroppedSpanStats.StatsKey, DroppedSpanStats.StatsValue>>, Recyclable {
+public class DroppedSpanStats implements Iterable<Map.Entry<DroppedSpanStats.StatsKey, DroppedSpanStats.Stats>>, Recyclable {
 
-    public static class StatsKey {
-        private final String destinationServiceResource;
-        private final Outcome outcome;
+    public static class StatsKey implements Recyclable {
+        private String destinationServiceResource;
+        private Outcome outcome;
 
-        public StatsKey(CharSequence destinationServiceResource, Outcome outcome) {
+        public StatsKey() {
+
+        }
+
+        public StatsKey init(CharSequence destinationServiceResource, Outcome outcome) {
             this.destinationServiceResource = destinationServiceResource.toString();
             this.outcome = outcome;
+            return this;
         }
 
         public String getDestinationServiceResource() {
@@ -45,6 +54,12 @@ public class DroppedSpanStats implements Iterable<Map.Entry<DroppedSpanStats.Sta
 
         public Outcome getOutcome() {
             return outcome;
+        }
+
+        @Override
+        public void resetState() {
+            destinationServiceResource = null;
+            outcome = null;
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/DroppedSpanStats.java
```diff
@@ -99,33 +136,39 @@ public void captureDroppedSpan(Span span) {
         stats.sum.addAndGet(span.getDuration());
     }
 
-    private StatsValue getStats(StatsKey statsKey) {
-        StatsValue statsValue = statsMap.get(statsKey);
-        if (statsValue != null) {
-            return statsValue;
+    private Stats getOrCreateStats(StatsKey statsKey) {
```
jackshirazi, line 139: I'm not sure you need the synchronized block, for example I think this works equally without any lock
```
    private Stats getOrCreateStats(StatsKey statsKey) {
        Stats stats = statsMap.get(statsKey);
        if (stats != null || statsMap.size() > 127) {
            statsKeyObjectPool.recycle(statsKey);
            return stats;
        }

        Stats newStats = statsObjectPool.createInstance();
        stats = statsMap.putIfAbsent(statsKey, newStats);
        if (stats == null) {
            return newStats;
        } else {
            statsKeyObjectPool.recycle(statsKey);
            statsObjectPool.recycle(newStats);
            return stats;
        }
    }
```
this doesn't guarantee the 128 limit, but the limit is arbitrary anyway, so exceeding it a little doesn't really matter

tobiasstadler: Yes, Ia also wanted to do something like this, but came to the conclusion that we might exit the limit.

tobiasstadler: I guess most applications won't hit the limit.

jackshirazi: The limit is an arbitrary number to stop it being something that could grow indefinitely in memory. Just in case the limit is expected on the server too in future (it isn't now), you can limit the JSON write loop to stop at 128

tobiasstadler: so, should we enforce the limit now (when serializing the data)?

jackshirazi: yes. may as well

tobiasstadler: Done

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/DroppedSpanStats.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        @Override
+        public void resetState() {
+            count.set(0);
+            sum.set(0L);
+        }
+    }
+
+    private static final ObjectPool<StatsKey> statsKeyObjectPool = QueueBasedObjectPool.<StatsKey>ofRecyclable(new MpmcAtomicArrayQueue<StatsKey>(512), false, new Allocator<StatsKey>() {
```
tobiasstadler, line 98: is 512 a reasonable default?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/DroppedSpanStats.java
```diff
@@ -0,0 +1,171 @@
...
+        @Override
+        public StatsKey createInstance() {
+            return new StatsKey();
+        }
+    });
+
+    private static ObjectPool<Stats> statsObjectPool = QueueBasedObjectPool.<Stats>ofRecyclable(new MpmcAtomicArrayQueue<Stats>(512), false, new Allocator<Stats>() {
```
tobiasstadler, line 105: is 512 a reasonable default?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/DroppedSpanStats.java
jackshirazi: I think that the majority of applications will be using the default values and not very large number of spans in transactions. So 512 is fine for these. A minority will drop a lot and might need a larger value for optimum recycling. If we get complaints in future about too much garbage with a high level of dropped spans, we might need to expose the value as config, but for now this is fine

jackshirazi (approved):

jackshirazi:

tobiasstadler:
Thank You!

PR #2508, opened 2022-03-10
[debug] vertx debugging

SylvainJuge:
Add debug statements to help investigation with Vert.x issue.


apm-agent-plugins/apm-vertx/apm-vertx3-plugin/src/main/java/co/elastic/apm/agent/vertx/v3/web/http1/HttpServerRequestImplStartInstrumentation.java
```diff
@@ -70,10 +77,14 @@ public static Object enter(@Advice.This HttpServerRequestImpl request) {
...
-                Transaction transaction = (Transaction) transactionObj;
-                transaction.captureException(thrown).deactivate();
+            if (!(transactionObj instanceof Transaction)) {
+                return;
             }
+            Transaction transaction = (Transaction) transactionObj;
+            transaction.captureException(thrown).deactivate();
+
+            HttpServerResponseImpl response = request.response();
+            response.endHandler(new ResponseEndHandlerWrapper(transaction, response, request));
```
SylvainJuge, line 87: this replaces the constructor instrumentation as we can access the response directly once the original method body has been executed.

SylvainJuge:
Notes on current state of this PR:

Trying to avoid the instrumentation of `io.vertx.core.http.impl.HttpServerResponseImpl` is definitely challenging, while it is always called from within `io.vertx.core.http.impl.HttpServerRequestImpl#handleBegin` on all versions we can't yet get rid of this as easily as it sounds.

We can avoid to instrument the constructor by calling `io.vertx.core.http.impl.HttpServerRequestImpl#response` when the `handleBegin` ends, however we can't call `HttpServerResponseImpl#endHandler` in version 3.6.x to register a handler to deal with ending the transaction due to response state being altered, this is fine in 3.9.x though.

If we can't rely on the `endHandler` to execute something extra when the response handling ends, we could in theory rely on a shared map that stores links from `HttpServerRequestImpl` to `Transaction`, however as parts of the processing is done directly within `endHandler` in 3.6.x, it does executes before we can write to the map.

Also, I had to alter a few tests because some 405 errors were returned when 404s where expected.

SylvainJuge:
Closing as not needed, the related issue was fixed in #2564

PR #2509, opened 2022-03-12
Fix classloader for cross-plugin dependencies

SylvainJuge:
## What does this PR do?

Fixes cross-plugin dependencies class-loading issues.
Fixing the issue is quite trivial, but the real challenge is creating a test to prevent similar issues.

Initially reported [in the forum](https://discuss.elastic.co/t/java-lang-noclassdeffounderror-javax-servlet-servletcontext-apm-agent-1-29-0-in-docker/298975).

- minor-refactor: move 'http.client' to 'httpclient' for consistency
- Implement test to prevent similar issue (parsing plugin dependencies)
- add CL root package customizer for all inter-plugin dependencies

## Checklist

- [x] This is a bugfix
  - [ ] Fixing/removing one exclusion for the log-shading plugin, won't be necessary once https://github.com/elastic/apm-agent-java/pull/2428 is merged.


felixbarny:
All plugins, except for the spring mvc one, don't have a `PluginClassLoaderRootPackageCustomizer` on purpose. The modules they depend on, such as `apm-httpclient-core`, `apm-cassandra-core-plugin`, and `apm-redis-common` don't depend on any specific library classes. Therefore, it's better to load them from the agent class loader so that the plugin class loaders don't contain redundant classes that increase the memory impact.

However, this makes it hard to have a heuristic and create an automated test when to or when not to require a PluginClassLoaderRootPackageCustomizer`.

SylvainJuge:
I agree with you @felixbarny , we should aim to minimize extra memory usage when we can, in this case I think the benefits/cost seems worth it:
- We can assume that loading multiple times those classes is quite rare, for example using multiple redis drivers at the same time is not that common. Also the number of classes and their size is usually small.
- when the test fails there is a clear indication of what to do, thus it's a huge gain for us as detecting this with integration tests would be too tedious as it would require to cover every plugin.

Maybe we could re-think this later with a better approach to provide cross-plugin visibility without duplication of loaded classes if we find that this memory overhead is becoming an issue.

felixbarny:
Good points.

Just a question on the feasibility of making the heuristic a bit smarter: Would it be possible to detect that `HttpClientHelper` is from the `apm-httpclient-core` plugin which doesn't have any dependencies, and therefore it's not required to add a `PluginClassLoaderRootPackageCustomizer`?

SylvainJuge:
Good idea, I have added this case in last commit (https://github.com/elastic/apm-agent-java/pull/2509/commits/603596cdea68c1ecc7617b3103377f8ffd357651) so it simplifies things a lot for HTTP client plugins.

felixbarny:
Nice 👏 

You can also remove the `apm-jdbc-plugin` dependency in the `apm-cassandra-core-plugin` as it is not needed. I forgot to remove it after I have moved some parts out of the JDBC plugin into the core.

As the `apm-redis-common` module also doesn't have any dependencies, you should be able to remove the related `PluginClassLoaderRootPackageCustomizer` that add the `co.elastic.apm.agent.redis` package.

apm-agent-plugins/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/springwebmvc/SpringServiceNameInstrumentation.java
```diff
@@ -83,14 +87,10 @@ public static void afterInitPropertySources(@Advice.This WebApplicationContext a
...
             ServletContext servletContext = applicationContext.getServletContext();
             if (servletContext != null) {
-                try {
-                    ClassLoader servletClassloader = servletContext.getClassLoader();
-                    if (servletClassloader != null) {
-                        classLoader = servletClassloader;
-                        fromServletContext = ServletServiceNameHelper.detectServiceInfo(JavaxServletApiAdapter.get(), servletContext, servletClassloader);
-                    }
-                } catch (UnsupportedOperationException e) {
-                    // silently ignored
```
SylvainJuge, line 93: [for reviewer] the try-catch block is similar in the Servlet adapter, thus removing duplication.

apm-agent/src/test/java/co/elastic/apm/agent/PackagingTest.java
```diff
@@ -18,83 +18,300 @@
...
+                .filter(d -> getGroupId(d).equals(mavenGroupId))
+                .filter(d -> d.getScope() == null || !d.getScope().equals("test"))
                 .map(Dependency::getArtifactId)
                 .collect(Collectors.toSet());
+        }
 
-            assertThat(dependenciesArtifacts).describedAs(createDescription(plugins)).containsAll(plugins);
+        public boolean hasDependencies() {
+            return !mavenModel.getDependencies().isEmpty();
+        }
```
felixbarny, line 211: [minor] just an idea to allow for the module having test dependencies

```suggestion
        public boolean hasNonTestDependencies() {
            return mavenModel.getDependencies().stream().anyMatch(d -> !d.getScope().equals("test"));
        }
```

apm-agent-plugins/apm-redis-plugin/apm-lettuce-plugin/src/main/java/co/elastic/apm/agent/lettuce/LettucePluginClassLoaderRootPackageCustomizer.java
```diff
@@ -0,0 +1,13 @@
+package co.elastic.apm.agent.lettuce;
+
+import co.elastic.apm.agent.bci.PluginClassLoaderRootPackageCustomizer;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+public class LettucePluginClassLoaderRootPackageCustomizer extends PluginClassLoaderRootPackageCustomizer {
```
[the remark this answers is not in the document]
SylvainJuge: Good catch, I've removed it now.

felixbarny (approved):
Excellent!

PR #2519, opened 2022-03-16
Fix asciidoc links in main branch

SylvainJuge:
Remove direct links with asciidoc links for better portability

This PR should not trigger any difference when building documentation (the links should remain the same).

Once this PR is merged, back-porting similar changes to the `1.x` branch will be required.


SylvainJuge:

PR #2513, opened 2022-03-15
Create a span for a nested action invocation via chain result

tobiasstadler:
## What does this PR do?
If the result of an action invocation is a action chain, start a new span for the new action invocation.

## Checklist
- [x] This is an enhancement of existing features, or a new feature in existing plugins
  - [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)
  - [x] I have added tests that prove my fix is effective or that my feature works
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] I have made corresponding changes to the documentation

PR #2522, opened 2022-03-17
Support for setting the remote address of a transaction via the public api

tobiasstadler:
## What does this PR do?
Adds support for setting the remote address of a transaction via the public api

## Checklist


apm-agent-api/src/main/java/co/elastic/apm/api/Transaction.java
```diff
@@ -229,6 +229,15 @@
      */
     Transaction setUser(String id, String email, String username, String domain);
 
+    /**
+     * Provides a way to manually set the {@code http.request.socket.remote_address} field.
```
tobiasstadler, line 233: Maybe `http.request.socket.remote_address` should be `client.source.ip` (see elastic/apm-server#6152)?

tobiasstadler:
ping

tobiasstadler:
ping

tobiasstadler:
superseded by #2513

tobiasstadler:
ping

tobiasstadler:
ping

jackshirazi:
Hi @tobiasstadler, could you give a little detail on why this feature addition is useful? We're more focused on enhancing our support of OpenTelemetry so that eventually that provides the API for the agent, but we can add to the existing API if it's useful and simple

tobiasstadler:
I would like to record the "client ip" in my ejb plugin (https://github.com/tobiasstadler/apm-wildfly-remote-ejb-plugin). From time time I have the need to filter my transactions based on the service (which in our case is mostly identical with the client ip) that is calling my service. right now this is hard to accomplish reliably fro me.

jackshirazi:
Thanks. Couple of things. 

We think that if the transaction api is used to set the remote address, then that should take precedence over any discovered address - even if the transaction sets it to null or empty string. I think atm the final value would just be down to the order of setting

Secondly, could you add a test?

PR #2525, opened 2022-03-18
Implement default type fallback

SylvainJuge:
## What does this PR do?

Implements https://github.com/elastic/apm/pull/602

## Checklist


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -675,4 +675,11 @@ public T withUserOutcome(Outcome outcome) {
         return thiz();
     }
 
+    protected String normalizeType(@Nullable String type) {
+        if (type == null || type.isEmpty()) {
+            return "custom";
+        }
+        return type;
+    }
```
eyalkoren, line 683: [minor] proposal - do that in a base implementation for `beforeEnd()` instead and call `super.beforeEnd()` in children

eyalkoren (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
SylvainJuge: I suggest to do it without any sub-classing: by moving the `type` attribute directly into `AbstractSpan`, while the field value has different semantics in `Transaction` and `Span` this makes sense and allows to keep this normalization private in `AbstractSpan`.

I'll push an implementation of this in this PR, let me know if that would be a good fit here.

SylvainJuge: I really don't know why it was like this in the first place, doing some git archaeology it has been like this since 2018.
Maybe @felixbarny remember something about having two distinct fields and one of them being `volatile`, as this is not urgent we can definitely discuss this during next weekly meeting.

PR #2530, opened 2022-03-21
Avoid multiple HttpUrlConnection span ending on nested calls

eyalkoren:
When there are nested calls instrumented by `HttpUrlConnectionInstrumentation`, there may be a state where the same span will be ended more than once.
For example, the `HttpUrlConnection#connect` instrumentation creates a span and doesn't end it, but instead adds it to the `inFlightSpans` map. Then, if there are nested calls to `HttpUrlConnection#getInputStream`, both will find it in the map in the enter advice, thus trying to end it in the exit advice.
This PR simply adds a `span.isFinished()` check before trying to end it.

Ideally, we would identify nested calls and only apply the logic on the outermost invocation (e.g. through `co.elastic.apm.agent.sdk.state.CallDepth`), but since we rely on instance fields in this instrumentation (both `connected` and `responseCode`), I am not sure what will be the implications in case of an HttpUrlConnection wrapper that delegates invocations to another object but doesn't get its fields updated, so I took the more defensive route.


SylvainJuge (approved):

eyalkoren:
Actually, it contains a race condition, where the span can be recycled and even reused when the outermost method is exited.
This can be fixed by deactivating the span as the last thing in the exit advice, but the actual problem is that we accept this non-deterministic span processing order.
I took another look to see if I can process only at the outermost (nesting) call, and it seems it wouldn't change the current behavior:
-  if the nested calls are from subclass to superclass (same object), then the fields are shared within the base class, so it should be OK to rely on them only in the outermost method (superclasss)
- if the nested calls are from a wrapper to a delegate, then the delegate will never find the span in the map (because it is a different `this`) and it will never succeed to create an exit span (because the parent is an exit span create by the caller method), which means it will always return a `null` span, which means the exit advice will never be executed.

So, I will change that accordingly and ask for a second review

eyalkoren:
Another thing to consider here is that we may instrument custom implementations of `HttpUrlConnection` that use the super methods. In that case, these custom classes may be loaded by a custom class loader, which means we will create a plugin class loader for them that includes the instrumentation class, so superclass and subclass may use different `inFlightSpans` maps. 
Annotating the instrumentation class with `@GlobalState` should resolve that.

I will take a look at this PR later on again and likely do some changes

eyalkoren:
run end-to-end tests

SylvainJuge (approved):

PR #2533, opened 2022-03-22
release: enable the disabled stages and run docker images publishing earlier

v1v:
## What does this PR do?
* Enable the stages that were disabled to test the Release pipeline
* Run docker push earlier
* Send slack message with the release notes URL in the comment.

## Why

Release Notes publishing to be the last action

### DIFF
--- a/.ci/release/Jenkinsfile
+++ b/.ci/release/Jenkinsfile
@@ -241,6 +241,19 @@ pipeline {
             }
           }
         }
+        stage('Build and push Docker images') {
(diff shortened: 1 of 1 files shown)


eyalkoren (approved):

eyalkoren:
Can we use this PR to move the `Build and push Docker images` to be before `Publish release on GitHub`?
The GH publish should be the final step as it tells everyone that all agent artifacts can be consume on all environments.

PR #2537, opened 2022-03-23
Log less frightening message when a plugin's class version is unsupported in the current VM

felixbarny:


apm-agent-core/src/main/java/co/elastic/apm/agent/util/DependencyInjectingServiceLoader.java
```diff
@@ -143,8 +143,8 @@ private T instantiate(ClassLoader classLoader, String implementation) {
         } catch (InstantiationException e) {
             String msg = String.format("unable to instantiate '%s', please check descriptor in META-INF", implementation);
             throw new ServiceConfigurationError(msg, e);
-        } catch(UnsupportedClassVersionError e) {
-            logger.debug(String.format("unable to instantiate '%s', unsupported class version error: %s", implementation, e.getMessage()));
+        } catch (UnsupportedClassVersionError e) {
+            logger.debug("Skipping {} because it only applies to more recent Java versions", implementation);
```
jackshirazi, line 147:
```suggestion
            logger.debug("Skipping {} because it only applies to more recent Java versions and the JVM running this app is too old to load it", implementation);
```

jackshirazi (approved):

SylvainJuge (approved):

PR #2551, opened 2022-03-28
Fixed AWS lambda support for missing AWS events classes

AlexanderWert:
When AWS handlers are using core types (e.g. String, Integer) as generic type for the handler class (instead of one of the AWS events types from the `com.amazonaws.services.lambda.runtime.events` package), then the current instrumentation throws a ClassNotFound exception because the `com.amazonaws.services.lambda.runtime.events` package might be not available.

This PR fixes this case.

- [x] successfully tested on AWS Lambda with the Hello world case in which the `events` package is not available

### DIFF
--- a/apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/RequestHandlerInstrumentation.java
+++ b/apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/RequestHandlerInstrumentation.java
@@ -74,7 +74,7 @@ public class RequestHandlerInstrumentation extends AbstractAwsLambdaHandlerInstr
                 Transaction transaction = (Transaction) transactionObj;
 
                 if (output != null && output.getClass().getName().startsWith("com.amazonaws.services.lambda.runtime.events")) {
-                    // handler uses aws events, it's save to assume that the AWS events classes are available
+                    // handler uses aws events, it's safe to assume that the AWS events classes are available
                     AWSEventsHelper.finalizeTransaction(transaction, output, thrown);
                 } else {
                     // Fallback instrumentation (AWS events classes might be not available)
--- a/apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/RequestHandlerInstrumentation.java
+++ b/apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/RequestHandlerInstrumentation.java
@@ -18,22 +18,11 @@
  */
 package co.elastic.apm.agent.awslambda;
 
-import co.elastic.apm.agent.awslambda.helper.APIGatewayProxyV1TransactionHelper;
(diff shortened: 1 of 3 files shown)


apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/RequestHandlerInstrumentation.java
```diff
@@ -67,26 +56,14 @@ public String getAdviceClassName() {
...
-                return SQSTransactionHelper.getInstance().startTransaction((SQSEvent) input, lambdaContext);
-            } else if (input instanceof SNSEvent) {
-                // SNS trigger
-                return SNSTransactionHelper.getInstance().startTransaction((SNSEvent) input, lambdaContext);
-            } else if (input instanceof S3Event) {
-                // S3 event trigger
-                return S3TransactionHelper.getInstance().startTransaction((S3Event) input, lambdaContext);
+        public static Object handlerEnter(@Nullable @Advice.Argument(value = 0) Object input, @Advice.Argument(value = 1) Context lambdaContext) {
+            if (input != null && input.getClass().getName().startsWith("com.amazonaws.services.lambda.runtime.events")) {
+                // handler uses aws events, it's save to assume that the AWS events classes are available
```
eyalkoren, line 61:
```suggestion
                // handler uses aws events, it's safe to assume that the AWS events classes are available
```

eyalkoren (approved):
I think this is definitely safe enough.
If we ever hear of a scenario where the new `AWSEventsHelper` gets loaded even when the `com.amazonaws.services.lambda.runtime.events` types are not present (due to JVM quirkiness), we can make a complete and full separation of dependencies then.

apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/RequestHandlerInstrumentation.java
```diff
@@ -95,12 +72,12 @@ public static void handlerExit(@Nullable @Advice.Enter Object transactionObj,
                                        @Nullable @Advice.Return Object output) {
             if (transactionObj instanceof Transaction) {
                 Transaction transaction = (Transaction) transactionObj;
-                if (output instanceof APIGatewayV2HTTPResponse) {
-                    APIGatewayProxyV2TransactionHelper.getInstance().finalizeTransaction(transaction, (APIGatewayV2HTTPResponse) output, thrown);
-                } else if (output instanceof APIGatewayProxyResponseEvent) {
-                    APIGatewayProxyV1TransactionHelper.getInstance().finalizeTransaction(transaction, (APIGatewayProxyResponseEvent) output, thrown);
+
+                if (output != null && output.getClass().getName().startsWith("com.amazonaws.services.lambda.runtime.events")) {
+                    // handler uses aws events, it's save to assume that the AWS events classes are available
```
AlexanderWert, line 77:
```suggestion
                    // handler uses aws events, it's safe to assume that the AWS events classes are available
```

PR #2552, opened 2022-03-28
fix premature recycling

jackshirazi:
## What does this PR do?
Moves the parent.decrementReferences() call from Span.beforeEnd() to Span.afterEnd()

## Checklist
- [x] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
@@ -305,13 +305,12 @@ public class Span extends AbstractSpan<Span> implements Recyclable {
         if (transaction != null && transaction.isSpanCompressionEnabled() && parent != null) {
             Span buffered = parent.bufferedSpan.get();
             if (!isCompressionEligible()) {
-                parent.decrementReferences();
                 if (buffered != null) {
                     if (parent.bufferedSpan.compareAndSet(buffered, null)) {
-                        parent.decrementReferences();
                         this.tracer.endSpan(buffered);
                     }
                 }
+                parent.decrementReferences();
                 this.tracer.endSpan(this);
                 return;
             }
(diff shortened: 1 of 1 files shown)


felixbarny:
> Moves the parent.decrementReferences() call from Span.beforeEnd() to Span.afterEnd()

I'm not sure if that's enough. As long as the parent is holding a buffered span, it's not safe to decrement references of the parent. Maybe there are more problematic cases but the most problematic one is when the parent is already ended but child spans are created and compressed. In this case, the compressed spans will never get reported. That's because a compressed span is reported when its parent ends (see [spec](https://github.com/elastic/apm/blame/main/specs/agents/handling-huge-traces/tracing-spans-compress.md#L164)). Therefore, we'll probably have to adjust the spec so that a span is only [eligible for compression](https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-compress.md#eligibility-for-compression) if its parent has not ended yet. However, the issue is that this condition may be racy. The parent and child may end concurrently in different threads.

jackshirazi:
I'm rethinking it all. Here's where I am at the moment
1. If there is a concurrent child, decrementReferences() is fine because the parent won't get recycled
2. If there is no concurrent child but there is a buffer, decrementReferences() is fine because the parent won't get recycled, there is still a reference
3. If there is no concurrent child and no buffer, we want to decrementReferences() EXCEPT if this child in the end() call becomes the buffer

The race case where the parent and child are concurrently ended needs the parent to check if it has a buffer and recycle that before recycling itself. You still have the condition where the parent has recycled the buffer before the buffer is ended, still trying to get that one

I'm wondering if the recycling is worth it

felixbarny (approved):
While there are open questions for the interactions with recycling and span compressions, I feel like this is an incremental improvement that's worth merging in.

jackshirazi:
@felixbarny what do you think about letting the recycle only happen if the bufferedSpan is null? The decrement reference happens after the bufferedSpan is cleared under normal operation, so that would address the edge race case though with the risk that the Span ultimately wouldn't get recycled?

jackshirazi:
This set of changes doesn't fix the issue, but all the changes are needed since the object shouldn't be recycled before the buffer is accessed - which is happening prior to this. I'll look in to reproducing the problem and applying a fix using a different branch. Meanwhile applying this set of changes and including it in the patch

PR #2553, opened 2022-03-28
Fix/refactor unsampled tests

SylvainJuge:
## What does this PR do?

- removes duplicated tests
- introduce more reliable testing for features that rely on APM server version

A further improvement could be to introduce a test-only apm server client implementation that does not have any asynchronous behavior and where APM server version is set by the tests and not inferred from the healthcheck request response.


apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ElasticApmTracerTest.java
```diff
@@ -56,17 +62,24 @@
...
     private TestObjectPoolFactory objectPoolFactory;
 
     @BeforeEach
     void setUp() {
         objectPoolFactory = new TestObjectPoolFactory();
         reporter = new MockReporter();
         config = SpyConfiguration.createSpyConfig();
+
+        apmServerClient = new ApmServerClient(config.getConfig(ReporterConfiguration.class), config.getConfig(CoreConfiguration.class));
+        apmServerClient = mock(ApmServerClient.class, delegatesTo(apmServerClient));
```
SylvainJuge, line 76: [for reviewer] this allows to delegate on default implementation and also enables us to mock some methods for easier testing.

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ElasticApmTracerTest.java
```diff
@@ -338,7 +351,19 @@ void testPauseMidTransaction() {
     }
 
     @Test
-    void testSamplingNone() throws IOException {
```
SylvainJuge, line 341: [for reviewer] this was the originally failing test that did not passed when executed in isolation. combined with the next modified block, it allows to cover the now removed test cases in other files.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
```diff
@@ -118,6 +121,12 @@ public ElasticApmTracerBuilder withLifecycleListener(LifecycleListener listener)
         return this;
     }
 
+    // protected only for testing
+    public ElasticApmTracerBuilder withApmServerClient(ApmServerClient apmServerClient) {
```
eyalkoren, line 125:
```suggestion
   ElasticApmTracerBuilder withApmServerClient(ApmServerClient apmServerClient) {
```

eyalkoren (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracerBuilder.java
SylvainJuge: It was actually required, thus this change was reverted in https://github.com/elastic/apm-agent-java/pull/2553/commits/7a718fbef47c6c8fd12be61e706e238e8078d6b4.

eyalkoren: So it's comment is wrong 🙂

eyalkoren: But the correct thing to do is leave it package-private.
The way I dealt with providing test access like that is through `TracerInternalApiUtils`. I'd encourage you to do that (even use the same class as it is in the same package) instead of making a test-only method `public`

PR #2558, opened 2022-03-29
workaround premature recycling (compression related)

jackshirazi:
## What does this PR do?
lose the compression rather than retry looking for strategy, so that the application proceeds with minimum delay - branch to see if this works for the customer as a temporary workaround

## Checklist

- [x] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
@@ -413,10 +413,11 @@ public class Span extends AbstractSpan<Span> implements Recyclable {
     }
 
     private boolean tryToCompressComposite(Span sibling) {
-        String compressionStrategy;
-        do {
-            compressionStrategy = composite.getCompressionStrategy();
-        } while (compressionStrategy == null);
+        String compressionStrategy = composite.getCompressionStrategy();
+        if (compressionStrategy == null) {
+            //lose the compression rather than retry, so that the application proceeds with minimum delay
+            return false;
+        }
 
         switch (compressionStrategy) {
             case "exact_match":


felixbarny (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -413,10 +413,11 @@ private boolean tryToCompressRegular(Span sibling) {
...
     private boolean tryToCompressComposite(Span sibling) {
-        String compressionStrategy;
-        do {
-            compressionStrategy = composite.getCompressionStrategy();
-        } while (compressionStrategy == null);
+        String compressionStrategy = composite.getCompressionStrategy();
+        if (compressionStrategy == null) {
+            //lose the compression rather than retry, so that the application proceeds with minimum delay
+            return false;
+        }
```
efdknittlfrank, line 420: Thank you for working on this! I'd be happy to test the workaround.

I was wondering if the retry could have a kind of fail-safe, aborting after a fixed number of retries. Something similar to:

```suggestion
        String compressionStrategy;
        for (int i = 0; i < 100 && compressionStrategy == null; ++i) {
            compressionStrategy = composite.getCompressionStrategy();
        }
        if (compressionStrategy == null) {
            // lose the compression rather than retry, so that the application proceeds with minimum delay
          return false;
        }
```

But I assume the workaround will be removed soon, once the root cause is found?

jackshirazi: Let's see how it looks when you try it. We might prefer to leave the fast fail so that there is minimum impact to the application. When we've eliminated premature recycling, this concurrent race should happen rarely. At most I think I'd retry only once

PR #2564, opened 2022-03-31
Vertx avoid wrapping

eyalkoren:
Reworking Vert.x 3.x instrumentation to avoid constructor instrumentation and handler wrapping.
In addition, added some debug logs to enable incorrect tracing flows.
Currently, this will be used for analysis purposes. If we decide to use the new mechanism, we can still improve as follows:
- This PR replaces the signal to end the transaction from response ending event to request ending event. While this should be conceptually equivalent for HTTP 1, it is not necessarily the case for HTTP 2. The full fix would be to do apply this sequence:
   1. start a transaction, map the request to the transaction
   2. as early as possible, map the response to the transaction (in HTTP 1 that would be at the end of `handleBegin`, in HTTP 2 this can be done as soon as we start the transaction, at the exit of `io.vertx.core.http.impl.Http2ServerConnection#createRequest()`)
   3. follow the route and whenever handled- activate and deactivate the transaction. The lookup can be done based on the request, and if not found, on the value of `request.response()`. This may replace the current instrumentation we have for `io.vertx.ext.web.impl.HttpServerRequestWrapper`, as it delegates the `response` to its `delegate`. If indeed we are worried about handling that occurs after Request is invalid in HTTP 2, we can look for the transaction based on `io.vertx.ext.web.RoutingContext#response()` (currently we only look based on `io.vertx.ext.web.RoutingContext#request()`)
   4. identify the request end event and end the transaction and remove the request mapping
   5. identify the response end event and remove the response mapping
- Look if we can benefit from instrumentation of `io.vertx.core.http.impl.HttpServerRequestImpl#handleException` and `io.vertx.core.http.impl.Http2ServerRequestImpl#handleException`


apm-agent-plugins/apm-vertx/apm-vertx3-plugin/src/main/java/co/elastic/apm/agent/vertx/v3/web/WebHelper.java
```diff
@@ -72,22 +76,22 @@ public Transaction setRouteBasedNameForCurrentTransaction(RoutingContext routing
...
     }
 
-    public void removeTransactionFromContext(HttpServerRequest request) {
-        if (request.getClass().getName().equals("io.vertx.ext.web.impl.HttpServerRequestWrapper")) {
-            request = request.endHandler(noopHandler);
-        }
-        requestTransactionMap.remove(request);
+    @Nullable
+    public Transaction removeTransactionFromContext(HttpServerRequest request) {
+        return requestTransactionMap.remove(request);
```
AlexanderWert, line 81: Why did you remove the handling of the wrapped request case?
The purpose of that was that we never have the a wrapped request object (`io.vertx.ext.web.impl.HttpServerRequestWrapper`) as a key in our context-map.

eyalkoren: We now instrument the `io.vertx.core.http.impl.HttpServerRequestImpl#handleEnd` and `io.vertx.core.http.impl.Http2ServerRequestImpl#handleEnd` so there is no need for that

eyalkoren: If these events are not guaranteed to run, we have a bigger problem than leftover in the map - we won't end the transaction, so we can't use this approach

AlexanderWert (approved):

eyalkoren:
I ended up only registering and looking up response objects for HTTP 2, where handling sequence starts only after both are registered. 

In HTTP 1, handling may start before we get the chance to register the response (assuming we are not instrumenting constructor). There are ways to deal with that, e.g. map the route details to the response and end the transaction only when the response ends. However, in HTTP 1 it seems sufficient to end the transaction and lose its references when the request end event is identified. Therefore, I left the `HttpServerRequestWrapper#endHandler` instrumentation in place. We can find an alternative only if and when we see a need for that.

As for the instrumentation of `HttpServerRequestImpl#handleException` and `HttpServerResponseImpl#handleException` - so far I didn't find a scenario where these APIs are invoked while there is an active request (and corresponding active transaction).
There are more central places to capture such out-of-transaction errors, like `io.vertx.core.net.impl.ConnectionBase#handleException`, but as long as we don't create a test scenario, we will wait with that.

apm-agent-plugins/apm-vertx/apm-vertx3-plugin/src/main/java/co/elastic/apm/agent/vertx/v3/web/HttpServerRequestImplEndInstrumentation.java
```diff
@@ -16,47 +16,63 @@
...
 
+        private static final Logger log = LoggerFactory.getLogger(HttpRequestEndAdvice.class);
+
         private static final WebHelper helper = WebHelper.getInstance();
 
         @Advice.OnMethodExit(suppress = Throwable.class, inline = false)
-        public static void exit(@Advice.This HttpServerRequestImpl request) {
-            helper.removeTransactionFromContext(request);
+        public static void exit(@Advice.This HttpServerRequest request,
+                                @Advice.FieldValue(value = "response", typing = Assigner.Typing.DYNAMIC) @Nullable HttpServerResponse response) {
```
AlexanderWert, line 68: We are using this Advice to instrument bot, `HttpServerRequestImpl` and `Http2ServerRequestImpl` (see type matcher above), BUT, here you use `HttpServerResponse` which is only valid for `HttpServerRequestImpl`. Response class for `Http2ServerRequestImpl` is called `Http2ServerResponseImpl`. So, with Http2 we'll never finalize the transaction with the actual response information with this instrumentation, right? So we may want to split this into a separate advice or do some casting of the response object

eyalkoren:
Based on a [user report](https://discuss.elastic.co/t/some-transaction-names-are-incorrect-falsely-reported/302475), there may be a situation where the obtained `path` is `null`. I was not able to reproduce that with and empty path, but added this null check either way so that in such scenarios the transaction name will remain the default - `unknown route`.

AlexanderWert (changes requested):
There's only one thing that is not clear to me (see comment). Everything else LGTM!

apm-agent-plugins/apm-vertx/apm-vertx3-plugin/src/main/java/co/elastic/apm/agent/vertx/v3/web/HttpServerRequestImplEndInstrumentation.java
eyalkoren: `io.vertx.core.http.HttpServerResponse` is the interface of both and I specified I allow dynamic typing for the inference of this field

AlexanderWert (approved):

Issue #2550, opened 2022-03-28
Elasticsearch instrumentation


AlexanderWert:

pugnascotia:
So far, the only change I've made to the agent is as follows:

```diff
diff --git a/apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/logging/LoggerFactory.java b/apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/logging/LoggerFactory.java
index c61257d21..4bcf2ee57 100644
--- a/apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/logging/LoggerFactory.java
+++ b/apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/logging/LoggerFactory.java
@@ -18,6 +18,9 @@
  */
 package co.elastic.apm.agent.sdk.logging;
 
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
 public class LoggerFactory {
 
     private static volatile ILoggerFactory iLoggerFactory;
@@ -32,11 +35,16 @@ public class LoggerFactory {
      * @param name The name of the logger.
      * @return logger
      */
-    public static Logger getLogger(String name) {
+    public static Logger getLogger(final String name) {
         if (iLoggerFactory == null) {
             return NoopLogger.INSTANCE;
         }
-        return iLoggerFactory.getLogger(name);
+        return AccessController.doPrivileged(new PrivilegedAction<Logger> () {
+            @Override
+            public Logger run() {
+                return iLoggerFactory.getLogger(name);
+            }
+        });
     }
 
     /**
```

This may not ultimately be what we want to do, but I got me further.

The problem with `java.lang.Throwable` not being found was (I believe) due to the ES code lacking the `accessSystemModules` permission.

CC @ChrisHegarty

felixbarny:
I've noticed that the `apm-agent` module doesn't have all the sources of the agent, which makes debugging a challenge.

May not be entirely in scope of this issue but having the ability to pull in all sources to that the agent can be debugged within a project that it's integrated in seems like an important capability. That's not only in the context of Elasticsearch but arguably specifically important there.

pugnascotia:
I tried running a 3 node cluster in docker-compose, and noticed that the service map doesn't break down the cluster, which I think would be useful if one was in a degraded state. I set a different `service_node_name` on each, which is reflected in the "JVMs" tab.

[...]

PR #2566, opened 2022-03-31
Fix agent class package definition

SylvainJuge:
## What does this PR do?

Fixes agent class package definition in shaded classloader

- relates to #2550 

## Checklist

- [x] This is a bugfix
  - [x] manually testing within ES that it fixes the issue.

### DIFF
--- a/elastic-apm-agent/src/main/java/co/elastic/apm/agent/premain/ShadedClassLoader.java
+++ b/elastic-apm-agent/src/main/java/co/elastic/apm/agent/premain/ShadedClassLoader.java
@@ -64,14 +64,14 @@ public class ShadedClassLoader extends URLClassLoader {
 
         MethodHandle methodHandle = null;
         try {
-            methodHandle = MethodHandles.lookup()
-                .findSpecial(ClassLoader.class,
+            methodHandle = MethodHandles.publicLookup()
+                .findVirtual(ShadedClassLoader.class,
                     "getDefinedPackage",
-                    MethodType.methodType(Package.class, String.class),
-                    ShadedClassLoader.class);
-        } catch (NoSuchMethodException|IllegalAccessException e) {
-            // ignored, the method is either absent or we don't have access to it.
-            // On IBM J9 an IllegalAccessException is thrown because the method is present but package-private
+                    MethodType.methodType(Package.class, String.class));
+        } catch (NoSuchMethodException e) {
+            // ignored as method is not present
+        } catch (IllegalAccessException e) {
+            throw new IllegalStateException(e);
         }
         getDefinedPackage = methodHandle;
     }
@@ -108,7 +108,7 @@ public class ShadedClassLoader extends URLClassLoader {
                 } else {
                     definePackage(packageName, null, null, null, null, null, null, null);
                 }
-            } catch (IllegalStateException e){
+            } catch (IllegalArgumentException e){
                 // The package may have been defined by a parent class loader in the meantime
                 if (isPackageNotDefined(packageName)) {
                     throw e;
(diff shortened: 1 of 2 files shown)


elastic-apm-agent/src/main/java/co/elastic/apm/agent/premain/ShadedClassLoader.java
```diff
@@ -80,18 +80,11 @@ public ShadedClassLoader(File jar, ClassLoader parent, String customPrefix) thro
...
-                // The package may have been defined by a parent class loader in the meantime
-                if (getPackage(packageName) == null) {
-                    throw e;
-                }
+        if (packageName != null && getDefinedPackage(packageName) == null) {
+            if (manifest != null) {
+                definePackage(packageName, manifest, jarUrl);
+            } else {
+                definePackage(packageName, null, null, null, null, null, null, null);
             }
```
felixbarny, line 88: we may still need that catch block

elastic-apm-agent/src/main/java/co/elastic/apm/agent/premain/ShadedClassLoader.java
```diff
@@ -80,23 +101,39 @@ public ShadedClassLoader(File jar, ClassLoader parent, String customPrefix) thro
...
+                if (isPackageNotDefined(packageName)) {
                     throw e;
                 }
             }
         }
         return defineClass(name, classBytes, 0, classBytes.length, ShadedClassLoader.class.getProtectionDomain());
     }
 
+    @SuppressWarnings("deprecation")
+    private boolean isPackageNotDefined(String packageName) {
```
felixbarny, line 122: Probably a matter of taste but I'd always use affirmative rather than negative method names.
See also https://softwareengineering.stackexchange.com/a/196837

elastic-apm-agent/src/main/java/co/elastic/apm/agent/premain/ShadedClassLoader.java
```diff
@@ -80,23 +101,39 @@ public ShadedClassLoader(File jar, ClassLoader parent, String customPrefix) thro
...
+        Package pkg;
+        if (getDefinedPackage != null) {
+            try {
+                pkg = (Package) getDefinedPackage.invokeExact(this, packageName);
```
felixbarny, line 126: Do we even need to call `getDefinedPackage` or is `getPackage` good enough? Maybe its preferable to have the same semantics regardless of the Java version?

On Slack you said

> f migrating the call to the deprecated `getPackage(...)` and using `getDefinedPackage(...)` the difference is that now it does not tries to delegate to parent nor Bootstrap CL when checking if the package is defined.
Given the classes that are loaded by the shaded CL can only be loaded by this CL, I don't think there is any need to ask the parent CL in this case.

So it seems there's not really a benefit of calling `getDefinedPackage`?

felixbarny: Ah, I misunderstood. The `getPackage` delegates to the parent which may be an issue in certain cases if the package has been defined in a parent CL but not the current CL. Probably not an issue either way in our case, as the parent (platform CL or bootstrap CL) will never have defined an agent package.

I'm fine either way, I guess. Calling `getDefinedPackage` is technically more correct and avoids calling a deprecated API but adds a bit of boiler plate.

SylvainJuge: That's a very good question !

In theory there is no case where the `ShadedClassLoader` will load classes that are defined in packages that have been defined by a parent classloader, so using `getDefinedPackage` should be enough and we can skip the extra lookups that are done by `getPackage`. Performance wise, it should be a trivial difference as the packages are stored in a concurrent map.

In practice, we still have cases where the call the `defineClass` throws an `IllegalArgumentException`, which means a parent CL is expected to load those classes. So if the classes can be defined elsewhere, then it would also be consistent to consider that the packages can also be defined elsewhere, hence delegation to parent CL does make sense.
On top of that, we only check for the package presence with a `null` check and don't use the returned `Package` instance, so there is no difference on the origin of the `Package` instance.

As a conclusion I now think it's not worth the extra complexity to use a method handle here, and we can always revisit this later. I'll add a comment to clarify the usage of `getPackage` vs `getDefinedPackage`.

felixbarny (approved):

PR #2568, opened 2022-04-01
Fix log4j init with SecurityManager

SylvainJuge:
## What does this PR do?

Relates to #2550 

Simply wraps the logger init in a `doPrivileged` block to get the proper permissions when a Security Manager is present.

Moving the `doPrivileged` wrapping down to the `getLogger`  or `getContext` methods in `Log4jFactoryBridge` does not work as expected and we get other permission errors with ES, so we keep it in the caller side.

## Checklist

- [x] This is a bugfix
  - [x] manually testing with ES

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4jLoggerFactoryBridge.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4jLoggerFactoryBridge.java
@@ -28,19 +28,15 @@ import org.apache.logging.log4j.spi.LoggerContextFactory;
 import org.apache.logging.log4j.util.StackLocatorUtil;
 
 /**
- * Based on {@code org.apache.logging.slf4j.Log4jLoggerFactory}
+ * Based on {@code org.apache.logging.slf4j.Log4jLoggerFactory}.
+ * <p>
+ * This class does not implement {@link ILoggerFactory} directly but through the super class implementation that has
+ * a method matching {@link ILoggerFactory#getLogger(String)}. Given this method is caller-sensitive then we can't
+ * implement it directly without side effects.
  */
 public class Log4jLoggerFactoryBridge extends AbstractLoggerAdapter<Logger> implements ILoggerFactory {
 
-    private static final String FQCN = Log4jLoggerFactoryBridge.class.getName();
-    private static final String PACKAGE = "co.elastic.apm.agent.sdk.logging";
-
-    @Override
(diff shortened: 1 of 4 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4jLoggerFactoryBridge.java
```diff
@@ -35,6 +35,13 @@
     private static final String FQCN = Log4jLoggerFactoryBridge.class.getName();
     private static final String PACKAGE = "co.elastic.apm.agent.sdk.logging";
 
+    @Override
+    public Logger getLogger(String name) {
+        // parent class AbstractLoggerAdapter already 'implements' the method because it has the same signature
+        // this provides an actual implementation of ILoggerFactory while delegating explicitly to the parent impl.
+        return super.getLogger(name);
```
SylvainJuge, line 42: [for reviewer] while completely not required, it makes navigation from `ILoggerFactory` class hierarchy more straightforward, otherwise we can miss that the parent class already implements the method.

felixbarny (approved):

pugnascotia:
cc @ChrisHegarty

ChrisHegarty (approved):
LGTM.

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/Log4jLoggerFactoryBridge.java
```diff
@@ -53,7 +56,12 @@ protected Logger newLogger(final String name, final LoggerContext context) {
...
     @Override
     protected LoggerContext getContext() {
-        final Class<?> anchor = StackLocatorUtil.getCallerClass(FQCN, PACKAGE);
-        return anchor == null ? LogManager.getContext() : getContext(StackLocatorUtil.getCallerClass(anchor));
+        // the logger context is defined by the class that calls LoggerFactory
+        final Class<?> factoryCaller = StackLocatorUtil.getCallerClass(LOGGER_FACTORY);
+        if (factoryCaller == null) {
+            return LogManager.getContext();
+        }
+        return getContext(factoryCaller);
```
SylvainJuge, line 64: @eyalkoren I do need your review on this as I'm not 100% sure it's the right way to fix it.
Without this, the added anonymous class makes `loggingLevelChangeTest` test fail.

pugnascotia (approved):

eyalkoren (approved):

Issue #2569, opened 2022-04-01
Missing documentation on how logging appenders are selected for ECS reformatting


fgabolde:
## Describe the bug

Missing documentation on how ECS reformatting works: how does it know which appenders to apply to. This would allow users to be confident that they aren't overriding e.g. their Sentry logger at the same time as their normal STDOUT stream.

Or my expectations for the reformatter might be too high and it just reformats ALL appenders, in which case this should also be documented as a warning to users :)

## Expected behavior

A list of conditions that must be met for the ECS reformatter to be applied to an appender (on top of enabling the reformatting option, which is already documented).

eyalkoren:
Thanks for reporting this.
Currently we reformat only `FileAppender` types (hardcoded) and plan to apply that to console appenders as well (see #2570).

We had a related [discussion](https://github.com/elastic/apm/pull/440) before, and we ended up only with an option for fine tuning is based on [formatter types](https://www.elastic.co/guide/en/apm/agent/java/current/config-logging.html#config-log-ecs-formatter-allow-list), but I don't think this is what you need.

@felixbarny I'd be happy to hear your opinion on this. To me it sounds reasonable to allow filtering based on appender type/name or even logger name pattern.

fgabolde:
> Thanks for reporting this. Currently we reformat only `FileAppender` types (hardcoded) and plan to apply that to console appenders as well (see #2570).

There is actually this bit of the docs on which appenders are affected:

```
[...] while SHADE and REPLACE options are only relevant to file log appenders, the OVERRIDE option is also valid for other appenders, like System out and console
```

The "other appenders" part had me concerned about stuff like the Sentry appender which would not be a file appender I guess, probably has an in-memory buffer and makes a bunch of network calls.

> We had a related [discussion](https://github.com/elastic/apm/pull/440) before, and we ended up only with an option for fine tuning is based on [formatter types](https://www.elastic.co/guide/en/apm/agent/java/current/config-logging.html#config-log-ecs-formatter-allow-list), but I don't think this is what you need.

No, this is actually what I was looking for. I missed it in the docs because the doc for `log_ecs_reformatting`, which is basically the entry point for this feature's documentation, doesn't mention it. Since the docs I was complaining about do exist you can close this issue if you wish.

The default value for this option does explain why my normal STDOUT appender is being reformatted (it normally uses a PatternLayout formatter) and does reassure me that my Sentry appender should be "safe" from reformatting.

[...]

PR #2571, opened 2022-04-04
Clarify that log_ecs_reformatting only applies to pattern layouts by default

felixbarny:
inspired by #2569


felixbarny:
@fgabolde does this clarify the concern you had in https://github.com/elastic/apm-agent-java/issues/2569?

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/LoggingConfiguration.java
```diff
@@ -176,7 +176,9 @@ static LogLevel mapLogLevel(LogLevel original) {
             " - OVERRIDE - same log output is used, but in ECS-compatible JSON format instead of the original format. \n" +
             "\n" +
             "NOTE: while `SHADE` and `REPLACE` options are only relevant to file log appenders, the `OVERRIDE` option \n" +
-            "is also valid for other appenders, like System out and console")
+            "is also valid for other appenders, like System out and console.\n" +
+            "This option only applies to pattern layouts/formatters by default.\n" +
+            "See also <<config-log-ecs-formatter-allow-list, `log_ecs_formatter_allow_list`>>.")
```
eyalkoren, line 181: This is true for all options, so I would add it to the general description rather than this note

fgabolde:
> @fgabolde does this clarify the concern you had in #2569?

Yep, looks good. Thanks!

PR #2578, opened 2022-04-11
Replace/extend resource dest. with service target.{name.type}

SylvainJuge:
## What does this PR do?

Implements the agent-side migration from sending `span.destination.service.resource` to `span.service.target.type` + `span.service.target.name`.

Impacted features:
- dropped spans stats
- compressed spans
- User API ability to set a custom value for `span.destination.service.resource`
- migration of all usages of `span.destination.service.resource`

The agent remains backwards compatible with existing APM servers:
- only new fields are added to the protocol
- existing fields values are inferred from the new internal data structure.

## Checklist

- [x] add a new API entry point to set the span service type and name.
- [x] add new fields to the existing resource destination JSON spec + back-port changes to upstream spec https://github.com/elastic/apm/pull/627
- [x] backport changes to JSON resource spec + OTel gherkin spec to upstream APM repo
- [x] sync with state of upstream specification https://github.com/elastic/apm/pull/627
- [x] create follow-up PR in https://github.com/elastic/apm/ to enhance target fields values & destination resource for non-relational databases: MongoDB (already impl in this PR), Elasticsearch, ... https://github.com/elastic/apm/issues/646


apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
```diff
@@ -466,4 +466,15 @@ public interface Span {
      */
     @Nonnull
     Span setDestinationService(@Nullable String resource);
```
felixbarny, line 468: Should this method be deprecated? What's our recommendation for users for when to use `setDestinationService` vs `setServiceTarget`?

SylvainJuge: Yes, we can deprecate it, the behavior will be:
- `setDestinationResource(null)` behavior is preserved: resource and service target fields are dropped
- `setDestinationResource("value")` will set `resource` = `"value"`, `service.target.type` = `""` (empty value), `service.target.name` = `"value"` : just reusing the behavior defined on server [here](https://github.com/elastic/apm/blob/c19f4e9de1bb46961169b63e5aa2d52feb4d479d/specs/agents/tracing-spans-service-target.md#phase-3)
- replacing `setDestinationResource("value")` with `setServiceTarget("type", "value")` will force the user to pick a value for `type` and will change the resource to `"type/value"`, if they really want to keep it as previously they can combine it with a call to `setDestinationResource` to preserve compatibility.

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/context/ServiceTargetTest.java
```diff
@@ -0,0 +1,217 @@
...
+    @Test
+    void userValuePriority() {
+
+        ServiceTarget serviceTarget = new ServiceTarget();
+        serviceTarget.withType("type").withName("name");
+        assertThat(serviceTarget).hasType("type").hasName("name").isNotSetByUser();
+
+        assertThat(serviceTarget.withUserType("user-type")).hasType("user-type").isSetByUser();
+        assertThat(serviceTarget.withUserName("user-name")).hasName("user-name").isSetByUser();
+    }
```
eyalkoren, line 92: Test in reverse order as well (user value first) and test `withHostPortName` as well

apm-agent-core/src/test/java/co/elastic/apm/agent/impl/transaction/FastExitSpanTest.java
```diff
@@ -70,8 +71,10 @@ void testExitSpanBelowDuration() {
         assertThat(spanCount.getDropped().get()).isEqualTo(1);
 
         DroppedSpanStats droppedSpanStats = reporter.getFirstTransaction().getDroppedSpanStats();
-        assertThat(droppedSpanStats.getStats("postgresql", Outcome.SUCCESS).getCount()).isEqualTo(1);
-        assertThat(droppedSpanStats.getStats("postgresql", Outcome.SUCCESS).getSum()).isEqualTo(49_999L);
+
+        ServiceTarget st = new ServiceTarget().withType("postgresql");
+        assertThat(droppedSpanStats.getStats(st, Outcome.SUCCESS).getCount()).isEqualTo(1);
+        assertThat(droppedSpanStats.getStats(st, Outcome.SUCCESS).getSum()).isEqualTo(49_999L);
```
eyalkoren, line 77: Those are a lot of changes instead of doing `ServiceTarget st = new ServiceTarget()` within `DroppedSpanStats#getStats` and leave all tests the same

apm-agent-plugins/apm-httpclient-core/src/main/java/co/elastic/apm/agent/httpclient/HttpClientHelper.java
```diff
@@ -76,31 +76,26 @@ public static void setDestinationServiceDetails(Span span, @Nullable String sche
...
-        }
-        destination.getService().withType(EXTERNAL_TYPE);
+        span.getContext().getDestination()
+            .withAddress(host)
+            .withPort(port);
+
+        span.getContext().getServiceTarget()
+            .withType("http")
+            .withHostPortName(host, port)
+            .withNameOnlyDestinationResource();
```
eyalkoren, line 98: Should't we test `resource` as well?

apm-agent-plugins/apm-httpclient-core/src/test/java/co/elastic/apm/agent/httpclient/AbstractHttpClientInstrumentationTest.java
```diff
@@ -188,7 +188,9 @@ protected Span verifyHttpSpan(String host, String path, int status, boolean requ
         int addressEndIndex = (host.endsWith("]")) ? host.length() - 1 : host.length();
         assertThat(destination.getAddress().toString()).isEqualTo(host.substring(addressStartIndex, addressEndIndex));
         assertThat(destination.getPort()).isEqualTo(port);
-        assertThat(destination.getService().getResource().toString()).isEqualTo("%s:%d", host, port);
+
+        assertThat(span.getContext().getServiceTarget())
+            .hasDestinationResource(String.format("%s:%d", host, port));
```
eyalkoren, line 193: What about testing `name` and `type`?

apm-agent-plugins/apm-httpclient-core/src/test/java/co/elastic/apm/agent/httpclient/HttpClientHelperTest.java
```diff
@@ -50,7 +50,9 @@ void testNonDefaultPort() throws URISyntaxException {
         Span httpSpan = reporter.getFirstSpan();
         assertThat(httpSpan.getContext().getHttp().getUrl().toString()).isEqualTo("http://testing.local:1234/path?query");
         Destination destination = httpSpan.getContext().getDestination();
-        assertThat(destination.getService().getResource().toString()).isEqualTo("testing.local:1234");
+        assertThat(httpSpan.getContext().getServiceTarget())
+            .hasType("http")
+            .hasName("testing.local:1234");
```
eyalkoren, line 55: test resource

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
```diff
@@ -264,9 +264,10 @@ private void verifyDestinationFields(Span span) {
                 assertThat(destination.getPort()).describedAs("destination port is required").isGreaterThan(0);
             }
         }
-        Destination.Service service = destination.getService();
         if (checkDestinationService) {
-            assertThat(service.getResource()).describedAs("service resource is required").isNotEmpty();
+            assertThat(span.getContext().getServiceTarget())
+                .describedAs("service target is required")
+                .isNotEmpty();
```
eyalkoren, line 270: Since we still send `resource` for everything, assert for that as well

eyalkoren (commented):
Nice!

Several comments, nothing major.
I didn't go over all plugins because I saw that you used `MockReporter#checkDestinationService` to see that you didn't miss, but worth changing this flag and related methods to something like `co.elastic.apm.agent.MockReporter#checkServiceTarget`.
I did see that in many plugin tests you either replaced resource test with name and type, you didn't, but unless I am missing something, we should test both. If not - ignore.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/ServiceTarget.java
```diff
@@ -0,0 +1,189 @@
...
+
+    // only used as a buffer to avoid re-writing
+    private final StringBuilder destinationResource;
+
+    private boolean onlyNameInResource = false;
+
+    public ServiceTarget() {
+        this.name = new StringBuilder();
+        this.destinationResource = new StringBuilder();
+        resetState();
```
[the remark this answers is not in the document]
SylvainJuge: good catch, it's not required as all `boolean` fields are false by default.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/ServiceTarget.java
```diff
@@ -0,0 +1,189 @@
...
+        this.destinationResource.setLength(0);
+        this.destinationResource.append(other.destinationResource);
+        this.typeSetByUser = other.typeSetByUser;
+        this.nameSetByUser = other.nameSetByUser;
+        this.onlyNameInResource = other.onlyNameInResource;
+    }
+
+    private void invalidateCache() {
+        this.destinationResource.setLength(0); // invalidate cached value
+    }
```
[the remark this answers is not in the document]
SylvainJuge: Yes, that's a very good point, for each span we use the destination resource only once per span, so this optimization is not relevant at all.
- dropping the span, for adding it to the dropped spans metrics
- serialize the span.

apm-agent-api/src/main/java/co/elastic/apm/api/Span.java
SylvainJuge: I've removed the doc + deprecated the API, the strategy with empty `""` type have also been added to the specification.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/ServiceTarget.java
```diff
@@ -154,13 +142,15 @@ public CharSequence getDestinationResource() {
...
-        if (destinationResource.length() == 0) {
-            if (type != null && type.length() > 0) {
-                destinationResource.append(type);
-                destinationResource.append("/");
-            }
-            destinationResource.append(name);
+        // will allocate a bit, but it's fine as it's only expected to be called once
+        // - when the span is dropped for dropped spans stats
+        // - when the span is serialized
+        StringBuilder destinationResource = new StringBuilder();
```
eyalkoren, line 148: Why not keep it as a final private field and do `destinationResource.setLength(0)` here?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/DroppedSpanStats.java
```diff
@@ -78,23 +86,35 @@ public Outcome getOutcome() {
...
         @Override
         public int hashCode() {
-            return Objects.hash(serviceType, serviceName, outcome);
+            Objects.requireNonNull(serviceType);
+            int result = serviceType.hashCode();
+            result = 31 * result + CharSequenceUtils.hashCode(serviceName);
+            result = 31 * result + CharSequenceUtils.hashCode(destinationResource);
+            result = 31 * result + outcome.hashCode();
+            return result;
         }
```
eyalkoren, line 119: Quite heavy calculations for that, but looks it was similar with `Objects.hash`. Let's cache the hash so that it is only calculated once (and clear on `reset`)

docs/api-elastic.asciidoc
```diff
@@ -792,14 +792,6 @@ Sets the start timestamp of this event.
...
 * `epochMicros`: the timestamp of when this event started, in microseconds (µs) since epoch
 
-[float]
-[[api-span-set-destination-resource]]
-==== `Span setDestinationService(String resource)` added[1.25.0]
-Provides a way to manually set the span's `destination.service.resource` field, which is used for the
-construction of service maps and the identification of downstream services.
-Any value set through this method will take precedence over the automatically inferred one.
-Using `null` or empty resource string will result in the omission of this field from the span context.
-
```
eyalkoren, line 804: We should release not at the same day as the stack to make sure this didn't break cross-repo links

eyalkoren (commented):
LGTM
I made a couple of suggestions for very small but very easy perf improvements.
In addition, there are some comments from the original review that were still not addressed, mainly if the null check in serializer is relevant and whether we want to keep `todo` in code rather than have a meta issue

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/DslJsonSerializer.java
```diff
@@ -1037,19 +1079,19 @@ private void serializeDestination(Destination destination) {
...
+            serializeService(resource);
             jw.writeByte(OBJECT_END);
             jw.writeByte(COMMA);
         }
     }
 
-    private void serializeService(boolean isServiceHasContent, Destination.Service service) {
-        if (isServiceHasContent) {
+    private void serializeService( @Nullable CharSequence resource) {
+        if (resource != null) {
```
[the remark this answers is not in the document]
SylvainJuge: Same as previous, it can be `null` when the user removes it through the agent API.

SylvainJuge:
The comments should be fixed, I've opened https://github.com/elastic/apm/issues/646 to enhance spec on other types of backends.

PR #2593, opened 2022-04-25
Fix/remove tests for non-sampled transactions

eyalkoren:


eyalkoren (commented):
Besides the comments, please make sure we DO maintain tests for non-sampled transactions like:
- context propagation works for non-sampled transactions
- at least one central place to verify that we actually don't send non-sampled transactions

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
```diff
@@ -340,10 +340,10 @@ public boolean supportsLogsEndpoint() {
     }
 
     public boolean supportsKeepingUnsampledTransaction() {
-        // supportsKeepingUnsampledTransaction is called from application threads
-        // return true instead of blocking the thread
+        // Method is called from application threads thus we have to return fast to avoid blocking application threads
+        // When server version is not known we assume that it's a recent one and discard unsampled transactions.
         if (apmServerVersion != null && !apmServerVersion.isDone()) {
-            return true;
+            return false;
```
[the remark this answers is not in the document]
felixbarny: I think paying this overhead is much preferable over potential loss of functionality.
++

While the agent hasn't determined the version of the server, it has to assume that the server may need upsampled transactions.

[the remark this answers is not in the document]
eyalkoren: Worth making sure other agents are aware of this case as well.
It is not only about startup time, but in general for unknown APM Server version. 
In one of the 7.x versions there was a change in the healthcheck response format, causing the agent to fail parsing the APM Server version. We should always use backwards compatible defaults where possible.

SylvainJuge:
> Besides the comments, please make sure we DO maintain tests for non-sampled transactions like:
> 
> * context propagation works for non-sampled transactions
> * at least one central place to verify that we actually don't send non-sampled transactions

Those are already covered by those tests, thus it should be fine:
- `co.elastic.apm.agent.impl.transaction.TraceContextTest#testPropagateTransactionIdForUnsampledSpan`
- `co.elastic.apm.agent.impl.transaction.TraceContextTest#testPropagateSpanIdForSampledSpan`
- `co.elastic.apm.agent.impl.ElasticApmTracerTest#testSamplingNone_sendUnsampled`
- `co.elastic.apm.agent.impl.ElasticApmTracerTest#testSamplingNone_dropUnsampled`

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java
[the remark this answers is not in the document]
SylvainJuge: Adding this to the spec for unsampled transactions: https://github.com/elastic/apm/pull/639

apm-opentracing/src/test/java/co/elastic/apm/opentracing/OpenTracingBridgeTest.java
```diff
@@ -487,9 +494,16 @@ void testManualSampling() {
             span.finish();
         }
         transaction.finish();
-        assertThat(reporter.getTransactions()).hasSize(1);
-        assertThat(reporter.getFirstTransaction().getContext().getLabel("foo")).isNull();
-        assertThat(reporter.getSpans()).isEmpty();
+
+        if(serverSupportsUnsampled){
```
felixbarny, line 498:
```suggestion
        if (serverSupportsUnsampled) {
```

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
```diff
@@ -592,21 +590,34 @@ public void awaitUntilAsserted(ThrowingRunnable runnable) {
...
-        if (thrown != null) {
-            throw new RuntimeException(String.format("Condition not fulfilled within %d ms", timeoutMs), thrown);
+        throw new RuntimeException(String.format("Condition not fulfilled within %d ms", timeoutMs), thrown);
+    }
+
+    /**
+     * @param timeoutMs timeout of the condition
+     * @param runnable  a runnable that throws an exception when the condition is not met
+     */
+    public void awaitUntilTimeout(long timeoutMs, ThrowingRunnable runnable) {
```
felixbarny, line 613: Isn't that effectively the same as
```
Thread.sleep(timeoutMs);
runnable.run();
```

felixbarny (approved):

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
SylvainJuge: That's quite close, with the minor difference that it's failing faster when the assertion throws.

felixbarny: ah, right

PR #2602, opened 2022-05-02
Added instrumentation group to `WarmupInstrumentation`

tobiasstadler:
## What does this PR do?
`WarmupInstrumentation` should have an instrumentation group, so it can be enabled when using `enable_instrumentations` configuration option.

## Checklist
- [x] This is something else


eyalkoren:
We have a special (non documented) config option for the warmup instrumentation - `warmup_byte_buddy`, which is enabled unless configured explicitly to `false`.
What is the use case? Is it that when you set `enable_instrumentations` the warmup ones are not applied? If that's the case, the fix is not to add an instrumentation group, but to make sure it is only decided based on its special config rather than enable/disable instrumentations.

tobiasstadler:
yes, if `enable_instrumentations` is set, `WarmupInstrumentation` is not applied anymore.

eyalkoren:
I agree that's a bug. As you noticed- the `enable_instrumentations` should not determine whether we warmup byte buddy or not.
Thanks for reporting!

PR #2603, opened 2022-05-02
Set service.version in ECS-reformatting

### LINKED ISSUE #2597: Set service.version in ECS-reformatting
Once https://github.com/elastic/ecs-logging-java/pull/168 is merged and released, set `service.version` when applying auto ECS-reformatting

NOTE: current [tests](https://github.com/elastic/apm-agent-java/blob/c8e7936b1a6a897aef609b820d20e254cd1440cc/apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/test/java/co/elastic/apm/agent/loginstr/LoggingInstrumentationTest.java#L67) add the `service.version` as a custom field. As part of this PR, change that to use the new APIs.

tobiasstadler:
## What does this PR do?
Fixes #2597 

## Checklist

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -32,6 +32,7 @@ allows for spans representing asynchronous handling of requests for which the co
 
 [float]
 ===== Features
+* Set the service version when using the ECS reformatting of the application logs: {pull}2603[#2603]
 
 [float]
 ===== Bug fixes
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -24,6 +24,13 @@ endif::[]
 ==== 1.30.2 - YYYY/MM/DD
 
 [float]
+===== Potentially breaking changes
+* Starting this version, when using <<config-log-ecs-reformatting>>, the agent will automatically set the `service.version` field.
+If you are using ECS-logging and set the `service.version` through a custom field, the behaviour is not strictly defined. Remove the
+custom `service.name` field setting and either allow the agent to automatically discover and set it, or use the
+<<config-service-version>> config option to set it manually.
+
+[float]
 ===== Refactorings
 * Vert.x 3.x instrumentation was refactored to remove constructor instrumentation as well as wrapping of response handler. In addition,
 Vert.x 3.x transactions are now ended when the request end event occurs, instead of the response end event. In HTTP 2 request handling,
--- a/apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/main/java/co/elastic/apm/agent/loginstr/reformatting/AbstractEcsReformattingHelper.java
+++ b/apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/main/java/co/elastic/apm/agent/loginstr/reformatting/AbstractEcsReformattingHelper.java
@@ -152,6 +152,9 @@ public abstract class AbstractEcsReformattingHelper<A, F> {
     private final String configuredServiceName;
 
     @Nullable
+    private final String configuredServiceVersion;
+
+    @Nullable
     private final String configuredServiceNodeName;
 
     @Nullable
@@ -167,6 +170,7 @@ public abstract class AbstractEcsReformattingHelper<A, F> {
             tracer.getConfig(ServerlessConfiguration.class)
         );
         configuredServiceName = service.getName();
+        configuredServiceVersion = service.getVersion();
         if (service.getNode() != null) {
             configuredServiceNodeName = service.getNode().getName();
         } else {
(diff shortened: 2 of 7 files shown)


eyalkoren (commented):
Thanks!
See the `NOTE` in #2597 - your new code is currently not tested because we set the `service.version` through [additional fileds](https://github.com/elastic/apm-agent-java/blob/main/apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/test/java/co/elastic/apm/agent/loginstr/LoggingInstrumentationTest.java#L67).
Part of this PR should be to replace the `service.version` in theses custom fields with another field and test that both custom fields are added.
The validation already tests that `service.version` is set, so existing tests should still pass, but my [JUL ECS-reformatting PR](https://github.com/elastic/apm-agent-java/pull/2591) would fail, as it should, until I fix it.

CHANGELOG.asciidoc
```diff
@@ -32,6 +32,7 @@ allows for spans representing asynchronous handling of requests for which the co
 
 [float]
 ===== Features
+* Set the service version when using the ECS reformatting of the application logs: {pull}2603[#2603]
```
tobiasstadler, line 35: Do you think this is a breaking change for people using additional field for setting the service version?

tobiasstadler:
> Thanks!
> See the `NOTE` in #2597 - your new code is currently not tested because we set the `service.version` through [additional fileds](https://github.com/elastic/apm-agent-java/blob/main/apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/test/java/co/elastic/apm/agent/loginstr/LoggingInstrumentationTest.java#L67).

Done

> Part of this PR should be to replace the `service.version` in theses custom fields with another field and test that both custom fields are added.

Done

CHANGELOG.asciidoc
eyalkoren: Good question, depends on what happens when we set both - would we serialize the same field twice? What would be the consequences? You can try to set two different versions and see what happens.

Either way, I think adding a bullet under a `Potentially breaking changes` section in the changelog can't harm.

felixbarny: It'll depend on the JSON processor and it's configuration. I've recently enhanced the Elasticsearch [json processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/json-processor.html) to add a flag for `allow_duplicate_keys` for this reason. I don't remember how Beats handles duplicate keys. It might allow them but I'm not quite sure.

eyalkoren: So do you suggest anything other than adding a sentence to a `Potentially breaking changes` section? Such sentence would say that such additional key should be removed and either rely on automatic discovery or on agent config.

felixbarny: SGTM

eyalkoren (approved):

apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/main/java/co/elastic/apm/agent/loginstr/reformatting/AbstractEcsReformattingHelper.java
```diff
@@ -439,6 +444,18 @@ private String getServiceName() {
         return configuredServiceName;
     }
 
+    /**
+     * We currently get the same service version that is reported in the metadata document.
+     * This may mismatch automatically-discovered service version (if not configured). However, we only set it
+     * once when configuring our appender, so we can have only one service version.
```
felixbarny, line 450: This is because the auto-detection for different web apps happens after their logging setup is initialized, right?

[minor, feel free to ignore] I think the auto-detected version from a standalone jar (`META-INF/MANIFEST.mf`) would still be applied, right? Therefore, the variable `configuredServiceVersion` may be misleading as it could also be auto-discovered. Maybe rename to `globalServiceVersion`. But I see the same also applies to `configuredServiceName` so probably out of scope for this PR.

felixbarny (approved):

apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/main/java/co/elastic/apm/agent/loginstr/reformatting/AbstractEcsReformattingHelper.java
eyalkoren: Yeah, I agree. IIRC, it was really only the configured value when added and changed later to globally-auto-detected.
I'll change.

tobiasstadler:
Thank You!

PR #2606, opened 2022-05-02
AWS SDK instrumentation for S3 and DynamoDB

### LINKED ISSUE #1636: [META 408] Instrumentation for DynamoDB
Provide instrumentation support for DynamoDB.

Meta issue: https://github.com/elastic/apm/issues/408
SPEC: https://github.com/elastic/apm/issues/409

### LINKED ISSUE #1637: [META 408] Instrumentation for AWS S3
Provide instrumentation for AWS S3.

Meta issue: elastic/apm#408
SPEC: elastic/apm#409

AlexanderWert:
## What does this PR do?
Instrumentation for AWS SDK covering S3 and DynamoDB

Closes #1636 
Closes #1637

## Checklist

- [x] This is a new plugin


apm-agent-plugins/apm-aws-sdk/apm-aws-sdk-1-plugin/src/main/java/co/elastic/apm/agent/awssdk/v1/helper/DynamoDbHelper.java
```diff
@@ -0,0 +1,66 @@
...
+import javax.annotation.Nullable;
+
+public class DynamoDbHelper extends AbstractDynamoDBInstrumentationHelper<Request<?>, ExecutionContext> {
+
+    static final WeakMap<Object, String> dynamoDbRequestToTableNameMap = WeakConcurrent.buildMap();
+
+    @Nullable
+    private static DynamoDbHelper INSTANCE;
+
+    public static DynamoDbHelper getInstance() {
```
jackshirazi, line 38: Although unlikely, there could be concurrent calls that would create multiple instances, which could produce different threads using different req-table maps. This would quickly be resolved for subsequent requests, so it is only a matter of some odd initial reqs being handled incorrectly in the worst case. So can leave it like this, especially if the lazy initialization is deliberate. An alternative is to initialize at class initialization time

apm-agent-plugins/apm-aws-sdk/apm-aws-sdk-1-plugin/src/main/java/co/elastic/apm/agent/awssdk/v1/helper/DynamoDbHelper.java
```diff
@@ -0,0 +1,66 @@
...
+        if (INSTANCE == null) {
```
jackshirazi, line 39: per comment in DynamoDbHelper getInstance()

apm-agent-plugins/apm-aws-sdk/apm-aws-sdk-1-plugin/src/main/java/co/elastic/apm/agent/awssdk/v1/helper/S3Helper.java
```diff
@@ -0,0 +1,44 @@
...
+import com.amazonaws.http.ExecutionContext;
+
+import javax.annotation.Nullable;
+
+public class S3Helper extends AbstractS3InstrumentationHelper<Request<?>, ExecutionContext> {
+
+    @Nullable
+    private static S3Helper INSTANCE;
+
+    public static S3Helper getInstance() {
```
jackshirazi, line 34: per comment in DynamoDbHelper getInstance()

apm-agent-plugins/apm-aws-sdk/apm-aws-sdk-1-plugin/src/main/java/co/elastic/apm/agent/awssdk/v1/AmazonHttpClientInstrumentation.java
```diff
@@ -0,0 +1,101 @@
...
+
+
+    public static class AdviceClass {
+
+        @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
+        @Nullable
+        public static Object enterInvoke(@Advice.Argument(value = 0) Request<?> request,
+                                         @Advice.Argument(value = 3) ExecutionContext executionContext) {
+            String awsService = request.getHandlerContext(HandlerContextKey.SERVICE_ID);
+            Span span = null;
```
jackshirazi, line 70: Why is this a Span rather than a Transaction?

jackshirazi: Ignore that, it should be a span. Should it be an exit span?

apm-agent-plugins/apm-aws-sdk/apm-aws-sdk-1-plugin/src/main/java/co/elastic/apm/agent/awssdk/v1/helper/S3Helper.java
jackshirazi: actually, I think this one doesn't matter if more than one is created, there's no concurrent data issue so ignore this comment thread

jackshirazi (approved):
nice!

apm-agent-plugins/apm-aws-sdk/apm-aws-sdk-1-plugin/src/main/java/co/elastic/apm/agent/awssdk/v1/helper/DynamoDbHelper.java
AlexanderWert: I think having the lazy initialization is good to not allocate unnecessary memory in case certain parts of the SDK are not used at all (e.g. S3 but not DynamoDB). 
Regarding race, I agree that it's possible but very unlikely. And actually we do the same in other plugins already. I think introducing a synchronized block is not worth the benefit compared to the performance impact. So I would leave it as is.

PR #2610, opened 2022-05-03
Add basic span links capabilities

### LINKED ISSUE #2489: [META 605] Add span links for message tracing
[issue details](https://github.com/elastic/apm/issues/605)

[issue details](https://github.com/elastic/apm/issues/606)

eyalkoren:
## What does this PR do?
Closes #2489:
- [x] Adding basic span links capabilities
- [ ] Add span link to polling spans in all messaging plugins
- [ ] Kafka: when there is an active transaction before the beginning of the iteration - create a message-handling span for each record with a single link to the sending span, if such exists. Otherwise- create a transaction per record
- [ ] [SpringAmqpBatchMessageListenerInstrumentation](https://github.com/elastic/apm-agent-java/blob/2a60d89e8d981c34b41cf4fd417eb938662c8873/apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-spring/src/main/java/co/elastic/apm/agent/rabbitmq/SpringAmqpBatchMessageListenerInstrumentation.java) - same as Kafka (if makes sense)
- [ ] AWS Lambda functions triggered by SNS/SQS: create one transaction always and add span links
- [ ] Add to changelog and elsewhere in documentation if required

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
@@ -84,6 +84,7 @@ public class ElasticApmTracer implements Tracer {
     private final ObjectPool<Transaction> transactionPool;
     private final ObjectPool<Span> spanPool;
     private final ObjectPool<ErrorCapture> errorPool;
+    private final ObjectPool<TraceContext> spanLinkPool;
     private final Reporter reporter;
     private final ObjectPoolFactory objectPoolFactory;
     // Maintains a stack of all the activated spans/contexts
@@ -144,6 +145,9 @@ public class ElasticApmTracer implements Tracer {
 
         // we are assuming that we don't need as many errors as spans or transactions
         errorPool = poolFactory.createErrorPool(maxPooledElements / 2, this);
+
+        // span links pool allows for 10X the maximum allowed span links per span
+        spanLinkPool = poolFactory.createSpanLinkPool(AbstractSpan.MAX_ALLOWED_SPAN_LINKS * 10, this);
 
         sampler = ProbabilitySampler.of(coreConfiguration.getSampleRate().get());
         coreConfiguration.getSampleRate().addChangeListener(new ConfigurationOption.ChangeListener<Double>() {
@@ -447,6 +451,10 @@ public class ElasticApmTracer implements Tracer {
         reporter.report(error);
     }
 
+    public TraceContext createSpanLink() {
+        return spanLinkPool.createInstance();
+    }
+
     public void recycle(Transaction transaction) {
         transactionPool.recycle(transaction);
     }
@@ -457,6 +465,10 @@ public class ElasticApmTracer implements Tracer {
 
     public void recycle(ErrorCapture error) {
         errorPool.recycle(error);
+    }
+
+    public void recycle(TraceContext traceContext) {
+        spanLinkPool.recycle(traceContext);
     }
 
     public synchronized void stop() {
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
@@ -31,7 +31,9 @@ import co.elastic.apm.agent.sdk.logging.LoggerFactory;
 import co.elastic.apm.agent.util.LoggerUtils;
 
 import javax.annotation.Nullable;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -46,6 +48,7 @@ public abstract class AbstractSpan<T extends AbstractSpan<T>> implements Recycla
     public static final int PRIO_DEFAULT = 0;
     private static final Logger logger = LoggerFactory.getLogger(AbstractSpan.class);
     private static final Logger oneTimeDuplicatedEndLogger = LoggerUtils.logOnce(logger);
+    private static final Logger oneTimeMaxSpanLinksLogger = LoggerUtils.logOnce(logger);
 
     protected static final double MS_IN_MICROS = TimeUnit.MILLISECONDS.toMicros(1);
     protected final TraceContext traceContext;
@@ -114,6 +117,10 @@ public abstract class AbstractSpan<T extends AbstractSpan<T>> implements Recycla
     protected volatile String type;
 
     protected final AtomicReference<Span> bufferedSpan = new AtomicReference<>();
+
+    // Span links handling
+    public static final int MAX_ALLOWED_SPAN_LINKS = 1000;
+    private final List<TraceContext> spanLinks = new ArrayList<>();
 
     @Nullable
     private OTelSpanKind otelKind = null;
@@ -362,6 +369,46 @@ public abstract class AbstractSpan<T extends AbstractSpan<T>> implements Recycla
         return traceContext;
     }
 
+    /**
+     * Adds a span link based on the tracecontext header retrieved from the provided {@code carrier} through the provided {@code
+     * headerGetter}.
+     * @param childContextCreator the proper tracecontext inference implementation, corresponding on the header and types
+     * @param headerGetter the proper header extractor, corresponding the header and carrier types
+     * @param carrier the object from which the tracecontext header is to be retrieved
+     * @param <H> the tracecontext header type - either binary ({@code byte[]}) or textual ({@code String})
+     * @param <C> the tracecontext header carrier type, e.g. Kafka record or JMS message
+     */
+    public <H, C> void addSpanLink(TraceContext.ChildContextCreatorTwoArg<C, HeaderGetter<H, C>> childContextCreator,
+                                    HeaderGetter<H, C> headerGetter, @Nullable C carrier) {
+        if (spanLinks.size() == MAX_ALLOWED_SPAN_LINKS) {
+            oneTimeMaxSpanLinksLogger.warn("Span links for {} has reached the allowed maximum ({}). No more spans will be linked.",
+                this, MAX_ALLOWED_SPAN_LINKS);
+            return;
+        }
+        try {
+            TraceContext childTraceContext = tracer.createSpanLink();
+            if (childContextCreator.asChildOf(childTraceContext, carrier, headerGetter)) {
+                spanLinks.add(childTraceContext);
+            } else {
+                tracer.recycle(childTraceContext);
+            }
+        } catch (Exception e) {
+            logger.error(String.format("Failed to add span link to %s from header carrier %s and %s", this, carrier,
+                headerGetter.getClass().getName()), e);
+        }
+    }
+
+    /**
+     * Returns a list of links from this span to other spans in the format of child {@link TraceContext}s, of which parent is the linked
+     * span. For each entry in the returned list, the linked span {@code traceId} can be retrieved through
+     * {@link TraceContext#getTraceId()} and the {@code spanId} can be retrieved through {@link TraceContext#getParentId()}.
+     *
+     * @return a list of child {@link TraceContext}s of linked spans
+     */
+    public List<TraceContext> getSpanLinks() {
+        return spanLinks;
+    }
+
     @Override
     public void resetState() {
         finished = true;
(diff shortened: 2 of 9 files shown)


eyalkoren:
run integration tests

eyalkoren:
run integration tests

AlexanderWert (commented):
LGTM

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MessagingConfiguration.java
```diff
@@ -95,9 +112,35 @@ public boolean shouldEndMessagingTransactionOnPoll() {
         return endMessagingTransactionOnPoll.get();
     }
 
-    public enum Strategy {
+    public enum JmsStrategy {
```
SylvainJuge, line 115: [question] While it's only used for JMS at the moment, could it become relevant to other messaging systems as well ?

apm-agent-plugins/apm-kafka-plugin/apm-kafka-base-plugin/src/main/java/co/elastic/apm/agent/kafka/KafkaConsumerInstrumentation.java
```diff
@@ -47,53 +52,70 @@ public ElementMatcher<? super MethodDescription> getMethodMatcher() {
...
+
+        public static class KafkaPollExitAdvice {
+            @SuppressWarnings("unused")
+            @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class, inline = false)
+            public static void pollEnd(@Advice.Thrown final Throwable throwable) {
+
+                Span span = tracer.getActiveSpan();
+                if (span != null &&
+                    span.getSubtype() != null && span.getSubtype().equals("kafka") &&
+                    span.getAction() != null && span.getAction().equals("poll")
```
SylvainJuge, line 114:
```suggestion
                    "kafka".equals(span.getSubtype()) &&
                    "poll".equals(span.getAction())
```

apm-agent-plugins/apm-kafka-plugin/apm-kafka-headers-plugin/src/main/java/co/elastic/apm/agent/kafka/NewKafkaPollExitInstrumentation.java
```diff
@@ -0,0 +1,65 @@
...
+
+        @SuppressWarnings("unused")
+        @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class, inline = false)
+        public static void pollEnd(@Advice.Thrown final Throwable throwable,
+                                   @Advice.Return @Nullable ConsumerRecords<?, ?> records) {
+
+            Span span = tracer.getActiveSpan();
+            if (span != null &&
+                span.getSubtype() != null && span.getSubtype().equals("kafka") &&
+                span.getAction() != null && span.getAction().equals("poll")
```
SylvainJuge, line 57:
```suggestion
                "kafka".equals(span.getSubtype()) &&
                "poll".equals(span.getAction())
```
Also, given the behavior here is relying on having the `poll` action, it could be relevant to introduce a constant to make sure that any later modification will make it obvious that the multiple `"poll"` strings are tied together.

apm-agent-plugins/apm-kafka-plugin/apm-kafka-headers-plugin/src/main/java/co/elastic/apm/agent/kafka/NewKafkaPollExitInstrumentation.java
```diff
@@ -0,0 +1,65 @@
...
+            ) {
+                helper.addSpanLinks(records, span);
+                span.captureException(throwable);
+                span.deactivate().end();
```
SylvainJuge, line 61:
```suggestion
                span.captureException(throwable)
                    .deactivate()
                    .end();
```

apm-agent-plugins/apm-kafka-plugin/apm-kafka-headers-plugin/src/test/java/co/elastic/apm/agent/kafka/KafkaClientVersionsIT.java
```diff
@@ -44,7 +44,7 @@ public static Iterable<Object[]> data() {
             // dependency version, probably due to some static initializations done by the first one loaded.
             // We use this test runner framework only so it runs from a class loader that is not the app class loader,
             // in order to verify that we don't have class visibility problems.
-            {"2.4.0"}
```
SylvainJuge, line 47: [question] would it be relevant to keep testing the `2.4.0` version ?

apm-agent-plugins/apm-kafka-plugin/apm-kafka-headers-plugin/src/test/java/co/elastic/apm/agent/kafka/KafkaIT.java
```diff
@@ -271,8 +290,8 @@ public void testIgnoreTopic() {
 
         // we expect only one span for polling the reply topic
         List<Span> spans = reporter.getSpans();
-        assertThat(spans).hasSize(1);
-        verifyPollSpanContents(spans.get(0));
+        assertThat(spans.size()).isGreaterThanOrEqualTo(1);
+        verifyPollSpanContents(spans);
```
SylvainJuge, line 294: [question] I'm not sure to understand here why we now have 1+ instead of exactly one previously.

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-spring/src/main/java/co/elastic/apm/agent/rabbitmq/SpringAmqpBatchMessageListenerInstrumentation.java
```diff
@@ -47,21 +57,73 @@ public String getAdviceClassName() {
...
-        public static List<Message> beforeOnBatch(@Advice.Argument(0) @Nullable final List<Message> messageBatch) {
-            if (!tracer.isRunning() || tracer.currentTransaction() != null || messageBatch == null) {
-                return messageBatch;
+        public static Object[] beforeOnBatch(@Advice.This Object thiz,
+                                             @Advice.Argument(0) @Nullable final List<Message> messageBatch) {
+
+            List<Message> processedBatch = messageBatch;
+            Transaction batchTransaction = null;
+
+            if (tracer.isRunning() && messageBatch != null && !messageBatch.isEmpty()) {
```
SylvainJuge, line 79: [minor] maybe return early in case where tracer is not running or batch is empty to prevent some if nesting

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-spring/src/main/java/co/elastic/apm/agent/rabbitmq/SpringAmqpBatchMessageListenerInstrumentation.java
```diff
@@ -47,21 +57,73 @@ public String getAdviceClassName() {
...
+                        batchTransaction.withType("messaging")
+                            .withName("Spring AMQP Message Batch Processing")
+                            .activate();
+                    }
+                } else {
+                    oneTimeTransactionCreationWarningLogger.warn("Unexpected active span during batch message processing start: {}",
+                        active);
+                }
+
+                active = tracer.getActive();
```
SylvainJuge, line 95: Isn't the `active` variable already set with the active span/transaction on line 80 ?

SylvainJuge (approved):
LGTM, mostly minor comments & questions.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -388,10 +438,19 @@ public void resetState() {
         userOutcome = null;
         hasCapturedExceptions = false;
         bufferedSpan.set(null);
+        recycleSpanLinks();
         otelKind = null;
         otelAttributes.clear();
     }
 
+    private void recycleSpanLinks() {
```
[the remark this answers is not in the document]
eyalkoren: This makes sense, but I think it makes more sense that extracting from the pool and recycling back are located together. Since this is a managed pool owned by the `tracer` and the logic for creating a span link fits here, I would leave it here and keep the list as "dumb" data structure as possible.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MessagingConfiguration.java
eyalkoren: It's here for VERY long, with a description saying it is specific to JMS. If we need to change it in the future, we can. Note that the config property name was not changed, even though it is not a documented one. This is only an internal `enum`

apm-agent-plugins/apm-kafka-plugin/apm-kafka-headers-plugin/src/test/java/co/elastic/apm/agent/kafka/KafkaClientVersionsIT.java
eyalkoren: See the comment above- we use this type of testing only to isolate the client class loading to provide more realistic runtime environment.
If we were to test more versions, we would need a separate module.
We have the `KafkaLegacyBrokerIT` for a similar purpose, that is disabled for this reason and in order to test that, you would need to not only disable it, but also use the same version in the dependencies.

apm-agent-plugins/apm-rabbitmq/apm-rabbitmq-spring/src/main/java/co/elastic/apm/agent/rabbitmq/SpringAmqpBatchMessageListenerInstrumentation.java
eyalkoren: Yes, but line 88 potentially activates a transaction. This basically says: "if there was an active span before, **or if we just created one and activated it**, I want to add span links in both cases".

eyalkoren: Because the flow of this method is a bit complicated and because the return value is complex one that may be affected by multiple decision branches, I prefer one return in such methods (related to the comment above), but it's definitely a mater of taste

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -459,6 +467,10 @@ public void recycle(ErrorCapture error) {
         errorPool.recycle(error);
     }
 
+    public void recycle(TraceContext traceContext) {
```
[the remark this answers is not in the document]
eyalkoren: Actually, it doesn't need to be specific for span links, it's for any potential use of pooled `TraceContext`, so I'll leave it just as is, similarly to all other `recycle` methods that differ by their argument type.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -369,6 +375,50 @@ public TraceContext getTraceContext() {
...
+     * @param childContextCreator the proper tracecontext inference implementation, corresponding on the header and types
+     * @param headerGetter the proper header extractor, corresponding the header and carrier types
+     * @param carrier the object from which the tracecontext header is to be retrieved
+     * @param <H> the tracecontext header type - either binary ({@code byte[]}) or textual ({@code String})
+     * @param <C> the tracecontext header carrier type, e.g. Kafka record or JMS message
+     * @return {@code true} if added, {@code false} otherwise
+     */
+    public <H, C> boolean addSpanLink(TraceContext.ChildContextCreatorTwoArg<C, HeaderGetter<H, C>> childContextCreator,
+                                    HeaderGetter<H, C> headerGetter, @Nullable C carrier) {
+        if (spanLinks.size() == MAX_ALLOWED_SPAN_LINKS) {
```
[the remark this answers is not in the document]
eyalkoren: I started moving it, but losing the ability to log which span has the problem seems like a big compromise only for this extreme corner case where we there are EXACTLY 1000 span links and repetitions in addition to that. Even if we encounter this somehow, this single logging is not so bad anyway.

PR #2611, opened 2022-05-03
Finalize Vert.x 3.x HTTP 1 transactions when response completes

eyalkoren:
## What does this PR do?
A followup on #2564 - in Vert.x HTTP 1 request processing, the engine considers response completion as the end of request handling. The logic introduced in #2564 ends the transaction based on request end event, thus causes mismatch between the reported transaction duration and the request handling duration measured by Vert.x. 
This PR tries to eliminate this discrepancy by ending the transaction at the same time as measured by Vert.x.

## Checklist
- [x] This is a bugfix
  - ~~[ ] I have added tests that would fail without this fix~~
  - [ ] This fix was verified by the reporting user


SylvainJuge:
Adding this for the next release as the previous PR #2564 introduced a regression on Vertx span measurements.

AlexanderWert (approved):

PR #2612, opened 2022-05-03
release: add opbeans stage

v1v:
## What does this PR do?

Stop running the `opbeans` stage in the main pipeline when a release tag is created and use the post-release pipeline to run the same stage.

### DIFF
--- a/.ci/jobs/apm-agent-java-mbp.yml
+++ b/.ci/jobs/apm-agent-java-mbp.yml
@@ -12,8 +12,8 @@
             discover-pr-forks-strategy: merge-current
             discover-pr-forks-trust: permission
             discover-pr-origin: merge-current
-            discover-tags: true
-            head-filter-regex: '(main|PR-.*|v\d+\.\d+\.\d+.*)'
+            discover-tags: false
+            head-filter-regex: '(main|PR-.*)'
             notification-context: 'apm-ci'
             repo: apm-agent-java
             repo-owner: elastic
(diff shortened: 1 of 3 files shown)


cachedout (approved):

SylvainJuge (approved):
LGTM, we can now remove building on tags for the main pipeline, if I remember correctly it's only changing `.ci/jobs/apm-agent-java-mbp.yml` to have `head-filter-regex: '(main|PR-.*)'`.

I have executed it on `1.30.1` one last time to fix the last missing opbeans update, thus it's no longer required after that (and it also creates extra useless work on CI on every release).

v1v:
> LGTM, we can now remove building on tags for the main pipeline, if I remember correctly it's only changing .ci/jobs/apm-agent-java-mbp.yml to have head-filter-regex: '(main|PR-.*)'.

Yes, let me do it

v1v:
https://github.com/elastic/apm-agent-java/pull/2612/commits/295b1601880e7741bd72ae189ed4b236d39fb341 is the proposal.

SylvainJuge:
:+1: for the proposal, it definitely helps keeping it simpler !

For the next steps, I'll suggest the following (feel free to comment if I've missed something)
- wait for this PR build to complete without errors
- merge it (actually we can auto-merge it)
- once merged check the opbeans steps and tag builds are removed on the main pipeline
- when doing the next release, double-check that opbeans is properly updated

v1v:
> tag builds are removed on the main pipeline

This wont' be immediately done, the automation takes a few minutes to apply the new changes. In any case, tags will be gone

PR #2627, opened 2022-05-10
Servlet version error catching

eyalkoren:
## What does this PR do?
Catches and orderly logs error trying to obtain Servlet API version, as reported in the [forum](https://discuss.elastic.co/t/java-apm-with-new-springboot-2-6-7-is-erroring-out-on-zuul-components/304248). This only affects agent logging for debugging purposes. Hopefully, if fails for one Servlet or only on `init`, it can still work with another Servlet or through the `service` instrumentation.

## Checklist

- [x] This is a bugfix
  - [ ] The reporting user confirmed the fix


eyalkoren:
Since this version-discovery instrumentation allocates an `Object[]` and since it is applied to `Servlet#service()`, which is heavily invoked, I also added a check that utilizes the `co.elastic.apm.agent.util.LoggerUtils.LogOnceLogger` capability to disable logging after the first usage.

PR #2628, opened 2022-05-10
Jakarta websocket upgrade to 2.1.0

eyalkoren:
Closes #2623 

The main issue it that the annotations we use in our tests were removed from `jakarta.websocket-api`, so I added a `test` dependency on `jakarta.websocket-client-api`.

The thing is that even without this issue, next upgrade of the library will fail as well due to the [static framework version check](https://github.com/elastic/apm-agent-java/blob/53204c6986e2e00ae0bc354d0aefc4259a61eb64/apm-agent-plugins/apm-jakarta-websocket-plugin/src/test/java/co/elastic/apm/agent/websocket/JakartaServerEndpointInstrumentationTest.java#L41). 
I assume the original intention was to avoid using the same `VersionUtils.getVersion()` logic in the assertion as in the instrumentation, but this will be an issue keeping it up to date. I think within the scope of this test, it is enough to see that the framework is set and has the same value you get from `VersionUtils.getVersion()`. The `VersionUtils` functionality is tested through `VersionUtilsTest`.

### DIFF
--- a/apm-agent-plugins/apm-jakarta-websocket-plugin/pom.xml
+++ b/apm-agent-plugins/apm-jakarta-websocket-plugin/pom.xml
@@ -25,8 +25,14 @@
         <dependency>
             <groupId>jakarta.websocket</groupId>
             <artifactId>jakarta.websocket-api</artifactId>
-            <version>2.0.0</version>
+            <version>2.1.0</version>
             <scope>provided</scope>
+        </dependency>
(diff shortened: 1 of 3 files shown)


jackshirazi (approved):
Agreed, the check was an extra differentiation check between javax and jakarta, and your dynamic version access using the artifact is definitely better, thanks!

PR #2629, opened 2022-05-10
added elasticsearch java client tests

### LINKED ISSUE #2172: Add a test for the new Elasticsearch Java client
We should start continuously test the new [Elasticsearch Java client](https://github.com/elastic/elasticsearch-java). 
It may be supported without any additional effort because it's built on top of the low level client.

kananindzya:
## What does this PR do?

closes #2172 

## Checklist

- [x] This is something else
 - [x] Check why error not captured when deleting non-existing index
 - [x] Found reason when searching via MSearchRequest: `reason -> {JsonStringImpl@7286} ""The msearch request must be terminated by a newline [\n]""` - I opened pull request(https://github.com/elastic/elasticsearch-java/issues/41). 
As I understand it, the error is due to the fact that in _search(and template) requests, the new client sends an array instead of new-line separated json values.
I have set the `@Ignore` annotation for these two tests so far.
 - [x] Add ElasticsearchAsyncClient to tests.
 - [x] Upgrade to the latest client version and restore the msearch tests
 - [x] Add the new client to supported-technologies page


eyalkoren:
run integration tests

apm-agent-plugins/apm-es-restclient-plugin/apm-es-java-client-test/src/test/java/co/elastic/apm/esjavaclient/AbstractElasticsearchJavaTest.java
```diff
@@ -0,0 +1,525 @@
...
+        }
+
+        assertThat(cause).isInstanceOf(ElasticsearchException.class);
+
+        ElasticsearchException ee = (ElasticsearchException) cause;
+        assertThat(ee.status()).isEqualTo(404);
+
+        assertThat(cause).isNotNull();
+
+//        no errors captured, because in this case event come to ResponseListener#onSuccess
```
eyalkoren, line 113: This is weird..
I'll check if this is not a client bug. 
In the meantime I'll leave commented out and extend span validation

eyalkoren (approved):
Thanks @kananindzya, this looks great!!
I will add some minor adjustments and test and I think we can merge this and extend/modify later based on user feedback

eyalkoren:
run elasticsearch-ci/docs

jackshirazi (approved):

apm-agent-plugins/apm-es-restclient-plugin/apm-es-java-client-test/src/test/java/co/elastic/apm/esjavaclient/AbstractElasticsearchJavaTest.java
eyalkoren: This is the response I got from the Java client team: Java API client explicitly ask the REST client to consider status codes 400,401,403,404,405 as non-error. They are handled at a higher level and may or may not lead to an exception to be thrown depending on the endpoint that was called.
Since we instrument the low level client, we can't capture such errors as the `onSuccess` will be invoked rather than `onFailure`. In order to capture that, we would need to add a different instrumentation to the Java API client, but this is out of scope for now and not necessarily something we want to do.

PR #2632, opened 2022-05-16
Non discardable api

eyalkoren:
## What does this PR do?
Adds APIs to make spans non-discardable

## Checklist
- [x] Added to public API
- [x] Added public API tests
- [x] Added to public API docs
- [x] Added to annotations API
- [x] Added annotations API tests
- [x] Added to annotations API docs
- [x] Added to OTel bridge through a custom attribute
- [x] Added OTel tests
- [x] Added to OTel docs
- [x] Added to CHANGELOG


apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/api/SpanDiscardingTest.java
```diff
@@ -0,0 +1,80 @@
...
+import co.elastic.apm.agent.configuration.CoreConfiguration;
+import co.elastic.apm.agent.configuration.converter.TimeDuration;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.doReturn;
+
+@SuppressWarnings("unused")
```
SylvainJuge, line 31: [minor] why do we have to suppress warnings here ?

apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/api/SpanDiscardingTest.java
```diff
@@ -0,0 +1,80 @@
...
+            parent(false);
+        }
+        List<co.elastic.apm.agent.impl.transaction.Span> spans = reporter.getSpans();
+        assertThat(spans).hasSize(2);
+        assertThat(spans).anyMatch(span -> span.getNameAsString().equals("parent-false"));
+        assertThat(spans).anyMatch(span -> span.getNameAsString().equals("not-discarded"));
+        transaction.end();
+    }
+
+    private void parent(boolean discardChild) {
```
SylvainJuge, line 49: [minor] maybe just for extra clarity, maybe rename it to `parentSpan`
```suggestion
    private void parentSpan(boolean discardChild) {
```

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/BehavioralAttributes.java
```diff
@@ -0,0 +1,27 @@
...
+ * under the License.
+ */
+package co.elastic.apm.agent.opentelemetry.sdk;
+
+/**
+ * An extension to the OTel API that allows us to provide special tracing settings.
+ * Such settings may affect tracing behavior internally, but they are not added as data and not persisted as span attributes.
+ */
+public class BehavioralAttributes {
+    public static final String NON_DISCARDABLE = "Elastic-OpenTelemetry-NonDiscardable";
```
SylvainJuge, line 26: [minor] OTel attribute keys are lower-cased and separated by dots, I would suggest to change the value to something like `elastic.otel.nondiscardable`
```suggestion
    public static final String NON_DISCARDABLE = "elastic.otel.nondiscardable";
```

Also, to prevent allocation it is recommended to store the attribute as an `AttributeKey` rather than use the `String` key.

SylvainJuge (approved):

apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/api/SpanDiscardingTest.java
eyalkoren: So that we don't get many unnecessary warnings about non-used scopes

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/BehavioralAttributes.java
eyalkoren: Good point! I started implementation from the builder where preventing this allocation didn't make sense, but it does in this API

SylvainJuge (approved):

docs/api-opentelemetry.asciidoc
```diff
@@ -84,6 +84,37 @@ you can use the OpenTelemetry API to get ahold of the current span and call `set
...
+
+We utilize the `setAttribute()` API not only to <<otel-set-attribute, add custom metadata>>, but also as a way to customize some
+special tracing features through corresponding custom attributes listed below. Such attributes are not added to span metadata. For example:
+
+[source,java]
+----
+Span.current().setAttribute("elastic.otel.nondiscardable", true);
+----
+
+The `co.elastic.apm.agent.opentelemetry.sdk.BehavioralAttributes` utility class provides supported attributes as keys. Using them is
```
felixbarny, line 99: We'd have to publish a dedicated module for users to be able to use this class, right?

eyalkoren: Yes, this makes no sense. I thought of making it nicer as an `enum` and realized that a big incentive for users to use OTel is avoid vendor-locking, so not very appealing.
This also worth considering about the usage of such custom "behavioral" attributes that we do not add as span metadata, but other OTel implementations would

SylvainJuge: Ideally yes, but maybe we could just simply support both:
- with a dependency on the not-yet-published class
- with an attribute that just follows the same naming convention (which allows to not have a strong dependency)

So, I suggest that for short term we only support the naming convention, then we handle this properly with a published artifact when deprecating our API and/or when we have defined more attributes in sync with other agents.

From what I remember in OTel the semantic attributes are defined in yaml, then code is generated for every platform, while we might not need to go that far and keep doing it manually, just choosing a proper name for the artifact might require some alignment with other agents.

felixbarny: The thing is that the "discardable" logic is something that's only in the Java agent. Other agents don't implement `span_min_duration`, only `exit_span_min_duration` and therefore don't need the more complex discardable handling.

felixbarny: I'd remove the section that mentions `co.elastic.apm.agent.opentelemetry.sdk.BehavioralAttributes` for now as there's no way for users to get a dependency on a module that contains this class.

eyalkoren: Sure, which means it makes more sense to be reverted back to be constant strings.
We'll followup later if we feel there is a need then with exposing somehow

eyalkoren:
@SylvainJuge please only take a look at the latest commits adding support for this in the annotations API and not reverting the exposure of the custom attributes as we discussed

SylvainJuge (approved):

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/BehavioralAttributes.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Can be used to make a span non-discardable,
+     * In some cases, spans may be discarded, for example if {@code span_min_duration} config option is set and the span does not exceed
+     * the configured threshold. Use this attribute to make sure the current span is not discarded.
+     *
+     * NOTE: making a span non-discardable implicitly makes the entire stack of active spans non-discardable as well. Child spans can
+     * still be discarded.
+     */
+    public static final String NON_DISCARDABLE = "elastic.otel.nondiscardable";
```
felixbarny, line 35: Per the [OTel naming schema recommendations](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md#recommendations-for-application-developers), I suggest using `co.elastic` as a prefix.
Also, I'd prefer discardable=false rather than nondiscardable=true
```suggestion
    public static final String DISCARDABLE = "co.elastic.discardable";
```

SylvainJuge: :+1: if you merge the suggestion make sure to also update/comment on https://github.com/elastic/apm/issues/645 so we keep a consistent proposal.

PR #2637, opened 2022-05-18
Added SQS Instrumentation

AlexanderWert:
Adds instrumentation for SQS messaging.

TODOs: handle span links


apm-agent-core/src/test/java/co/elastic/apm/agent/report/serialize/DslJsonSerializerTest.java
```diff
@@ -425,6 +425,7 @@ void testSpanDestinationContextSerialization() {
         span.getContext().getDestination().withAddress("whatever.com").withPort(80)
             .getService()
             .withResource("whatever.com:80");
+        span.getContext().getDestination().getCloud().withRegion("us-east-1");
```
jackshirazi, line 428: Because the JSON writing output is multi-branched (`hasPort || hasServiceContent || hasCloudContent`) with each of combination of true/false on these taking a different path especially on how objects terminate, I think you need to expand the testing to multiple combinations of these. Otherwise it would be easy for a maintenance change to result in the wrong JSON termination in edge cases, better to catch that here

apm-agent-plugins/apm-aws-sdk/apm-aws-sdk-common/src/main/java/co/elastic/apm/agent/awssdk/common/IAwsSdkDataSource.java
```diff
@@ -20,16 +20,30 @@
...
-    String getRegion(R sdkRequest, C context);
+    public abstract String getFieldValue(String fieldName, R sdkRequest);
 
     @Nullable
-    String getFieldValue(String fieldName, R sdkRequest, C context);
+    public String getQueueNameFromQueueUrl(@Nullable String queueUrl) {
+        if (queueUrl != null) {
+            int lastSlashIdx = queueUrl.lastIndexOf('/');
+            if (lastSlashIdx < queueUrl.length()) {
+                return queueUrl.substring(queueUrl.lastIndexOf('/') + 1);
```
jackshirazi, line 44:
```suggestion
                return queueUrl.substring(lastSlashIdx + 1);
```

apm-agent-plugins/apm-aws-sdk/apm-aws-sdk-1-plugin/src/main/java/co/elastic/apm/agent/awssdk/v1/helper/SQSHelper.java
```diff
@@ -0,0 +1,237 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class SQSHelper extends AbstractSQSInstrumentationHelper<Request<?>, ExecutionContext, Message> implements TextHeaderSetter<Map<String, MessageAttributeValue>> {
+
+    @Nullable
+    private static SQSHelper INSTANCE;
+
+    public static SQSHelper getInstance() {
+        if (INSTANCE == null) {
```
jackshirazi, line 53: why is this lazily instantiated?

AlexanderWert: removed lazy instantiation

apm-agent-plugins/apm-aws-sdk/apm-aws-sdk-1-plugin/src/main/java/co/elastic/apm/agent/awssdk/v1/helper/SQSHelper.java
```diff
@@ -0,0 +1,234 @@
...
+        }
+    }
+
+    @Override
+    protected String getMessageBody(Message sqsMessage) {
+        return sqsMessage.getBody();
+    }
+
+    @Nullable
+    protected Long getMessageAge(Message message) {
```
jackshirazi, line 95: Is this Long rather than long for a reason? The null check could be replaced with Long.MIN_VALUE then we wouldn't have any autoboxing concern (escape analysis might get rid of the box, but it might not on some JVMs)

AlexanderWert: replaced it with `long` now

jackshirazi:
Excellent work! Just need to add a CHANGELOG entry (and possibly a supported tech entry) and you're done!

jackshirazi (approved):

PR #2642, opened 2022-05-23
parse db name from JDBC string

### LINKED ISSUE #1841: Parse db.instance from JDBC URL

This is a follow-up issue on #1800 (use either `Connection#getCatalog` or `Connection#getSchema` for `db.instance` field, which is covered by pr #1820 .

In some cases mentioned in https://github.com/elastic/apm-agent-java/issues/1800#issuecomment-850195815, we can't use `getCatalog` and proper JDBC URL parsing seems the right option to populate the `db.instance` field.

- try to parse the `db.instance` value from JDBC URL.
- for Oracle, rely in the `SERVICE_NAME` attribute in JDBC URL if provided, with fallback to `SID` or `INSTANCE_NAME`
- when it can't be parsed, then we should fallback to using `Connection#getConnection` at the time when the connection is created.

While `Connection#setCatalog` (or `Connection#setSchema`) could be called at runtime to change the current catalog on the connection, this feature is assumed to be quite rarely used, and would further complicate things. Also, keeping a stable value over time that would be consistent with application configuration (here the JDBC URL) seems a reasonable option.

## Additional context

- few other Oracle examples https://github.com/elastic/apm-agent-java/issues/1800#issuecomment-854380240
- Oracle documentation https://docs.oracle.com/cd/E11882_01/network.112/e41945/concepts.htm#NETAG253

SylvainJuge:
## What does this PR do?

Fixes #1841 

This would be a "potentially breaking change" as the granularity for dependencies will be higher and include the database name, the `span.destination.service.resource` includes the database name when it is part of the JDBC connection string.

On the implementation side, there are quite a few variants to parse and because the information we collect is not always captured together (for example host name and port), it's more convenient to have a builder.

## Checklist

## Things to clarify during review
- [x] should the captured database name be case-sensitive ? For Oracle the parsing currently makes it lowercase, but other similar variables like `SID` are case-sensitive (related [stackoverflow post](https://dba.stackexchange.com/questions/267752/is-an-oracle-sid-uppercase-or-lowercase)).
  - answer: keep case-sensitive for database name, keep normalizing the host name

## Things to discuss for follow-up issues / PRs.
- [x] we already normalize the host name to lowercase (which makes sense as DNS is usually not case sensitive), should we also normalize IPv6 addresses ? we can currently have `::1` `[::1]`
- [x] should we also capture the database user from JDBC url ? answer: let's do that later maybe
- [x] We don't capture DB metadata from "standard JDBC properties" that are passed to `java.sql.DriverManager#getConnection(java.lang.String, java.util.Properties)` : not discussed, but let's implement this only if required later.
- [x] What is the best fallback when there is a parsing error ? Should we prefer a "missing database" or a "dummy name database" ? 
  - answer: better to not report any misleading db name + issue a proper error/warn in the logs..


apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaDataTest.java
```diff
@@ -33,297 +35,333 @@ class ConnectionMetaDataTest {
...
-        testUrl("jdbc:ingres://localhost:567/testdb;UID=userid;PWD=password", "ingres", "localhost", 567);
+        testUrl("jdbc:ingres://myhost:II7/testdb;UID=userid;PWD=password", "ingres", "myhost", -1, "testdb");
+        testUrl("jdbc:ingres://localhost:II7/testdb;UID=userid;PWD=password", "ingres", "localhost", -1, "testdb");
+        testUrl("jdbc:ingres://localhost:567/testdb;UID=userid;PWD=password", "ingres", "localhost", 567, "testdb");
     }
 
     @Test
     void testMysql() {
         // https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-jdbc-url-format.html
         // Single host:
```
jackshirazi, line 119: `testUrl("jdbc:p6spy:mysql://myhost:666/database", "mysql", "myhost", 666, "database");`
would fail. I know it's an edge case and why are they using p6spy when they're using our agent, but it's possible

SylvainJuge: I'm not very familiar with P6Spy implementation and need to test that to be sure, but it would likely behave exactly like the Testcontainers JDBC URLs : https://www.testcontainers.org/modules/databases/jdbc/#jdbc-url-examples (we are actually using them in `JdbcDbIT`) where the URL that we capture is provided by a call to `java.sql.Connection#getMetaData`, not the URL that is provided through the `java.sql.DriverManager#getConnection(java.lang.String, java.util.Properties)`

SylvainJuge: I just manually tested within the opbeans java app, and the agent is able to properly capture the `h2` when setting `spring.datasource.url=jdbc:p6spy:h2:mem:opbeans` in `application.properties` so I think we are good to go, but it's definitely something that we will have to change if we start to instrument at the driver level.

jackshirazi: excellent!

apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaDataTest.java
```diff
@@ -33,297 +35,333 @@ class ConnectionMetaDataTest {
...
+        testUrl("jdbc:hsqldb:http://dbserver.somedomain.com", "hsqldb", "dbserver.somedomain.com", 9001, null);
+        testUrl("jdbc:hsqldb:mem:", "hsqldb", "localhost", -1, null);
+        testUrl("jdbc:hsqldb:res:/adirectory/dbname", "hsqldb", "localhost", -1, "/adirectory/dbname");
     }
 
     @Test
     void testInvalid() {
-        testUrl("postgresql://myhost:666/database", "unknown", null, -1);
+        // invalid URL, just checking the defaults, assuming "localhost" and no port as if it would be an in-memory database
+        testUrl("", "unknown", "localhost", -1, null);
```
jackshirazi, line 343:
```suggestion
        testUrl("", "unknown", "localhost", -1, null);
        testUrl("wrong", "unknown", "localhost", -1, "wrong");
```

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
```diff
@@ -266,100 +226,135 @@ private HostPort parseAddressList(String connectionUrl) {
...
             @Override
-            ConnectionMetaData parse(String connectionUrl, String instance, String user) {
-                return ConnectionUrlParser.defaultParse(connectionUrl, dbVendor, 50000, instance, user);
+            Builder parse(String vendorUrl, Builder builder) {
+                return defaultParse(vendorUrl, builder.withPort(50000));
+            }
+
+            @Override
+            protected int getPropertiesStart(String vendorUrl) {
+                int start = 2; // ignore the leading '//' prefix
```
jackshirazi, line 290: Do you need to verify this, in case of incorrectly formatted urls?

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
```diff
@@ -678,24 +748,162 @@ static int toNumericPort(String url, String portString, int defaultPort) {
...
+
+        public ConnectionMetaData build() {
+            return new ConnectionMetaData(vendor, host, port, instance, user);
+        }
+
+        public Builder withParsingError() {
+            host = null;
+            port = -1;
+            instance = null;
+            return this;
```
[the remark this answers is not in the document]
jackshirazi: I think it needs to be present as an "unresolved" entry so that customers notice and raise it

jackshirazi (approved):
Wow that parsing looks like it was painful to implement! Well done

apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaDataTest.java
SylvainJuge: The fact that we have `wrong` as a parsed database name is not intentional, but the implementation makes it work this way.
I will leave it as-is for now as it is really a corner case, and we can revisit later if required.

apm-agent-plugins/apm-jdbc-plugin/src/main/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaData.java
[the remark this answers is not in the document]
SylvainJuge: As discussed in weekly meeting, when things get wrong it's simpler to have the following scenario from user perspective: something is missing > look in the logs for warnings/errors > ask support or open an issue

namannigam:
> ## Things to discuss for follow-up issues / PRs.
> * [x]  should we also capture the database user from JDBC url ? answer: let's do that later maybe

@SylvainJuge Wanted to understand for a [use case](https://stackoverflow.com/questions/79392003/customise-the-span-name-or-subtype-for-db-over-elastic-apm-metrics) of observability which relies on a similar approach as far as I could think of. 
Additionally, what if the database names across different SQL clusters is same? Do we recommend to ingest metrics to different APM servers?
Underlying users and tables can differ. Can IPs/domains of these cluster be a crucial metrics for the spans captured?

SylvainJuge:
The database name that is being captured is the one we get from the application perspective, so if the underlying database or the network alias (for example in docker) is identical we don't have any way to distinguish from the application side where the agent is running.

Using an opentelemetry-based agent like https://github.com/elastic/elastic-otel-java/ can help to capture more attributes on the span, however the default mapping you will see in the map and the dependencies in APM are likely to remain the same.

What you could do however is combining extra otel attributes with an ingest pipeline, or even an intermediate otel collector/processor to transform the captured data. For example, you could compute a dedicated value for `db.namespace` ([see semconv](https://opentelemetry.io/docs/specs/semconv/attributes-registry/db/)) that should be used in the APM map.

namannigam:
@SylvainJuge Thank you for the response. Please illustrate *from the application perspective*. In my case, both driver URL have different IPs, but the same database name. Is that not a primary consideration when identifying a cluster?
```
default.datasource.jdbc-url=jdbc:mysql://20.20.20.20:3306/commonDatabaseName?autoReconnect=true&useSSL=false&allowPublicKeyRetrieval=true&user=ommon
```
and 
```
default.datasource.jdbc-url=jdbc:mysql://10.10.10.10:3306/commonDatabaseName?autoReconnect=true&useSSL=false&allowPublicKeyRetrieval=true&user=common
```

SylvainJuge:
In this case the URL is different and should have a distinct network host, as we can see in the unit tests that relate to this JDBC url parsing here : https://github.com/elastic/apm-agent-java/blob/main/apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/helper/ConnectionMetaDataTest.java

However, the host part is not used when building the map and the dependencies, only the database name, hence what you see on the UI side.

What I am suggesting here is that maybe using an opentelemetry agent can help here as it would capture similar attributes (which are defined in otel semantic conventions), and because otel makes it quite easy to modify the data through a custom collector/processor then you could easily transform the attribute that currently contains the database name to include the host.

As an alternative without switching to OTel, using [ingest-pipelines with APM](https://www.elastic.co/guide/en/observability/current/apm-ingest-pipelines.html) could also be done, but this is trickier to implement as you will have to modify the traces and the metrics for this to work properly, so I would recommend trying it with OTel first.

namannigam:
@SylvainJuge Thank you again for your response. Though OTel doesn't seem straightforward for me to try out as this might need integration changes across applications, I would try out the route via ingest-pipeline.

PR #2659, opened 2022-06-02
Adding support for `latest` version to attcher cli

### LINKED ISSUE #2614: Add special support for latest version to remote attacher
Our [`AgentDownloader`](https://github.com/elastic/apm-agent-java/blob/main/apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentDownloader.java) currently constructs maven URL in the form: `https://repo1.maven.org/maven2/co/elastic/apm/elastic-apm-agent/<VERSION>`, where the `VERSION` comes from the input arguments. We want to add special handling to get the latest version. Requirements are:
1. When the requested version is `latest` - use the following URL for agent downloading: https://oss.sonatype.org/service/local/artifact/maven/redirect?r=releases&g=co.elastic.apm&a=elastic-apm-agent&v=LATEST
2. In the slim CLI, when the version argument is not provided, default to `latest`.

Related to our Fleet integration - see https://github.com/elastic/kibana/issues/129866 for more context

eyalkoren:
## What does this PR do?
Closes #2614 

## Checklist
  - [x] I have tested manually end-to-end


apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/AgentDownloaderTest.java
```diff
@@ -91,4 +94,20 @@ void testDownloadAgentAndFailVerification() throws Exception {
...
+        System.out.println("latestVersion = " + latestVersion);
+        assertThat(latestVersion).isNotEmpty();
+    }
+
+    @Test
+    void testAgentArtifactMavenPageParsing() throws IOException {
+        TreeSet<Version> versions = AgentDownloader.parseMavenArtifactHtml(AgentDownloaderTest.class.getResourceAsStream(
+            "/MavenCentral_agent_artifact.html"));
+        assertThat(versions).hasSize(50);
+        assertThat(versions.first().toString()).isEqualTo("0.5.1");
```
felixbarny, line 110:
```suggestion
        assertThat(versions).doesNotContain(Version.of("1.0.0.RC1");
        assertThat(versions.first().toString()).isEqualTo("0.5.1");
```

felixbarny (approved):

apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/AgentDownloaderTest.java
eyalkoren: No need, there are multiple lines as such and the number of parsed versions is 50, excluding those. In addition, if we would capture those, the highest version would be 1.32.0.

As an anecdote, your proposed assertion may only pass due to being doubly-erroneous 🙂 :
1. `Version.of("1.0.0.RC1")` would yield a version equals to `1.0.0`
2. The assertion would still pass because `doesNotContain` relies on `equals`, which `Version` does not implement

felixbarny: > No need, there are multiple lines as such and the number of parsed versions is 50, excluding those.

I think it would be a bit more expressive than just asserting on the number of parsed versions

> In addition, if we would capture those, the highest version would be 1.32.0.

This would be just to ensure that when we release a new RC in the future, such as 1.33.0.RC1, this will not be the `latest` version.

But it's already implicitly tested by checking the number of parsed versions, as you said.

> As an anecdote, your proposed assertion may only pass due to being doubly-erroneous 🙂 :

Ouch, that's surprising. A little strange that `Version` implements `Comparable` but not `equals`.

Don't take this as blocking this PR. I though it was an easy addition which would have made the test a little more expressive. But it seems it would require more work than justified.

eyalkoren: > This would be just to ensure that when we release a new RC in the future, such as 1.33.0.RC1, this will not be the latest version.

Unless I misunderstand what you mean, I think that's what I did by adding `1.32.0-RC1` and `1.32.0.RC1` and asserting that the highest version is 1.31.0.

> A little strange that Version implements Comparable but not equals.

Somewhat indeed. I assume it's because I naturally link `equals` to the ability to use within hashing data structures. Adding `equals` is as easy as checking whether `compareTo(other) == 0`, but better add it together with `hashCode` and some unit tests, so I agree not justified for this PR

PR #2660, opened 2022-06-02
remove transaction.status from otel transactions

SylvainJuge:
## What does this PR do?

Removes the `transaction.result` field value for transactions that are created through the OTel bridge.

When the value is set in the intake protocol, APM server currently keeps the provided value, removing it allows to infer the value from OTel attributes values and stay consistent with native OTLP intake.

## Checklist

- [x] This is a bugfix
  - [x] backport changes to Spec + OTel bridge gherkin to upstream APM repo : https://github.com/elastic/apm/pull/649

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -38,7 +38,8 @@ the agent will now include the database name in the dependency, thus `mysql/my-d
 
 [float]
 ===== Bug fixes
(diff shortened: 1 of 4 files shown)


felixbarny (approved):
LGTM. Does this require changes in the spec?

SylvainJuge:
> Does this require changes in the spec?

Yes : https://github.com/elastic/apm/pull/649

PR #2665, opened 2022-06-07
Maven version parse xml rather than html

felixbarny:
Consider the second commit optional. I can remove it if you don't like it.


apm-agent-common/src/test/java/co/elastic/apm/agent/util/VersionTest.java
```diff
@@ -19,31 +19,49 @@
...
+    @CsvSource(value = {
+        "1.0.0,                          <, 1.0.1",
+        "1.2.3,                          =, 1.2.3",
+        "1.2.3-SNAPSHOT,                 <, 1.2.3",
+        "4.5.13,                         =, 4.5.13",
+        "4.5.13.redhat-00001,            <, 4.5.13",
+        "httpclient-4.5.13,              =, 4.5.13",
+        "httpclient-4.5.13.redhat-00001, <, 4.5.13",
+        "httpclient.4.5.13.redhat-00001, =, 4.5.13.redhat-00001",
+        "httpclient.4.5.13.redhat,       <, 4.5.13",
```
eyalkoren, line 39: Why? I am not sure about the rationale of this

apm-agent-common/src/main/java/co/elastic/apm/agent/util/Version.java
```diff
@@ -70,7 +74,7 @@ public int compareTo(Version another) {
                 return left < right ? -1 : 1;
             }
         }
-        return 0;
+        return another.suffix.compareTo(suffix);
```
eyalkoren, line 77: I don't think this is something to rely on.
At most, consider one that `hasSuffix()` as lower than one that doesn't, but even that is not so solid.

eyalkoren (commented):
I do like the parsing and testing touches, but I think relying on suffix for ordering is not solid.
While this is obvious:
```
Version.of(1.2.3-SNAPSHOT).equals(Version.of(1.2.3-RC0)) == false
```
This is just arbitrary:
```
Version.of(1.2.3-SNAPSHOT).compareTo(Version.of(1.2.3-RC0)) </> 0
```
The problem is that making those inconsistent could cause trouble specifically in sorted sets (see the [`Comparable`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Comparable.html) javadoc).

Note that the parsing logic we have now in `Version` is not necessarily about identifying version prefixes and suffixes as you think about them, but a way to extract the version part out of a string, for example, such that comes from a jar manifest. The redhat httpclient example is real. 

Bottom line, I think it's best to make both equality check and ordering rely only on the version numbers, but still maintain the `hasSuffix` and use it as you already did.

apm-agent-common/src/test/java/co/elastic/apm/agent/util/VersionTest.java
felixbarny: Any suffix is considered lower than no suffix, as it usually means a pre-release. But yeah, in this case, we probably want `httpclient.4.5.13.redhat = 4.5.13`, right? Because that's important for determining whether or not to instrument a library.

felixbarny:
> The problem is that making those inconsistent could cause trouble specifically in sorted sets (see the [Comparable](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Comparable.html) javadoc).

I don't quite get this part as the sort is consistent with equals. While 1.2.3-SNAPSHOT > 1.2.3-RC0 may be arbitrary, it is consistent.

But yeah, overall it seems it doesn't quite work out due to the `.redhat` suffix thing.

eyalkoren:
> I don't quite get this part as the sort is consistent with equals. While 1.2.3-SNAPSHOT > 1.2.3-RC0 may be arbitrary, it is consistent.

Sorry for not being clear enough. IMO, the best way to reflect versioning equality/comparison logic is:
```
Version.of(1.2.3-SNAPSHOT).equals(Version.of(1.2.3-RC0)) == false;
Version.of(1.2.3-SNAPSHOT).compareTo(Version.of(1.2.3-RC0)) == 0;
```
However, I think we should avoid that due to the inconsistency issue.

apm-agent-common/src/test/java/co/elastic/apm/agent/util/VersionTest.java
eyalkoren: > Any suffix is considered lower than no suffix, as it usually means a pre-release

`Any` doesn't fit well with `usually`, because of the exceptions 🙂 .
The thing is that suffix is not necessarily `RC1` or `SNAPSHOT`, in which case ordering will be arbitrary. Moreover, as you say - for the main purpose we use it for (determine behavior based on version), `1.2.3.RC1 == 1.2.3`.

felixbarny:
Gotcha, I created a method that lets you remove suffixes. I think that's the best of both worlds. We can ignore suffixes in the element matcher where we know we have to deal with unorthodox `.redhat` versions.

eyalkoren (approved):

apm-agent-common/src/test/java/co/elastic/apm/agent/util/VersionTest.java
felixbarny: Our Version class supports both semver and non-semver-compliant versions now. Semver specifies the [ordering of versions](https://semver.org/#spec-item-11), which we comply with.

If a version is known to potentially not be semver compliant, such as 4.5.13.redhat, we can just call `Version::withoutSuffix` and rely on only comparing the version numbers.

PR #2677, opened 2022-06-20
Mark Java 7 support as deprecated

eyalkoren:


docs/supported-technologies.asciidoc
```diff
@@ -42,30 +42,35 @@ the agent does not capture transactions.
 [[supported-java-versions]]
 === Java versions
 
+
+NOTE: As of version 1.33.0, Java 7 support is deprecated and will be removed in a future release
```
felixbarny, line 46: We have a very similar note at the top and at the bottom of the table now. Maybe merge the two?

eyalkoren: Yeah, that's on purpose. The `NOTE` section is what we really need, but as we know people may skip those, so I marked 7 versions within the table itself, which any user would probably look in more carefully.
Since dropping support for a Java version is the most breaking change we can think of, I don't mind having this duplication, but if you feel we better off without it after this explanation, I'll remove the second one.

felixbarny (approved):

PR #2674, opened 2022-06-16
Fix some tests on Windows

SylvainJuge:
## What does this PR do?

Fix some test failures on Windows

## Checklist

- [x] `co.elastic.apm.agent.jul.reformatting.JulEcsReformattingHelperTest`
- [ ] ~~`co.elastic.apm.agent.log4j2.Log4j2_17_1InstrumentationTest`~~ DISABLED ON WINDOWS
- [ ] ~~`co.elastic.apm.agent.log4j2.Log4j2_6SInstrumentationTest`~~ DISABLED ON WINDOWS
- [x] `git-commit-id-maven-plugin` failure - not failing the build when the plugin fails

PR #2680, opened 2022-06-22
Fix windows tests 2

eyalkoren:
Supersedes #2674 
Testing on a clean branch to see if the git failure on the former is related to branch history


eyalkoren:
Nope, same error, continue to investigate...

PR #2682, opened 2022-06-23
Fix API annotations `discardable` instrumentation

eyalkoren:
## What does this PR do?
Fixes an issue that was introduced through #2632 by the addition of the `discardable` attributes to the `@CaptureSpan` and `@Traced` annotations.

The application error looks as follows:
```
Caused by: java.lang.VerifyError: Bad type on operand stack
Exception Details:
...  
Reason:
    Type null (current frame, stack[5]) is not assignable to integer
```
The annotations instrumentation is tolerant to addition of annotation attributes, assigning such with `null`. The issue in this case is that the new attributes are of `boolean` type, which is not assignable from `null`.

## Checklist

- [x] This is a bugfix
  - [x] I have updated the unit tests and Javadoc
  - [x] I have updated the API documentation
  - [ ] I have added integration tests that use fixed old API version

### DIFF
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/Traced.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/Traced.java
@@ -83,12 +83,13 @@ public @interface Traced {
      * <p>
      * By default, spans are discardable. This attribute is only relevant if the annotated method results with a span.
      * In some cases, spans may be discarded, for example if {@code span_min_duration} config option is set and the span does not exceed
-     * the configured threshold. Set this attribute to {@code false} to make the current span non-discardable.
+     * the configured threshold. Set this attribute to {@code "false"} (note that this needs to be a {@code String} and not a
+     * {@code boolean})  to make the current span non-discardable.
      * </p>
      * <p>
      * NOTE: making a span non-discardable implicitly makes the entire stack of active spans non-discardable as well. Child spans can still
      * be discarded.
      * </p>
      */
-    boolean discardable() default true;
+    String discardable() default "true";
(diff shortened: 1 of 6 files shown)


apm-agent-api/src/main/java/co/elastic/apm/api/Traced.java
```diff
@@ -83,12 +83,13 @@
...
+     * the configured threshold. Set this attribute to {@code "false"} (note that this needs to be a {@code String} and not a
+     * {@code boolean})  to make the current span non-discardable.
      * </p>
      * <p>
      * NOTE: making a span non-discardable implicitly makes the entire stack of active spans non-discardable as well. Child spans can still
      * be discarded.
      * </p>
      */
-    boolean discardable() default true;
+    String discardable() default "true";
```
felixbarny, line 94: This would be a breaking change.
Maybe there's a way to use a non-constant stack manipulation here:
https://github.com/elastic/apm-agent-java/blob/ee8c07b22816cd6ac996925825d80cbc9aa04ac0/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/bytebuddy/AnnotationValueOffsetMappingFactory.java#L50

And use the wrapper type `Boolean` as a method argument here:
https://github.com/elastic/apm-agent-java/blob/ee8c07b22816cd6ac996925825d80cbc9aa04ac0/apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/TracedInstrumentation.java#L75

felixbarny (approved):

PR #2685, opened 2022-06-24
Implement otel spanlinks api

jackshirazi:
## What does this PR do?
Implements the span links part of the OTel bridge

## Checklist
- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
+++ b/apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
@@ -98,7 +98,9 @@ class OTelSpanBuilder implements SpanBuilder {
     @Override
     public SpanBuilder addLink(SpanContext spanContext, Attributes attributes1) {
         addLink(spanContext);
-        attributes1.forEach((AttributeKey<?> k, Object v) -> attributes.put((AttributeKey<? super Object>) k, (Object) v));
+        if (attributes1 != null && !attributes1.isEmpty()) {
+            addLinkLogger.warn("Adding attributes to links is currently unsupported - the links have been added but with no attributes, the following attributes have been ignored: %s",attributes1);
+        }
         return this;
     }
 
--- a/apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
+++ b/apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
@@ -189,9 +189,6 @@ class OTelSpanBuilder implements SpanBuilder {
         span.withUserOutcome(Outcome.UNKNOWN);
 
         // Add the links to the span
-        // Presumably you'd only start the span once, so no need to
-        // worry about too much byte[] garbage ... but maybe we should
-        // pool across builders? Or maybe the OtelSpanBuilder should be recyclable?
         byte[] header = new byte[TraceContext.BINARY_FORMAT_EXPECTED_LENGTH];
         for (int i = 0; i < links.size(); i++) {
             span.addSpanLink(TraceContext.fromParentContext(), links.get(i));
--- a/apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
+++ b/apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
@@ -20,9 +20,11 @@ package co.elastic.apm.agent.opentelemetry.sdk;
 
 import co.elastic.apm.agent.impl.ElasticApmTracer;
 import co.elastic.apm.agent.impl.transaction.AbstractSpan;
+import co.elastic.apm.agent.impl.transaction.BinaryHeaderGetter;
 import co.elastic.apm.agent.impl.transaction.MultiValueMapAccessor;
 import co.elastic.apm.agent.impl.transaction.OTelSpanKind;
 import co.elastic.apm.agent.impl.transaction.Outcome;
+import co.elastic.apm.agent.impl.transaction.TraceContext;
 import co.elastic.apm.agent.impl.transaction.Transaction;
 import co.elastic.apm.agent.sdk.logging.Logger;
 import co.elastic.apm.agent.sdk.logging.LoggerFactory;
@@ -41,9 +43,12 @@ import io.opentelemetry.context.Context;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.function.BiConsumer;
 
 class OTelSpanBuilder implements SpanBuilder {
 
@@ -57,6 +62,7 @@ class OTelSpanBuilder implements SpanBuilder {
     private AbstractSpan<?> parent;
     @Nullable
     private Context remoteContext;
+    private List<byte[]> links = new ArrayList<>();
 
     @Nullable
     private SpanKind kind;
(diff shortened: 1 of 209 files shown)


apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
```diff
@@ -182,9 +190,47 @@ public Span startSpan() {
...
+        byte[] header = new byte[TraceContext.BINARY_FORMAT_EXPECTED_LENGTH];
+        for (int i = 0; i < links.size(); i+=2) {
+            byte[] traceID = links.get(i);
+            byte[] spanID = links.get(i+1);
+            TraceContext.filIdOutgoingTraceParentBinaryHeader(header, traceID, spanID);
+            span.addSpanLink(
+                TraceContext.getFromTraceContextBinaryHeaders(),
+                BinaryByteArrayAccessor.INSTANCE,
+                header);
+        }
```
eyalkoren, line 206: I think this is forcing complicated logic (and allocation) only to match the current `AbstractSpan#addSpanLink` API, which assumes the caller is dealing with a header carrier and a dedicated accessor that knows how to get it based on name.
However, if you add other `AbstractSpan#addSpanLink()` APIs, that could get much simpler.

When we looked together, I thought you could add `AbstractSpan#addSpanLink(String/byte[] traceId, String/byte[] spanId)`, that could be, in which case the implementation would grab a `TraceContext` from the object pool and fill it with the required details (probably requiring a new `TraceContext#fromParent(String/byte[] traceId, String/byte[] parentId)`).

However, now that I looked at it again, it seems it may get even simpler than that - the `OTelSpanBuilder#addLink()` APIs receive a `io.opentelemetry.api.trace.SpanContext` as the parent. IIANM, this is always expected to be our implementation - `co.elastic.apm.agent.opentelemetry.sdk.OTelSpanContext`, which holds a `TraceContext` already.  So if you add an `AbstractSpan#addSpanLink(TraceContext parent)`, you're basically done. Since `co.elastic.apm.agent.opentelemetry.sdk.OTelSpanContext` seems immutable and non recyclable, you can add the `TraceContext` it refers to directly to the links list. Otherwise, if you fear it could be leaked, you can take a new `TraceContext` from the object pool and fill it through `copyFrom`.

On the other hand, I didn't really try anything, so orry if I am missing something....

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
```diff
@@ -182,9 +190,47 @@ public Span startSpan() {
...
+        links.clear();
```
eyalkoren, line 207: Why is this required?

eyalkoren (commented):
See my comment inside- I think it can get more straightforward.
If I am wrong, I apologize.

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
eyalkoren: I just realized that since the span links are saved as children of the links, you cannot use the context itself. 
So the better API to add to `AbstractSpan` would be: 
```
public <T> boolean addSpanLink(ChildContextCreator<T> creator, T parent)
```
Of which implementation would still be extremely simple and very similar to the existing `addSpanLink`:
1. grab a `TraceContext` instance from the pool
2. initialize it through `creator.asChildOf()`

And to use it with a `TraceContext` you would do:
```
span.addSpanLink(TraceContext.fromParentContext(), parentTraceContext)
```

However, criticizing my own work- keeping the span links as children of the linked spans was a compromise of doing something less intuitive in order to avoid complicating `TraceContext`. However, I now think it makes sense to add the ability to create `TraceContext` from headers not as children and makes things more straightforward.
Let's talk about it...

jackshirazi: I've applied that advice (thanks!), have a new look

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
```diff
@@ -182,6 +188,11 @@ public Span startSpan() {
         // user outcome to provide higher priority and avoid inferring outcome from any reported exception
         span.withUserOutcome(Outcome.UNKNOWN);
 
+        // Add the links to the span
+        for (int i = 0; i < links.size(); i++) {
+            span.addSpanLink(TraceContext.fromParentContext(), links.get(i));
```
eyalkoren, line 193: We can delay the use of `((OTelSpanContext) spanContext).getElasticTraceContext()` until here.
I think it better reflects who's responsible to what - the user is responsible to make sure that at this point all OTel states are correct and it is our responsibility to take it from here.

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
```diff
@@ -86,13 +91,14 @@ public SpanBuilder setNoParent() {
...
+        links.add(((OTelSpanContext) spanContext).getElasticTraceContext());
         return this;
     }
 
     @Override
-    public SpanBuilder addLink(SpanContext spanContext, Attributes attributes) {
-        addLinkLogger.warn("The addLink API is not supported at the moment");
+    public SpanBuilder addLink(SpanContext spanContext, Attributes attributes1) {
+        addLink(spanContext);
+        attributes1.forEach((AttributeKey<?> k, Object v) -> attributes.put((AttributeKey<? super Object>) k, (Object) v));
```
eyalkoren, line 101: Is that the intention? In the OTel javadoc it only says: "the attributes of the Link". I don't know what this means - are those intended to be added to the built span attributes?

eyalkoren (commented):
Thanks for applying the suggestion, looks good!
Only made a tiny proposal and a question about link attributes

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
felixbarny: These shouldn't be added to the span attributes as they're scoped to a particular span link. However, we don't have support for span link attributes (an no OTel auto instrumentation seems to use span link attributes yet). So we can just ignore the span link attributes for now.

jackshirazi: In this case I think this method should revert to the API not supported rather than add the link and ignore the attributes. That way it's clear to the user they shouldn't use it

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
```diff
@@ -58,6 +61,8 @@ class OTelSpanBuilder implements SpanBuilder {
     @Nullable
     private Context remoteContext;
 
+    private List<TraceContext> links = new ArrayList<>();
```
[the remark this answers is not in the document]
jackshirazi: okay, will do

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
jackshirazi: agreed

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-plugin/src/main/java/co/elastic/apm/agent/opentelemetry/sdk/OTelSpanBuilder.java
eyalkoren: The only down side to that is that users the use it currently with `null` or empty attributes won't be able to migrate to using us.
What about logging a warning once, saying these attributes will be ignored?

jackshirazi: Yes, agree. I went with logging the warning every time, it's what we did previously and better to make that clear alongside the call when investigating

eyalkoren (approved):

PR #2688, opened 2022-06-27
Allow failing due to `git-commit-id-maven-plugin`

eyalkoren:
For failure reproducing purpose

### DIFF
--- a/apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
+++ b/apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
@@ -39,6 +39,7 @@ import com.networknt.schema.JsonSchema;
 import com.networknt.schema.JsonSchemaFactory;
 import com.networknt.schema.ValidationMessage;
 import org.awaitility.core.ThrowingRunnable;
+import org.junit.jupiter.api.condition.OS;
 import org.stagemonitor.configuration.ConfigurationRegistry;
 import specs.TestJsonSpec;
 
@@ -96,6 +97,8 @@ public class MockReporter implements Reporter {
     private final boolean verifyJsonSchema;
 
     private boolean closed;
+
+    private static final long DEFAULT_ASSERTION_TIMEOUT = OS.WINDOWS.isCurrentOs() ? 3000 : 1000;
 
     private static final JsonNode SPAN_TYPES_SPEC = TestJsonSpec.getJson("span_types.json");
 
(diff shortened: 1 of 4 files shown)


elastic-apm-agent/pom.xml
```diff
@@ -197,7 +197,7 @@
                         </goals>
                         <configuration>
                             <verbose>true</verbose>
-                            <failOnUnableToExtractRepoInfo>false</failOnUnableToExtractRepoInfo>
+                            <failOnUnableToExtractRepoInfo>true</failOnUnableToExtractRepoInfo>
```
SylvainJuge, line 200: The default is `true`, thus we can probably remove the config option completely.

Jenkinsfile
```diff
@@ -71,8 +71,8 @@ pipeline {
       steps {
         pipelineManager([ cancelPreviousRunningBuilds: [ when: 'PR' ] ])
         deleteDir()
-        gitCheckout(basedir: "${BASE_DIR}", githubNotifyFirstTimeContributor: true, shallow: false,
-                    reference: '/var/lib/jenkins/.git-references/apm-agent-java.git')
+        // reference repo causes issues while running on Windows with the git-commit-id-maven-plugin
```
SylvainJuge, line 74: [question] the `reference` option seems to be an optimization for faster clone/checkout, is there any noticeable impact here ? If there is one, then maybe using a separate checkout on Windows with the same commit ID could be an option.

apm-agent-core/src/test/java/co/elastic/apm/agent/MockReporter.java
```diff
@@ -97,6 +98,8 @@ public class MockReporter implements Reporter {
 
     private boolean closed;
 
+    private static final long DEFAULT_ASSERTION_TIMEOUT = OS.WINDOWS.isCurrentOs() ? 3000 : 1000;
```
SylvainJuge, line 101: [minor] just add a comment to help future codebase archeology on the intent here.
```suggestion
    // we have to use a longer timeout on Windows to help reduce flakyness
    private static final long DEFAULT_ASSERTION_TIMEOUT = OS.WINDOWS.isCurrentOs() ? 3000 : 1000;
```

SylvainJuge (approved):

elastic-apm-agent/pom.xml
eyalkoren: Yes, but if and when we bump into issues with this plugin again, we won't have to look it up again

PR #2692, opened 2022-06-28
Removing jakarta from dependabot

eyalkoren:
Closes #2691


SylvainJuge (approved):
LGTM, we only expect 6.x versions from now on thus it makes sense to avoid all those dependabot updates.

PR #2693, opened 2022-07-04
Enforcing correct CI workers for benchmarks

eyalkoren:
Context: benchmarks script failing on new CI processor - https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/main/285/pipeline:
```
CPU_MODEL='12th Gen Intel(R) Core(TM) i5-12500 '
...
Cannot determine base frequency for CPU model [12th Gen Intel(R) Core(TM) i5-12500 ]. Please adjust the build script.
```

Based on Intel's processor's [spec](https://ark.intel.com/content/www/us/en/ark/products/96144/intel-core-i512500-processor-18m-cache-up-to-4-60-ghz.html).

@kuisathaverat Do you think a `ci:benchmarks` label to tell CI to run benchmarks for PRs would be useful?

### DIFF
--- a/scripts/jenkins/run-benchmarks.sh
+++ b/scripts/jenkins/run-benchmarks.sh
@@ -28,6 +28,10 @@ function setUp() {
     then
         CORE_INDEX=7
         BASE_FREQ="1.9GHz"
+    elif [ "${CPU_MODEL}" == "12th Gen Intel(R) Core(TM) i5-12500 " ]
+    then
+        CORE_INDEX=11
+        BASE_FREQ="3.0GHz"
     else
         >&2 echo "Cannot determine base frequency for CPU model [${CPU_MODEL}]. Please adjust the build script."
         exit 1
--- a/scripts/jenkins/run-benchmarks.sh
+++ b/scripts/jenkins/run-benchmarks.sh
@@ -28,10 +28,6 @@ function setUp() {
     then
         CORE_INDEX=7
         BASE_FREQ="1.9GHz"
-    elif [ "${CPU_MODEL}" == "12th Gen Intel(R) Core(TM) i5-12500 " ]
-    then
-        CORE_INDEX=11
-        BASE_FREQ="3.0GHz"
     else
         >&2 echo "Cannot determine base frequency for CPU model [${CPU_MODEL}]. Please adjust the build script."
         exit 1
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -55,6 +55,7 @@ pipeline {
     booleanParam(name: 'end_to_end_tests_ci', defaultValue: false, description: 'Enable APM End-to-End tests')
 
     // disabled by default, not required for merge
+    // opt-in with 'ci:benchmarks' tag on PR
     booleanParam(name: 'bench_ci', defaultValue: false, description: 'Enable benchmarks')
 
     // disabled by default, not required for merge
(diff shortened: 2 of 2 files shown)


scripts/jenkins/run-benchmarks.sh
```diff
@@ -28,6 +28,10 @@ function setUp() {
     then
         CORE_INDEX=7
         BASE_FREQ="1.9GHz"
+    elif [ "${CPU_MODEL}" == "12th Gen Intel(R) Core(TM) i5-12500 " ]
+    then
+        CORE_INDEX=11
```
eyalkoren, line 33: Based on the other types, it seems to count the effective threads rather than physical cores, but I am not sure

eyalkoren:
run benchmark tests

felixbarny:
Why are the benchmarks suddenly executed on a new hardware? This would make any comparisons with past results meaningless.

eyalkoren:
> Why are the benchmarks suddenly executed on a new hardware? This would make any comparisons with past results meaningless.

Does this mean we cannot upgrade? Or is it enough to manually compare now and start accumulating new history?

felixbarny:
I suppose it doesn't mean that we can never upgrade but we should carefully review doing so. For example, is the new worker a dedicated bare metal machine like the old one or is it a cloud VM that may have noisy neighbours? Which background processes are running on the new worker that might interfere with the benchmark?

felixbarny:
@cachedout do you know why the benchmarks aren't executed on the dedicated bare metal machines anymore?

kuisathaverat:
> Do you think a ci:benchmarks label to tell CI to run benchmarks for PRs would be useful?

I am not a fan of using labels to trigger CI jobs, I preferred it to be intentional using a GitHub comment because these additional jobs usually do not need to run on every commit, and if you do it you will waisting CI time.

cc @v1v

v1v:
> Do you think a ci:benchmarks label to tell CI to run benchmarks for PRs would be useful?

It's not supported, https://github.com/elastic/apm-agent-java/blob/32ee207ccafe347d3fe77a65172ef809307c29f2/Jenkinsfile#L57-L58

You can enable it by using something like:

```
                expression { matchesPrLabel(label: 'ci:benchmarks') }
```

in https://github.com/elastic/apm-agent-java/blob/32ee207ccafe347d3fe77a65172ef809307c29f2/Jenkinsfile#L300-L304


> do you know why the benchmarks aren't executed on the dedicated bare metal machines anymore?

It uses [`metal`](https://github.com/elastic/apm-agent-java/blob/main/Jenkinsfile#L292) and, recently there are a bunch of new CI workers for being used by the `prodfiler`.

Unfortunately, the `metal` label was added in the CI provisioner, and the CI Jenkinsifle for this project should use something else to be sure it uses the dedicated ones.

What baremetals can you use?

- [linux&&metal](https://apm-ci.elastic.co/label/linux&&metal/)
- [benchmarks](https://apm-ci.elastic.co/label/benchmarks/ <--- that matches one worker -> https://apm-ci.elastic.co/computer/worker-1213919/ therefore I wonder if this is the one to be used instead

v1v:
@eyalkoren , I just added some changes to this PR so:
- I reverted the change for supporting a new CPU
- Fixed the agent labels to filter only those workers that were used in the past.
- Supported `ci:benchmarks` tag 
- Delete the workspace in the baremetals

Let's see how it goes, I didn't want to change the description of this PR, feel free to do it so 🙇

PR #2700, opened 2022-07-07
Enabling agent build and support for Vert.x 4.3.2

eyalkoren:
A followup on #2699


SylvainJuge:
I have fixed the instrumentation and it now works as expected for Vert.x 4.3.2.
The downside is that the 4.x plugin is now compiled and tested against only this version, and not 4.3.1 anymore.

Thus maybe the next step (maybe as a follow-up PR) here is :
- add an integration test that cover the 3.x and 4.x versions, which would allow keep testing 4.3.1.
- (optional) remove some duplicated tests for 3.x

SylvainJuge (approved):

PR #2703, opened 2022-07-07
ci: use the microbenchmarks-pool

v1v:
### What

From now on we will use the same CI workers for the benchmarking

### DIFF
--- a/.ci/load/Jenkinsfile
+++ b/.ci/load/Jenkinsfile
@@ -164,7 +164,7 @@ pipeline {
         stage('execute test') {
             parallel {
                 stage('Load generation') {
-                    agent { label 'metal' }
+                    agent { label 'linux && metal' }
                     steps {
                         echo 'Preparing load generation..'
                         unstash('scripts')
--- a/.ci/load/Jenkinsfile
+++ b/.ci/load/Jenkinsfile
@@ -164,7 +164,7 @@ pipeline {
         stage('execute test') {
             parallel {
                 stage('Load generation') {
-                    agent { label 'linux && metal' }
+                    agent { label 'microbenchmarks-pool' }
                     steps {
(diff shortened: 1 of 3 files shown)


scripts/jenkins/run-benchmarks.sh
```diff
@@ -28,6 +28,10 @@ function setUp() {
     then
         CORE_INDEX=7
         BASE_FREQ="1.9GHz"
+    elif [ "${CPU_MODEL}" == "12th Gen Intel(R) Core(TM) i5-12500 " ]
```
v1v, line 31: @eyalkoren , I cherry-picked your changes from https://github.com/elastic/apm-agent-java/pull/2693/commits/60593129312c624941832468e0b1fa60fda91714

v1v:
I assume these errors are unrelated to the nature of this change, since it ran the `Benchmarks` without any trouble

jackshirazi:
I think that one has become a flaky test, I'll re-run and investigate if it stays

.ci/load/Jenkinsfile
```diff
@@ -164,7 +164,7 @@ pipeline {
         stage('execute test') {
             parallel {
                 stage('Load generation') {
-                    agent { label 'metal' }
+                    agent { label 'linux && metal' }
```
cachedout, line 167: Why would this not go into the microbenchmarks pool?

jackshirazi (approved):

.ci/load/Jenkinsfile
v1v:
```suggestion
                    agent { label 'microbenchmarks-pool' }
```

jackshirazi (approved):

cachedout (approved):

PR #2709, opened 2022-07-11
also support SIDs when getting instance from Oracle jdbc connections

wolframhaussig:
## What does this PR do?
I found that the Kibana service map only shows services of oracle database with its name when using the service_name. When connecting using the SID it only shows as `oracle`.
Possible workaround: Set the servicename for the database and use the servicename

This PR gathers the low-hanging fruit: It reads the instance from the SID the same way as from SERVICE_NAME.  
  
To discuss: This really was the lowhanging fruit - as the SID (e.g. DB instead of servicename DB.COMPANY.COM) may not be unique. I am not sure if we should add the hostname to the instance to make it unique - e.g. host:port/SID - what o you think?

## Checklist

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,7 +28,7 @@ endif::[]
 
 [float]
 ===== Features
-* Cool new feature: {pull}2526[#2526]
+* Capture Oracle SID in connection string - {pull}2709[#2709]
 
 [float]
 ===== Bug fixes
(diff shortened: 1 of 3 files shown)


SylvainJuge (approved):
I think for now it should be fine to just capture the SID to provide better
  granularity when only SID is used.

  Regarding the non-uniqueness of the SID and the visible impact in Kibana map,
  it is something that also happens with other databases when the same
  database name is used on two distinct hosts or in a clustered database, they
  will be displayed as if it was a single node.
  This is a known limitation of the current approach, thus I think that
  merging this contribution as-is should already provide some benefit for your
  use-case.

SylvainJuge:

PR #2717, opened 2022-07-19
document the metric_set_limit option

jackshirazi:
## What does this PR do?
document the metric_set_limit option per comments in https://github.com/elastic/apm-agent-java/pull/2148 - making the option not `internal` and expand the documentation


eyalkoren (changes requested):
1. Transaction names is only one specific case of what's affecting the number of metricsets. The general factor that determine metricsets cardinality is the metrics labels (see `MetricRegistryTest`). So I propose talking about labels, but saying that it's better to reduce cardinality rather than increasing the limit is good, in which case you can mention transaction name as one option
2. You need to run `ConfigurationExporterTest` for this to be actually added to the documentation

jackshirazi:
Thanks @eyalkoren , is the latest version good enough, or still further changes needed?

eyalkoren (approved):

PR #2718, opened 2022-07-20
Fix build break lettuce505 dependency

jackshirazi:
The transitive dependency just stopped being pulled for no reason I can figure out, so added it explicitly


jackshirazi:
I pulled in extra commits from the wrong branch cos of a rebase, so redoing this

PR #2719, opened 2022-07-20
add transitive dependency which for no reason has stopped being pulled in

jackshirazi:
so added it explicitly to the test


eyalkoren (approved):

felixbarny:
> add transitive dependency which for no reason has stopped being pulled in

In the VersionIT tests, we're pulling dependencies with ivy. While ivy can use a pom.xml to pull in dependencies, doesn't support all of the features supported by maven, such as `pluginManagement` and possibly (I'm not certain if that's the case) also using property substitution/interpolation for version numbers. That's why we sometimes have to pull in transitive dependencies manually.

jackshirazi:
but doesn't explain why it suddenly stopped working when it was fine the week before and no commits to our repo could have changed how it worked

PR #2723, opened 2022-07-25
Avoid context propagation from agent threads and Executors

eyalkoren:
Disabling context propagation from agent threads and when submitting tasks/runnables/callables to agent executors.

This started by investigation of CI flakiness at `CommonsExecAsyncInstrumentationTest` (example: https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/job/main/313/) that is apparently related to our test causing metadata to start threads on which they execute processes for metadata discovery (eg `uname -n`), which invokes our process instrumentation and so forth.

But besides eliminating test issues, IIANM such exclusion of context propagation on agent threads is the right thing to do anyway.


SylvainJuge (approved):
LGTM, maybe worth a few assertions to `ExecutorUtilsTest.java` but that might be slightly overkill : `assertThat(ExecutorUtils.isAgentExecutor(...)).isTrue();`.

PR #2724, opened 2022-07-25
JUL log correlation

### LINKED ISSUE #2492: Add log correlation to JUL

We currently don't have any log correlation capabilities for JUL

JUL doesn't implement an MDC-like concept internally, so we'll have to implement it.
The idea is to add an MDC-equivalent to JUL ECS-logging, which `co.elastic.logging.jul.EcsFormatter` would use to add the trace context data to logged events. The agent will look for such within the plugin CL, so that it will be supported only if JUL is used with ECS-logging of the appropriate version.
We may or may not extend JUL ECS-logging [`MdcSupplier`](https://github.com/elastic/ecs-logging-java/blob/e8dce7ef4242862f9236cecc1aacca986615d1f9/jul-ecs-formatter/src/main/java/co/elastic/logging/jul/MdcSupplier.java).

An alternative is to look for slf4j MDC through the plugin CL and if there is such - update it. It would then be used through the ECS-logging [slf4j MDC integration](https://github.com/elastic/ecs-logging-java/blob/e8dce7ef4242862f9236cecc1aacca986615d1f9/jul-ecs-formatter/src/main/java/co/elastic/logging/jul/EcsFormatter.java#L70).
## Additional context

A followup for #2490

SylvainJuge:
## What does this PR do?

Fixes #2492 
Depends on https://github.com/elastic/ecs-logging-java/pull/190 and https://github.com/elastic/ecs-logging-java/pull/193

## Checklist
- [x] merge dependency PR on ecs-logging-java : https://github.com/elastic/ecs-logging-java/pull/193
- [x] release ecs-logging-java + use release version in this PR (build will fail until this is done)
- [x] decide if the native MDC implementation is enough for now or we have to invest more efforts on this (which will likely require another update in ecs-logging-java if we change this): for now we avoid having any MDC impl and delegate to actual tracer state through `CorrelationIdMapAdapter`.
- [x] Decide if we want/need to explicitly exclude instrumentation of internal plugins classes (currently breaks javalin tests): delegated to https://github.com/elastic/apm-agent-java/issues/2757
- [x] Test with simple app that relies on JUL (with/without agent + ECS lib)


apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/src/main/java/co/elastic/apm/agent/jul/correlation/JulEcsLogCorrelationHelper.java
```diff
@@ -0,0 +1,35 @@
...
+package co.elastic.apm.agent.jul.correlation;
+
+import co.elastic.apm.agent.loginstr.correlation.AbstractLogCorrelationHelper;
+import co.elastic.logging.jul.JulMdc;
+
+public class JulEcsLogCorrelationHelper extends AbstractLogCorrelationHelper.DefaultLogCorrelationHelper {
+
+    @Override
+    protected void addToMdc(String key, String value) {
+        JulMdc.getInstance().put(key, value);
```
felixbarny, line 28: Maybe make these methods static, similar to other MDC/ThreadContext implementations?

```suggestion
        JulMdc.put(key, value);
```

apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/src/test/java/co/elastic/apm/agent/jul/error/JulLoggerErrorCapturingInstrumentationTest.java
```diff
@@ -23,10 +23,9 @@
 import org.junit.jupiter.api.Test;
 
 import java.util.logging.Level;
+import java.util.logging.LogRecord;
 import java.util.logging.Logger;
 
-// todo: add support and enable
-@Disabled
```
eyalkoren, line 29: ❤️

apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/src/main/java/co/elastic/apm/agent/jul/error/JulLoggerErrorCapturingInstrumentation.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public static class LoggingAdvice {
+
+        private static final LoggerErrorHelper helper =new LoggerErrorHelper(LoggingAdvice.class, tracer);
+
+        @Nullable
+        @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
+        public static Object onEnter(@Advice.Argument(0) LogRecord record, @Advice.Origin Class<?> clazz) {
+            return helper.enter(record.getThrown(), clazz);
```
eyalkoren, line 67: Should be dependent on the `java.util.logging.LogRecord#getLevel`, I believe only when it is `java.util.logging.Level#SEVERE`

apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/src/test/java/co/elastic/apm/agent/jul/error/JulLoggerErrorCapturingInstrumentationTest.java
```diff
@@ -36,4 +35,12 @@ void captureErrorExceptionWithStringMessage() {
         logger.log(Level.SEVERE, "exception captured", new RuntimeException("some business exception"));
         verifyThatExceptionCaptured(1, "some business exception", RuntimeException.class);
     }
+
+    @Test
+    void captureErrorExceptionWithLogRecord() {
+        LogRecord lr = new LogRecord(Level.WARNING, "exception captured");
```
eyalkoren, line 41: Good test, but it should assert that an error is not reported 🙂 
And an additional with `java.util.logging.Level#SEVERE` would assert for an error being reported

apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/main/java/co/elastic/apm/agent/loginstr/error/LoggerErrorHelper.java
```diff
@@ -0,0 +1,58 @@
...
+                if (error != null) {
+                    error.activate();
+                }
+                return error;
+            }
+        }
+        return null;
+    }
+
+    public void exit(@Nullable Object errorCapture){
```
eyalkoren, line 51: Since it is outside the advice now, please add javadoc that this needs to be called with the value returned from the corresponding `enter` invocation.

eyalkoren (commented):
Since the log correlation instrumentation is independent of the ECS-reformatting one, we need to distinguish between two use cases:
1. ECS-logging-JUL is used as an application dependency, in which case we need to update the `JulMdc` that is loaded by the agent CL. 
2. ECS-reformatting - where we need to add to the `JulMdc` that is loaded by the agent plugin CL.

I think the proper way to achieve that is by having two instrumentations (within two plugins): 
1. one that is only applied if the application CL can load `JulMdc`. Its advice adds to the `JulMdc` that is loaded by the application class loader. In order to achieve that, you may need to create a separate plugin (so that the plugin class loader does not contain the ECS logging classes bundled with the agent) AND you may need to update [the list](https://github.com/elastic/apm-agent-java/blob/cb870b072523f83c56669ba8f4cabc3b48f79c44/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/classloading/IndyPluginClassLoader.java#L71) that says what's loaded from the agent class loader and what's not.
2. the second is only applied if the application CL cannot load `JulMdc`, in which case the advice would add to the agent-bundled `JulMdc`. This one needs to have the `JulMdc` loaded by the plugin CL, which is the case with the current logging plugin, so your `JulEcsLogCorrelationInstrumentation` should be fit for this purpose.

Sorry if I confused or mislead you, it is all based on static code analysis and better be verified with tests of both scenarios.

SylvainJuge:
> Since the log correlation instrumentation is independent of the ECS-reformatting one, we need to distinguish between two use cases:
> 
> 1. ECS-logging-JUL is used as an application dependency, in which case we need to update the `JulMdc` that is loaded by the agent CL.
> 2. ECS-reformatting - where we need to add to the `JulMdc` that is loaded by the agent plugin CL.
> 
> I think the proper way to achieve that is by having two instrumentations (within two plugins):
> 
> 1. one that is only applied if the application CL can load `JulMdc`. Its advice adds to the `JulMdc` that is loaded by the application class loader. In order to achieve that, you may need to create a separate plugin (so that the plugin class loader does not contain the ECS logging classes bundled with the agent) AND you may need to update [the list](https://github.com/elastic/apm-agent-java/blob/cb870b072523f83c56669ba8f4cabc3b48f79c44/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/classloading/IndyPluginClassLoader.java#L71) that says what's loaded from the agent class loader and what's not.
> 2. the second is only applied if the application CL cannot load `JulMdc`, in which case the advice would add to the agent-bundled `JulMdc`. This one needs to have the `JulMdc` loaded by the plugin CL, which is the case with the current logging plugin, so your `JulEcsLogCorrelationInstrumentation` should be fit for this purpose.
> 
> Sorry if I confused or mislead you, it is all based on static code analysis and better be verified with tests of both scenarios.

The way I understand it, there are three cases:
- `jul-ecs-formatter` is used by the application, without agent, the `JulMdc` is not read and is always empty
- `jul-ecs-formatter` is used by the application with the agent with log reformatting, in this case we should update the `JulMdc` that is used by the application and ignore the one internal to the agent.
- `jul-ecs-formatter` is not used by the application with the agent + log reformatting, in this case we just rely on agent JulMdc and agent EcsFormatter.

So, in a sense, we just need to alter which instance of the `JulMdc` we update depending on the scenario.

eyalkoren:
> So, in a sense, we just need to alter which instance of the JulMdc we update depending on the scenario.

Exactly, but this may be trickier than it seems. 
What I wrote above is what it would take to do in the "proper" way, where your code can directly deal with both `JulMdc` classes as types.
Another alternative is to extend your current `JulEcsLogCorrelationInstrumentation` to add the ability to update the one loaded by the app class loader by loading it explicitly through the application class loader based on class name and get `MethodHandle`s for the `put` and `remove` methods. Then you can always update the one loaded by the agent and **in addition** update the one loaded by the app CL if there is such.

SylvainJuge:
If we manage to update the `JulMdc` of the application, it means that we can make ECS logging output ((of the application) contain the `trace.id` `error.id` and `transaction.id` fields that are populated by the agent.

However, if we limit ourselves to the MDC attributes, the other ECS logging attributes `service.name`, `service.version`, ... will remain with the values configured at the application level.

In the case of overriding logs, I *think* it would be simpler to just replace the `EcsFormatter` of the application with the internal one of the agent, in other words it could be as simple as allowing `EcsFormatter` for Jul to allow delegating to another instance that would be set by the agent through instrumentation.

eyalkoren:
> However, if we limit ourselves to the MDC attributes, the other ECS logging attributes service.name, service.version, ... will remain with the values configured at the application level.

Why is that a problem? User configured values are always the best, we only decorate the logs with the tracing IDs if they user chooses that. Logging correlation is an independent feature from ECS-reformatting, even if they can cooperate.
BTW, there is also a `apm-ecs-logging-plugin` that automatically sets the discovered service name, though not implemented to JUL. 

> In the case of overriding logs, I think it would be simpler to just replace the EcsFormatter of the application with the internal one of the agent, in other words it could be as simple as allowing EcsFormatter for Jul to allow delegating to another instance that would be set by the agent through instrumentation.

But the `EcsFormatter` that the application sees is a different class than the one loaded by the agent, as they are loaded by different class loaders.

SylvainJuge:
PR updated with a different approach:
- agent has its own MDC : `AgentMDC`, which currently duplicates the one in ECS Logging
- calls to `JulMdc` within ECS logging are re-directed to `AgentMDC` through instrumentation.
- log-correlation (within the agent) only uses `AgentMDC` for reading & writing, thus there is no confusion from having two copies of `JulMdc` in two distinct classloaders.

As a result, when using ECS log formatter in the application, the agent can now automatically inject the log correlation attributes into the logs.

apm-agent-core/src/main/java/co/elastic/apm/agent/logging/AgentMDC.java
```diff
@@ -0,0 +1,77 @@
...
+import javax.annotation.Nullable;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Agent internal MDC for log-correlation to be used when instrumented library does not provide one.
+ */
+@GlobalState
+public class AgentMDC {
```
SylvainJuge, line 32: [for reviewer] this is a duplication of the `JulMdc` class within `ecs-logging-java`, duplicating allows to instrument `JulMdc` to use this class when agent is present, otherwise dealing with two distinct copies of `JulMdc` in two different classloaders (one in the app, one in the agent for ecs reformatting) is more complex and less clear.

apm-agent-attach-cli/pom.xml
```diff
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>co.elastic.logging</groupId>
             <artifactId>log4j2-ecs-layout</artifactId>
-            <version>${version.log4j2-ecs-layout}</version>
```
SylvainJuge, line 45: [for reviewer] using a single version of `ecs-logging-java` (or one of its sub-modules) should be controlled with a single property in `pom.xml`

apm-agent-plugins/apm-ecs-logging-plugin/pom.xml
```diff
@@ -19,13 +19,19 @@
...
-            <version>1.4.0</version>
+            <version>${version.ecs.logging}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>co.elastic.logging</groupId>
+            <artifactId>jul-ecs-formatter</artifactId>
+            <version>${version.ecs.logging}</version>
             <scope>provided</scope>
         </dependency>
```
SylvainJuge, line 36: [for reviewer] the dependency on `jul-ecs-formatter` might only be required at runtime or for testing, and there is no compile-time dependency on it (excepted for tests), thus maybe `test` scope would be more appropriate.

apm-agent-plugins/apm-ecs-logging-plugin/src/test/java/co/elastic/logging/jul/JulMdcAccessor.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.Map;
+
+/**
+ * Test-only accessor for {@link JulMdc#getEntries()}
+ */
+public class JulMdcAccessor {
+
+    public static Map<String, String> getEntries(){
+        return JulMdc.getEntries();
```
SylvainJuge, line 29: [for reviewer] the `getEntries` method being package-protected, this allows to provide access for tests.

apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/src/main/java/co/elastic/apm/agent/jul/correlation/JulEcsLogCorrelationInstrumentation.java
```diff
@@ -0,0 +1,57 @@
...
+            .or(named("java.util.logging.ConsoleHandler"));
+    }
+
+    @Override
+    public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+        return isPublic().and(named("publish"));
+    }
+
+    public static class AdviceClass {
+        private static final JulEcsLogCorrelationHelper helper = new JulEcsLogCorrelationHelper();
```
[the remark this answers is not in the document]
SylvainJuge: this should be covered with the new current approach with an internal MDC implementation `AgentMDC`.

apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/src/main/java/co/elastic/apm/agent/jul/reformatting/JulEcsReformattingHelper.java
```diff
@@ -86,7 +87,12 @@ protected String getAppenderName(StreamHandler handler) {
     protected Formatter createEcsFormatter(String eventDataset, @Nullable String serviceName, @Nullable String serviceVersion,
                                            @Nullable String serviceNodeName, @Nullable Map<String, String> additionalFields,
                                            Formatter originalFormatter) {
-        EcsFormatter ecsFormatter = new EcsFormatter();
+        EcsFormatter ecsFormatter = new EcsFormatter(){
+            @Override
+            protected Map<String, String> getMdcEntries() {
+                return AgentMDC.getEntries();
+            }
+        };
```
SylvainJuge, line 95: [for reviewer] this is the reason we need to allow this through subclassing in ecs-logging-java: https://github.com/elastic/ecs-logging-java/pull/193

pom.xml
```diff
@@ -113,7 +113,7 @@
         (https://nvd.nist.gov/vuln/detail/CVE-2020-9488#vulnCurrentDescriptionTitle), the SMTP appender is
          excluded from the build and not packaged into the agent artifacts -->
         <version.log4j>2.12.4</version.log4j>
-        <version.log4j2-ecs-layout>1.4.0</version.log4j2-ecs-layout>
+        <version.ecs.logging>1.4.1-SNAPSHOT</version.ecs.logging>
```
SylvainJuge, line 116: [for reviewer] using a single property for all `ecs-logging-java` artifacts, will require to be updated with a non-snapshot once ecs-logging-java is released.

apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/pom.xml
```diff
@@ -26,6 +26,11 @@
             <artifactId>jul-ecs-formatter</artifactId>
             <version>${version.ecs.logging}</version>
         </dependency>
+        <dependency>
+            <groupId>co.elastic.apm</groupId>
+            <artifactId>apm-ecs-logging-plugin</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
eyalkoren, line 33: Should this be only a testing dependency?

apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/src/main/java/co/elastic/apm/agent/jul/reformatting/JulEcsReformattingHelper.java
```diff
@@ -18,6 +18,7 @@
  */
 package co.elastic.apm.agent.jul.reformatting;
 
+import co.elastic.apm.agent.loginstr.correlation.CorrelationIdMapAdapter;
```
eyalkoren, line 21:
```suggestion
```

apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/src/main/java/co/elastic/apm/agent/jul/error/JulLoggerErrorCapturingInstrumentation.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public String getAdviceClassName() {
+        return "co.elastic.apm.agent.jul.error.JulLoggerErrorCapturingInstrumentation$LoggingAdvice";
+    }
+
+    public static class LoggingAdvice {
+
+        private static final LoggerErrorHelper helper =new LoggerErrorHelper(LoggingAdvice.class, tracer);
```
eyalkoren, line 62:
```suggestion
        private static final LoggerErrorHelper helper = new LoggerErrorHelper(LoggingAdvice.class, tracer);
```

eyalkoren:
Forgot to add to the review comment - what about trace correlation when applying ECS-reformatting? 
Do we rely on instrumenting our embedded ECS-logging classes? 
I think it would be safer to explicitly exclude ECS logging classes from instrumentation and instead create a custome `EcsFormatter` in [ECS-reformatting](https://github.com/elastic/apm-agent-java/blob/b7f997baccceb0b3b0394984187209af226a539a/apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/src/main/java/co/elastic/apm/agent/jul/reformatting/JulEcsReformattingHelper.java#L89) that implements `getMdcEntries()`.

eyalkoren (commented):
One more thing - please add JUL support for correlation and error capturing to the [supported technologies](https://www.elastic.co/guide/en/apm/agent/java/current/supported-technologies-details.html#supported-logging-frameworks) page.

apm-agent-core/src/main/java/co/elastic/apm/agent/util/ClassLoaderUtils.java
```diff
@@ -52,4 +34,14 @@ public static boolean isAgentClassLoader(@Nullable ClassLoader classLoader) {
     public static boolean isBootstrapClassLoader(@Nullable ClassLoader classLoader) {
         return classLoader == null;
     }
+
+    public static boolean isPluginClassLoader(@Nullable ClassLoader classLoader) {
+        if(classLoader == null){
```
felixbarny, line 39:
```suggestion
        if (classLoader == null) {
```

apm-agent-core/src/main/java/co/elastic/apm/agent/util/ClassLoaderUtils.java
```diff
@@ -52,4 +34,14 @@ public static boolean isAgentClassLoader(@Nullable ClassLoader classLoader) {
...
+            return false;
+        }
+        String clName = classLoader.getClass().getName();
+        return (IndyPluginClassLoader.class.getName().equals(clName)
+            || ExternalPluginClassLoader.class.getName().equals(clName));
```
felixbarny, line 44: I think we still want to instrument classes within an external plugin. Otherwise, the public API wouldn't work in the context of an external plugin.

eyalkoren: Yes, we do, so also this method should be called `isInternalPluginClassLoader`

eyalkoren: https://github.com/elastic/apm-agent-java/pull/2376#issue-1090341350 is what I vaguely remembered - so it seems there was indeed an issue with instrumenting classes that were loaded by the IndyPlugingCL.
The part that makes sure we load he API classes through the External plugin CL, so that we can instrument them is: https://github.com/elastic/apm-agent-java/blob/c8e7936b1a6a897aef609b820d20e254cd1440cc/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java#L383-L388

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -684,6 +683,7 @@ public Iterable<? extends List<Class<?>>> onError(int index, List<Class<?>> batc
                 ? new LruTypePoolCache(TypePool.Default.ReaderMode.FAST).scheduleEntryEviction()
                 : AgentBuilder.PoolStrategy.Default.FAST)
             .ignore(any(), isReflectionClassLoader())
+//            .ignore(any(), CustomElementMatchers.isInternalPluginClassLoader()) // ignore classes loaded by plugin for instrumentation
```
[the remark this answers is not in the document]
felixbarny: Then maybe for now just ignore the internal plugin class loader in the JUL instrumentation and open an issue to exclude it in general.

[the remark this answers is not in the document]
SylvainJuge: Good point, I've opened #2757 to track this.

docs/supported-technologies.asciidoc
```diff
@@ -591,8 +591,13 @@ JBoss LogManager - 1.30.0
 |JUL - `java.util.logging`
 |All supported Java versions
 |When <<config-log-ecs-reformatting, `log_ecs_reformatting`>> is enabled, logs will be automatically reformatted into
-ECS-compatible format
-|1.31.0
+ECS-compatible format.
+
+Trace correlation is only supported when used with <<config-log-ecs-reformatting, `log_ecs_reformatting`>> as JUL does
+not provide any MDC implementation.
```
felixbarny, line 597: Don't we also support log correlation when using ecs-logging-java (`co.elastic.logging.jul.EcsFormatter`) with JUL?

SylvainJuge: Yes, good point, that's exactly for this kind of small details that I love your reviews :-).

felixbarny (approved):

PR #2720, opened 2022-07-21
guard against NPE in bb warmup error handling

jackshirazi:
## Checklist

- [ ] This is an enhancement of existing features, or a new feature in existing plugins
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] I have made corresponding changes to the documentation
- [ ] This is a bugfix
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)
  - [ ] I have added tests that would fail without this fix
- [ ] This is a new plugin
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)
  - [ ] My code follows the [style guidelines of this project](https://github.com/elastic/apm-agent-java/blob/main/CONTRIBUTING.md#java-language-formatting-guidelines)
  - [ ] I have made corresponding changes to the documentation
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/main/CONTRIBUTING.md#testing) pass locally with my changes
  - [ ] I have updated [supported-technologies.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/docs/supported-technologies.asciidoc)
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] Added an instrumentation plugin? Describe how you made sure that old, non-supported versions are not instrumented by accident.
- [ ] This is something else
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)

PR #2728, opened 2022-07-28
Making ShadedClassLoader child-first

eyalkoren:
## What does this PR do?
Making the `ShadedClassLoader` work in a child-first mode, so that any agent dependency would be loaded locally even if it is somehow present in the parent CL (which is the bootstrap CL).
This should work around issues such as the one described in https://github.com/elastic/apm-agent-java/pull/2720#issuecomment-1192521307.

## A note for reviewer
I think this should be solid enough, but all possible execution routes make it complicated, so please look carefully for things I may have missed.
What I don't like is the multiple redundant lookups caused by doing child first and then, instead of directly calling to parent, calling to super, which does parent-first and ends with a local search. 
The reason I did that is not trying to deal with `null` parent (ie boot CL).

My `loadClass` implementation, for example, may go through this execution path for non-found classes:
- `loadClass`
   - `findClass`
      - `getResourceAsStream`
         - `getResource`
            - `findResource`
            - `super.getResource`
               - `parent.getResource`
               - `findResource` (second time)
- `super.loadClass`
   - `parent.loadClass`
   - `findClass` (second time)
      - `getResourceAsStream`
         - `getResource`
            - `findResource`
            - `super.getResource`
               - `parent.getResource`
               - `findResource` (third time)

That said, ending class loading with a `ClassNotFoundException` (i.e. such that is not eventually found) is a very rare occasion, so normally there would be no more than one lookup in each CL in practice

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [x] This is something else


elastic-apm-agent/src/main/java/co/elastic/apm/agent/premain/ShadedClassLoader.java
```diff
@@ -116,8 +171,8 @@ public String getPackageName(String className) {
         return null;
     }
 
-    private byte[] getClassBytes(String name) throws ClassNotFoundException {
-        try (InputStream is = getResourceAsStream(name.replace('.', '/') + CLASS_EXTENSION)) {
+    private byte[] getShadedClassBytes(String name) throws ClassNotFoundException {
```
felixbarny, line 174: Can there ever be a case where this receives a `name` that is already shaded via `getShadedResourceName`?

felixbarny (approved):

elastic-apm-agent/src/main/java/co/elastic/apm/agent/premain/ShadedClassLoader.java
eyalkoren: I don't see how. This is part of the `loadClass` execution path, where `name` is the name of the class to be loaded. Since `agent` is not part of the actual package name and since we do not relocate classes, noone should look for a class that starts with `agent.`.
`getShadedResourceName` is only called within the `getResource` path, where `name` refers to the file resource, rather than the class name.

eyalkoren:
run integration tests

eyalkoren:
run integration tests

eyalkoren:
run end-to-end tests

elastic-apm-agent/src/main/java/co/elastic/apm/agent/premain/ShadedClassLoader.java
```diff
@@ -68,9 +74,58 @@ public ShadedClassLoader(File jar, ClassLoader parent, String customPrefix) thro
...
+        synchronized (getClassLoadingLock(name)) {
+            try {
+                // First, check if the class has already been loaded
+                Class<?> c = findLoadedClass(name);
+                if (c == null) {
+                    c = findClass(name);
+                    if (resolve) {
+                        resolveClass(c);
+                    }
+                }
```
SylvainJuge, line 118: In the case where the class name does not start with `customPrefix`, could we avoid attempting the local lookup and directly call the parent instead ?

eyalkoren: Class name should never start with the custom prefix, as there is no `agent.` package in reality.
Note that everything outside this class loader should not be aware that classes are shaded, all classes are referred with their original name.

eyalkoren:
As explain [above](https://github.com/elastic/apm-agent-java/pull/2728#issue-1321200628), the full call tree described should not actually happen unless it ends with a `ClassNotFoundException`, which is rare and presents a fur bigger problem than resource lookup overhead. However, only within the `ShadedClassLoader#findClass` implementation there can be a duplicated lookup because we rely on `getResourceAsStream`, which relies on `getResource`, which starts by looking locally, then delegating to super, which starts with parent and then goes back to local lookup.
So my last commit proposes a simple and cheap way to avoid that.

PR #2735, opened 2022-08-08
Fix OTel bridge behavior with multiple classloaders + improved activation

SylvainJuge:
## Summary

When there are multiple external plugins using OTel API, or there are multiple OTel APIs that are instrumented in distinct classloaders (CL), each copy is instrumented within an isolated CL, which makes all those calls to OTel APIs unable to properly work together in the OTel bridge as they can't rely on type equality due to having isolated CLs.

For example, as it was reported in the forum, creating a transaction with an external plugin and creating spans with another external plugin would make using both not work as expected and unable to properly create child spans when the transaction is active.

On top of that, the "span upgrade" process that is currently used when wrapping Elastic spans to make them visible to OTel is flawed:
- The active Elastic span is deactivated, then wrapped for OTel, wrapped is then activated in place
- as a result the original span activation is not preserved and may lead to activation leaks: the original span is no longer active which is the opposite of what the plugin that created it might expect, and the wrapped span won't be deactivated.

## Solution proposal

- Make Elastic spans (or rather implementations of `ElasticContext`) able to store wrapped  objects with a `Map<Class<?>,ElasticContet<?>>` with wrapped object type as the key, and wrapped object as value.
- Make the "upgrade process" only create a wrapper when none is registered, this should preserve the Elastic span activation workflow.
- For allocation, we expect at most one entry per copy of the OTel API, which would be at least the number of external plugins so only a few entries in most cases.

## Checklist

- [x] validate implementation manually with test application
- [x] adding integration tests to cover this properly (might be optional depending on complexity/cost), ideally testing that all OTel APIs are instrumented and allow to see each-other spans transparently.
- [x] This is a bugfix


integration-tests/external-plugin-otel-test/external-plugin-otel-test-app/src/test/java/ExternalPluginOTelIT.java
```diff
@@ -0,0 +1,101 @@
...
+            .withCommand(String.format("java -javaagent:/tmp/agent.jar %s -jar /tmp/app.jar --wait", getAgentArgs()))
+            .waitingFor(Wait.forLogMessage(".*app end.*", 1));
+
+        try {
+            app.start();
+
+            List<String> logLines = Arrays.asList(app.getLogs().split("\n"));
+            assertThat(logLines).isNotEmpty();
+
+            assertThat(logLines).containsExactly(
```
SylvainJuge, line 57: [for review] this is a very simple way to validate proper plugin execution by relying on the fact that instrumentation plugin will print to output, proper parsing of the generated event could added later with a mock server.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java
```diff
@@ -128,6 +128,9 @@
 
     private final Map<String, Object> otelAttributes = new HashMap<>();
 
+    // TODO properly size to limit allocation
+    private final Map<Class<?>, ElasticContext<?>> contextWrappers = new HashMap<>();
```
felixbarny, line 132: The same span can be active in multiple threads at the same time. The map of context wrappers needs to be maintained by the `activeStack` in `ElasticApmTracer`. Maybe on the first time an OTel context is requested, upgrade the `Span` `ElasticContext` to a `ElasticContext` that has a context wrapper map.

API wise, it could look like `tracer.getActive(OTelBridgeContext.class, span -> OTelBridgeContext.wrapElasticActiveSpan(tracer, currentSpan))`.

SylvainJuge: The wrapped span won't be used for activation/deactivation (and we could even throw an exception if we ever try to), so in theory we should not have issues with creating or overwriting more than once for the same span (we might have to switch to a concurrent-safe collection though).

felixbarny: Consider the following scenario:

Thread 1 activates context#1 with context objects: span:span A, foo:42
Thread 2 activates context#2 with context objects: span:span A, foo:21
Thread 1 gets active context
-- expected: context#1
-- actual: context#2

The activation of context#2 overrides context#1 in the Span#contextWrappers map.
As a span can be active in multiple contexts, it's not the proper place to store the context wrapper objects.

SylvainJuge: Thanks for the very clear example here. While the context objects do not hold any state and should *in theory* be interchangeable and strictly equivalent as long as they belong to the same classloader, this would break if there is any condition or assumption using reference equality, which seems to be a common usage pattern with OTel, hence making it thread-local would avoid this potential issue.

felixbarny: > While the context objects do not hold any state 

Context objects do hold state. In my example, they contain the same span and a `ContextKey` named `foo`. Thread 1 would have expected the value `42` for the `foo` context object but it only gets half of the answer (`21`).

> making it thread-local would avoid this potential issue.

The thread local map should, however, not be in the Span object. We should only have a single place where we store thread-local state: `co.elastic.apm.agent.impl.ElasticApmTracer#activeStack`.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/ElasticContextWrapper.java
```diff
@@ -0,0 +1,122 @@
...
+    private final ElasticContext<T> context;
+
+    /**
+     * Contains other stored wrappers, one entry per type
+     */
+    private final WeakMap<Class<?>, ElasticContext<?>> contextWrappers;
+
+    public ElasticContextWrapper(int initialSize, ElasticContext<T> context) {
+        this.contextWrappers = WeakConcurrent.<Class<?>, ElasticContext<?>>weakMapBuilder()
+            .withInitialCapacity(initialSize)
```
felixbarny, line 51: This might still cause re-sizing the map as the default load factor is 0.75

felixbarny (approved):
LGTM

eyalkoren (approved):
Thanks for walking me through it @SylvainJuge 🙏

PR #2740, opened 2022-08-11
Adding null check to `AmazonHttpClientInstrumentation`

### LINKED ISSUE #2739: NullPointerException in AmazonHttpClientInstrumentation
Hi, 

We currently get an exception in the AmazonHttpClientInstrumentation class at line 71 in v 1.33.0:

```
at co.elastic.apm.agent.awssdk.v1.AmazonHttpClientInstrumentation$AdviceClass.enterInvoke(AmazonHttpClientInstrumentation.java:71)
```

Apparently the `awsService` is null in our case after `String awsService = request.getHandlerContext(HandlerContextKey.SERVICE_ID);`. I am not sure why this is and I will investigate further. Currently we can't use the latest version because of this.

eyalkoren:
Closes #2739 

Assuming that when the handler context map doesn't contain the `ServiceId` key, it is neither S3 nor DynamoDB related request. If this assumption is wrong, this doesn't close it and further investigation reqired.


felixbarny (approved):
Seems like a valid assumption

PR #2747, opened 2022-08-21
Make traceparent header creation thread-safe

eyalkoren:
## What does this PR do?
Makes the creation of the outgoing `traceparent` header thread safe.
Context: attempts to fix an issue [reported in the forum](https://discuss.elastic.co/t/trace-id-is-inserting-prohibited-character-in-header/312270).


jackshirazi (approved):
Fix matches the symptom seen of the interleaved double write of the header and will correctly prevent it. Looks good to me, nicely caught

PR #2748, opened 2022-08-22
Weblogic empty servlet path fallback

SylvainJuge:
## What does this PR do?

Improves Servlet path naming on Weblogic when `use_path_as_transaction_name=true` is set.

## Checklist

- [x] This is a bugfix

### DIFF
--- a/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletApiAdvice.java
+++ b/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletApiAdvice.java
@@ -239,10 +239,23 @@ public abstract class ServletApiAdvice {
                     }
                 }
 
-                servletTransactionHelper.onAfter(transaction, t == null ? t2 : t, adapter.isCommitted(httpServletResponse), adapter.getStatus(httpServletResponse),
-                    overrideStatusCodeOnThrowable, adapter.getMethod(httpServletRequest), parameterMap, adapter.getServletPath(httpServletRequest),
-                    adapter.getPathInfo(httpServletRequest), contentTypeHeader, true
-                );
+                String contextPath = adapter.getContextPath(adapter.getServletContext(httpServletRequest));
+                String pathInfo = adapter.getPathInfo(httpServletRequest);
+                String requestURI = adapter.getRequestURI(httpServletRequest);
+
+                String normalizedServletPath = servletTransactionHelper.normalizeServletPath(requestURI, contextPath, adapter.getServletPath(httpServletRequest), pathInfo);
+
+                servletTransactionHelper.onAfter(
+                    transaction, t == null ? t2 : t,
+                    adapter.isCommitted(httpServletResponse),
+                    adapter.getStatus(httpServletResponse),
+                    overrideStatusCodeOnThrowable,
+                    adapter.getMethod(httpServletRequest),
+                    parameterMap,
+                    normalizedServletPath,
+                    pathInfo,
+                    contentTypeHeader,
+                    true);
             }
         }
         if (span != null) {
--- a/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletApiAdvice.java
+++ b/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletApiAdvice.java
@@ -239,11 +239,14 @@ public abstract class ServletApiAdvice {
                     }
                 }
 
-                String contextPath = adapter.getContextPath(adapter.getServletContext(httpServletRequest));
+                ServletContext servletContext = adapter.getServletContext(httpServletRequest);
+                String servletPath = adapter.getServletPath(httpServletRequest);
                 String pathInfo = adapter.getPathInfo(httpServletRequest);
-                String requestURI = adapter.getRequestURI(httpServletRequest);
-
-                String normalizedServletPath = servletTransactionHelper.normalizeServletPath(requestURI, contextPath, adapter.getServletPath(httpServletRequest), pathInfo);
+                if (servletPath == null || servletPath.length() == 0 && servletContext != null) {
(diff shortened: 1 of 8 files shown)


apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletApiAdvice.java
```diff
@@ -239,10 +239,26 @@ public static <HttpServletRequest, HttpServletResponse, ServletContext, ServletC
...
                 }
 
-                servletTransactionHelper.onAfter(transaction, t == null ? t2 : t, adapter.isCommitted(httpServletResponse), adapter.getStatus(httpServletResponse),
-                    overrideStatusCodeOnThrowable, adapter.getMethod(httpServletRequest), parameterMap, adapter.getServletPath(httpServletRequest),
-                    adapter.getPathInfo(httpServletRequest), contentTypeHeader, true
-                );
+                ServletContext servletContext = adapter.getServletContext(httpServletRequest);
+                String servletPath = adapter.getServletPath(httpServletRequest);
+                String pathInfo = adapter.getPathInfo(httpServletRequest);
+                if (servletPath == null || servletPath.length() == 0 && servletContext != null) {
```
felixbarny, line 245:
```suggestion
                if ((servletPath == null || servletPath.length() == 0) && servletContext != null) {
```

IINM, it would be equivalent to `if (servletPath == null || (servletPath.length() == 0 && servletContext != null))` otherwise.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletApiAdvice.java
```diff
@@ -239,10 +239,26 @@ public static <HttpServletRequest, HttpServletResponse, ServletContext, ServletC
...
+                    String contextPath = adapter.getContextPath(servletContext);
+                    String requestURI = adapter.getRequestURI(httpServletRequest);
+                    servletPath = servletTransactionHelper.normalizeServletPath(requestURI, contextPath, servletPath, pathInfo);
+                }
+
+                servletTransactionHelper.onAfter(
```
felixbarny, line 251: Much cleaner to have every argument in a separate line. I like!

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
```diff
@@ -144,13 +144,38 @@ public static void setUsernameIfUnset(@Nullable String userName, TransactionCont
         }
     }
 
+    public String normalizeServletPath(String requestURI, @Nullable String contextPath, @Nullable String servletPath, @Nullable String pathInfo) {
+        String path = servletPath;
+        if (path == null || path.isEmpty()) {
```
felixbarny, line 149: We'd expect this case to be relatively rare, right? It does cause some allocations but if it's an edge case I think that's fine.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletApiAdvice.java
SylvainJuge: Good catch!

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
SylvainJuge: Yes, this is definitely an edge case as most of the time we have `servletPath` provided as we usually expect.

integration-tests/simple-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -27,6 +27,15 @@
         <url-pattern>/servlet</url-pattern>
     </servlet-mapping>
 
+    <servlet-mapping>
+        <!--
+        allows testing servlet mapping with wildcard, can make servlet path to be empty/null according to spec
+        but in practice this behavior relies on app server implementation and might differ.
+        -->
+        <servlet-name>TestServlet</servlet-name>
+        <url-pattern>/test-servlet/*</url-pattern>
```
felixbarny, line 36: Does this reproduce the exact issue we're seeing on Weblogic?

SylvainJuge: Not directly, if I understand the servlet spec correctly this should in theory make `servletPath` to be empty, but in practice with Tomcat this is not the case. for example we get `servletPath` = `/test-servlet` and `pathInfo` = `/info` when request URL is `/test-servlet/info` which is also a bit confusing. I am not sure it's really worth adding an integration test for that case though, I've kept it mostly for manual testing.

felixbarny (approved):
LGTM, but let's wait for the feedback before merging.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
```diff
@@ -146,30 +146,35 @@ public static void setUsernameIfUnset(@Nullable String userName, TransactionCont
...
+        if (path != null && !path.isEmpty()) {
+            return path;
+        }
 
-            if (start == end) {
-                // use complete request URI instead of an empty string as fallback
-                path = requestURI;
-            } else {
-                path = requestURI.substring(start, end);
+        logger.debug("Empty servlet path fallback applied. requestURI = {}, contextPath = {}, servletPath = {}, pathInfo = {}", requestURI, contextPath, servletPath, pathInfo);
```
felixbarny, line 153: nit: I'd guard this with a if so that the varargs array is only allocated when the debug log level is enabled.

PR #2755, opened 2022-08-24
Fix compressed span NPE

SylvainJuge:
## What does this PR do?

Attempt to fix NPE reported [in the forum](https://discuss.elastic.co/t/npe-using-apm-agent-1-33-0/312702).

## Checklist

- [x] This is a bugfix

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
@@ -275,19 +275,20 @@ public class Span extends AbstractSpan<Span> implements Recyclable {
                 return;
             }
             if (buffered == null) {
+                parent.decrementReferences();
                 if (!parent.bufferedSpan.compareAndSet(null, this)) {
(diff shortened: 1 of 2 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -284,14 +284,14 @@ protected void afterEnd() {
                 return;
             }
             if (!buffered.tryToCompress(this)) {
+                parent.decrementReferences();
                 if (parent.bufferedSpan.compareAndSet(buffered, this)) {
                     this.tracer.endSpan(buffered);
                 } else {
                     // the failed update would ideally lead to a compression attempt with the new buffer,
                     // but we're dropping the compression attempt to keep things simple
                     this.tracer.endSpan(this);
```
SylvainJuge, line 293: [for reviewer] `endSpan(this)` might indirectly trigger span recycling, which in turns makes `parent` to be set to `null`, hence triggering an NPE.

jackshirazi (approved):

PR #2760, opened 2022-08-26
implementation for Trace Continuation Strategy

jackshirazi:
## What does this PR do?
Implements Trace Continuation Strategy per https://github.com/elastic/apm/blob/main/specs/agents/trace-continuation.md

Note for reviewers. The choice of Transaction.onTransactionStart() as the sole location to apply the strategy is because

1. The headers are processed in TraceContext, so would be the ideal place to apply the strategy, but it doesn't have access to 
the span for adding span links, so the options were to
1a. Reimplement all the trace context header handling to enable that or
1b. Move up the stack - which is what I chose
2. Moving up the stack, I analyzed all the callers of TraceConext.getFromTraceContextTextHeaders() and getFromTraceContextBinaryHeaders(). These were either in tests (ignorable), or from plugins adding spanlinks (ignorable). Of the remaining two
2a ExternalSpanContextInstrumentation.parseTextMap is for reading headers only, so can be ignored
2b JmsInstrumentationHelper.makeChildOf which uses JMS headers so I've assumed that the strategy doesn't apply here (THIS MAY BE AN INCORRECT ASSUMPTION)
3. Consequently places in Transaction which call those two TraceContext methods both funnel through Transaction.onTransactionStart() so this handles both binary and text headers

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -34,11 +34,17 @@ version of classes that are available also in the parent (bootstrap) class loade
 * Implemented span links in the OTel bridge. - {pull}2685[#2685]
 * Added support for MongoDB's 4.x Sync Driver - {pull}2241{#2241}
 * Capture keyspace in `db.instance` for Cassandra database - {pull}2684[#2684]
+* Added support for AWS SQS - {pull}2637[#2637]
 * Add Trace Continuation Strategy per https://github.com/elastic/apm/blob/main/specs/agents/trace-continuation.md - {pull}2760[#2760]
 
 [float]
 ===== Bug fixes
 * Fix unexpected side-effects of `toString` calls within reactor instrumentation - {pull}2708[#2708]
+* Fix Vert.x instrumentation for 4.3.2 - {pull}2700[#2700]
+* Fix `NullPointerException` in `AmazonHttpClientInstrumentation` - {pull}2740[#2740]
+* Fix stack frame exclusion patterns - {pull}2758[#2758]
+* Fix `NullPointerException` with compressed spans - {pull}2755[#2755]
+* Fix empty servlet path with proper fallback - {pull}2748[#2748]
 
 [[release-notes-1.x]]
 === Java Agent version 1.x
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -34,7 +34,7 @@ version of classes that are available also in the parent (bootstrap) class loade
 * Implemented span links in the OTel bridge. - {pull}2685[#2685]
 * Added support for MongoDB's 4.x Sync Driver - {pull}2241{#2241}
 * Capture keyspace in `db.instance` for Cassandra database - {pull}2684[#2684]
-
+* Add Trace Continuation Strategy per https://github.com/elastic/apm/blob/main/specs/agents/trace-continuation.md - {pull}2760[#2760]
 
 [float]
 ===== Bug fixes
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -33,6 +33,8 @@ version of classes that are available also in the parent (bootstrap) class loade
 * Capture Oracle SID in connection string - {pull}2709[#2709]
 * Implemented span links in the OTel bridge. - {pull}2685[#2685]
 * Added support for MongoDB's 4.x Sync Driver - {pull}2241{#2241}
+* Capture keyspace in `db.instance` for Cassandra database - {pull}2684[#2684]
+
 
 [float]
 ===== Bug fixes
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -35,7 +35,7 @@ version of classes that are available also in the parent (bootstrap) class loade
 * Added support for MongoDB's 4.x Sync Driver - {pull}2241{#2241}
 * Capture keyspace in `db.instance` for Cassandra database - {pull}2684[#2684]
 * Added support for AWS SQS - {pull}2637[#2637]
-* Add Trace Continuation Strategy per https://github.com/elastic/apm/blob/main/specs/agents/trace-continuation.md - {pull}2760[#2760]
+* Add <<config-trace-continuation-strategy, `trace_continuation_strategy`>> configuration option - {pull}2760[#2760]
 
 [float]
 ===== Bug fixes
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -730,6 +730,25 @@ public class CoreConfiguration extends ConfigurationOptionProvider {
         .dynamic(true)
         .buildWithDefault(Collections.<WildcardMatcher>emptyList());
 
+    private final ConfigurationOption<TraceContinuationStrategy> traceContinuationStrategy = ConfigurationOption.enumOption(TraceContinuationStrategy.class)
+        .key("trace_continuation_strategy")
+        .tags("added[1.34.0]")
+        .configurationCategory(CORE_CATEGORY)
+        .description("This option allows some control over how the APM agent handles W3C trace-context headers on incoming requests. " +
+            "By default, the `traceparent` and `tracestate` headers are used per W3C spec for distributed tracing. " +
+            "However, in certain cases it can be helpful to not use the incoming `traceparent` header. Some example use cases:\n\n" +
+            "* An Elastic-monitored service is receiving requests with `traceparent` headers from unmonitored services.\n" +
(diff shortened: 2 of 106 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceState.java
```diff
@@ -304,4 +304,13 @@ private StringBuilder getTracestateBuffer() {
             return buffer;
         }
     }
+
+    public boolean includesVendor(){
```
felixbarny, line 308:
```suggestion
    public boolean includesElasticVendor(){
```

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
```diff
@@ -134,8 +137,26 @@ private void onTransactionStart(boolean startedAsChild, long epochMicros, Sample
...
+                tracer.getConfig(CoreConfiguration.class).getTraceContinuationStrategy();
+
+            if (traceContinuationStrategy.equals(RESTART) ||
+                (
+                    traceContinuationStrategy.equals(RESTART_EXTERNAL) &&
+                        !traceContext.getTraceState().includesVendor()
+                )
+            ) {
+                //Add a span link to the parent
+                addSpanLink(TraceContext.fromParentContext(), traceContext.copy().setIdToParentId());
```
felixbarny, line 154: Why is `setIdToParentId()` needed? Isn't the id of a span link irrelevant? We're only serializing the trace id and span id of span links. See also `co.elastic.apm.agent.report.serialize.DslJsonSerializer#serializeSpanLinks`.

```suggestion
                addSpanLink(TraceContext.fromParentContext(), traceContext.copy());
```

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
```diff
@@ -730,6 +730,25 @@ public String toString(Collection<String> value) {
...
+            "* An Elastic-monitored service is publicly exposed, and does not want tracing data (trace-ids, sampling decisions) to possibly be spoofed by user requests.\n\n" +
+            "Valid values are:\n" +
+            "* 'continue': The default behavior. An incoming `traceparent` value is used to continue the trace and determine the sampling decision.\n" +
+            "* 'restart': Always ignores the `traceparent` header of incoming requests. A new trace-id will be generated and the sampling decision" +
+            " will be made based on transaction_sample_rate. A span link will be made to the incoming `traceparent`.\n" +
+            "* 'restart_external': If an incoming request includes the `es` vendor flag in `tracestate`, then any `traceparent` will be considered " +
+            "internal and will be handled as described for 'continue' above. Otherwise, any `traceparent` is considered external and will be handled as described for 'restart' above.\n\n" +
+            "Starting with Elastic Observability 8.2, span links are visible in trace views.\n\n" +
+            "This option is case-insensitive.")
+        .buildWithDefault(TraceContinuationStrategy.CONTINUE);
```
felixbarny, line 750: As per the spec 

```suggestion
        .dynamic(true)
        .buildWithDefault(TraceContinuationStrategy.CONTINUE);
```

CHANGELOG.asciidoc
```diff
@@ -35,6 +35,7 @@ version of classes that are available also in the parent (bootstrap) class loade
 * Added support for MongoDB's 4.x Sync Driver - {pull}2241{#2241}
 * Capture keyspace in `db.instance` for Cassandra database - {pull}2684[#2684]
 * Added support for AWS SQS - {pull}2637[#2637]
+* Add Trace Continuation Strategy per https://github.com/elastic/apm/blob/main/specs/agents/trace-continuation.md - {pull}2760[#2760]
```
felixbarny, line 38:
```suggestion
* Add <<config-trace-continuation-strategy, `trace_continuation_strategy`>> configuration option - {pull}2760[#2760]
```

felixbarny (commented):
Looks good 👏 
Just a couple of small suggestions

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Transaction.java
jackshirazi: You are right the ID of the span link is irrelevant but we need the call because ... (this is part of the add span links implementation that Eyal wasn't that happy with). The span link added is the CHILD of the trace context passed here, so the resulting span link's parent ID is the ID of the tracecontext, not the parent ID. ie the general case is
  `TraceContext(traceID=A, ID=B, ParentID=C) -> spanlink(traceID=A, ID=D, ParentID=B)`

felixbarny: But for a span link, the span.id is irrelevant, right? When calling `setIdToParentId()`, we'd have `spanlink(traceID=A, ID=B, ParentID=B)` i.e. the ID and the ParentID being the same value which also doesn't seem quite right.

jackshirazi: Yes. But the span link parent needs to point to the trace context parent. So let me work through it

1. The external span is "X-OtherServiceSpan". We want the span link to point to that
2. First we construct a TraceContext with `TraceID=X, ID=NewSpanID-A, ParentID=OtherServiceSpan`
3. If we construct a span link from this TraceContext as is, we would have a SpanLink with `TraceID=X, ID=NewSpanID-B, ParentID=NewSpanID-A` - because the span link creates a child of that TraceContext. If you send this to the APM server, you send `SpanLink(TraceID=X, ParentID=NewSpanID-A)` - the span link is pointing at a span which doesn't exist anywhere in the APM server. We want the span link to point at the incoming request, which is `SpanLink(TraceID=X, ParentID=OtherServiceSpan)`
4. So in order to achieve this, I overwrite the ID, ie we start with `TraceContext(TraceID=X, ID=NewSpanID-A, ParentID=OtherServiceSpan)`, then I overwrite the ID to get `TraceContext(TraceID=X, ID=OtherServiceSpan, ParentID=OtherServiceSpan)`, then I construct the span link (which is a child of that) to get `SpanLink(TraceID=X, ID=DontCare, ParentID=OtherServiceSpan)`.
5. Then this `SpanLink(TraceID=X, ID=DontCare, ParentID=OtherServiceSpan)` is serialized (just `TraceID=X, ParentID=OtherServiceSpan` ie a link pointing at "X-OtherServiceSpan") and the APM server has the span link we want, pointing at that incoming span

felixbarny: Ah, got it now. The part that I was missing is that `addSpanLink` creates a child context of the provided context. Since you don't have the original context that has been provided in the headers of the incoming request, you re-create a "fake" parent so that you can create another child.

It's a few more hoops than I would have expected.

In a nutshell, what we currently have is:
- Make transaction a child of the remote parent
- If the trace should be re-started
  - Modify the context so that the transaction is a new trace root
  - Re-create the remote parent context
  - Add a span link, creating a child context of the fake remote context

I think we could simplify by moving the logic from `onTransactionStart` to `start`:

- If the trace should be re-started
  - Add a span link, creating a child context of the remote context
- Else
  - Make transaction a child of the remote parent

jackshirazi: I started with start(), but it has an implementation for binary and text, which means doing it there the code is partially duplicated, and to create the span link there I have to handle the separate implementations. I'm not sure it simplifies the code. I'll check

felixbarny: Right, the `start` method is overloaded. But it's not binary vs text based. The first one is used only to create a root transaction and the second `start` method is only used when there's a remote context. I'd suggest renaming the methods to make that more clear. The span link thing is only relevant for the second `start` method.

jackshirazi: The start -> startRoot for the first is much better, thanks! Otherwise, I've reimplemented per your suggestion, I don't actually think it's clearer. But it's not any worse either

felixbarny: I made some suggestions to make the implementation a bit cleaner by avoiding instanceof checks and casts: https://github.com/jackshirazi/apm-agent-java/pull/3

jackshirazi: Very nice, much cleaner like that!

felixbarny (approved):

PR #2766, opened 2022-08-31
Ignore few other agents

SylvainJuge:
## What does this PR do?

AppDynamics agent is now using another package `com.appdynamics.` and thus our heuristic to ignore it should be updated, and we need to exclude at classloader level rather than class name package only.

While we are at it, we might also enhance heuristic for other known agents.

## Checklist

- [x] This is a bugfix
  - [x] manually testing with a problematic other vendor agent


apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -694,6 +694,7 @@ public Iterable<? extends List<Class<?>>> onError(int index, List<Class<?>> batc
             .or(nameStartsWith("com.dynatrace."))
             // AppDynamics
             .or(nameStartsWith("com.singularity."))
+            .or(nameStartsWith("com.appdynamics."))
```
felixbarny, line 697: Do you know if they also have a dedicated agent class loader? If so, we could ignore all classes loaded by that class loader so that we'll not instrument non-shaded classes that area loaded by other agents. For example:

```
.or(any(), classLoaderWithNameStartsWith("com.appdynamics."))
```

SylvainJuge: I don't know if such CL is involved, but in this particular case we know the module name which is AppD specific, analyzing the agent binary could help here but unfortunately it's not publicly available.

felixbarny: Related: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/fcda760ad504aacf32eefc95c95eff5c53cf339c/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ignore/GlobalIgnoredTypesConfigurer.java

felixbarny (approved):

PR #2767, opened 2022-09-05
Capture user name with Azure SSO

SylvainJuge:
## What does this PR do?

When using Azure Tomcat service, the user name is not properly captured by Java agent.
This PR adds a fallback heuristic to capture the provided user name based on [official MS code samples](https://docs.microsoft.com/en-us/azure/app-service/configure-language-java?pivots=platform-linux#tomcat-1) and [documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens).

## Checklist

- [x] validate implementation manually with Azure/Tomcat deployment
  - [x] ~~Added an API method or config option? Document in which version this will be introduced~~ n/a
  - [x] ~~I have made corresponding changes to the documentation~~ n/a no doc change required.

### DIFF
--- a/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
+++ b/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
@@ -169,9 +169,12 @@ public class ServletTransactionHelper {
         // https://docs.microsoft.com/en-us/azure/app-service/configure-language-java?pivots=platform-linux#tomcat-1
         Object entry = map.get(key);
         if (entry instanceof List) {
-            Object first = ((List<?>) entry).get(0);
-            if (first instanceof String) {
-                return (String) first;
+            List<?> list = (List<?>) entry;
+            if (!list.isEmpty()) {
+                Object first = ((List<?>) entry).get(0);
+                if (first != null) {
+                    return first.toString();
+                }
             }
         }
         return null;
--- a/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
+++ b/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
@@ -35,8 +35,10 @@ import co.elastic.apm.agent.servlet.adapter.ServletRequestAdapter;
 import co.elastic.apm.agent.util.TransactionNameUtils;
 
 import javax.annotation.Nullable;
+import java.security.Principal;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
```diff
@@ -137,6 +139,44 @@ private void startCaptureBody(Transaction transaction, String method, @Nullable
...
+    @Nullable
+    private static String getFirstClaim(Map<?, ?> map, String key) {
+        // entries are stored in nested collection, even when there is a single entry
+        // https://docs.microsoft.com/en-us/azure/app-service/configure-language-java?pivots=platform-linux#tomcat-1
+        Object entry = map.get(key);
+        if (entry instanceof List) {
+            Object first = ((List<?>) entry).get(0);
+            if (first instanceof String) {
+                return (String) first;
+            }
```
eyalkoren, line 175: [minor] if we find a value for the looked-up key, don't we want to always use it?
```suggestion
            if (first != null) {
                first.toString();
            }
```

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
```diff
@@ -137,6 +139,44 @@ private void startCaptureBody(Transaction transaction, String method, @Nullable
...
+
+        return userName;
+    }
+
+    @Nullable
+    private static String getFirstClaim(Map<?, ?> map, String key) {
+        // entries are stored in nested collection, even when there is a single entry
+        // https://docs.microsoft.com/en-us/azure/app-service/configure-language-java?pivots=platform-linux#tomcat-1
+        Object entry = map.get(key);
+        if (entry instanceof List) {
```
eyalkoren, line 171: [questions] 
If it is not a list we don't want it? 
Is it never possible to be something else?
Can the obtained list be empty, making `get(0)` throw `IndexOutOfBoundsException`?

eyalkoren (approved):
Thanks!
Minor proposal and questions - please decide if they require addition/change/more-checks based on what you know.

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletTransactionHelper.java
SylvainJuge: good point on checking the list size, I'll add that.
The only information I have on those collections comes from experimentation, the code samples refer to those as generic `Collection` but they are actually `ArrayList` so we can avoid creating an iterator, I have not seen any entry which is not a list, so I think we can revisit this later if that proves to be an issue.

PR #2768, opened 2022-09-05
Fix windows tests

SylvainJuge:
## What does this PR do?

- Fixes a few test failures on Windows

## Checklist

- [x] This is something else

### DIFF
--- a/elastic-apm-agent/src/test/java/co/elastic/apm/agent/premain/ShadedClassLoaderTest.java
+++ b/elastic-apm-agent/src/test/java/co/elastic/apm/agent/premain/ShadedClassLoaderTest.java
@@ -28,6 +28,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.net.URLClassLoader;
 import java.util.List;
+import java.util.Objects;
 import java.util.jar.JarEntry;
 import java.util.jar.JarOutputStream;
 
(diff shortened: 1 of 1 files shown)


elastic-apm-agent/src/test/java/co/elastic/apm/agent/premain/ShadedClassLoaderTest.java
```diff
@@ -36,20 +38,31 @@
...
+    @Nullable
+    private ShadedClassLoader cl = null;
+
+    @AfterEach
+    void cleanup() throws IOException {
+        // If classloader isn't properly closed it will make tests fail on Windows as the temporary folder
+        // can't be deleted, unlike on Linux/MacOS where files can be deleted from filesystem even if they are in use.
+        if (cl != null) {
+            cl.close();
+        }
```
eyalkoren, line 50:
```suggestion
        cl.close();
```

[minor] any chance that `cl == null` at the end of a test, or is it only to remove the IDE warning? I think it's safer to fail on `NullPointerException`, so that if I add a test and test it on my mac, I cannot go along and merge it without storing the `ShadedClassLoader` reference in the dedicated field.

eyalkoren (approved):
❤️

elastic-apm-agent/src/test/java/co/elastic/apm/agent/premain/ShadedClassLoaderTest.java
SylvainJuge: If any test does not use the field then it can be `null`, however that would likely mean another potential failure on Windows, thus I'll add an explicit requirement to have it set with an NPE just to future-proof this :-).

PR #2777, opened 2022-09-12
release: configure git user

v1v:
## What does this PR do?
Use [setupAPMGitEmail](https://github.com/elastic/apm-pipeline-library/blob/main/vars/setupAPMGitEmail.groovy) within the release context

## Why

withGitRelease configures the git user pointed to the last committer, but in this case we want to force the user


cachedout (approved):

felixbarny (approved):

kuisathaverat (approved):

eyalkoren:
that was a race condition and you won @v1v 😄

PR #2778, opened 2022-09-12
ci: docs changeset against the previous commit

v1v:
## What does this PR do?

Override the default behaviour when comparing the current change set with the previous one.

Avoid aborting any `ongoing` builds if a new commit is pushed and it does only contain docs changes.

## Why

Give a higher precedence to the `GIT_PREVIOUS_COMMIT` env variable, then if there are a subsequent commits within the same Pull Request, it will compare the both commits rather than the whole changeset between the PR and the main.

Ideally, this should help with running faster builds by detecting what to run smarter.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -79,6 +79,11 @@ pipeline {
             // Compare with the previous git commit if possible, otherwise the target branch or the git base commit.
             def from = env.GIT_PREVIOUS_COMMIT?.trim() ? "${env.GIT_PREVIOUS_COMMIT}"
                                                        : "${env.CHANGE_TARGET?.trim() ? 'origin/${env.CHANGE_TARGET}' : env.GIT_BASE_COMMIT}"
+            // for debugging purposes only
+            echo "env.GIT_PREVIOUS_COMMIT=${env.GIT_PREVIOUS_COMMIT?.trim()}"
+            echo "env.CHANGE_TARGET=${env.CHANGE_TARGET?.trim()}"
+            echo "env.GIT_BASE_COMMIT=${env.GIT_BASE_COMMIT?.trim()}"
+            echo "isGitRegionMatch(from: ${from})"
             // Skip all the stages except docs for PR's with asciidoc and md changes only
             env.ONLY_DOCS = isGitRegionMatch(patterns: [ '.*\\.(asciidoc|md)' ], shouldMatchAll: true, from: from)
             // Prepare the env variables for the benchmark results
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -77,8 +77,11 @@ pipeline {
         stash allowEmpty: true, name: 'source', useDefaultExcludes: false
         script {
           dir("${BASE_DIR}"){
+            // Compare with the previous git commit if possible, otherwise the target branch or the git base commit.
+            def from = env.GIT_PREVIOUS_COMMIT?.trim() ? "${env.GIT_PREVIOUS_COMMIT}"
+                                                       : "${env.CHANGE_TARGET?.trim() ? 'origin/${env.CHANGE_TARGET}' : env.GIT_BASE_COMMIT}"
             // Skip all the stages except docs for PR's with asciidoc and md changes only
-            env.ONLY_DOCS = isGitRegionMatch(patterns: [ '.*\\.(asciidoc|md)' ], shouldMatchAll: true)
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -77,8 +77,11 @@ pipeline {
         stash allowEmpty: true, name: 'source', useDefaultExcludes: false
         script {
           dir("${BASE_DIR}"){
+            // Compare with the previous git commit if possible, otherwise the target branch or the git base commit.
+            def from = env.GIT_PREVIOUS_COMMIT?.trim() ? "${env.GIT_PREVIOUS_COMMIT}"
```
felixbarny, line 81: @SylvainJuge brought up a good point in the Java agent weekly today: what if a docs update commit cancels an ongoing build? Would we then still skip the tests, even if the cancelled build has non-docs changes?

v1v: That's a good question. 
The current implementation for `cancelOnGoingBuilds` and `gitRegionMatch` are not overlapped hence it could be the case.

Hence, we might need to revisit the `cancelOnGoingBuilds` with `gitRegionMatch` to evaluate if only-docs, then rerun everything too.  Only in case we wanna go smarter, otherwise, the existing implementation (the one in `main`) will not cause any problems.

v1v: I've just pushed https://github.com/elastic/apm-agent-java/pull/2778/commits/b5a8ed52af07b7ea1e12ac5b574dc4dddc15c284

Builds on PRs for only-doc changes will never cancel any of the old existing builds in the CI, this should avoid the corner case mentioned above.

v1v:
I've just realised the `from` transformation is not working as expected, so I'm debugging what's going on, sorry for the spam!

Jenkinsfile
```diff
@@ -70,15 +70,22 @@ pipeline {
...
         // reference repo causes issues while running on Windows with the git-commit-id-maven-plugin
         gitCheckout(basedir: "${BASE_DIR}", githubNotifyFirstTimeContributor: true, shallow: false)
         stash allowEmpty: true, name: 'source', useDefaultExcludes: false
         script {
           dir("${BASE_DIR}"){
+            // Compare with the previous git commit if possible, otherwise the target branch or the git base commit.
+            def from = env.GIT_PREVIOUS_COMMIT?.trim() ? "${env.GIT_PREVIOUS_COMMIT}"
+                                                       : "${env.CHANGE_TARGET?.trim() ? 'origin/${env.CHANGE_TARGET}' : env.GIT_BASE_COMMIT}"
+            // for debugging purposes only
+            echo "env.GIT_PREVIOUS_COMMIT=${env.GIT_PREVIOUS_COMMIT?.trim()}"
```
v1v, line 83:
```suggestion
            echo "env.GIT_PREVIOUS_SUCCESSFUL_COMMIT=${env.GIT_PREVIOUS_SUCCESSFUL_COMMIT?.trim()}"
            echo "env.GIT_PREVIOUS_COMMIT=${env.GIT_PREVIOUS_COMMIT?.trim()}"
```

Jenkinsfile
v1v:
```suggestion
            def from = env.GIT_PREVIOUS_COMMIT?.trim() ? "HEAD~1"
```

Jenkinsfile
```diff
@@ -70,15 +70,23 @@ pipeline {
...
       steps {
-        pipelineManager([ cancelPreviousRunningBuilds: [ when: 'PR' ] ])
         deleteDir()
         // reference repo causes issues while running on Windows with the git-commit-id-maven-plugin
         gitCheckout(basedir: "${BASE_DIR}", githubNotifyFirstTimeContributor: true, shallow: false)
         stash allowEmpty: true, name: 'source', useDefaultExcludes: false
         script {
           dir("${BASE_DIR}"){
+            // Compare with the previous git commit if possible, otherwise the target branch or the git base commit.
+            def from = env.GIT_PREVIOUS_COMMIT?.trim() ? "HEAD~1"
```
SylvainJuge, line 80: When pushing commits one by one, this will trigger a build for each commit, but not when we have more than one commit. I think here we need to have "the git commit of the previous build" if there is such.

For example, if I'm doing the following, will it work as expected ?
- create commit1 with code change
- create commit2 with doc change
- push them

From what I understand, the build will be triggered but will only detect a doc change.

v1v:
I'll close this for now as we are evaluating other CI options in the mid term, for now, we won't put much effort on this for avoiding the corner cases if no detecting changes when needed correctly

PR #2783, opened 2022-09-13
Limit activation stack depth

eyalkoren:
## What does this PR do?
Limits the activation stack we maintain from growing without boundaries.
The chosen solution still tries to facilitate proper management of all activations that happen within the allowed boundary.

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -101,6 +101,12 @@ protected Deque<ElasticContext<?>> initialValue() {
         }
     };
 
+    private final ThreadLocal<StackOverflowCounter> activeStackOverflowCounter = new ThreadLocal<StackOverflowCounter>() {
```
felixbarny, line 104: I think we could make this a bit cleaner by creating a dedicated `ActiveStack` class that has both an `ArrayDeque<ElasticContext<?>>` and an overflow counter. This abstracts the overflow handling so that the activation/deactivation methods don't need to deal with that.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -814,6 +832,13 @@ public void deactivate(ElasticContext<?> context) {
         if (logger.isDebugEnabled()) {
             logger.debug("Deactivating {} on thread {}", context, Thread.currentThread().getId());
         }
+
+        StackOverflowCounter stackOverflowCounter = activeStackOverflowCounter.get();
+        if (stackOverflowCounter.getOverflowCount() > 0) {
```
felixbarny, line 837: I wonder whether we should re-size the `ArrayDeque` to it's default size at when de-activating the last span. Otherwise, it may hold a very large array in memory per thread if there's one rogue transaction that gets called once in a while which has a very deep activation stack.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -782,6 +788,18 @@ public void activate(ElasticContext<?> context) {
             logger.debug("Activating {} on thread {}", context, Thread.currentThread().getId());
         }
 
+        if (activeStack.get().size() == coreConfiguration.getTransactionMaxSpans()) {
```
felixbarny, line 791: As this issue can also happen with a misconfigured trace_methods setting, I wonder if we should also explicitly check whether we've already reached the span limit in the trace methods advice before creating a span.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
eyalkoren: I don't think we should overthink this. What I added now is only a safety valve to avoid OOM, not a fix. 
This stack should never grow this much. In the rare cases that they do, it indicates that something is not working as expected for the user. Because we think this is a very rare case, we issue an error-level log and in these rare cases.

Besides, the real effect on memory is when these stacks refer to something, not when they are empty. If you compare a thread with, say 1K-size empty array, to another thread with 500 kept-alive spans through the stack, then this is negligible. And if these spans are kept alive elsewhere as well, then the memory issue is a lot worse and the OOM will occur anyway...

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
eyalkoren: Good idea! I would say even doing so when the number of spas reaches 90-95% of max-spans, so to allow room for the really important spans.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -928,4 +831,20 @@ public ScheduledThreadPoolExecutor getSharedSingleThreadedPool() {
...
+        }
+
+        public void incrementOverflowCount() {
+            this.overflowCounter++;
+        }
+
+        public void decrementOverflowCount() {
+            this.overflowCounter--;
+        }
+    }
```
felixbarny, line 849:
```suggestion
```

felixbarny (approved):
I love how this change makes `ElasticApmTracer` slimmer!

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ActiveStack.java
```diff
@@ -0,0 +1,203 @@
...
+     * In case the stack reaches its maximum allowed capacity, it won't allow for activation of any more spans.
+     * Instead, it would only keep this state of the size of the overflow. This enables both limited logging of only once
+     * per overflowing and the ability to restore proper activation/deactivation capabilities for the allowed stack size.
+     */
+    private long overflowCounter = 0;
+
+    /**
+     * Maintains a stack of all the activated spans/contexts.
+     * This way it's easy to retrieve the bottom of the stack (the transaction).
+     * Also, the caller does not have to keep a reference to the previously active span, as that is maintained by the stack.
```
felixbarny, line 56:
```suggestion
     * Also, the caller does not have to keep a reference to the previously active span, as that is maintained by the stack.
     * This makes activating a span allocation-free (assuming the stack has enough pre-allocated slots available).
```

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
eyalkoren: Misconfiguration of `trace_methods` is indeed a worthy issue to tackle, but I prefer to leave it out of this PR.
The current fix is dealing more with activation leak symptoms, where the activation stack increases to really monstrous size, whereas in common misconfiguration of `trace_methods`, the number of active spans should still be limited by thread stacks depths, so the JVM is likely to get suffocated by other parameters.
We currently issue a warning once per 5 minutes when the span limit is reached, normally due to this improper configuration.

felixbarny: I agree that this should be handled in a separate PR.

Maybe something else we could think about: issue a warning or even clear the stack if the previously active span has a different trace id. Could be expensive though.

PR #2785, opened 2022-09-14
ci: cache release stages

v1v:
## What does this PR do?

Support to rerun the same build for the same commit without running the stages that already passed.

This is handy when the release pipeline can fail for some unexpected reasons, but the previous stages that went through successfully should not be executed again.

Additionally, the behaviour can be changed by the new input parameter called `force_build`

## Caveats

It uses the git commit for the existing given branch, if for some reason the branch is updated this pipeline will not be able to cache anything.

## Actions

- [ ] Configure the bucket permissions
- [ ] Implement the env variable caching

### DIFF
--- a/.ci/release/Jenkinsfile
+++ b/.ci/release/Jenkinsfile
@@ -31,6 +31,7 @@ pipeline {
     string(name: 'branch_specifier', defaultValue: 'stable', description: "What branch to release from?")
     booleanParam(name: 'check_branch_ci_status', defaultValue: true, description: "Check for failing tests in the given branch (if no stable branch)?")
     booleanParam(name: 'publish_aws_lambda', defaultValue: true, description: "Whether to upload the AWS lambda")
+    booleanParam(name: 'force_build', defaultValue: false, description: "Whether to force the release build to run from scratch")
   }
   stages {
     stage('Initializing'){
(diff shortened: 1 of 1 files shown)


.ci/release/Jenkinsfile
```diff
@@ -309,3 +343,11 @@ def notifyStatus(def args = [:]) {
                       subject: args.subject,
                       body: args.body)
 }
+
+
+def cacheStage(def args = [:], Closure body){
+  // credentials are tied to the internal-ci controller that should have access to the beats-ci-temp bucket
+  stageStatusCache(id: args.id, bucket: 'beats-ci-temp', credentialsId: 'internal-ci-gcs-plugin-file-credentials', runAlways: params.force_build){
```
felixbarny, line 350: Rather than persisting the status and running resuming from the build from there, could we just add a build parameter that allows to run from a specific stage? For example `start_from=release-wait-artifact-available-central` would skip the first few stages and resume from release-wait-artifact-available-central onwards.

This seems to be less complex and more predictable rather than persisting and reading from an object store.

v1v: I just put another PR regarding an interactive approach:
- https://github.com/elastic/apm-agent-java/pull/2789. (still in progress, as it requires an extra input for gathering some parameters, and verify it will work as expected locally)

Rather than rerunning from a particular stage, it will ask for confirmation for each stage if needed.

v1v: I also raised another PR with a different approach, for now I'll close this one and gather details in the private GitHub issue

PR #2790, opened 2022-09-15
release: less interactive and allow running stages individually

v1v:
### UI

image

Then the version will be prompted earlier within the checkout.

### Issues

Superseedes https://github.com/elastic/apm-agent-java/pull/2789

### DIFF
--- a/.ci/release/Jenkinsfile
+++ b/.ci/release/Jenkinsfile
@@ -31,7 +31,7 @@ pipeline {
     string(name: 'branch_specifier', defaultValue: 'stable', description: "What branch to release from?")
     booleanParam(name: 'check_branch_ci_status', defaultValue: true, description: "Check for failing tests in the given branch (if no stable branch)?")
     booleanParam(name: 'check_oss_sonatype_org', defaultValue: true, description: "Check for the oss.sonatype.org?")
-    booleanParam(name: 'check_changelog', defaultValue: true, description: "Check for the CHANGELOG.asciidoc has been updated?")
+    booleanParam(name: 'check_changelog', defaultValue: true, description: "Check if the CHANGELOG.asciidoc has been updated?")
     booleanParam(name: 'set_release', defaultValue: true, description: "Set the release version?")
     booleanParam(name: 'create_update_major_branch', defaultValue: true, description: "Create or update Major Branch?")
     booleanParam(name: 'create_github_draft', defaultValue: true, description: "Create the GitHub release draft?")
--- a/.ci/release/Jenkinsfile
+++ b/.ci/release/Jenkinsfile
@@ -30,7 +30,17 @@ pipeline {
   parameters {
     string(name: 'branch_specifier', defaultValue: 'stable', description: "What branch to release from?")
     booleanParam(name: 'check_branch_ci_status', defaultValue: true, description: "Check for failing tests in the given branch (if no stable branch)?")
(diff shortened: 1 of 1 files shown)


jackshirazi (approved):
yes please!

.ci/release/Jenkinsfile
```diff
@@ -30,7 +30,17 @@ pipeline {
   parameters {
     string(name: 'branch_specifier', defaultValue: 'stable', description: "What branch to release from?")
     booleanParam(name: 'check_branch_ci_status', defaultValue: true, description: "Check for failing tests in the given branch (if no stable branch)?")
+    booleanParam(name: 'check_oss_sonatype_org', defaultValue: true, description: "Check for the oss.sonatype.org?")
+    booleanParam(name: 'check_changelog', defaultValue: true, description: "Check for the CHANGELOG.asciidoc has been updated?")
```
cachedout, line 34:
```suggestion
    booleanParam(name: 'check_changelog', defaultValue: true, description: "Check if the CHANGELOG.asciidoc has been updated?")
```

cachedout (approved):
💯 Nice one.

SylvainJuge:
Hi,

From what I understand this PR allows to conditionally execute all the steps in the release process, which is already a huge win when things go bad, so there is already a benefit to merge it as-is.

However, I still think that the end goal should be to have a release pipeline with idempotent steps (some are already like the `create_update_major_branch` that only do something if required), do you know anything that would prevent us to achieve this ? I think that we could even reuse the "restart from stage" feature of Jenkins, but I don't have any experience with it.

v1v:
> do you know anything that would prevent us to achieve this?

IIUC, the issue is mainly the release process in java that requires to push changes with the release and the new version, and generate the changelog on the fly and push it. Those steps might be the ones causing conflicts with the idempotent, though we might need to look into this deeply as I might be wrong.

I personally prefer to delegate that responsibility to a Java Maven plugin in charge of the whole post-release process, this will avoid the caveats with any CI.

In any case, I think this proposal will help with the idea of running the build from a particular stage onwards in the future if that's needed.

>  "restart from stage" feature of Jenkins,

IIRC, it didn't work well with ephemeral workers. Additionally there are some [pre-requisites](https://www.jenkins.io/doc/book/pipeline/running-pipelines/#restarting-or-rerunning-a-pipeline):

> All inputs to the Pipeline will be the same. This includes SCM information, build parameters, and the contents of any stash step calls in the original Pipeline, if specified.

We could give a go the restart and verify whether it works as expected, no strong opinion, though I prefer to play safe for the shake of reliability in the release process, so this PR might be enough, even though if the overhead is when filling the form, on the other side, there are no many releases per year - so far 7 in this year, so maybe it's not worth the invest to major changes knowing there are certain WiP to use a different CI vendor and we have not decided much.

What do you think?

SylvainJuge:
> What do you think?

I think it's fine to keep the proposal (this PR) as-is, even if the Jenkins-specific "restart from stage" might work with the known limitations.

What I wonder though (and could definitely be discussed separately) is how we could build the steps to make them 1) idempotent and 2) maybe reusable if we switch to another CI vendor. We can't delegate all the steps to maven as it's only in charge of the build (for example the github release is not part of it). In my opinion, an ideal release pipeline would only have the target branch and release version as parameters and would execute only the steps that are needed.

PR #2796, opened 2022-09-20
Capture Elasticsearch cluster name

SylvainJuge:
## What does this PR do?

 __Updated plan__

Just capture the cluster name from response HTTP header when we can.

----
__Initial plan (obsolete)__

Implements Elasticsearch cluster name capture based on heuristic defined in [specification](https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-db.md#cluster-name) and with a few extra implementation details:
- The extra request to ES has to be performed only once the application request completes as we have to get the response headers: if using an ES cloud instance the cluster name is provided in HTTP response headers, we can't guess cloud deployment before issuing the request.
- Instrumentation has to store and use the application request credentials, which is trivial for synchronous requests, but require propagation to the request end part in order to issue the extra request to ES to retrieve cluster name
- In order to retrieve the cluster name, we have to use an asynchronous call to ES because using the synchronous variant introduces a concurrent access which makes the cluster name retrieval never complete before the original request (which means it would be always late and never available for the first request). One benefit of using the asynchronous variant is that is allows to set an arbitrary maximum time budget to prevent delaying the initial request too much (as of writing it's set to 30ms).
- Because cluster name retrieval is not guaranteed, it is expected to have a few ES  span that do not have their cluster name set, and testing of this feature can't be strict to avoid introducing flakyness.

## Checklist

- [x] get spec change PR for common opt-in configuration merged https://github.com/elastic/apm/pull/688
- [x] implement the spec change
- [x] ~~Avoid extra request when run in FAAS environment as discussed in https://github.com/elastic/apm/pull/688~~ (not required anymore)
  - [x] ~~I have added tests that prove my fix is effective or that my feature works~~ probably not worth testing the impl. details of getting a simple HTTP header.
  - [x] ~~Added an API method or config option? Document in which version this will be introduced~~ no config option added
  - [x] ~~I have made corresponding changes to the documentation~~ no changes required.


SylvainJuge:
Currently blocked by specification PR https://github.com/elastic/apm/pull/688

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/esrestclient/ElasticsearchRestClientInstrumentationHelper.java
```diff
@@ -119,12 +119,17 @@ public void finishClientSpan(@Nullable Response response, Span span, @Nullable T
...
             int port = -1;
+            String cluster = null;
             if (response != null) {
                 HttpHost host = response.getHost();
                 address = host.getHostName();
                 port = host.getPort();
                 url = host.toURI();
                 statusCode = response.getStatusLine().getStatusCode();
+
+                // response header has higher priority than fetched/cached value
```
eyalkoren, line 130: [minor] Is it leftover, or I just didn't understand it properly?

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/test/java/co/elastic/apm/agent/esrestclient/AbstractEsClientInstrumentationTest.java
```diff
@@ -135,13 +139,13 @@ protected void validateDbContextContent(Span span, List<String> possibleContents
...
     protected void validateSpanContent(Span span, String expectedName, int statusCode, String method) {
-        validateSpanContentWithoutContext(span, expectedName, statusCode, method);
+        validateSpanContentWithoutContext(span, expectedName);
         validateHttpContextContent(span.getContext().getHttp(), statusCode, method);
         validateDestinationContextContent(span.getContext().getDestination());
 
         assertThat(span.getContext().getServiceTarget())
             .hasType(ELASTICSEARCH)
-            .hasNoName()
+            .hasNoName() // we can't validate cluster name here as there is no simple way to inject that without reverse-proxy
```
eyalkoren, line 148: Maybe we can replace the ES testcontainers with ESS deployments?
Not sure how fun it would be to set up and maintain CI secrets to store the credentials though...

eyalkoren (approved):

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/esrestclient/ElasticsearchRestClientInstrumentationHelper.java
SylvainJuge: this is definitely a leftover, thanks for catching this !

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/test/java/co/elastic/apm/agent/esrestclient/AbstractEsClientInstrumentationTest.java
SylvainJuge: As this header is injected by cloud proxy, doing a complete cloud deployment would definitely be overkill.
Another rather reasonable approach would be to use a reverse proxy that adds this header in response, but that's still quite a lot for testing a simple HTTP header, this is why I went to testing directly our helper implementation.

eyalkoren: I didn't think it worth the effort only for this PR, it is a nice alternative to testcontainers due to multiple reasons.
Adding a reverse proxy for testing doesn't satisfy the purpose of testing through ESS, which is to know if this logic ever breaks (eg header name changes). 
Either way, this is certainly out of scope for this PR, only a thought.

PR #2806, opened 2022-09-26
Capture MongoDB statement

SylvainJuge:
## What does this PR do?

Implements statement capture for MongoDB 4.x drivers (partial support for 3.x)

- refactor span/db assertions
- capture mongob statement
- test for statement capture in 4.x (in 3.x not all operations allow capturing it)

For now, this capture will trigger extra memory allocation to serialize the BSON object into a String object, optimizing this part will likely be done in a follow-up PR if that becomes an issue.

## Checklist
- [x] add new configuration option from https://github.com/elastic/apm/pull/698
- [x] modify tests to ensure no regression 
- [x] maybe anticipate https://github.com/elastic/apm/issues/714 and at least make this option dynamic in the agent to allow updating it at runtime without agent restart.
  - [x] ~~Added an API method or config option? Document in which version this will be introduced~~ n/a


SylvainJuge:
Partially blocked by https://github.com/elastic/apm/pull/695 for adding a common option to opt-out of DB statement capture.

If we decide to proceed without this common option, it could be an acceptable compromise with the two following caveats:
- we can't disable the feature
- there is an extra overhead (haven't measured it, but adds an extra for existing MongoDB users)

If we wait for the common configuration option to be defined, then it means we can provide a way to "opt-out" in case it does not behaves as expected or users want to revert to the previous behavior.

SylvainJuge:
Is now blocked by https://github.com/elastic/apm/pull/698, we can still make the implementation fit the current state of the spec (disabled by default, opt-in required).

SylvainJuge:
Now that https://github.com/elastic/apm/pull/698 is merged, moving forward in this PR is unblocked.

apm-agent-plugins/apm-mongodb/apm-mongodb-common/src/test/java/co/elastic/apm/agent/mongodb/AbstractMongoClientInstrumentationIT.java
```diff
@@ -70,10 +76,12 @@ public void endTransaction() throws Exception {
             currentTransaction.deactivate().end();
         }
 
-        // drop outside of transaction to prevent creating any span
-        dropCollection();
-
-        reporter.reset();
+        try {
+            // drop outside of transaction to prevent creating any span
+            dropCollection();
```
SylvainJuge, line 81: [for reviewer] this might throw an exception when the test did not create a collection, the try/finally allows to make sure that we always reset the reporter.

apm-agent-plugins/apm-mongodb/apm-mongodb3-plugin/src/main/java/co/elastic/apm/agent/mongodb/v3/ConnectionInstrumentation.java
```diff
@@ -88,7 +88,7 @@ public static Object onEnter(@Advice.This Connection thiz,
                 namespace.getDatabaseName(),
                 namespace.getCollectionName(),
                 command,
-                address.getHost(), address.getPort());
+                address.getHost(), address.getPort(), null);
```
SylvainJuge, line 91: [for reviewer] this is the case where the statement can't be captured with some 3.x drivers.

apm-agent-plugins/apm-mongodb/apm-mongodb4-plugin/src/main/resources/META-INF/services/org.stagemonitor.configuration.ConfigurationOptionProvider
```diff
@@ -0,0 +1 @@
+co.elastic.apm.agent.mongodb.MongoConfiguration
```
SylvainJuge, line 1: [for reviewer] the reason this file is duplicated for 3.x and 4.x instrumentations and not created once in the common module is because those service definition files are used as heuristic to know which maven modules need to be added as dependencies in `apm-agent` (this is being tested by `PackagingTest.checkPluginDependencies`).

jackshirazi (approved):
I really liked the assertion refactoring!

PR #2815, opened 2022-09-29
Added support for spring-kafka batch consumers

### LINKED ISSUE #2601: Add support of kafka batch consumer
when using Spring kafka batch listener (`batch = "true"` parameter), database requests are not captured by the java agent.
```java
@KafkaListener(topics = "${my-topic}", batch = "true")
public void consume(List<String> consumerRecords, Acknowledgment acknowledgment) {
    //SQL database requests are not captured
}
```
Note that when batching is disabled (`batch = "false"`) then database requests are captured by the java agent. 

Add kafka batch consumer tracing in order to capture database requests in `consumer` method.

An alternative is to add `@CaptureTransaction` annotation on `consume` method.
```java
@CaptureTransaction
@KafkaListener(topics = "${my-topic}", batch = "true")
public void consume(List<String> consumerRecords, Acknowledgment acknowledgment) {
    //SQL database requests are captured
}
```

JonasKunz:
## What does this PR do?
Closes #2601.

In reference to [Eyal's comment](https://github.com/elastic/apm-agent-java/issues/2601#issuecomment-1117009243):
> The current plan is to create a span (child of the batch transaction) for each record with a link to the sender span. Another option is to not create a span per record, but to add span links for all batch records (10 in your case) to the enclosing transaction.

The existing instrumentation for Kafka assumes that the iteration over the messages contained by a `ConsumerRecords` happens within an existing transaction for batch-processing.
This is not the case for spring-kafka batch listeners: There spring iterates over the messages, transforms them and afterwards passes them to the listener in user-code. As a result, the batch-listeners are not instrumented.

I therefore decided to add an enclosing transaction for message collection and listener invocation and to go with the second option eyal proposed:
> Another option is to not create a span per record, but to add span links for all batch records (10 in your case) to the enclosing transaction.

Another difference to eyals comment is that I chose to not follow this decision:

> As for automatic instrumentation of the Spring wrapper, I am not sure about that. If we add it, then we make the decision for our user to have single batch-processing transaction instead of transaction per record.

Due to the fact that spring first collects all messages in a list and afterwards invokes the batch-listener means that without the instrumentation the code within the batch-listener will not be part of the transaction. Instead one would only get a transaction per message capturing just the `List.add` when spring collects the messages for subsequently passing them to the actual listener.

I verified that this instrumentation works with all spring-kafka versions since `2.0`.

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else


apm-agent-plugins/apm-kafka-plugin/apm-kafka-spring-plugin/src/test/java/co/elastic/apm/agent/kafka/KafkaSpringIT.java
```diff
@@ -0,0 +1,227 @@
...
+        Span listenerSpan = reporter.getSpanByName("batchListener");
+        assertThat(listenerSpan)
+            .isNotNull()
+            .hasParent(batchReceive);
+    }
+
+    // This test has nothing to do with the SpringKafkaBatchListenerInstrumentation,
+    // it just verifies that non-batch listeners are caught by the standard kafka instrumentation
+    @Test
+    public void testSingleReceive() {
```
SylvainJuge, line 123: [very minor] maybe rename the test to something like `testNonBatchSingleReceive` so it's a bit more explicit that it does not relate to spring batch instrumentation without reading the comment above.

apm-agent-plugins/apm-kafka-plugin/apm-kafka-spring-plugin/src/main/java/co/elastic/apm/agent/kafka/SpringKafkaBatchListenerInstrumentation.java
```diff
@@ -0,0 +1,100 @@
...
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+
+/**
+ * Spring internally does the batch processing in two steps:
+ * 1. Iterate over all Messages and store them in a List
+ * 2. Pass the list to the registered batch KafkaListener
+ * <p>
+ * In order to catch both operations in a single transaction with span links,
+ * this instrumentation creates a transaction which spans both of the actions listed above.
```
SylvainJuge, line 42: [minor] it's usually quite convenient to list the instrumented method with javadoc `{@link }` for easy navigation to instrumented code, adding `@SuppressWarnings("JavadocReference")` might also be used when linking to a method/class that is not visible.

apm-agent-plugins/apm-kafka-plugin/apm-kafka-spring-plugin/src/main/java/co/elastic/apm/agent/kafka/SpringKafkaBatchListenerInstrumentation.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        @Nullable
+        @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
+        public static Object onEnter(@Advice.This Object thiz) {
+            Transaction transaction = tracer.startRootTransaction(thiz.getClass().getClassLoader());
+            if (transaction != null) {
+                transaction
+                    .withType("messaging")
+                    .withName("Spring Kafka Message Batch Processing")
```
SylvainJuge, line 78: [minor] I think here using a simpler naming like `Spring Kafka Batch` would be enough to keep things short. The `Message` is quite implicit from `Kafka`, and the `Processing` does not add any relevant information.

apm-agent-plugins/apm-kafka-plugin/apm-kafka-spring-plugin/src/main/java/co/elastic/apm/agent/kafka/SpringKafkaBatchListenerInstrumentation.java
```diff
@@ -0,0 +1,100 @@
...
+                oneTimeTransactionCreationWarningLogger.warn("Failed to start Spring Kafka transaction for batch processing");
+            }
+            //we don't need to add span links here, they will be added by the KafkaConsumerRecordsInstrumentation
+            return transaction;
+        }
+
+        @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class, inline = false)
+        public static void onExit(
+            @Nullable @Advice.Enter Object transactionObj
+        ) {
```
SylvainJuge, line 91:
```suggestion
        public static void onExit(@Nullable @Advice.Enter Object transactionObj {
```

SylvainJuge (approved):

apm-agent-plugins/apm-kafka-plugin/apm-kafka-spring-plugin/src/main/java/co/elastic/apm/agent/kafka/SpringKafkaBatchListenerInstrumentation.java
JonasKunz: As discussed on slack: We will stay with the current name to keep it consistent with the spring AMQP batch listener transactions.

eyalkoren:
Nice! 
Consider using [`message_batch_strategy`](https://github.com/elastic/apm-agent-java/blob/37e66e0465fb9220483e8fcde75d14a166f8be4f/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MessagingConfiguration.java#L47) to opt in to treat each message in its own transaction. 
I agree that the default should be to treat it as a batch now that we have span links, but this is sort-of a breaking change for current users, who may have been happy with the way it was so far.

JonasKunz:
> Consider using [message_batch_strategy](https://github.com/elastic/apm-agent-java/blob/37e66e0465fb9220483e8fcde75d14a166f8be4f/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MessagingConfiguration.java#L47) to opt in to treat each message in its own transaction.

I considered that, but decided that the current behaviour for spring-batch-consumers was a bit broken in regards to the timing of the kafka-transactions generated by our standard kafka instrumentation.

Spring implements the batch consumer as follows:
> There spring iterates over the messages, transforms them and afterwards passes them to the listener in user-code. As a result, the batch-listeners are not instrumented.

This means with our default kafka instrumentation the Kafka-Transactions will merely capture the timing of the spring internal `ArrayList.add` operation and nothing else.

If we want to have correct support for `message_batch_strategy= SINGLE_HANDLING` we would need to instrument/wrap the transformed `List` of events which is actually passed to the user batch listener and remember for each entry to which Kafka-message it corresponds.

That's the reasoning why I classified the previous behaviour rather as a feature-gap / buggy and therefore assumed that this breaking change is okay.

eyalkoren:
Ahh, I looked only at implementation and not the description, makes perfect sense, thanks for explaining!

PR #2816, opened 2022-09-29
Implement ECS shading for console output

### LINKED ISSUE #2570: Apply log_ecs_reformatting to console/System.out appenders
We currently restrict ecs-reformatting only to file appenders (hardcoded in the implementations of [`createAndStartEcsAppender()`](https://github.com/elastic/apm-agent-java/blob/main/apm-agent-plugins/apm-logging-plugin/apm-logging-plugin-common/src/main/java/co/elastic/apm/agent/loginstr/reformatting/AbstractEcsReformattingHelper.java#L423) implementations).

We should apply `SHADE` and `OVERRIDE` [options](https://www.elastic.co/guide/en/apm/agent/java/current/config-logging.html#config-log-ecs-reformatting) to standard output logging as well, especially since this is a common logging output in containerized environments.

SylvainJuge:
## What does this PR do?

When `log_ecs_reformatting` = `shade`, the file-based log appenders have their output duplicated to an ECS JSON formatted file, but the console-based ones are not.

This pull-request adds proper shading for console-base appenders, the `override` option already work as expected.

Closes #2570 

## Implementation details
- most of the the changes are in `createAndStartEcsAppender` implementations
- in tests, the files that might be written to current working directory are moved to `./target` to prevent polluting the project.
- for testing, the existing file-based logger has been modified to also append to console appender, hence that allows to easily test shading by comparing the console-based and file-based ECS files.

## to be discussed/clarified

- [ ] should we gather all console-based appenders to a single ECS-formatted file ? That could somehow simplify the naming of the ECS-formatted console log file, by having for example `console.err.ecs.json` and `console.out.ecs.json` for `System.err` and `System.out` respectively.
- [ ] should we normalize the values of `event.dataset` ?
- [ ] when `log_ecs_reformatting_dir` is empty or not set, the console ECS log file will be written to current working directory
- [ ] when `log_ecs_reformatting_dir` is set to a relative path, the console ECS log file will be written to a relative path of the current working directory which could be confusing for the end-user, but the file-based ECS log files will be relative to the original log file location.
  - For example with `log_ecs_reformatting_dir=ecs-logs`, a file-based log in `/var/logs/app.log` will be written to `/var/logs/ecs-logs/app.log.ecs.json`, whereas the console-based one will be written to `./ecs-logs/console.ecs.json` when the user could expect to have it in `/var/logs/ecs-logs/console.ecs.json`, using an absolute path for `log_ecs_reformatting_dir` allows to write all log files to a single folder.
  - option 1 : leave it as-is and document it, which probably means suggesting using an absolute path and expect some confused users.
  - option 2 : change the meaning of "relative path" for this setting: describe it as "the folder where all the ECS log files will be written", when a relative path is used it will be relative to the current working directory and not relative to the original log file location. While this is a breaking change for the users that already use the reformatting feature (which is experimental), I think this is the best option. I think there are only a few rare cases where the original log files would conflict when reformatted to a single folder, for example with `/logs/server/service.log` and `/logs/app/service.log` both would be written to `service.ecs.json`.
- [ ] should we make console logging optional and/or disabled by default ?
- [ ] should we also implement `replace` option for console-based appenders as mentionned in https://github.com/elastic/apm-agent-java/issues/2570#issuecomment-1255977672 ?

## Checklist


SylvainJuge:
After discussing it today with @felixbarny we think the extra complications here probably mean it's not worth pushing this PR further, or we might have to take a different approach here.

Maybe revisiting this when the agent is able to stream logs would be an interesting option as it removes the need to pick a file name for the console-based loggers.

SylvainJuge:
Dropping this PR for now, we might re-evaluate adding this feature later once we have proper feedback with current implementation.

PR #2839, opened 2022-10-10
ECS reformat support for Tomcat

SylvainJuge:
Makes Tomcat logging (which relies on `java.util.logging`) work with ECS reformatting.

## Checklist


apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/src/main/java/co/elastic/apm/agent/jul/reformatting/JulEcsReformattingHelper.java
```diff
@@ -148,7 +81,7 @@ static String computeEcsFileHandlerPattern(String pattern, Path originalFilePath
             pattern = pattern + ".%g";
         }
         int lastPathSeparatorIndex = pattern.lastIndexOf('/');
-        if (lastPathSeparatorIndex > 0 && pattern.length() > lastPathSeparatorIndex) {
+        if (lastPathSeparatorIndex > 0 && lastPathSeparatorIndex < pattern.length() - 1) {
```
SylvainJuge, line 84: [for reviewer] this was a bug in existing implementation

apm-agent-plugins/apm-logging-plugin/apm-tomcat-logging-plugin/src/main/java/co/elastic/apm/agent/tomcatlogging/reformatting/FileHandlerPublishAdvice.java
```diff
@@ -0,0 +1,50 @@
...
+package co.elastic.apm.agent.tomcatlogging.reformatting;
+
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.implementation.bytecode.assign.Assigner;
+import org.apache.juli.FileHandler;
+
+import java.util.logging.Handler;
+import java.util.logging.LogRecord;
+
+public class FileHandlerPublishAdvice {
```
SylvainJuge, line 28: [for reviewer] this is where the main difference between `java.logging.FileHandler` and `org.apache.juli.FileHanler` implementation is due to different class fields.

apm-agent-plugins/apm-logging-plugin/apm-tomcat-logging-plugin/src/test/java/co/elastic/apm/agent/tomcatlogging/TomcatLoggingTestIT.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.BiPredicate;
+import java.util.stream.Collectors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.testcontainers.containers.output.OutputFrame.OutputType.STDOUT;
+
+@Disabled // disabled because it's slow and might introduce flakyness
```
felixbarny, line 52: This is the only test for this plugin and it's disabled. Would it be feasible to implement something analogous to the `JulInstrumentationTest` for Tomcat JUL?

apm-agent-plugins/apm-logging-plugin/apm-tomcat-logging-plugin/src/test/java/co/elastic/apm/agent/tomcatlogging/TomcatLoggingTestIT.java
```diff
@@ -0,0 +1,160 @@
...
+public class TomcatLoggingTestIT {
```
felixbarny, line 53:
```suggestion
public class TomcatLoggingIT {
```

apm-agent-plugins/apm-logging-plugin/apm-tomcat-logging-plugin/src/test/java/co/elastic/apm/agent/tomcatlogging/TomcatLoggingTestIT.java
SylvainJuge: Very good point, we might also need to test the Helper class too. I'll add that.

apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/src/main/java/co/elastic/apm/agent/jul/reformatting/AbstractJulEcsReformattingHelper.java
```diff
@@ -0,0 +1,117 @@
...
+                shadeHandler = new FileHandler(pattern, maxLogFileSize, 2, true);
+                shadeHandler.setFormatter(ecsFormatter);
+            } catch (Exception e) {
+                logger.error("Failed to create Log shading FileAppender. Auto ECS reformatting will not work.", e);
+            }
+        }
+        return shadeHandler;
+    }
+
+    protected abstract boolean shouldCreateEcsAppender(Handler originalHandler);
```
JonasKunz, line 115: [minor suggestion]: As someone new to our logging stuff, the name confused me here a little.
I was expecting this to somehow read the `log_ecs_reformatting`, but this actually happens in `createAndMapShadeAppenderFor`.

Is there a reason for not naming this more specific `isFileAppender`? Because these are the two only implementations, right?

apm-agent-plugins/apm-logging-plugin/apm-tomcat-logging-plugin/src/main/java/co/elastic/apm/agent/tomcatlogging/reformatting/TomcatReformattingHelper.java
```diff
@@ -0,0 +1,99 @@
...
+    private static final ThreadLocal<String> currentSuffix = new ThreadLocal<>();
+
+    TomcatReformattingHelper() {
+    }
+
+    public boolean onAppendEnter(FileHandler fileHandler, String directory, String prefix, String suffix) {
+        try {
+            currentDirectory.set(directory);
+            currentPrefix.set(prefix);
+            currentSuffix.set(suffix);
```
JonasKunz, line 49: [personal preference] I don't know why, but using ThreadLocals to just pass parameters along a code-path we 100% control feels a bit hacky to me.
An alternative would be to extend the `AbstractEcsReformattingHelper` methods `onAppendEnter` and `getShadeFilePatternAndCreateDir` with a generic `C context` parameter which gets passed through. However, because this Helper is the only case where it is needed so far, this might not really be worth it.

apm-agent-plugins/apm-logging-plugin/apm-tomcat-logging-plugin/src/test/java/co/elastic/apm/agent/tomcatlogging/TomcatLoggingIT.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.BiPredicate;
+import java.util.stream.Collectors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.testcontainers.containers.output.OutputFrame.OutputType.STDOUT;
+
+@Disabled // disabled because it's slow and might introduce flakyness
+public class TomcatLoggingIT {
```
JonasKunz, line 53: Maybe add a comment that the agent needs to be built before running the test? I just ran them and initially got confused why they are failing :D

JonasKunz (approved):

apm-agent-plugins/apm-logging-plugin/apm-tomcat-logging-plugin/src/main/java/co/elastic/apm/agent/tomcatlogging/reformatting/TomcatReformattingHelper.java
SylvainJuge: I agree with you, this is a bit more hack-ish that it should, the challenge here is that:
- we can only read the class internal fields from the advice code, which is the same for regular JUL and Tomcat Juli.
- we can avoid allocating "context" objects in the case where the log events are not actually written (this depends on configuration).

So, while it definitely does not meet most code sanity standards, I would say this is fine for this case.
Alternatively, we could also use context objects as you suggested, but that would require to pool them to avoid allocation.

apm-agent-plugins/apm-logging-plugin/apm-jul-plugin/src/main/java/co/elastic/apm/agent/jul/reformatting/AbstractJulEcsReformattingHelper.java
SylvainJuge: That's a very good point, I have renamed it to `isFileAppender` for clarity.

apm-agent-plugins/apm-logging-plugin/apm-tomcat-logging-plugin/src/main/java/co/elastic/apm/agent/tomcatlogging/reformatting/TomcatReformattingHelper.java
JonasKunz: Fully agree that it's fine as is, especially because the use of the `ThreadLocals` is just locally here and therefore doesn't hinder understandability :)

PR #2843, opened 2022-10-12
Ignore exclude of co.elastic.apm for instrumentation

JonasKunz:
## What does this PR do?

A customer excluded `co.elastic.apm` from the instrumentation, which caused unwanted side-effects, such as that the public-api stopped working.

This change prevents exclusions of the entire `co.elastic.apm` package, concrete subpackages / classes may still be excluded.

Question: Even though this is a very small change, it still somewhat changes the behaviour of the agent. Therefore we should add it to the changelog, right?

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
@@ -36,7 +36,6 @@ import co.elastic.apm.agent.impl.ElasticApmTracer;
 import co.elastic.apm.agent.impl.ElasticApmTracerBuilder;
 import co.elastic.apm.agent.impl.GlobalTracer;
 import co.elastic.apm.agent.matcher.MethodMatcher;
-import co.elastic.apm.agent.matcher.WildcardMatcher;
 import co.elastic.apm.agent.sdk.ElasticApmInstrumentation;
 import co.elastic.apm.agent.sdk.logging.Logger;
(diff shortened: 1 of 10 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -711,10 +712,22 @@ public Iterable<? extends List<Class<?>>> onError(int index, List<Class<?>> batc
             .or(nameContains("javassist"))
             .or(nameContains(".asm."))
             .or(anyMatch(coreConfiguration.getDefaultClassesExcludedFromInstrumentation()))
-            .or(anyMatch(coreConfiguration.getClassesExcludedFromInstrumentation()))
+            .or(anyMatch(sanitizeConfiguredExcludes(logger, coreConfiguration.getClassesExcludedFromInstrumentation())))
             .disableClassFormatChanges();
     }
 
+    private static List<WildcardMatcher> sanitizeConfiguredExcludes(Logger logger, List<WildcardMatcher> classesExcludedFromInstrumentation) {
```
felixbarny, line 719: I'd move this into `CoreConfiguration::getClassesExcludedFromInstrumentation` so that there's no way to circumvent the sanitization.

JonasKunz: As discussed in slack, I also adopted the mechanism to log only once per config change in order to prevent log-spam.

PR #2859, opened 2022-10-24
Avoid NPE in afterEndSpan

SylvainJuge:
## What does this PR do?

Fixes NPE reported [in our forum](https://discuss.elastic.co/t/npe-using-apm-agent-1-34-1/317179).

## Checklist

- [x] This is a bugfix

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
@@ -285,8 +285,8 @@ public class Span extends AbstractSpan<Span> implements Recyclable {
             }
             if (!buffered.tryToCompress(this)) {
                 if (parent.bufferedSpan.compareAndSet(buffered, this)) {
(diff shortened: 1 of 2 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -285,8 +285,8 @@ protected void afterEnd() {
             }
             if (!buffered.tryToCompress(this)) {
                 if (parent.bufferedSpan.compareAndSet(buffered, this)) {
-                    this.tracer.endSpan(buffered);
                     parent.decrementReferences();
```
felixbarny, line 288: Should we store `parent` in a local variable so that we have a stable reference even if the instance variable changes?

jackshirazi: no, this is recycling and just the order here was wrong. better not to hang on to objects being recycled

jackshirazi (approved):

PR #2863, opened 2022-10-26
Shortcut attacher discovery and fork

### LINKED ISSUE #2713: Shortcut agent attach CLI discovery and run-as-user
Following up on https://github.com/elastic/apm-server/pull/8590:
1. If the attacher CLI is invoked with only `--include-pid` discovery rules, there's no need for discovery and for applying discovery rules, only try to attach to the requests PIDs. If the attacher is not invoked in a `--continuous` mode, no need to even look whether this PID was encountered before. This avoids a lot of noise related to discovery, especially such that requires actual attachment for extracting data.
2. Add a `--switch-user` option to the attacher CLI, which defaults to `true`, that makes it possible to shortcut the user registry and `su` route. Even though it is currently protected through `isCurrentUser()`, this check may not work if we avoid the full JVM discovery (e.g. due to 1, where discovery is not required).

Checklist:
- [ ] if discovery rules only include `--include-pid` rules, don't do any discovery, invoke BB attachment for the PID directly
- [ ] add a `--switch-user` config option that will default to `true` and will be set by the APM Server integration to `false`, in which case the attacher will not do any user registry things and will not invoke a sub attacher process
- [ ] add missing tests and update existing ones

eyalkoren:
## What does this PR do?
Closes #2713 

## Checklist

  - [x] I have added unit tests
  - [x] I have tested manually

Tested manually on:
- *macOS*: using Java 17.0.2, attachment succeeded and agent functions as expected, but the attacher ends with an error related to JNA library loading: `Caused by: java.lang.UnsatisfiedLinkError: Can't load library: /Users/eyalkoren/Library/Caches/JNA/temp/jna5306318620723725866.tmp`
- *Amazon Linux 2*: using Java 11.0.15 - works as expected

### DIFF
--- a/apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/DiscoveryRulesTest.java
+++ b/apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/DiscoveryRulesTest.java
@@ -76,6 +76,16 @@ class DiscoveryRulesTest {
     }
 
     @Test
+    void testDiscoveryRequired() {
+        assertThat(discoveryRules.isDiscoveryRequired()).isFalse();
+        discoveryRules.includePid("1");
+        discoveryRules.includePid("2");
+        assertThat(discoveryRules.isDiscoveryRequired()).isFalse();
+        discoveryRules.excludePid("3");
+        assertThat(discoveryRules.isDiscoveryRequired()).isTrue();
+    }
+
+    @Test
     void testExcludeNotIncluded() {
         discoveryRules.includePid("1");
         discoveryRules.excludePid("2");
--- a/docs/setup-attach-cli.asciidoc
+++ b/docs/setup-attach-cli.asciidoc
@@ -97,7 +97,9 @@ If provided, this program continuously runs and attaches to all running and star
 *--no-fork* added[1.35.0]::
 +
 --
-If provided, this program will not fork and attempt to attach directly as the current user.
+By default, when the attacher program is ran by user A and the target process is ran by user B, the attacher will attempt to start another
+process as user B.
+If this configuration option is provided, the attacher will not fork. Instead, it will attempt to attach directly as the current user.
 --
 
 *--include-all*::
--- a/docs/setup-attach-cli.asciidoc
+++ b/docs/setup-attach-cli.asciidoc
@@ -94,7 +94,7 @@ NOTE: The JVM arguments may contain sensitive information, such as passwords pro
 If provided, this program continuously runs and attaches to all running and starting JVMs which match the `--exclude` and `--include` filters.
 --
 
-*--no-fork*::
+*--no-fork* added[1.35.0]::
 +
 --
 If provided, this program will not fork and attempt to attach directly as the current user.
--- a/docs/setup-attach-cli.asciidoc
+++ b/docs/setup-attach-cli.asciidoc
@@ -94,6 +94,12 @@ NOTE: The JVM arguments may contain sensitive information, such as passwords pro
 If provided, this program continuously runs and attaches to all running and starting JVMs which match the `--exclude` and `--include` filters.
 --
 
+*--no-fork*::
++
+--
+If provided, this program will not fork and attempt to attach directly as the current user.
+--
+
 *--include-all*::
 +
 --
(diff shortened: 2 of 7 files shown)


docs/setup-attach-cli.asciidoc
```diff
@@ -94,6 +94,12 @@ NOTE: The JVM arguments may contain sensitive information, such as passwords pro
 If provided, this program continuously runs and attaches to all running and starting JVMs which match the `--exclude` and `--include` filters.
 --
 
+*--no-fork* added[1.35.0]::
++
+--
+If provided, this program will not fork and attempt to attach directly as the current user.
```
SylvainJuge, line 100: Do you think we should elaborate on reasons the end-user might need this ? Probably the most common example is when the attachment process is started by `root` user and the target JVM is using a non-root user. Because this is probably the only reason we have to fork, then maybe renaming this to `--no-fork-as-user` could make this a bit more explicit.

apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/DiscoveryRulesTest.java
```diff
@@ -75,6 +75,16 @@ void testIncludeMultiple() {
...
     }
 
+    @Test
+    void testDiscoveryRequired() {
+        assertThat(discoveryRules.isDiscoveryRequired()).isFalse();
+        discoveryRules.includePid("1");
+        discoveryRules.includePid("2");
+        assertThat(discoveryRules.isDiscoveryRequired()).isFalse();
+        discoveryRules.excludePid("3");
+        assertThat(discoveryRules.isDiscoveryRequired()).isTrue();
```
SylvainJuge, line 85: The way I understand this is when we include PIDs, we assume that the PID was provided from somewhere, and thus we don't have to attempt discoverey. However, I'm not sure to get why we shouldn't do the same for excluding PIDs.

SylvainJuge (approved):
LGTM, only a few questions.

apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/DiscoveryRulesTest.java
eyalkoren: Exclude PID doesn't give any advice as to what process to attach. It doesn't make sense to provide it without additional discovery rules, which means we need to do discovery.

eyalkoren: Note that even if you use `--include-pid` with any additional discovery rule, it will be executed after discovery and in the order defined by the configuration.
Don't treat this combination of rules as such that make sense in reality, they only unit-test the logic

docs/setup-attach-cli.asciidoc
eyalkoren: Since it is a command line argument, I prefer to keep it short (and somewhat a familiar one).
But I will add this explanation to the description. Thanks for the advice

SylvainJuge: Alternatively, you suggested the  `--switch-user` in https://github.com/elastic/apm-agent-java/issues/2713 also.

eyalkoren: Yes, but when implementing I thought it makes more sense that the config option name reflects the opt-in purpose, which is to avoid the forking command. 
It seems weird that the only reason to use a config option named `--switch-user` is to stop switching users...

apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/DiscoveryRulesTest.java
SylvainJuge: I see, we only allow to take the shortcut when including a specific PID because in this case we know other discovery rule is necessary.

docs/setup-attach-cli.asciidoc
eyalkoren: See extended description

apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/DiscoveryRulesTest.java
eyalkoren: More accurately: we only _**can**_ take the shortcut when including _**only**_ specific PIDs because in this case discovery is not required _**and we have enough info to attach**_

Issue #2543, opened 2022-03-24
[META 625] Agent health metrics


felixbarny:
Meta issue: 
- https://github.com/elastic/apm/issues/625

Spec issue:
- https://github.com/elastic/apm/issues/626

Define a standard set of metrics that each agent should eventually send
- Metrics about event handling, queuing, and sending to APM Server
- Agent CPU usage (stretch)

Creating a custom dashboard that shows the metrics and publish it to elastic/apm-contrib

jackshirazi:
I think the right time period would be per minute? Or would we just send cumulative stats?
- bytes sent in the last time period
- max queue size (of documents sent to server) in the last time period
- spans started in the last time period
- spans ended in the last time period
- transactions started in the last time period
- transactions ended in the last time period
- failed connection attempts in the last time period
- instrumentation count

I think we need much more input from customers. @alex-fedotyev @chrisdistasio @akhileshpok we'd ideally have this customer driven, do we have any requests/feedback from customers on agent health?

PR #2864, opened 2022-10-28
Added reporter health metrics

JonasKunz:
## What does this PR do?
Adds a tech preview for agent health metrics (#2543). Overhead metrics will be added via a separate PR.

## Checklist

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MetricsConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MetricsConfiguration.java
@@ -53,7 +53,7 @@ public class MetricsConfiguration extends ConfigurationOptionProvider {
     private final ConfigurationOption<Boolean> reporterHealthMetricsEnabled = ConfigurationOption.booleanOption()
         .key("agent_reporter_health_metrics")
         .configurationCategory(METRICS_CATEGORY)
-        .description("Enables metrics which capture the health state of the event reporting mechanism.\n" +
+        .description("Enables metrics which capture the health state of the agent's event reporting mechanism.\n" +
             "Disabled by default because this is a tech preview feature.")
         .tags("added[not officially added yet]", "internal")
         .dynamic(false)
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MetricsConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MetricsConfiguration.java
@@ -50,11 +50,24 @@ public class MetricsConfiguration extends ConfigurationOptionProvider {
         .dynamic(false)
         .buildWithDefault(1000);
 
+    private final ConfigurationOption<Boolean> reporterHealthMetricsEnabled = ConfigurationOption.booleanOption()
+        .key("agent_reporter_health_metrics")
+        .configurationCategory(METRICS_CATEGORY)
+        .description("Enables metrics which capture the health state of the event reporting mechanism.\n" +
+            "Disabled by default because this is a tech preview feature.")
+        .tags("added[not officially added yet]", "internal")
+        .dynamic(false)
+        .buildWithDefault(false);
+
     public boolean isDedotCustomMetrics() {
         return dedotCustomMetrics.get();
     }
 
     public int getMetricSetLimit() {
         return metricSetLimit.get();
+    }
+
+    public boolean isReporterHealthMetricsEnabled() {
+        return reporterHealthMetricsEnabled.get();
     }
 }
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MetricsConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MetricsConfiguration.java
@@ -53,9 +53,8 @@ public class MetricsConfiguration extends ConfigurationOptionProvider {
     private final ConfigurationOption<Boolean> reporterHealthMetricsEnabled = ConfigurationOption.booleanOption()
         .key("agent_reporter_health_metrics")
         .configurationCategory(METRICS_CATEGORY)
-        .description("Enables metrics which capture the health state of the agent's event reporting mechanism.\n" +
-            "Disabled by default because this is a tech preview feature.")
-        .tags("added[not officially added yet]", "internal")
+        .description("Enables metrics which capture the health state of the agent's event reporting mechanism.")
+        .tags("added[1.35.0]", "experimental")
         .dynamic(false)
         .buildWithDefault(false);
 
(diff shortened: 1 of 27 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MetricsConfiguration.java
```diff
@@ -50,11 +50,24 @@ public class MetricsConfiguration extends ConfigurationOptionProvider {
         .dynamic(false)
         .buildWithDefault(1000);
 
+    private final ConfigurationOption<Boolean> reporterHealthMetricsEnabled = ConfigurationOption.booleanOption()
+        .key("agent_reporter_health_metrics")
+        .configurationCategory(METRICS_CATEGORY)
+        .description("Enables metrics which capture the health state of the event reporting mechanism.\n" +
+            "Disabled by default because this is a tech preview feature.")
+        .tags("added[not officially added yet]", "internal")
+        .dynamic(false)
```
jackshirazi, line 59: We do periodic reporting, so it seems like it wouldn't be any overhead to check the config each time - which means this could be dynamic

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
```diff
@@ -43,6 +45,8 @@ public class MetricRegistry {
     private static final Logger logger = LoggerFactory.getLogger(MetricRegistry.class);
     private final WriterReaderPhaser phaser = new WriterReaderPhaser();
     private final ReporterConfiguration reporterConfiguration;
+
+    private final Set<MetricsProvider> metricsProviders = Collections.newSetFromMap(new ConcurrentHashMap<MetricsProvider, Boolean>());
```
jackshirazi, line 49: what limits the size of this set?

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricSet.java
```diff
@@ -46,6 +47,9 @@ public class MetricSet implements Recyclable {
     // low load factor as hash collisions are quite costly when tracking breakdown metrics
     private final ConcurrentMap<String, Timer> timers = new ConcurrentHashMap<>(32, 0.5f, Runtime.getRuntime().availableProcessors());
     private final ConcurrentMap<String, AtomicLong> counters = new ConcurrentHashMap<>(32, 0.5f, Runtime.getRuntime().availableProcessors());
+
+    //no need for concurrency because this field is only updated from the metrics reporter thread.
+    private final Map<String, Double> rawValues = new HashMap<>();
```
jackshirazi, line 52: But concurrency is not only about updates being from one thread, it's about reads and updates. So for example one known concurrency bug with using HashMap when it's updated in only one thread is that the update could rehash the table while the other thread is iterating. Just switch to a ConcurrentHashMap, true it has overheads until about Java 9, but by 11 it's really low overhead (both CPU and memory) compared to HashMap, so I think it's fine to use

apm-agent-core/src/main/java/co/elastic/apm/agent/report/AbstractIntakeApiHandler.java
```diff
@@ -200,25 +211,22 @@ protected boolean isApiRequestTimeExpired() {
         return System.nanoTime() >= requestStartedNanos + TimeUnit.MILLISECONDS.toNanos(reporterConfiguration.getApiRequestTime().getMillis());
     }
 
-    protected void onRequestError(Integer responseCode, InputStream inputStream, @Nullable IOException e) {
-        // TODO read accepted, dropped and invalid
-        onConnectionError(responseCode, currentlyTransmitting, 0);
+    private void onRequestError(Integer responseCode, long bytesWritten, InputStream inputStream, @Nullable IOException e) {
+        String responseBody = null;
+        try {
```
jackshirazi, line 216: is there a reason for removing the `isWarnEnabled` guard?

apm-agent-core/src/main/java/co/elastic/apm/agent/report/ReportingEventCounter.java
```diff
@@ -0,0 +1,115 @@
...
+    private final AtomicLongArray counters;
+
+    public ReportingEventCounter() {
+        ReportingEvent.ReportingEventType[] types = ReportingEvent.ReportingEventType.values();
+        counters = new AtomicLongArray(types.length);
+    }
+
+    public ReportingEventCounter(ReportingEventCounter toCopy) {
+        ReportingEvent.ReportingEventType[] types = ReportingEvent.ReportingEventType.values();
+        counters = new AtomicLongArray(types.length);
```
jackshirazi, line 34:
```suggestion
        this();
```

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MetricsConfiguration.java
JonasKunz: Two reasons why chose to not make it dynamic:
 * `disable_metrics`, which is used to disable individual metrics is not dynamic. This makes this config feel less useful / inconsistent to be dynamic.
 * The config option will be removed after the tech preview, because then the metrics will be opt-out. Therefore I didn't bother with the effort of allowing the metric enabled state to change at runtime

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricRegistry.java
JonasKunz: `MetricsProvider` are very much like our gauges: There's a finite set defined by our agent implementation (=all calls to `MetricRegistry.addMetricsProvider`). Therefore there is no risk of this set growing out of control.

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricSet.java
JonasKunz: My comment here was inaccurate, `rawValues` is updated and read from a single thread.
However, I agree that we should change it to `ConcurrentMap` as this assumption might not hold in the future and is not worth the headache.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/AbstractIntakeApiHandler.java
JonasKunz: Yes, previously the `isWarnEnabled` guard was there because we only parsed the APM server response body in case the log level was `WARN`.
Now we always parse the response independently of the log level, because we read the `accepted` field from it.
Therefore the only thing which would be left within the guard would be the `log.warn`s, which make the guard pointless.

jackshirazi (approved):

jackshirazi (approved):

PR #2865, opened 2022-10-31
Reference counting for buffered spans

eyalkoren:
## What does this PR do?
Fixing the concurrency issue reported in [our forum](https://discuss.elastic.co/t/npe-using-apm-agent-1-34-1/317179)

## Checklist

- [x] This is a bugfix
  - [ ] ~~I have added tests that would fail without this fix~~ Our existing multi-threaded tests fail to capture this


jackshirazi:
not sure how this is addressing the NPE which appears to me to be that `transaction` is null in `Span` line 369?

eyalkoren:
The addition of the reference-counting atomic reference is just because it is the right thing to do - as long as something holds a reference to an `AbstractSpan`, we should count it otherwise something else may look into it and think it is valid although it is recycled. It doesn't fix this NPE, but may avoid future concurrency issues.

The actual fix of the possible race condition that I saw is related to `afterEnd(Span.java:286)` in the stack trace and I believe it should be fixed through `RefCountingAtomicReference#getAndIncrementReferences()` and its usage in `afterEnd()`. Otherwise, two threads may `get()` a reference to the buffered span, and then one may end it, thus causing it to be recycled, after which the second thread may try to evaluate `tryToCompress()` on an already-recycled span.

eyalkoren:
`RefCountingAtomicReference#getAndIncrementReferences()` contained a race condition as it could allow for span removal and recycling between `get()` and `incrementReferences()`, meaning that the returned span was guaranteed to have a ref count > 0, but not guaranteed not to be recycled.
I fixed that in the last commit by replacing it with `RefCountingAtomicReference#incrementReferencesAndGet()`, which guarantees the returned span has not been recycled.

However, even in the current implementation, there is a race condition - if `RefCountingAtomicReference#incrementReferencesAndGet()` identifies that the span has been removed from the map during execution, it corrects the improper reference increment by decrementing it. The problem is that the removed span could potentially have been sent for recycling (as it momentarily had ref count of 0), in which case the ref count state is unknown. We can further fix that through ref count checks, but this may bring other issues.

I think I will replace the fix of the originally described race condition with something else. 
Two options:
1. implement `RefCountingAtomicReference` without relying on `AtomicReference`, making actions like store/replace/set/get atomic with reference count updates, e.g. using locks
2. adding the ability to mark a span as non-recyclable and use it for buffered spans

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -263,43 +263,49 @@ public void beforeEnd(long epochMicros) {
...
+                    if (parent.bufferedSpan.compareAndSet(parentBuffered, this)) {
+                        parent.decrementReferences();
+                        this.tracer.endSpan(parentBuffered);
+                    } else {
+                        parent.decrementReferences();
+                        // the failed update would ideally lead to a compression attempt with the new buffer,
+                        // but we're dropping the compression attempt to keep things simple
+                        this.tracer.endSpan(this);
+                    }
+                } else if (isSampled()) {
```
JonasKunz, line 297: We never decrement the `parent` references if `isSampled()==false` here. It was the same prior to your change, but is this intentional?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -263,43 +263,49 @@ public void beforeEnd(long epochMicros) {
...
     protected void afterEnd() {
         if (transaction != null && transaction.isSpanCompressionEnabled() && parent != null) {
-            Span buffered = parent.bufferedSpan.get();
-            if (parent.isFinished() || !isCompressionEligible()) {
-                if (buffered != null) {
-                    if (parent.bufferedSpan.compareAndSet(buffered, null)) {
-                        this.tracer.endSpan(buffered);
+            Span parentBuffered = parent.bufferedSpan.incrementReferencesAndGet();
+            try {
+                if (parent.isFinished() || !isCompressionEligible()) {
```
JonasKunz, line 268: This code can be reached even if this span is asynchronous to it's parent, correct?
If this is the case, we can't rely on `parent.isFinished()`, because it will always be racy.

So if the code outside this branch relies on the parent NOT being finished, this might cause problems.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -263,43 +263,49 @@ public void beforeEnd(long epochMicros) {
...
                 }
-                return;
-            }
-            if (!buffered.tryToCompress(this)) {
-                if (parent.bufferedSpan.compareAndSet(buffered, this)) {
+                if (parentBuffered == null) {
                     parent.decrementReferences();
-                    this.tracer.endSpan(buffered);
-                } else {
+                    if (!parent.bufferedSpan.compareAndSet(null, this)) {
```
JonasKunz, line 280: We access `parent.bufferdSpan` directly after a `parent.decrementReferences()`.
Doesn't this mean that `parent` might have already been recycled and reused. Therefore we might access the `bufferedSpan` of a totally unrelated Span?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/Span.java
```diff
@@ -262,50 +262,66 @@ public void beforeEnd(long epochMicros) {
...
+                    if (!parentBuffered.tryToCompress(this)) {
+                        if (parent.bufferedSpan.compareAndSet(parentBuffered, this)) {
+                            this.tracer.endSpan(parentBuffered);
+                        } else {
+                            // the failed update would ideally lead to a compression attempt with the new buffer,
+                            // but we're dropping the compression attempt to keep things simple
+                            this.tracer.endSpan(this);
+                        }
+                    } else {
+                        if (isSampled() && transaction != null) {
```
JonasKunz, line 305: [For Reviewer]:
Small change of behaviour here. Previously, if the compression was successfull but the span is **not** sampled, we would never
 * decrementReferences() on this
 * decrementReferences() on the parent
 
 I assumed that this was a bug and fixed it here. This behaviour was introduced [here](https://github.com/elastic/apm-agent-java/pull/2477/commits/40b709f5d4ad0fdfd043572ed169be6315946062) and was left uncommented, which is why I assume it was not intentional.´

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpanAtomicReference.java
```diff
@@ -38,8 +38,8 @@ public V incrementReferencesAndGet() {
             if (ref != null) {
                 ref.incrementReferences();
             }
+            return ref;
```
JonasKunz, line 41: [For Reviewer]
The return needs to be inside the synchronized, otherwise `ref` might have already been updated by someone else.

This was discovered thanks to the `SpanCompressionIT`.

jackshirazi: We may need a concurrent load test for this class, the intention for span compression was to avoid locks so that highly concurrent applications don't get penalized and I worry that this might become a bottleneck in that scenario

JonasKunz (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpanAtomicReference.java
jackshirazi: I spent 2 days before the break trying to setup a load test specifically for this, it's difficult. I'm going to review without load testing, and keep a background task for load testing

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpanAtomicReference.java
```diff
@@ -0,0 +1,77 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package co.elastic.apm.agent.impl.transaction;
+
+import javax.annotation.Nullable;
+
+public class AbstractSpanAtomicReference<V extends AbstractSpan<?>> {
```
jackshirazi, line 23: Why is this called `Abstract...` when it's not abstract?

jackshirazi: AtomicReferenceForAbstractSpan ?

jackshirazi (approved):
thanks, fixes several bugs and better code paths, nice

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpanAtomicReference.java
eyalkoren: Given that `Transaction` _is a_ `Span`, I think `SpanAtomicReference` would be appropriate. Let me know if you object, I can rename it.

PR #2869, opened 2022-11-11
Ensured that a connection is closed on connection failure

JonasKunz:
## What does this PR do?
We had a support case where a customer reported a leakage caused by agent conneciton failures.
In this particular case it was an IBM JVM where apparently each `HTTPUrlConnection` creates a thread. It looked like those `HttpUrlConnection` would leak upon connection failure, as at some point the application ran out of Threads.
Though normally this should be handled better by the JVM upon garbagecollection of the `HttpUrlConnection`s, it won't hurt to manually disconnect them.

## Checklist

- [x] This is a bugfix


eyalkoren (approved):
Good call!

eyalkoren:
@JonasKunz any reason not to go ahead with this?

PR #2870, opened 2022-11-14
action: use github actions

v1v:
## Description

* Use GitHub Actions instead of Jenkins for some stages:
  * Build
  * License
  * Unit Tests
  * Non-Application Server integration tests
  * Application Server integration tests
  * Javadoc
  * Build & Test Windows
  * JDK Compatibility Tests
* The project build is done once where possible.
* The CI use maven caching where possible. 

## Implementation

* Build Stage
  * It should always be triggered.
* License Stage
  * It should use build stage results.
  * It should always be triggered.
* Unit Tests Stage
  * It should use build stage results.
  * It should be triggered if the `test_ci` flag is true.
* Non-Application Server integration tests Stage
  * It should use build stage results.
  * It should be triggered if there is a label `ci:agent-integration`.
  * It should be triggered if the pull request isn't a draft.
  * It should be triggered if the `agent_integration_tests_ci` flag is true.
* Application Server integration tests Stage
  * It should use build stage results.
  * It should be triggered if there is a label `ci:agent-integration`.
  * It should be triggered if the pull request isn't a draft.
  * It should be triggered if the `agent_integration_tests_ci` flag is true.
* Javadoc Stage
  * It should use build stage results.
  * It should always be triggered.
* Build & Test Windows Stage
  * It should be triggered if the `test_ci` flag is true and :
    * the `windows_ci` flag is true.
    * or there is a label `ci:windows`.
* JDK Compatibility Tests Stage
  * It should use build stage results.
  * It should be triggered if there is a label `ci:jdk-compatibility`.
  * It should be triggered if the `jdk_compatibility_ci` flag is true.

## Results

Transition CI pipeline in Jenkins to GitHub actions

Screenshot 2023-02-28 at 09 23 38

## Next Steps

* [x] Snapshoty -> https://github.com/elastic/apm-agent-java/pull/2986
* Migration of release stage
* Migration of benchmark stage


eyalkoren:
@v1v are you planning to followup on this?

v1v:
> are you planning to followup on this?

Yes, we are working on the migration :)

v1v:
@amannocci , what's the reason for doing `force-push`? As a contributor to this PR, it makes the review harder and conflicts with my local branch.

amannocci:
> @amannocci , what's the reason for doing `force-push`? As a contributor to this PR, it makes the review harder and conflicts with my local branch.

Sorry about that, I will switch to the other method.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,247 @@
...
+    paths-ignore:
+      - '**.md'
+      - '**.asciidoc'
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
+      - ready_for_review
```
v1v, line 48: What's the reason for those types? IIRC, in Jenkins there are not labels-build events

.github/workflows/build/action.yml
```diff
@@ -0,0 +1,36 @@
...
+      with:
+        java-version: ${{ inputs.java-version }}
+        distribution: ${{ inputs.distribution }}
+        cache: 'maven'
+
+    - name: Fetch dependencies and build project
+      run: ./mvnw clean install -DskipTests=true -Dmaven.javadoc.skip=true
+      shell: ${{ inputs.shell }}
+
+    - run: ${{ inputs.command }}
```
v1v, line 34: I'm not sure if `clean` will produce an overhead when running with the unstash build artifact.

What's the reason for using `command` and the step `fetch dependencies` rather than running the given command instead?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,247 @@
...
+    paths-ignore:
+      - '**.md'
+      - '**.asciidoc'
+
+env:
```
v1v, line 53: I have not looked it at but maybe these env variables can be defined within the composite action? In case they are not used outside of the `uses: ./.github/workflows/build`

By doing so, we can remove a few lines of code

.github/workflows/ci.yml
```diff
@@ -0,0 +1,247 @@
...
+      - name: mvn license
+        uses: ./.github/workflows/build
+        with:
+          command: ./mvnw org.codehaus.mojo:license-maven-plugin:aggregate-third-party-report -Dlicense.excludedGroups=^co\\.elastic\\.
+          java-version: ${{ env.JAVA_VERSION }}
+          distribution: ${{ env.JAVA_DIST }}
+
+  unit-tests:
+    name: Unit Tests
+    if: inputs.test_ci == true || inputs.test_ci == null
```
v1v, line 99: when does `inputs.test_ci == null` happen? Probably a comment might help for the future 🙏

v1v (commented):
LGTM, with a few comments, in addition https://github.com/elastic/apm-agent-java/blob/main/.github/workflows/opentelemetry.yml needs to be updated, `test` is not the correct workflow name but `ci`

.mvn/maven.config
```diff
@@ -0,0 +1,6 @@
+-V
```
v1v, line 1: Super neat! https://github.com/elastic/apm-agent-java/blob/974ca32266fefb599d5560e93f4dbdce349473fd/.github/workflows/snapshoty.yml#L15-L21 is using the old fashion way, would you mind to update it too? 🙏

.github/workflows/build/action.yml
```diff
@@ -0,0 +1,36 @@
+---
+
+name: common build tasks
```
v1v, line 3: I wonder whether https://github.com/elastic/apm-agent-java/blob/974ca32266fefb599d5560e93f4dbdce349473fd/.github/workflows/snapshoty.yml could also use this composite action, what do you think?

.github/workflows/ci.yml
amannocci: We have some specific jobs to run if specific labels are present or when a PR is ready for review.
It avoids running the PR manually in some cases.

.github/workflows/build/action.yml
amannocci: It shouldn't make any difference with the stash/unstash action in place.
I removed the `fetch dependencies` step.

.github/workflows/main.yml
```diff
@@ -0,0 +1,255 @@
...
+          name: build
+          path: ${{ github.workspace }}
+
+  license:
+    name: License
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/workflows/unstash
```
JonasKunz, line 87: [Question] Doesn't the unstash override everything from the checkout above? can't the checkout be omitted then? (Same for the other jobs)

JonasKunz (approved):
Looks very clean and understandable! Great work!

.github/workflows/test-reporter.yml
```diff
@@ -0,0 +1,20 @@
+---
+## Workflow to process the JUnit test results and add a report to the checks.
+name: test-reporter
+on:
+  workflow_run:
+    workflows:
+      - ci
```
JonasKunz, line 7: Just noticed after approving: Wasn't the `ci` workflow renamed?

.github/workflows/main.yml
```diff
@@ -0,0 +1,255 @@
...
+        with:
+          name: build
+          path: ${{ github.workspace }}
+      - uses: ./.github/workflows/maven-goal
+        with:
+          command: ./mvnw test
```
JonasKunz, line 116:
```suggestion
          command: ./mvnw test -DargLine="-Delastic.apm.overwrite.config.docs=false"
```

This required argument was just added with a #3027.

.github/workflows/main.yml
amannocci: You are right, the `unstash` action will override the local content.
Unfortunately, we can't use any local action without checkout the repository.

.github/workflows/main.yml
```diff
@@ -0,0 +1,239 @@
...
+      - uses: ./.github/workflows/unstash
+        with:
+          name: build
+          path: ${{ github.workspace }}
+      - name: install artifacts
+        uses: ./.github/workflows/maven-goal
+        with:
+          # Need to install packages in m2 local repository
+          command: ./mvnw install -DskipTests=true -Dmaven.javadoc.skip=true
+      - name: mvn license
```
v1v, line 90: missed `uses: ./.github/workflows/maven-goal`?

.github/workflows/snapshoty.yml
```diff
@@ -13,35 +13,20 @@ on:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
 
-      - name: Set up JDK ${{ env.JAVA_VERSION }}
-        uses: actions/setup-java@v3
+      - uses: ./.github/workflows/maven-goal
         with:
           java-version: ${{ env.JAVA_VERSION }}
           distribution: ${{ env.JAVA_DIST }}
```
v1v, line 28: I'd say to use the default values and remove the env variable above

v1v (commented):
A couple of NITs

.github/workflows/main.yml
amannocci: AFAIS, we need to install project jars in the local maven repository to be able to run the license step.
This step isn't required in most cases.

v1v: I meant whether `mvn license` should use the reusable action called ` ./.github/workflows/maven-goal`, as far as I see it uses `run: ./mvnw org.codehaus.mojo:license-maven-plugin...`

v1v (commented):
LGTM, though I cannot approve it as I was the former author

.github/workflows/main.yml
v1v: Gotcha, thanks for the explanation! all ✅ in my end!

amannocci (approved):

PR #2871, opened 2022-11-16
Fix security manager issues

SylvainJuge:
## What does this PR do?

Fixes #2520 

- fixes `NoClassDefFound` errors triggered in classloading due to URL validation in JDK outside of a `doPrivileged` wrapping.
- fixes few missing `doPrivileged` calls to make the agent able to start properly (reading environment variables, get classloader reference of an existing class and resolve method handles).
- surprisingly, some of the calls to `System.getenv` trigger exception when not called in a privileged action, whereas the `System.getProperty` or `System.getProperties` do not, as a result I added a `PrivilegedActionUtils` class to properly wrap those calls with minimal code duplication and verbosity.

## Checklist

- [x] This is a bugfix
  - [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)
  - [x] I have added tests that would fail without this fix. tests will be provided by https://github.com/elastic/apm-agent-java/pull/2883.
  - [x] Manual tests with Tomcat 9.0.38 + JDK 11.0.8 + security manager enabled, started with `./bin/catalina.sh run -security` + enabling most agent features (log reformatting, inferred spans, ...)

Issue #2520, opened 2022-03-17
NoClassDefFound with Tomcat + Security Manager


SylvainJuge:
When a security manager is configured in Tomcat (might also happen with other server/apps), the standard output is polluted with lots of `NoClassDefFoundError` errors.

How to reproduce:
- grant agent every permission in `conf/catalina.policy` , for example: 
  ```
  grant codeBase "file:/path/to-agent/elastic-apm-agent-1.29.1-SNAPSHOT.jar" {
    permission java.security.AllPermission;
  };
  ```
- start tomcat with `bin/catalina.sh run -security`

Agent version: 1.29.0 + same on main as of writing
Tomcat version: 9.0.x

Setting `warmup_byte_buddy=false` does not makes a difference here.
While the symptoms might be similar, the fix for #2509 does not cover this case, thus it might not be related but similar.

<details>
  <summary>catalina.out output</summary>
 
  ```
  2022-03-17 10:31:22,251 [main] INFO  co.elastic.apm.agent.util.JmxUtils - Found JVM-specific OperatingSystemMXBean interface: com.sun.management.OperatingSystemMXBean
2022-03-17 10:31:22,290 [main] INFO  co.elastic.apm.agent.configuration.StartupInfo - Starting Elastic APM 1.29.1-SNAPSHOT.UNKNOWN as tomcat-application on Java 11.0.8 Runtime version: 11.0.8+10 VM version: 11.0.8+10 (AdoptOpenJDK) Linux 5.13.0-30-generic
2022-03-17 10:31:22,290 [main] INFO  co.elastic.apm.agent.configuration.StartupInfo - warmup_byte_buddy: 'false' (source: Java System Properties)
2022-03-17 10:31:22,290 [main] WARN  co.elastic.apm.agent.configuration.StartupInfo - To enable all features and decrease startup time, please configure application_packages
2022-03-17 10:31:23,417 [main] INFO  co.elastic.apm.agent.impl.ElasticApmTracer - Tracer switched to RUNNING state
2022-03-17 10:31:23,450 [elastic-apm-server-healthcheck] INFO  co.elastic.apm.agent.report.ApmServerHealthChecker - Elastic APM server is available: {  "build_date": "2022-02-03T16:50:25Z",  "build_sha": "6afb387ae002e5a8dc2427b53225aa42681e4208",  "publish_ready": true,  "version": "8.0.0"}
[...]

PR #2883, opened 2022-11-22
Enable security manager in Servlet container tests

eyalkoren:
Adds servlet container tests with security manager enabled

Currently, the non-jakarta Tomcat tests reproduce the errors observed in https://github.com/elastic/apm-agent-java/issues/2520#issue-1172146921.

Currently blocked. Once #2871 is merged, we should bring this one to a functional state.


apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -753,48 +755,64 @@ public static File getAgentJarFile() {
...
-                        ElementMatcher.Junction<? super TypeDescription> typeMatcher = getTypeMatcher(classToInstrument, apmInstrumentation.getMethodMatcher(), none());
-                        if (typeMatcher != null && isIncluded(apmInstrumentation, config)) {
-                            agentBuilder = applyAdvice(tracer, agentBuilder, apmInstrumentation, typeMatcher.and(apmInstrumentation.getTypeMatcher()));
-                        }
+                final Set<Collection<Class<? extends ElasticApmInstrumentation>>> updatedAppliedInstrumentations =
+                    dynamicallyInstrumentedClasses.get(classToInstrument);
+                if (!updatedAppliedInstrumentations.contains(instrumentationClasses)) {
+                    if (System.getSecurityManager() == null) {
+                        applyInstrumentation(classToInstrument, instrumentationClasses, updatedAppliedInstrumentations, tracer);
+                    } else {
```
SylvainJuge, line 774: [minor] do you remember what method(s) triggered security manager checks here ? Maybe a small comment might help to understand why it's required here (and could also help answer questions we may have about all the "privileged code" we have).

integration-tests/external-plugin-test/external-plugin/src/main/java/co/elastic/apm/agent/plugin/PluginInstrumentation.java
```diff
@@ -47,12 +47,24 @@ public class PluginInstrumentation extends ElasticApmInstrumentation {
 
     @Override
     public ElementMatcher<? super TypeDescription> getTypeMatcher() {
-        return named(System.getProperty("elastic.apm.plugin.instrumented_class", "co.elastic.apm.plugin.test.TestClass"));
+        String customType = null;
+        try {
+            customType = System.getProperty("elastic.apm.plugin.instrumented_class");
+        } catch (Exception e) {
```
SylvainJuge, line 53: [minor] why not catch `SecurityException` and only ignore those ?

integration-tests/external-plugin-test/external-plugin/src/main/java/co/elastic/apm/agent/plugin/PluginInstrumentation.java
```diff
@@ -47,12 +47,24 @@ public class PluginInstrumentation extends ElasticApmInstrumentation {
...
+        return named(customType != null ? customType : "co.elastic.apm.plugin.test.TestClass");
     }
 
     @Override
     public ElementMatcher<? super MethodDescription> getMethodMatcher() {
-        return named(System.getProperty("elastic.apm.plugin.instrumented_method", "traceMe"));
+        String customMethod = null;
+        try {
+            customMethod = System.getProperty("elastic.apm.plugin.instrumented_method");
+        } catch (Exception e) {
```
SylvainJuge, line 64: [minor] same as previous, we could catch `SecurityException`.

integration-tests/application-server-integration-tests/src/test/resources/catalina.policy
```diff
@@ -0,0 +1,274 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more
```
SylvainJuge, line 1: Is this tomcat policy file the same for all Tomcat versions ? I would be quite surprised if it was the case, and it might defeat the purpose of testing with "the tomcat server as the users will experience it". Also, it would make it harder to do the same with other containers.

If possible, I would prefer if we could just append the last few lines to an existing policy file, maybe creating a policy on the fly in `enableSecurityManager`.

```
grant codeBase "file:/elastic-apm-agent.jar" {
    permission java.security.AllPermission;
};
```

SylvainJuge (approved):
LGTM, maybe only modify the existing policy instead of replacing it, but that could be left as an improvement for later.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
eyalkoren: I'll look for it

integration-tests/application-server-integration-tests/src/test/resources/catalina.policy
eyalkoren: > Is this tomcat policy file the same for all Tomcat versions ? 

Not 100%, there are some minor small additions of specific permissions through versions. Does it matter? 

> I would be quite surprised if it was the case, and it might defeat the purpose of testing with "the tomcat server as the users will experience it". 

How would it defeat the purpose? The purpose is to see whether the agent functions as expected without causing issues when the security manager is enabled. The permissions within the policy are only allowing basic Tomcat stuff and web apps to function properly.
The way I know it does not defeat the purpose is that it helped me identify those places where additional changes in the agent code were still missing. The tests fail without these fixes.

> Also, it would make it harder to do the same with other containers.

This one is only adding to Tomcat and it already proved very efficient as it helped identifying additional required fixes. It is not a base for adding to other Servlet containers anyway as it relies on `JAVA_OPTS`, which is specific to Tomcat's startup scripts.

> If possible, I would prefer if we could just append the last few lines to an existing policy file, maybe creating a policy on the fly in enableSecurityManager.

The effective policy at runtime is a superset of all used policies. To my understanding, this includes policies added through `-Djava.security.policy=...` and using `-Djava.security.policy==...` would override other policies. I also wanted to do the same, only adding a policy that grants permissions to our agent, assuming like you that the Tomcat image includes a default policy file, but apparently it doesn't, as this didn't work without providing a full policy.

eyalkoren:
> maybe only modify the existing policy instead of replacing it

What we do here is not replacing, it should be additive, and I am not sure what it means to modify the existing policy - we use Tomcat Docker images.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
eyalkoren: The first one to fail is:
```
[tomcat] java.security.AccessControlException: access denied ("java.io.FilePermission" "/elastic-apm-agent.jar" "read")
[tomcat] at java.security.AccessControlContext.checkPermission(AccessControlContext.java:472)
[tomcat] at java.security.AccessController.checkPermission(AccessController.java:684)
[tomcat] at java.lang.SecurityManager.checkPermission(SecurityManager.java:549)
[tomcat] at java.lang.SecurityManager.checkRead(SecurityManager.java:888)
[tomcat] at java.util.zip.ZipFile.<init>(ZipFile.java:212)
[tomcat] at java.util.zip.ZipFile.<init>(ZipFile.java:151)
[tomcat] at java.util.jar.JarFile.<init>(JarFile.java:154)
[tomcat] at java.util.jar.JarFile.<init>(JarFile.java:118)
[tomcat] at net.bytebuddy.dynamic.ClassFileLocator$ForJarFile.of(ClassFileLocator.java:782)
[tomcat] at co.elastic.apm.agent.bci.ElasticApmAgent.getAgentBuilder(ElasticApmAgent.java:665)
[tomcat] at co.elastic.apm.agent.bci.ElasticApmAgent.applyInstrumentation(ElasticApmAgent.java:791)
[tomcat] at co.elastic.apm.agent.bci.ElasticApmAgent.ensureInstrumented(ElasticApmAgent.java:772)
...
```
I assume if I add a specific permission for that I would find the second, however I think it is enough to add a general comment, which I will.

PR #2885, opened 2022-11-22
Possible Reactor context loss in TracedSubscriber and WebClientSubscriber fix

MichaelGT:
After activating/attaching of agent to java process all custom information stored by Mono.contextWrite(..) becomes unreachable later in subscriber

## Steps to reproduce
Store something in the context
```
webClient
  .get()
  .uri(...)
  .exchangeToMono(...)
  .contextWrite(ctx -> ctx.put("KEY_1", "VAL_1"));
```

And later if we'll try to get `context.get("KEY_1")` this info from the context in a subscriber we'll get `NoSuchElementException`

Without activating/attaching of agent we'll get this info from the context.

Looks like the problem is in `co.elastic.apm.agent.reactor.TracedSubscriber` and `co.elastic.apm.agent.springwebclient.WebClientSubscriber`. 
In `TracedSubscriber` the real subscriber's context is stored in the constructor but should be obtained each time in currentContext() 
```
@Override
public Context currentContext() {
    final AbstractSpan<?> context = getContext();
    if (context != null) {
        return subscriber.currentContext().put(AbstractSpan.class, context);
    } else {
        return subscriber.currentContext();
    }
}
```

In `WebClientSubscriber` default implementation of method `currentContext()` in parent `CoreSubscriber` recreates empty context but should delegate this call to a real subscriber:
```
@Override
public Context currentContext() {
    return subscriber.currentContext();
}
```

### DIFF
--- a/apm-agent-plugins/apm-reactor-plugin/src/main/java/co/elastic/apm/agent/reactor/TracedSubscriber.java
+++ b/apm-agent-plugins/apm-reactor-plugin/src/main/java/co/elastic/apm/agent/reactor/TracedSubscriber.java
@@ -54,20 +54,20 @@ public class TracedSubscriber<T> implements CoreSubscriber<T> {
 
     private final Tracer tracer;
 
-    private final Context context;
-
     TracedSubscriber(CoreSubscriber<? super T> subscriber, Tracer tracer, AbstractSpan<?> context) {
         this.subscriber = subscriber;
         this.tracer = tracer;
         contextMap.put(this, context);
-
-        // store our span/transaction into reactor context for later lookup without relying on active tracer state
-        this.context = subscriber.currentContext().put(AbstractSpan.class, context);
     }
 
     @Override
     public Context currentContext() {
-        return context;
+        final AbstractSpan<?> context = getContext();
+        if (context != null) {
+            return subscriber.currentContext().put(AbstractSpan.class, context);
+        } else {
+            return subscriber.currentContext();
+        }
     }
 
     /**
--- a/apm-agent-plugins/apm-spring-webflux/apm-spring-webclient-plugin/src/main/java/co/elastic/apm/agent/springwebclient/WebClientSubscriber.java
+++ b/apm-agent-plugins/apm-spring-webflux/apm-spring-webclient-plugin/src/main/java/co/elastic/apm/agent/springwebclient/WebClientSubscriber.java
@@ -21,7 +21,6 @@ package co.elastic.apm.agent.springwebclient;
 import co.elastic.apm.agent.collections.WeakConcurrentProviderImpl;
 import co.elastic.apm.agent.impl.Tracer;
 import co.elastic.apm.agent.impl.context.web.ResultUtil;
-import co.elastic.apm.agent.impl.transaction.Outcome;
 import co.elastic.apm.agent.impl.transaction.Span;
 import co.elastic.apm.agent.sdk.weakconcurrent.WeakMap;
 import org.reactivestreams.Subscription;
@@ -29,6 +28,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.web.reactive.function.client.ClientResponse;
 import reactor.core.CoreSubscriber;
+import reactor.util.context.Context;
 
 import javax.annotation.Nullable;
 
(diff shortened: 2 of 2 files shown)


SylvainJuge:
Hi @MichaelGT , thanks for opening this PR.

One thing that is missing are the tests to ensure this issue is properly fixed, from what I understand the issue is in reactor instrumentation, but the only known way to trigger it so far is in webclient instrumentation.

I am really not an expert in Webflux, so maybe I am missing something obvious here.

I tried to reproduce it directly into our unit tests on your branch, but the `currentContext` method that you modified does not seem to be called when I modified `co.elastic.apm.agent.springwebclient.WebClientInstrumentationTest.RequestStrategy#EXCHANGE_TO_MONO` to look like this:

```
        EXCHANGE_TO_MONO {
            @Override
            void execute(WebClient client, String uri) {
                client.get().uri(uri).exchangeToMono(response -> response.bodyToMono(String.class))
                    // test context value storage
                    .contextWrite(ctx -> ctx.put("key", "value"))
                    .doOnEach(signal -> {
                        String signalName = signal.getType().name();
                        ContextView contextView = signal.getContextView();
                        assertThat(contextView.isEmpty())
                            .describedAs("context should not be empty for signal '%s'", signalName)
                            .isFalse();

                        Object contextValue = contextView.get("key");
                        assertThat(contextValue)
                            .describedAs("value stored in context should be preserved")
                            .isEqualTo("value");
                    })
                    .block();
            }
        },
```

Also, you will have to sign the CLA agreement so we can move forward with your contribution.

Last but not least, do you mind if we push changes directly into your PR if required ? That could help avoid some back & forth with minor details like formatting or updating the branch with latest version of `main`. Given that you opened this PR using your own fork `main`  branch, that means that it could conflict with other changes you would be pushing to this branch (if this PR is the only reason you have this fork, this won't be an issue).

eyalkoren:
@MichaelGT this is getting stale, please respond to @SylvainJuge's [questions above](https://github.com/elastic/apm-agent-java/pull/2885#issuecomment-1332116278) or close this issue if it is not needed anymore.
Thanks!

akrepon:
Hi,

I am having a similar issue. In my case I have registered a reactor Hook.onLastOperator to pass context data. During my tests I noticed that my context data was missing while elastic-apm context was added every single time. At first I thought the reactor instrumentation was causing this. After checking the implementation, I concluded that TracedSubscriber was passing the context correctly. Then I decided to disable instrumentation for reactor plugin. After disabling it the elastic apm Context disappeared. The final context was completely empty, which suggested me there is something that was overwriting the context. Then I started disabling plugins one by one until I disabled spring-webclient plugin. The problem was gone and context was correctly propagated. Then I started checking the code for this plugin and indeed the context is overwritten by default CoreSubscriber.currentContext which is empty. 

Conclusion: 

co.elastic.apm.agent.springwebclient.WebClientSubscriber needs to pass the original context.

tomholik:
Hi @eyalkoren  @SylvainJuge ,
I believe this problem and fix is still relevant. We have similar issue as described by @akrepon 

Spring boot 3.1.0. with webflux and kotlin coroutines, elastic apm-agent-attach 1.39.0. We pass custom context data and update MDC context with Hooks.onEachOperator. This works fine and APM context data are always present but our custom context in one case missing. When is Elastic instrumentation (ElasticApmAttacher.attach()) disabled it works without issue. 

I tried locally to fix it in version 1.39.0 - update of `WebClientSubscriber` to pass original context as it currently creates emptyContext (https://github.com/elastic/apm-agent-java/pull/2885/files#diff-7848da190e69a2f2bb53228f99a879a47d1c4cb6cd25ec97fe5db3bc1f7a76f0). With those changes our custom context data were propagated correctly. For our use case update of `TracedSubscribed` is not necessary.

Custom context data is missing in `customJackson2JsonEncoder` which is used before HTTP call to external service. 
```
WebClient.builder()
            .codecs { codecConfigurer ->
                with(codecConfigurer.defaultCodecs()) {
                    jackson2JsonEncoder(customJackson2JsonEncoder)
                }
            }.build
```

Can you please fix it in another PR in @MichaelGT is not responding for CLA - it's been almost a year and it's a small change. 

@kananindzya Can you please check it as original author? thanks

v1v:
run docs-build

PR #2888, opened 2022-11-23
Added agent background overhead metrics

### LINKED ISSUE #2543: [META 625] Agent health metrics

JonasKunz:
## What does this PR do?
Adds a tech preview for agent background overhead metrics (https://github.com/elastic/apm-agent-java/issues/2543).
Closes #2543.

## Checklist


apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/AgentOverheadMetrics.java
```diff
@@ -0,0 +1,453 @@
...
+    /**
+     * Contains all started, running and stopped threads since the last metrics reporting.
+     */
+    private final ConcurrentHashMap<Thread, ThreadInfo> lastThreadInfo = new ConcurrentHashMap<>();
+
+    private static class ThreadInfo {
+        final String threadPurpose;
+
+        @Nullable
+        volatile Long cpuTime;
```
jackshirazi, line 69: Would be better to avoid the `Long` object boxing and unboxing if it's straightforward

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/AgentOverheadMetrics.java
```diff
@@ -0,0 +1,442 @@
...
+            if (!cpuTimeMeasurementEnabled) {
+                cpuOverheadMetricEnabled = false;
+                cpuUsageMetricEnabled = false;
+            }
+        }
+
+        if (anyMetricEnabled()) {
+            //initialize values for first report
+            lastReportedProcessCpuTime = getProcessCpuTime();
+            ExecutorUtils.setThreadStartListener(this);
```
jackshirazi, line 153: How do you ensure that this is run before any agent threads have been created? Or do you just ignore threads created before this as not counted and they don't matter?

JonasKunz: > Or do you just ignore threads created before this as not counted and they don't matter

Partly this. To be more precise we ignore any cpu usage / allocations which have occured prior to this line.
However, we still count threads which were started before, we just start the counting at this point in time.

I thought that this was an okay limitation, because it is not really possible to capture the entire startup overhead:
E.g. for the overhead metrics, the configuration and logging needs to be initialized already. We could do the measurement right after that by relying on manual instantiation instead of `AbstractLifecycleListener`, but I thought that this is just not worth the introduced spaghetti by manual wiring.

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/AgentOverheadMetrics.java
```diff
@@ -0,0 +1,442 @@
...
+            }
+        }
+    }
+
+    private void collectActiveThreadsMetric(MetricCollector collector) {
+        if (!threadCountMetricEnabled) {
+            return;
+        }
+        Map<String, AtomicLong> threadCountByPurpose = new HashMap<>();
+
```
jackshirazi, line 246: can we reuse the Map at least for these 3 calls, so one map created and dumped instead of 3. I don't think it's necessary to reuse it across different collectAndReset calls since that is multiple seconds apart and one map per call is acceptable. 3 maps probably are too, but it's so easy to reuse that one we might as well cut down that garbage

jackshirazi (approved):
nice work

Issue #2887, opened 2022-11-23
Micrometer test flakyness


JonasKunz:
## Test method

`MicrometerInstrumentationStepTest.testStepVsCumulativeMultiple`

## Test logs

<details>
<summary>Click to expand</summary>

```
java.lang.RuntimeException: Condition not fulfilled within 5000 ms
	at co.elastic.apm.agent.MockReporter.awaitUntilAsserted(MockReporter.java:636)
	at co.elastic.apm.agent.micrometer.MicrometerInstrumentationStepTest.testStepVsCumulative(MicrometerInstrumentationStepTest.java:142)
	at co.elastic.apm.agent.micrometer.MicrometerInstrumentationStepTest.testStepVsCumulativeMultiple(MicrometerInstrumentationStepTest.java:81)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
[...]

PR #2897, opened 2022-12-01
increase timeout for MicrometerInstrumentationStepTest.testStepVsCumulativeMultiple

jackshirazi:
increase timeout for MicrometerInstrumentationStepTest.testStepVsCumulativeMultiple micrometer flaky test. I think it's just slow on the more constrained CI vCPUs so bump the timeout right up to where if it fails it's likely not a CPU load issue

### DIFF
--- a/apm-agent-plugins/apm-micrometer-plugin/src/test/java/co/elastic/apm/agent/micrometer/MicrometerInstrumentationStepTest.java
+++ b/apm-agent-plugins/apm-micrometer-plugin/src/test/java/co/elastic/apm/agent/micrometer/MicrometerInstrumentationStepTest.java
@@ -139,7 +139,7 @@ public class MicrometerInstrumentationStepTest {
                 }
             }).start();
         }
-        reporter.awaitUntilAsserted(5000, () ->
+        reporter.awaitUntilAsserted(20000, () ->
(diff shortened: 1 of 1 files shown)


JonasKunz (approved):
Increasing the timeout won't hurt, but I'm pretty sure it won't solve the issue.
This excerpt from the logs in #2887 shows that the value of a counter is actually higher than expected IINM:
```
Caused by: org.opentest4j.AssertionFailedError: 
expected: FooBar{fooCount=0, barCount=21}
 but was: FooBar{fooCount=0, barCount=22}
```

The only reasonable explanation coming to my mind is either (a) something notbeing cleaned up from a previous run or (b) some race condition

jackshirazi:
there's a bunch of reasons this can fail because of the 3 different concurrent loops happening: micrometer meters getting updated, the agent reporting sampling that and flushing, and the test loop reading the reporting output. I ran through thousands of tests and analyzed the failures (roughly one in 500 I think) and they were non-functional ones related to the 3 loops interacting, so I decided to go with this. I'm hoping that increasing the timeout will get the CI to the same level. 

The failure rate is potentially real in that it could happen in prod, with interval sampled values slightly off occasionally, this is noted in the class comments and deemed acceptable (at least for now)

PR #2914, opened 2022-12-06
Change to DEBUG (from INFO) 'Not re-registering MeterRegistry ...'

jackshirazi:
## What does this PR do?
Change to DEBUG (from INFO) 'Not re-registering MeterRegistry ...' message as it is too verbose

### DIFF
--- a/apm-agent-plugins/apm-micrometer-plugin/src/main/java/co/elastic/apm/agent/micrometer/MicrometerMetricsReporter.java
+++ b/apm-agent-plugins/apm-micrometer-plugin/src/main/java/co/elastic/apm/agent/micrometer/MicrometerMetricsReporter.java
@@ -114,7 +114,7 @@ public class MicrometerMetricsReporter implements Runnable, Closeable {
         }
         long step = getStep(meterRegistry);
         if (step >= 0 && step < 1000) {
-            logger.warn("Not registering unsupported step interval of {} milliseconds (1 seconds is the minimum supported) for Micrometer MeterRegistry: {}", step, meterRegistry);
+            logger.debug("Not registering unsupported step interval of {} milliseconds (1 seconds is the minimum supported) for Micrometer MeterRegistry: {}", step, meterRegistry);
             return;
         }
         Step newStep = new Step(step);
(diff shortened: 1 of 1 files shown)


apm-agent-plugins/apm-micrometer-plugin/src/main/java/co/elastic/apm/agent/micrometer/MicrometerMetricsReporter.java
```diff
@@ -120,7 +120,7 @@ void registerMeterRegistry(MeterRegistry meterRegistry) {
         Step newStep = new Step(step);
         Step hopefullyNull = meterRegistries.putIfAbsent(meterRegistry, newStep);
         if (hopefullyNull != null) {
-            logger.info("Not re-registering MeterRegistry as it is already registered from another compound meter registry: {}", meterRegistry);
+            logger.debug("Not re-registering MeterRegistry as it is already registered from another compound meter registry: {}", meterRegistry);
```
JonasKunz, line 123: If my understanding is correct, this code path is executed basically on every interaction with a meter, which in turn causes the large amounts of logging.

I feel like `DEBUG`-level is still to much: It would clutter the logs we use for support cases.
I would move it to trace or even remove this statement completely, because I don't know hat it would be used for.

JonasKunz: Can't comment on unchanged code, so adding this here:
```
        if (step >= 0 && step < 1000) {
            logger.warn("Not registering unsupported step interval of {} milliseconds (1 seconds is the minimum supported) for Micrometer MeterRegistry: {}", step, meterRegistry);
            return;
        }
```
This will also cause a spam of warnings IINM. Maybe we should make sure to log only once per registry?

JonasKunz: And another one I'm not sure about:
```
logger.warn("Identified Micrometer SimpleConfig: {}", config);
```
Shouldn't this one be `debug`?

jackshirazi: yes agree, will move to trace. I don't want to add additional memory cost to track which have been logged, trace is fine. It is useful when there is a problem tracing which meters are being processed.

Also agree that other should be debug

JonasKunz (approved):

siddharth76:
@JonasKunz  @jackshirazi  Hey i am using apm-agent-attach/1.35.0, latest version of apm-agent-attach, still getting above logs undefinetely

code4mankind:
@JonasKunz @jackshirazi any idea when this fix will be released?   We are currently blocked on upgrading until that time - presumably we will be seeing a `1.35.1` release soon that includes this fix.

jackshirazi:
Targeting this month

mrodal:
this couldnt make it to 1.36.0? 😥

PR #2915, opened 2022-12-07
Support daily snapshots

amannocci:
Signed-off-by: Adrien Mannocci <adrien.mannocci@elastic.co>

## What does this PR do?

* Add a stage to publish snapshots in a GCS bucket.

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

Screenshot 2022-12-07 at 12 05 04

### DIFF
--- a/.ci/snapshoty.yml
+++ b/.ci/snapshoty.yml
@@ -1,0 +1,53 @@
+---
+
+# Version of configuration to use
+version: '1.0'
+
+# You can define a Google Cloud Account to use
+account:
+  # Project id of the service account
+  project: '${GCS_PROJECT}'
+  # Private key id of the service account
+  private_key_id: '${GCS_PRIVATE_KEY_ID}'
+  # Private key of the service account
+  private_key: '${GCS_PRIVATE_KEY}'
+  # Email of the service account
+  client_email: '${GCS_CLIENT_EMAIL}'
+  # URI token
+  token_uri: 'https://oauth2.googleapis.com/token'
+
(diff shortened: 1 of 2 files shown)


kuisathaverat (approved):

SylvainJuge:
The snapshots are also published daily to sonatype repository (in internal CI), would it be something that we could also automate from the GCP storage (or bundle them together) to ensure having the same snapshot both in the private GCP and public repositories synchronized ?

While the risk of inconsistency between the two is quite low, having the whole publication in the same pipeline would likely make things simpler.

amannocci:
The mechanism isn't exactly the same. We are publishing each snapshot artifact to a specific GCS bucket to make them available. In the other, Jenkins job we are publishing each snapshot artifact daily. The syntax isn't the same between both Jenkins jobs.
It's possible to adapt the daily job in the other CI to also publish on the GCS bucket.
Do we want each snapshot artifact published on Sonatype nexus on every commit?

SylvainJuge:
> Do we want each snapshot artifact published on Sonatype nexus on every commit?
I don't think that is required, but if we could that would be awesome as it means that once something is merged then the next artifact available publicly would already contain the change. Right now we have to wait until the next day.

We sometimes use snapshots for customer support, but most of the time we can rely on direct links to the CI before the PR is merged. In short that would be a nice bonus but we can definitely live without it.

SylvainJuge (approved):
LGTM, let's leave sonatype repository as-is for now.

PR #2926, opened 2022-12-13
Add Activation method for telemetry on how the agent was started

jackshirazi:
## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [x] This is something else


apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
```diff
@@ -0,0 +1,134 @@
...
+    K8S_ATTACH, // https://github.com/elastic/apm-mutating-webhook
+    PROGRAMMATIC_SELF_ATTACH, // ElasticApmAttacher.attach();
+    JAVAAGENT_FLAG, // -javaagent:... command-line option used
+    AWS_LAMBDA_LAYER, // Only if installed by using layers (as other metadata already identifies lambda)
+    ENV_ATTACH, // JAVA_TOOL_OPTIONS env var used to specify -javaagent option
+    APM_AGENT_ATTACH_CLI, // 'java -jar apm-agent-attach-cli.jar ...' used
+    FLEET, // Fleet using 'java -jar apm-agent-attach-cli.jar ...' to attach (directly or through webhook)
+
+    AZURE_FUNCTIONS, //Not defined yet how this is identified
+    UNKNOWN;
```
eyalkoren, line 40: If you move this to the `apm-agent-common` module, it would be visible to attacher, attacher-cli and core agent, which may improve reuse and save some conversions. You should be able to use it directly in `CoreConfiguration` and benefit from the automatic mapping.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
```diff
@@ -0,0 +1,134 @@
...
+                //if there is more than one, this will return the first, which is the correct algorithm
+                for (String javaArg : javaArgs) {
+                    if (javaArg.startsWith("-javaagent:") && javaArg.contains(agentJarFile)) {
+                        return javaArg;
+                    }
+                }
+            }
+        }
+        return null;
+    }
```
eyalkoren, line 133: Since this basically finds out whether the `-javaagent` was used, can we just rely on whether the agent was started through `co.elastic.apm.agent.premain.AgentMain#premain()` or `co.elastic.apm.agent.premain.AgentMain#agentmain()`? Would we miss anything?
Note that this information already permeates to `co.elastic.apm.agent.bci.ElasticApmAgent#initialize`, so if this is good enough, it should be much easier to get.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        //The next picks up `Container image' lambdas but only if the layer is used, which we
+        //aren't sure is good or bad. We could also check ELASTIC_APM_AWS_LAMBDA_HANDLER
+        //but have deliberately decided not to (for now)
+        String lambdaWrapper = PrivilegedActionUtils.getEnv("AWS_LAMBDA_EXEC_WRAPPER");
+        logger.debug("AWS_LAMBDA_EXEC_WRAPPER: {}", lambdaWrapper);
+        if (null != lambdaWrapper && lambdaWrapper.equals("/opt/elastic-apm-handler")) {
+            return AWS_LAMBDA_LAYER;
+        }
```
eyalkoren, line 82: Since you already added `activation_method` as a config option, wouldn't setting 
```
export ELASTIC_APM_ACTIVATION_METHOD=AWS_LAMBDA_LAYER
``` 
in [`elastic-apm-handler`](https://github.com/elastic/apm-agent-java/blob/c8c8769d91efd2c153b96fb472cb601e2e8c0b7f/elastic-apm-agent/src/main/assembly/elastic-apm-handler) just do?

If it's in order to support existing customers, I would say it doesn't worth the heuristics.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
```diff
@@ -0,0 +1,134 @@
...
+
+        String elasticJavaagentOnTheCommandline = getElasticJavaagentOnTheCommandline();
+        logger.debug("Agent on command line: {}", elasticJavaagentOnTheCommandline);
+        String javaToolOptions = PrivilegedActionUtils.getEnv("JAVA_TOOL_OPTIONS");
+        logger.debug("JAVA_TOOL_OPTIONS: {}", javaToolOptions);
+        if (null != javaToolOptions) {
+            // fallback k8s test, this is the pattern used by the webhook, but it's
+            // a fragile check as the webhook could easily be changed
```
eyalkoren, line 90: Similar to AWS lambda - since there is a configuration option and since we own the webhook, instead of such heuristics we can simply set this configuration option through the webhook. We already [set our own env variables](https://github.com/elastic/apm-mutating-webhook/blob/main/patch.go#L112-L165), so IIANM, we should only set `ELASTIC_APM_ACTIVATION_METHOD= K8S_ATTACH` to get this very deterministic.

apm-agent-core/src/test/resources/junit-platform.properties
```diff
@@ -0,0 +1 @@
+junit.jupiter.execution.parallel.enabled=true
```
eyalkoren, line 1: Why here? Could there be negative implications like increased flakyness? (if not - I am all for it of course 🙂 )

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
jackshirazi: I deliberately kept the `ActivationType` - which is the type in the spec and reported by the agent - separate from `CoreConfiguration.ActivationMethod` because the reported spec/type has more options than the possible values passed to the agent, eg `ENV_ATTACH` is something that should be inferred by the agent by looking at the env rather than a passed value. I'm not wedded to this separation, but I still currently think it's a better option

I can move it to common, though I don't think anything else needs it so not sure it's better to widen the scope

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
jackshirazi: yes, the reasoning was lower touchpoints in other areas for a very tiny higher effort here, plus I didn't want to burden the lambda user with yet another env var in their system - in particular one that is entirely non-functional for them (ie keep their cognitive load down, and avoid superfluous questions to us and support ("why is ELASTIC_APM_ACTIVATION_METHOD set and what does it do blah blah blah")

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
jackshirazi: This is actually already set in the command line for the webhook, I just left this fallback because I had this before we added the command line value setting (which is picked up as the first option in the switch above). I can drop this clause entirely or leave it, there is no need to make further changes to the webhook

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
```diff
@@ -0,0 +1,134 @@
...
+    AZURE_FUNCTIONS, //Not defined yet how this is identified
+    UNKNOWN;
+
+    private static final Logger logger = LogManager.getLogger(ActivationType.class);
+
+    public String toReferenceString() {
+        return toString().replace('_', '-').toLowerCase();
+    }
+
+    public static ActivationType findType() {
```
eyalkoren, line 48: See above - I think it should be called only once and not every time we serialize metadata

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
eyalkoren: OK, I never considered env variables as such.
It's just that this is just too sensitive to `/opt/elastic-apm-handler` exactly.
Why was it decided not to rely on `ELASTIC_APM_AWS_LAMBDA_HANDLER`?

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
eyalkoren: Ahh - [here](https://github.com/elastic/apm-mutating-webhook/blob/85b342dd0f08f4c6a82cec87479dc85c6ce0a08e/charts/apm-attacher/values.yaml#L32).
OK then.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
jackshirazi: Worst case overhead is 1 millisecond (in interpreted mode) and 2 additional objects each time the metadata is serialized (which I think is once every 5 seconds). Do we want to add a cache for that low an overhead? I'm asking rather than objecting

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
jackshirazi: I am going to go with your recommendation, but I do still think these are logically two different things - an input type, and a reported type - which are connected rather than the same thing. But I agree it is simpler if I use one enum

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
jackshirazi: I don't see that env var in the docs https://www.elastic.co/guide/en/apm/agent/java/current/aws-lambda.html ?

eyalkoren: The only one that is relevant to doc here is `AWS_LAMBDA_EXEC_WRAPPER`, which sets [elastic-apm-handler](https://github.com/elastic/apm-agent-java/blob/c8c8769d91efd2c153b96fb472cb601e2e8c0b7f/elastic-apm-agent/src/main/assembly/elastic-apm-handler) as the wrapper script that will be executed during the lambda runtime startup. This script does what we want it to do, for example - setting  configuration options with pre-known values. Those do not require documentation as they are set automatically with known values. I proposed to add your new config option as well:
```
export ELASTIC_APM_ACTIVATION_METHOD=AWS_LAMBDA_LAYER
```
Maybe that's why you thought it causes more installation burden and I didn't- this will not change the installation flow at all.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
eyalkoren: This is not about overhead, it is conceptually wrong. As far as I understand, this is just like any other metadata information that we discover once and then use for the rest of the agent lifespan. And as such, it should be just another discovery step that is done by our `MetaData`, where it also needs to reside.
I wouldn't call it caching, it is part of the immutable discoverable information we collectively call `MetaData`.
TBH, the most proper treatment for the metadata is not only discover it once, but also serializing it once and forget about it, I just didn't get to that as it requires some refactoring in our serialization.

jackshirazi: If the right place conceptually to cache the metadata *is* at the serialization, then this is the wrong place to add another cache, especially since in that case this would be called just once.

eyalkoren: But I am not proposing to move it to the serialization in this PR. As I said, it will require a lot of refactoring because of how it is done now - the thing is that the metadata collection process involves some actions that we wanted to do in an unblocking way, asynchronously to other processes that may want to start reporting data, which requires the metadata to be ready. If you dive into it, you will see what I mean - it is doable, but possibly too difficult and risky to be worth the tiny improvement.
What you did, however, defies the metadata concept altogether as you invoke a discovery of immutable information every time we create a new request. So instead of adding another piece of info to this immutable data "cache", you propose to introduce a new way of getting metadata information.

Anyway, to conclude- I think that the activation method would be a very natural field addition to `co.elastic.apm.agent.impl.metadata.Agent` and that the most natural way to populate it would be when this object is created. 
If you follow the code and related flows and think otherwise- feel free to place elsewhere.

jackshirazi: Here you are asserting that metadata never changes. If, as we've discussed we could do, we enable the agent to be disabled so that it can be upgraded in place, then feasibly the activation could start as `JAVAAGENT_FLAG` and then later become `APM_AGENT_ATTACH_CLI`. In this scenario caching the activation now has introduced a bug that only shows later when everyone has forgotten this is cached

eyalkoren: I am saying that metadata is immutable **_during the lifespan of an agent_**. If we ever add a hot-upgrade capability then this would be considered a new agent instance with newly created metadata. 
`co.elastic.apm.agent.impl.metadata.Agent#version` would change for sure and probably `co.elastic.apm.agent.impl.metadata.Agent#ephemeralId` and other metadata fields as well. And this is all irrelevant because our intention is to completely remove all agent classes, class loaders and threads and start over again with almost all new agent code, including new `MetaData` class with possibly different fields.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
jackshirazi: Had a go at this, but the class as is depends on core classes (like `ElasticApmTracer`) and trying to avoid the dependencies makes for a ridiculously convoluted implementation. And of course we can't have `common` depending on `core`

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
jackshirazi: cached now

apm-agent-core/src/test/resources/junit-platform.properties
jackshirazi: I did it here just being cautious. Maybe we should discuss in the team about this capability/option?

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
eyalkoren: Sorry, I was too haste to 👍 yesterday. I only meant adding a bare enum to the common module, without any dependency whatsoever. Any discovery logic can (and should) be part of all other metadata discovery.
Besides, looking at the `PrivilegedActionUtils` imports, it would perfectly fit the common module, though this is not required here if the discovery stays in core.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ActivationType.java
eyalkoren: Not the way I see it. Being consistent in what you do in each platform makes much more sense to me. Either way, it is not consistent in this sense already- in fleet you set a config option through the CLI (which makes the most sense and easy to find) and in k8s you set a system properties (as opposed to all others that are set as env variables). 

Speaking on k8s, I think this indeed needs to change. Why appending it to an environment variable if it is a config option that can be set as a separate value? 
I propose making it consistent with the NodeJS and .NET settings - see https://github.com/elastic/apm-mutating-webhook/pull/70

apm-agent-core/src/test/resources/junit-platform.properties
eyalkoren: Definitely. I would treat that in a separate PR though, where we can isolate related test execution changes.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/metadata/Agent.java
```diff
@@ -75,4 +85,56 @@ public String getVersion() {
     public String getEphemeralId() {
         return ephemeralId;
     }
+
+    public String getActivationMethod() {
```
eyalkoren, line 89: Please do the discovery during instance construction unless there is a good reason to do lazily. 
All metadata discovery is happening on construction and everything that needs metadata is blocking until it is ready. This introduces concurrency in discovery that we don't have otherwise.

jackshirazi: Full metadata discovery doesn't happen at instance construction. The agent is part of the service object which does discovery in a queue of various discoveries. At the time of the Agent object creation, the tracer is not yet initialized, but that is needed for the activation discovery. The lazy initialization is needed, and guaranteed to work as the metadata is first called when it is serialized which explicitly happens after discovery is complete

PR #2930, opened 2022-12-19
Improve JMS listener heuristic

SylvainJuge:
## What does this PR do?

Improves JMS `MessageListener` instrumentation matching to make it faster.
Adds an internal `jms_listener_packages` that allows strict filtering for even better matcher performance.

## Checklist


apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+        .stringsOption()
+        .key("jms_listener_packages")
+        .configurationCategory("JMS")
+        .tags("internal")
+        .description("Defines which packages contain JMS MessageListener implementations for instrumentation." +
+            "\n" +
+            "When set to a non-empty value, only the classes matching configuration will be instrumented.\n" +
+            "This configuration option helps to make MessageListener type matching faster and improve application startup performance.\n" +
+            "\n" +
+            WildcardMatcher.DOCUMENTATION
```
eyalkoren, line 40: Shouldn't be added here, as the config option uses `String` collection, right?

SylvainJuge: good catch, this was from an earlier version, will remove it.

apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+package co.elastic.apm.agent.jms;
+
+import org.stagemonitor.configuration.ConfigurationOption;
+import org.stagemonitor.configuration.ConfigurationOptionProvider;
+
+import java.util.Collection;
+import java.util.Collections;
+
+public class JmsConfiguration extends ConfigurationOptionProvider {
```
eyalkoren, line 27: We already have `co.elastic.apm.agent.configuration.MessagingConfiguration`, no need to add a new one.

apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsMessageListenerInstrumentation.java
```diff
@@ -46,6 +53,28 @@ public class JmsMessageListenerInstrumentation extends BaseJmsInstrumentation {
...
+            .or(nameContains("/")) // lambdas
+            .or(nameContains("Message"))
+            .or(nameContains("Listener"));
+
+        Collection<String> listenerPackages = configuration.getJmsListenerPackages();
+        if (listenerPackages.isEmpty()) {
+            return nameMatcher;
+        } else {
+            return nameMatcher.and(isInAnyPackage(listenerPackages, ElementMatchers.<NamedElement>none()));
+        }
```
eyalkoren, line 75: This can be a breaking change. We should document this very clearly and add to the CHANGELOG as such.

apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsMessageListenerInstrumentation.java
```diff
@@ -46,6 +53,28 @@ public class JmsMessageListenerInstrumentation extends BaseJmsInstrumentation {
...
     public static final Logger logger = LoggerFactory.getLogger(JmsMessageListenerInstrumentation.class);
 
+    @Nullable
+    private JmsConfiguration configuration;
+
+    @Override
+    public ElementMatcher<? super NamedElement> getTypeMatcherPreFilter() {
+        if (configuration == null) {
+            configuration = GlobalTracer.getTracerImpl().getConfig(JmsConfiguration.class);
+        }
```
eyalkoren, line 63: [minor] Instrumentations can declare a constructor that receives a `co.elastic.apm.agent.impl.ElasticApmTracer`, consider using that instead of such lazy initialization.

apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsConfiguration.java
SylvainJuge: :+1: on top of that, adding a dedicated section for JMS where there is only a single option is probably overkill. The fact that the option is prefixed with `jms_` makes it quite clear that this would be only relevant to JMS.

apm-agent-plugins/apm-jms-plugin/apm-jms-plugin-base/src/main/java/co/elastic/apm/agent/jms/JmsMessageListenerInstrumentation.java
SylvainJuge: :+1: in the case where it actually prevents instrumentation of a listener that was previously covered, do you think we should add now another internal option that would allow to completely bypass the naming check ? It might be a case where we'd better wait for feedback and know if it's actually breaking something before adding extra complexity.

eyalkoren: What I thought we can do is use EITHER the suffix prefilter OR the package-based prefilter. This way, you use these defaults and if someone defines packages, it overrides this, thus provides a workaround.
I think that if you provide a package where to look for a listener, it should be very quick and the other optimization is not necessarily required.
Just a thought.

SylvainJuge: That's a very good idea !
- by default (`jms_listener_packages` is empty) we optimize with the new heuristic (and that makes a breaking change)
- when a non-empty value is set to `jms_listener_packages`, then we just use this package name as part of the heuristic, it would be up to the end-user to select a value that does not make the agent scan all classes.

eyalkoren (approved):
Looks good!

Issue #2328, opened 2021-12-08
Building & testing project with JDK 18


SylvainJuge:
For Java/JVM versions compatibility, we have two distinct ways to define compatibility
1. Ability to use a given JDK to build and test the agent, 
2. Ability to use the packaged agent with an application, 

Ideally we should test both, but we often consider 1) to be a first step for 2).

The goal of this issue is to track all the known incompatibilities when trying to build with JDK 18, either discovered through CI or by running and build locally.

### Known issues
- [ ] Jacoco support for Java 18, should be included in version 0.8.8 https://github.com/jacoco/jacoco/pull/1198

eyalkoren:
Superseded by #2936 that will add weekly build and test with JDK 19

PR #2936, opened 2022-12-25
Fix instrumentation for WebFlux with Spring 6.x

### LINKED ISSUE #2935: Fix support for Spring boot 3.x
Is your feature request related to a problem? Please describe.
Support for the moment is limited to Spring Boot 2.x

Describe the solution you'd like
The agent should support Spring Boot 3.x

Describe alternatives you've considered
Use another APM.

Additional context
```
2022-12-08 09:49:37.205 WARN 385599 --- [or-tcp-epoll-13] transaction.id=5b0c1281268dd31d, trace.id=a074b71f72622f22bb82fe8ccae7e6da reactor.core.Exceptions : throwIfFatal detected a jvm fatal exception, which is thrown and logged below:

java.lang.NoSuchMethodError: 'org.springframework.http.HttpStatus org.springframework.http.server.reactive.ServerHttpResponse.getStatusCode()'
at co.elastic.apm.agent.springwebflux.WebfluxHelper.fillResponse(WebfluxHelper.java:272)
at co.elastic.apm.agent.springwebflux.WebfluxHelper.endTransaction(WebfluxHelper.java:171)
at co.elastic.apm.agent.springwebflux.TransactionAwareSubscriber.onComplete(TransactionAwareSubscriber.java:177)
```

eyalkoren:
## What does this PR do?
Fixes #2935 
I locally managed to successfully run the entire WebFlux instrumentation plugin tests with spring-web 6.0.3 and spring-boot 3.0.1 with only this small change in the instrumentation, so I think it may be enough.

Adding the tests require some more efforts and we should probably add a dedicated spring-boot 3.x test and possibly upgrade the spring-boot we test with in other modules, so if this fix is verified, tests will be included in a separate PR.

## Checklist
- [x] This is a bugfix
  - [x] I have added tests that would fail without this fix - only added unit tests to verify proper loading of the right implementation per Spring version. Full WebFlux and other Spring tests will be covered through #2942


eyalkoren:
And a few more minutes thinking about it - it is insufficient to change the source, the bytecode would still contain the wrong interface 🤦‍♂️ 
I'll add a separate module that compiles with spring-web 6 and only abstract out the ` int getStatusCode()` functionality

kuisathaverat (approved):

eyalkoren:
I used this fix to set up the stage to switch to the following state:
- [x] build and test on JDK 17 by default
- [x] enforce Java 11 syntax compatibility by default
- [x] build and test weekly on JDK 11
- [ ] build and test weekly on JDK 19 (requires upgrade of jacoco and closes #2328)
- [x] have an infrastructure that allows modules that must be compiled with Java 17

The latter was achieved through maven profile, as follows:
- Java-17-dependent code will reside under a top-level module called `apm-agent-java-17`
- The `apm-agent-java-17` module is not meant to contain code, only to serve as a parent to other modules
- One example for a module that requires compilation with Java 17 is `apm-agent-spring-6`, a child of `apm-agent-java-17`
- We use the `non-java-11` profile to exclude the `apm-agent-java-17` module entirely from build when not running on Java 11. **NOTE: this is a restriction as it assumes that the only pre-Java-17 we build with is 11**
- We use the `non-java-11` profile also to include the modules with actual code, such as `apm-agent-spring-6`, when packaging

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-plugin/src/main/java/co/elastic/apm/agent/springwebflux/WebfluxHelper.java
```diff
@@ -270,8 +269,12 @@ private static void fillRequest(Transaction transaction, ServerWebExchange excha
 
     private static void fillResponse(Transaction transaction, ServerWebExchange exchange) {
         ServerHttpResponse serverResponse = exchange.getResponse();
-        HttpStatus statusCode = serverResponse.getStatusCode();
-        int status = statusCode != null ? statusCode.value() : 200;
+        int status = 0;
+        try {
+            status = SpringWebVersionUtils.getStatusCode(serverResponse);
+        } catch (Exception e) {
+            // todo : log
```
jackshirazi, line 276: when is this log to be implemented?

eyalkoren: Should be now, thanks!
I will do that with other post-review changes.

jackshirazi (approved):

eyalkoren:
For now I treated this as a bug fix, postponing the documentation of it as supported to #2942

PR #2941, opened 2023-01-01
Windows tests

eyalkoren:


jackshirazi (approved):

jackshirazi:
I think the log file name needs to be different for each test for this to work

eyalkoren:
> I think the log file name needs to be different for each test for this to work

It could be the case, I just tried this one thing before disabling it only for Windows. As long as it works for other OSs, I am not sure it worth the effort.

jackshirazi (approved):

PR #2951, opened 2023-01-05
Add bootstrap check for disabling agent when required

eyalkoren:
## What does this PR do?
- allow to disable JVM bootstrap checks with `ELASTIC_APM_DISABLE_BOOTSTRAP_CHECKS` environment variable, this is currently only possible with `-Delastic.apm.disable_bootstrap_checks` JVM argument (using JVM system properties), hence allowing to disable bootstrap checks when JVM arguments modification is not possible.
- ensure that JVM tools detection issues an error and prevents the agent to start.
- provide an additional way to exclude some processes from instrumentation by relying on JVM system properties.

## Checklist

- [x] This is an enhancement of existing feature - bootstrap checks
  - [x] I tested the added functionalities manually
  - [x] I added documentation for bootstrap checks


elastic-apm-agent/src/main/java/co/elastic/apm/agent/premain/ExcludeJvmBootstrapCheck.java
```diff
@@ -0,0 +1,174 @@
...
+
+        AccessController.doPrivileged(new PrivilegedAction<Void>() {
+            @Override
+            public Void run() {
+                doExcludeListCheck(result, excludeSystemProperties);
+                return null;
+            }
+        });
+
+        doExcludeListCheck(result, excludeSystemProperties);
```
SylvainJuge, line 131: leftover ? we already execute it after checking if SM is present or not.

elastic-apm-agent/src/test/java/co/elastic/apm/agent/premain/ExcludeJvmBootstrapCheckTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import javax.annotation.Nullable;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ExcludeJvmBootstrapCheckTest {
+
+    public static final String ALLOWLIST_PROPERTY = "elastic.apm.bootstrap_allowlist";
```
SylvainJuge, line 11: [very minor] what about moving those constants to `BootstrapCheck` class directly and remove some duplication ? at least that allows to easily see when things are related to each other

SylvainJuge (approved):

elastic-apm-agent/src/main/java/co/elastic/apm/agent/premain/ExcludeJvmBootstrapCheck.java
eyalkoren: Definitely looks like a leftover, thanks!

elastic-apm-agent/src/test/java/co/elastic/apm/agent/premain/ExcludeJvmBootstrapCheckTest.java
eyalkoren: Much better, will do 👍

docs/setup.asciidoc
```diff
@@ -65,6 +65,47 @@ expansion] capability for specifying the agent jar path.
...
+The allowlist option expects a comma-separated list of wild-card patterns. Such patterns may contain wildcards (`\*`), which match zero
+or more characters. Examples: `foo\*bar\*baz\*, \*foo\*`. Matching is case-insensitive by default. Prepending an element with
+`(?-i)}` makes the matching case-sensitive.
+
+Some examples:
+
+1. Allow JVM attachment only on Tomcat and a proprietary Java app: +
+   `-Delastic.apm.bootstrap_allowlist=\*org.apache.catalina.startup.Bootstrap*,my.cool.app.*`
+2. Disable when some custom System properties are set: +
+   `-Delastic.apm.bootstrap_exclude_list=custom.property.1,custom.property.2`
```
SylvainJuge, line 106: [question] Does the patterns are matched against the whole JVM command line or only the main class and the system properties ?

eyalkoren: There is an explanation above for the allow list and in the table.
Is it not clear from those what is used in each?
As for the allowlist - it is only the main class or jar and JVM parameters, basically the contents of the `sun.java.command` system property. I will add this info. Let me know if anything else is required

eyalkoren:
@SylvainJuge I removed the default exclude list, as it doesn't seem required at this point.
Please only review that and the added documentation. Thanks!

elastic-apm-agent/src/test/java/co/elastic/apm/agent/premain/ExcludeJvmBootstrapCheckTest.java
```diff
@@ -78,22 +66,29 @@ void testEnvVariableNames() {
         assertThat(EXCLUDE_LIST_ENV_VARIABLE).isEqualTo(EXCLUDE_LIST_SYSTEM_PROPERTY.replace('.', '_').toUpperCase());
     }
 
-    private BootstrapCheck.BootstrapCheckResult runCheckWithConfigurationProperty(@Nullable String configPropKey, @Nullable String configPropValue,
-                                                                                  @Nullable String systemProp, String cmd) {
+    private BootstrapCheck.BootstrapCheckResult runCheckWithConfigurationProperty(@Nullable String allowlistValue,
+                                                                                  @Nullable String excludeListValue,
```
SylvainJuge, line 70: [question] why do we still have an exclude list if the option was removed ?

SylvainJuge (approved):

elastic-apm-agent/src/test/java/co/elastic/apm/agent/premain/ExcludeJvmBootstrapCheckTest.java
eyalkoren: The configurable option was not removed, only the default list of exclusions

PR #2958, opened 2023-01-11
don't capture exceptions to workaround AWS coretto crashes

jackshirazi:
## Checklist

- [x] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else


jackshirazi:
[Discussion thread](https://discuss.elastic.co/t/sigsegv-jvm-crash/322812/8) that gives rise to this issue.

What we think may be happening here is that Amazon Coretto specializes how throwables are handled. Possibly, when running on an AWS server, the throwable is held in native space (maybe to give XRay access to it). The throwable is de-reffed and GCed when the application doesn't see it - from the point of view of the [AWS interceptors](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/core/interceptor/ExecutionInterceptor.html). But the agent instrumentation - which the interceptor doesn't see - uses the throwable data for reporting span exceptions. So there is a race condition where a throwable in native space may have been removed, and the instrumentation tries to access it, and this causes a SIGSEGV (instead of an NPE). It seems like most of the time the throwable hasn't been removed yet when the instrumentation accesses it, but it just needs to have been one time for the SIGSEGV crash to happen - which it does eventually if the application runs for long enough

apm-agent-plugins/apm-aws-sdk/apm-aws-sdk-2-plugin/src/main/java/co/elastic/apm/agent/awssdk/v2/BaseSyncClientHandlerInstrumentation.java
```diff
@@ -47,6 +47,11 @@
 import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 
 public class BaseSyncClientHandlerInstrumentation extends TracerAwareInstrumentation {
+    //Coretto causes sigsegv crashes when you try to access a throwable if it thinks
+    //it went out of scope, which it seems to for the instrumented throwable access
+    private static final boolean IS_CORRETTO = System.getProperty("java.vendor.version") == null ?
+        false :
+        System.getProperty("java.vendor.version").contains("Corretto");
```
eyalkoren, line 54: If eventually we end up doing such check, it should go into `co.elastic.apm.agent.common.JvmRuntimeInfo`

apm-agent-plugins/apm-aws-sdk/apm-aws-sdk-2-plugin/src/main/java/co/elastic/apm/agent/awssdk/v2/BaseSyncClientHandlerInstrumentation.java
```diff
@@ -47,6 +48,9 @@
 import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 
 public class BaseSyncClientHandlerInstrumentation extends TracerAwareInstrumentation {
+    //Coretto causes sigsegv crashes when you try to access a throwable if it thinks
+    //it went out of scope, which it seems to for the instrumented throwable access
+    private static Throwable REDACTED = new Throwable("Unable to provide details of the error");
```
eyalkoren, line 53: We are going to send an error with this throwable, which means that the stack trace will be visible. We need to take care of that somehow. Two options I thought of:
1. use a somewhat-useful stack trace: create this throwable lazily in the right spot (meaning- when you use it for the first time) so that it captures a more useful stack trace. Clone the stack trace without its top part that contains our code (for example by traversing it until you reach the AWS SDK package) and use it instead. It doesn't have to be synchronized because we don't mind doing it more than once.
2. don't use a stack trace: use the `Throwable` constructor that doesn't fill the stack trace.

jackshirazi: We can't do 1 cos that's exactly what can trigger the crash. Although it seems like the first time has low risk, but since this is to avoid the crash, better to skip that. So 2 is the option. Good call, thanks

eyalkoren: > We can't do 1 cos that's exactly what can trigger the crash

If any exception is unsafe to handle, then this fix is insufficient. We still treat all other exceptions everywhere the same.

jackshirazi: Yup, I did misunderstand. I think it's just about worth it. Implementation added, please review

eyalkoren: Actually, the way you implemented it made me think there is small further improvement available within reach - if you add a `@Advice.This thiz` argument to `exitDoExecute()`, then `thiz.getClass().getName()` would give you the **actual** class name, instead of using `AWS_CLASS_NAME` always, and then your `RedactedException` would be a multiton with an instance per instrumented class.

jackshirazi: another good call, thanks! added now

eyalkoren (approved):

PR #2968, opened 2023-01-16
Added OpenTelemetry Metrics SDK instrumentation.

JonasKunz:
## What does this PR do?

First of two PRs for adding OpenTelemetry metrics support.
This PR adds support for exporting metrics from user-created instances of the Otel Metrics SDK.
The second PR will add support for capturing metrics from users who only use the OTel API.

This PR is based on the current state of the unmerged spec elastic/apm#742.
The feature is marked as experimental, therefore we can add changes resulting from changes in the spec later on.

Documentation of the entire OTel Metrics feature will be added with the second PR.

## Checklist


apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-metrics-exporter/pom.xml
```diff
@@ -0,0 +1,29 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>apm-opentelemetry</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.35.1-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>apm-opentelemetry-metrics-exporter</artifactId>
```
JonasKunz, line 12: [For Reviewer]
Why is the exporter in a separate project? This is required for the API-bridge that follows in the next PR: There we will shade & embedded an Otel metrics SDK. In order to install our exporter there, we will need to add a shaded copy of the exporter-project as well.

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-metrics-exporter/src/main/java/co/elastic/apm/agent/otelmetricexport/MetricSetGenerator.java
```diff
@@ -0,0 +1,288 @@
...
+
+    private static final int INITIAL_BUFFER_SIZE = 2048;
+
+    private static final DslJson<Object> DSL_JSON = new DslJson<>(new DslJson.Settings<>());
+
+    private final StringBuilder replaceBuilder;
+    private final JsonWriter jw;
+    private boolean anySamplesWritten;
+
+    public MetricSetGenerator(Attributes attributes, CharSequence instrumentationScopeName, long epochMicros, StringBuilder replaceBuilder) {
```
JonasKunz, line 57: [For Reviewer]:
We now have metricset serialization in three places: The agent metrics, Micrometer and Otel. The `MetricSetGenerator` class is designed to allow a refactoring in the future to have the serialization logic in one place again.
The idea is to extract a common superclass which delegates just the `tags` generation to the subclasses.

If you agree that this is a good Idea, I'll create a tech-debt issue for this after this PR is merged.

apm-agent-core/src/test/java/co/elastic/apm/agent/testutils/assertions/metrics/MetricsetJson.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+import javax.annotation.Nullable;
+import java.util.HashMap;
+import java.util.Map;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class MetricsetJson {
```
SylvainJuge, line 29: Good idea to use jackson object mapping to make tests easier, however I think that having more "explicit" mapping of the attributes is more robust to object attributes renaming (when it happens it's a pain to debug). Alternatively, can we make jackson throw an error when a field in JSON is not mapped to this object ?

apm-agent-core/src/test/java/co/elastic/apm/agent/testutils/assertions/metrics/MetricsetEventJson.java
```diff
@@ -0,0 +1,24 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package co.elastic.apm.agent.testutils.assertions.metrics;
+
+public class MetricsetEventJson {
+
+    public MetricsetJson metricset;
```
SylvainJuge, line 23: can we use explicit mapping here (or make jackson throw an error when there is a field not mapped) ?

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-metrics-exporter/src/main/java/co/elastic/apm/agent/otelmetricexport/MetricSetGenerator.java
SylvainJuge: I think that's a very good idea, the duplication with micrometer metrics was somehow acceptable, but now it's clear we really need to prioritize this. As part of this refactor, we should include properly testing this class independently from the otel/micrometer/internal implementations as it is currently only tested indirectly.

apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/MetricsConfiguration.java
```diff
@@ -37,6 +42,20 @@ public class MetricsConfiguration extends ConfigurationOptionProvider {
...
+        .description("Defines the default bucket boundaries to use for OpenTelemetry histograms.")
+        .dynamic(true)
+        .tags("added[1.36.0]", "experimental")
+        .buildWithDefault(Arrays.asList(
+            0.00390625, 0.00552427, 0.0078125, 0.0110485, 0.015625, 0.0220971, 0.03125, 0.0441942,
+            0.0625, 0.0883883, 0.125, 0.176777, 0.25, 0.353553, 0.5, 0.707107, 1.0, 1.41421, 2.0,
+            2.82843, 4.0, 5.65685, 8.0, 11.3137, 16.0, 22.6274, 32.0, 45.2548, 64.0, 90.5097, 128.0,
+            181.019, 256.0, 362.039, 512.0, 724.077, 1024.0, 1448.15, 2048.0, 2896.31, 4096.0, 5792.62,
+            8192.0, 11585.2, 16384.0, 23170.5, 32768.0, 46341.0, 65536.0, 92681.9, 131072.0
+        ));
```
SylvainJuge, line 57: [minor] As those values are histogram buckets and it's a bit verbose, it could be relevant to assert that values are at least properly sorted and fallback on default if that's not the case (it could also be something worth adding to spec).

apm-agent-core/src/test/java/co/elastic/apm/agent/testutils/assertions/metrics/MetricsetJson.java
JonasKunz: >Alternatively, can we make jackson throw an error when a field in JSON is not mapped to this object

That is actually the default behaviour of jackson. For this reason I added the `@JsonIgnoreProperties(ignoreUnknown = true)` here, but you are right, we should be more explicit: I will replace `ignoreUnknown=true` with `value` with an explicit list of allowed attribute names taken from the metricset spec.
This way, we will get an error if any unknown attribute is present.

apm-agent-core/src/test/java/co/elastic/apm/agent/testutils/assertions/metrics/MetricsetEventJson.java
JonasKunz: See https://github.com/elastic/apm-agent-java/pull/2968#discussion_r1087746370: This will throw an error if any unmapped fields are found by default.

apm-agent-core/src/test/java/co/elastic/apm/agent/testutils/assertions/metrics/MetricsetsAssert.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+public class MetricsetsAssert extends BaseAssert<MetricsetsAssert, List<MetricsetJson>> {
```
SylvainJuge, line 35: [minor] rename to `MetricSetAssert` for consistency: others similar assertions are in singular form + case.

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-metricsdk-plugin/src/test/java/co/elastic/apm/agent/otelmetricsdk/AbstractOtelMetricsTest.java
```diff
@@ -0,0 +1,624 @@
...
+
+        long timestampMicros = System.currentTimeMillis() * 1000L;
+        counter.add(42);
+
+        resetReporterAndFlushMetrics();
+        assertThatMetricSets(reporter.getBytes())
+            .hasMetricsetCount(1)
+            .first()
+            //check for a range of +- 1 second to be safe against CI slowness / clock inaccuracy
+            .hasTimestampInRange(timestampMicros - 1_000_000L, timestampMicros + 1_000_000L);
```
SylvainJuge, line 171: [minor] why not use `Instant.now()` before and after the metric collection and use that as narrower interval boundaries ?

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-metricsdk-plugin/src/test/java/co/elastic/apm/agent/otelmetricsdk/AbstractOtelMetricsTest.java
```diff
@@ -0,0 +1,624 @@
...
+            .gaugeBuilder("my_other_gauge").ofLongs()
+            .buildWithCallback((obs) -> {
+                obs.record(43);
+                obs.record(143, Attributes.of(stringKey("foo"), "bar"));
+            });
+
+        resetReporterAndFlushMetrics();
+        assertThatMetricSets(reporter.getBytes())
+            .hasMetricsetCount(2)
+            .hasMetricsetWithLabelsSatisfying(
```
SylvainJuge, line 193: [minor] maybe for assertion readability I would split this in two distinct parts: `hasLabels(Map.of(....))` and `satisfies(metrics->metrics. [...])`.

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-metricsdk-plugin/src/test/java/co/elastic/apm/agent/otelmetricsdk/AbstractOtelMetricsTest.java
```diff
@@ -0,0 +1,624 @@
...
+        assertThatMetricSets(reporter.getBytes())
+            .hasMetricsetCount(0);
+    }
+
+
+    /**
+     * Smoke test which just checks that nothing breaks with the default histogram buckets.
+     */
+    @Test
+    public void testDefaultHistogramBuckets() {
```
SylvainJuge, line 592: :star_struck: That's a very good check to ensure the values we have in the spec make sense, I wonder if we could also do the same to estimate the relative accuracy of the histogram.

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-metricsdk-plugin/src/main/java/co/elastic/apm/agent/otelmetricsdk/SdkMeterProviderBuilderInstrumentation.java
```diff
@@ -0,0 +1,96 @@
...
+        @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
+        public static void onEnter(@Advice.This SdkMeterProviderBuilder thiz) {
+            if (checkMetricsSdkVersionSupported()) {
+                if (ALREADY_REGISTERED_BUILDERS.add(thiz)) {
+                    ElasticOtelMetricsExporter.createAndRegisterOn(thiz, tracer);
+                }
+            }
+        }
+
+        private static boolean checkMetricsSdkVersionSupported() {
```
SylvainJuge, line 81: This will very likely spam the logs with warnings, thus ensuring it is only triggered once would be better.

apm-agent-core/src/test/java/co/elastic/apm/agent/testutils/assertions/metrics/MetricsetsAssert.java
JonasKunz: I fixed the casing. Regarding plural/singular: We have both `MetricSetAssert` and `MetricSetsAssert`: The first one has verifications for a single metricset, the second one for collections of metricsets.

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-metricsdk-plugin/src/test/java/co/elastic/apm/agent/otelmetricsdk/AbstractOtelMetricsTest.java
JonasKunz: I don't think this would give any benefit here, as the upper bound for the relative error is defined by how the boundaries are defined in the spec: I used a nexponential histogram with base `sqrt(2)=~1.41`, which means we have a worst case relative accuracy of 41% for values which lie within the maximum / minimum of the boundaries. This is still not great, in the future we should switch to either T-Digests or HDRHistogram...

apm-agent-plugins/apm-opentelemetry/apm-opentelemetry-metricsdk-plugin/src/test/java/co/elastic/apm/agent/otelmetricsdk/AbstractOtelMetricsTest.java
JonasKunz: Unfortunately it is not possible to switch this easily into two operations:
`reporter.getBytes()` returns all reported metricsets, which should be two in this case.
We could do a `hasMetricsetSatisfying(metricset -> metricset.hasLabels(...).hasMetricValue(..)`, but I decided against it: What do you print if no metricset satisfies the provided condition? With my approach, we can distinguish based on the error message wheterh (a) no metricset with the given labels exist and (b) the metricset with the labels has other bad contents.

SylvainJuge (approved):

PR #2977, opened 2023-01-19
Readded ldap plugin, added instrumentation module opening support

### LINKED ISSUE #2966: Instrumentation of LDAP requests

JonasKunz:
## What does this PR do?
Adds back the LDAP plugin which was removed via #2431 .
The open issues were solved by adding the capabilities for instrumentation plugins to open modules via instrumentation.

Closes #2966.
## Checklist

- [x] This is a new plugin
  - [x] My code follows the [style guidelines of this project](https://github.com/elastic/apm-agent-java/blob/main/CONTRIBUTING.md#java-language-formatting-guidelines)
  - [x] New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/main/CONTRIBUTING.md#testing) pass locally with my changes
  - [x] I have updated [supported-technologies.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/docs/supported-technologies.asciidoc)
  - [ ] ~Added an instrumentation plugin? Describe how you made sure that old, non-supported versions are not instrumented by accident.~ Ldap implementation is ancient


jamesdowell2022:
Forwarded to the customer asking to test and feedback. Thanks so much for the incredible turnaround time!!

jamesdowell2022:
Feedback from Barclays They will update the case too with screen shots but excellent work team:
 
Good news – the snapshot agent is working as we wanted, we can see the LDAP backends as dependencies now, and on the service map.
I will update the case now with screenshots etc.
Thanks for helping to turn this around so quickly.

JonasKunz:
@eyalkoren This PR should be ready to review now, we got positive feedback from a customer who tested this PR snapshot.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
```diff
@@ -428,13 +432,19 @@ private static ConstantCallSite internalBootstrap(MethodHandles.Lookup lookup, S
...
+                // can't use MethodHandle.lookup(), see also https://github.com/elastic/apm-agent-java/issues/1450
+                MethodHandles.Lookup indyLookup = (MethodHandles.Lookup) lookupExposer.getMethod("getLookup").invoke(null);
+                // When calling findStatic now, the lookup class will be one that is loaded by the plugin class loader
+                MethodHandle methodHandle = indyLookup.findStatic(adviceInPluginCL, adviceMethodName, adviceMethodType);
+                return new ConstantCallSite(methodHandle);
+            } else {
+                // must not be a static field as it would initialize logging before it's ready
+                LoggerFactory.getLogger(IndyBootstrap.class).error("Cannot bootstrap advice because required modules could not be opened!");
+                return null;
+            }
```
eyalkoren, line 447: This really confused me- it looks as if the lookup handling is dependent on the module opening.
Since you already `return null`, I think it will be a clearer from this way:
```
if (!addRequiredModuleOpens(requiredModuleOpens, targetClassLoader, pluginClassLoader)) {
  // must not be a static field as it would initialize logging before it's ready
  LoggerFactory.getLogger(IndyBootstrap.class).error("Cannot bootstrap advice because required modules could not be opened!");
  return null;
}
// do the lookup exposer thing and get a method handle         
```

Furthermore, it will be even more readable if `addRequiredModuleOpens` is called when modules are supported and `!requiredModuleOpens.isEmpty()`. Then the `boolean` result returned from it really represents success or failure, which is clearer why we log an error and abort

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
```diff
@@ -444,11 +454,28 @@ private static ConstantCallSite internalBootstrap(MethodHandles.Lookup lookup, S
         }
     }
 
-    private static List<String> getClassNamesFromBundledPlugin(String adviceClassName, ClassLoader adviceClassLoader) throws IOException, URISyntaxException {
-        if (!adviceClassName.startsWith(EMBEDDED_PLUGINS_PACKAGE_PREFIX)) {
-            throw new IllegalArgumentException("invalid advice class location : " + adviceClassName);
+    private static boolean addRequiredModuleOpens(Map<String, List<String>> requiredModuleOpens, ClassLoader targetClassLoader, ClassLoader pluginClassLoader) {
```
eyalkoren, line 457: javadoc required, especially what the return value means.
I assume you'll find it difficult to explain, which indicates it is not very clear success/failure. I propose to either change what the method does, so that the return value represents success vs. error, or not returning a value and relying on an exception instead.

Sorry for nitpicking, this code is already complex as it is....

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
```diff
@@ -462,4 +489,12 @@ private static List<String> getClassNamesFromBundledPlugin(String adviceClassNam
         return pluginClasses;
     }
 
+    private static String getPluginPackageFromAdviceClass(String adviceClassName) {
+        if (!adviceClassName.startsWith(EMBEDDED_PLUGINS_PACKAGE_PREFIX)) {
+            throw new IllegalArgumentException("invalid advice class location : " + adviceClassName);
+        }
+        String pluginPackage = adviceClassName.substring(0, adviceClassName.indexOf('.', EMBEDDED_PLUGINS_PACKAGE_PREFIX.length()));
+        return pluginPackage;
+    }
```
eyalkoren, line 498: Why does this duplicate [what we have in the customizer](https://github.com/elastic/apm-agent-java/blob/9575d839280b71e6ef771be45e5436d241b68c04/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/PluginClassLoaderRootPackageCustomizer.java#L35-L41)? Can't we reuse it?

apm-agent-core/src/test/java/co/elastic/apm/agent/bci/modules/ModuleOpenerTest.java
```diff
@@ -0,0 +1,106 @@
...
+    public static class IllegalAccessor {
+
+        public static void doAccess() throws Exception {
+            //com.sun.jndi.ldap.LdapResult is not accessible according to module rules
+            Class.forName("com.sun.jndi.ldap.LdapResult").getConstructor().newInstance();
+        }
+
+    }
+
+    private static class ChildFirstCopyClassloader extends ClassLoader {
```
eyalkoren, line 67: Would [ChildFirstURLClassLoader](https://github.com/elastic/apm-agent-java/blob/9575d839280b71e6ef771be45e5436d241b68c04/apm-agent-core/src/test/java/co/elastic/test/ChildFirstURLClassLoader.java#L39) be useful here?

eyalkoren (approved):
Absolutely brilliant! 👋 
Including code, comments and test.
I added some comments, mostly about readability or reuse of code, but all minor enough to approve regardless. Feel free to decide which to use and which to ignore,
Well done!

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
JonasKunz: Well this code was simply extracted from the existing local `getClassNamesFromBundledPlugin` method. What would be the correct place for this logic in your opinion? A static method within `PluginClassLoaderRootPackageCustomizer`? Or somewhere else?

apm-agent-core/src/test/java/co/elastic/apm/agent/bci/modules/ModuleOpenerTest.java
JonasKunz: I actually looked at it and the `TestClassWithDependenciesRunner`: I decided against it because it requires extracting the bytecode of the class into a temporary jar.
Instead I used this implementation because:
 * It does not involve disk access coding
 * The code is not longer and complexer than reusing the `ChildFirstURLClassLoader` because it required packaging the target class in a temp jar first

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
eyalkoren: Yes, this definitely seems suitable for a static method and `PluginClassLoaderRootPackageCustomizer` seems like a reasonable common place to compute the value that is both its base field AND is used as its mapping key. 
It will probably allow to remove `PluginClassLoaderRootPackageCustomizer` from `IndyBootstrap` and have only one such as well.

PR #2979, opened 2023-01-19
Fix SQS instrumentation within messaging transactions

AlexanderWert:
Fixes a bug in the SQS instrumentation. Adds tests for that scenario.

## Buggy behavior

When an SQS span was created **within** a `messaging` type transaction (as parent), the transaction was ended immediately independent of the SQS action type.

While this behavior is expected as the default behavior of the config option `end_messaging_transaction_on_poll` for actions that POLL a message (i.e. `ReceiveMessage` request), this should not happen for all types of SQS requests (e.g. SendMessage, etc.).

This PR limits the above behavior to `ReceiveMessage` requests.


jackshirazi:
Test with multiple messages?

jackshirazi (approved):

PR #2918, opened 2022-12-07
Fixed AgentOverheadMetricsTest flakyness

JonasKunz:
## What does this PR do?
Attempts to fix the flakyness #2903.
 I reran the build of this PR on CI several times, so far no more failures.

Issue #2903, opened 2022-12-05
[Flaky test] `AgentOverheadMetricsTest` 


eyalkoren:
## Test methods
[co.elastic.apm.agent.metrics.builtin.AgentOverheadMetricsTest.checkCpuMetrics](https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/job/main/479/testReport/junit/co.elastic.apm.agent.metrics.builtin/AgentOverheadMetricsTest/checkCpuMetrics/)
[co.elastic.apm.agent.metrics.builtin.AgentOverheadMetricsTest.checkAllocationMetric](https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/job/main/479/testReport/junit/co.elastic.apm.agent.metrics.builtin/AgentOverheadMetricsTest/checkAllocationMetric/)

## Test logs

<details>
<summary>checkCpuMetrics</summary>

```
org.assertj.core.error.AssertJMultipleFailuresError: 

Multiple Failures (1 failure)
-- failure 1 --
Expecting actual:
  0.14638835
to be greater than:
  0.14638835
at AgentOverheadMetricsTest.lambda$checkCpuMetrics$3(AgentOverheadMetricsTest.java:120)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at co.elastic.apm.agent.metrics.builtin.AgentOverheadMetricsTest.lambda$checkCpuMetrics$7(AgentOverheadMetricsTest.java:118)
	at co.elastic.apm.agent.metrics.builtin.AgentOverheadMetricsTest.lambda$reportAndCheckMetrics$35(AgentOverheadMetricsTest.java:392)
	at co.elastic.apm.agent.metrics.MetricRegistry.flipPhaseAndReport(MetricRegistry.java:201)
	at co.elastic.apm.agent.metrics.builtin.AgentOverheadMetricsTest.reportAndCheckMetrics(AgentOverheadMetricsTest.java:391)
[...]

PR #2984, opened 2023-01-25
Disable CPU overhead test due to flakyness

JonasKunz:
Sad PR :( looks like #2918 didn't fix everything, the CPU-metrics test is still flaky in the CI for unclear reasons (See https://github.com/elastic/apm-agent-java/issues/2903#issuecomment-1402060114)


eyalkoren (approved):
Sure, let's disable first.
We may decide we are OK with not system-testing this functionality, but then the disabling comment should say this is kept only for running manual tests or something of that sort.
Otherwise, this PR needs to be accompanied by an issue that proposes what we do next:
- Attempt to fix this test
- Rewrite this test and either keep this one for manual testing or not
- Something else?

JonasKunz:
There is already an issue for it: #2903 , I just forgot to close it with the last PR where I attempted to fix the flakyness. I also added the logs of the most recent failure there. Maybe it would make sense to revive #2554 for such tests?

PR #2993, opened 2023-01-30
Broadened catch clause when extracting es query body

JonasKunz:
## What does this PR do?
A user reported that our instrumentation of the elasticsearch client triggers an error when using hiberantesearch:
```
java.lang.UnsupportedOperationException: Not implemented! Expected to produce content only over produceContent(), or writeTo(OutputStream) if blocking calls are acceptable for your use case.
at org.hibernate.search.elasticsearch.util.impl.GsonHttpEntity.getContent(GsonHttpEntity.java:164)
at co.elastic.apm.agent.esrestclient.ElasticsearchRestClientInstrumentationHelper.createClientSpan(ElasticsearchRestClientInstrumentationHelper.java:105)
at co.elastic.apm.agent.esrestclient.v6_4.ElasticsearchClientAsyncInstrumentation$ElasticsearchRestClientAsyncAdvice.onBeforeExecute(ElasticsearchClientAsyncInstrumentation.java:68)
```
This error is not caught by our current instrumentation. With this PR, we simply make our code more defensive to only fail the recording of the query and not the entire adivce.

I would suggest not implementing any special case handling for hibernatesearch, as the issue has been fixed there since version 6:
https://github.com/hibernate/hibernate-search/commit/8d603b112f8ce634e5c06ca72a96acc1653efffe

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/esrestclient/ElasticsearchRestClientInstrumentationHelper.java
+++ b/apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/esrestclient/ElasticsearchRestClientInstrumentationHelper.java
@@ -18,17 +18,18 @@
  */
 package co.elastic.apm.agent.esrestclient;
 
+import co.elastic.apm.agent.common.util.WildcardMatcher;
 import co.elastic.apm.agent.impl.ElasticApmTracer;
 import co.elastic.apm.agent.impl.GlobalTracer;
 import co.elastic.apm.agent.impl.transaction.AbstractSpan;
 import co.elastic.apm.agent.impl.transaction.Outcome;
 import co.elastic.apm.agent.impl.transaction.Span;
-import co.elastic.apm.agent.common.util.WildcardMatcher;
 import co.elastic.apm.agent.objectpool.Allocator;
 import co.elastic.apm.agent.objectpool.ObjectPool;
 import co.elastic.apm.agent.sdk.logging.Logger;
 import co.elastic.apm.agent.sdk.logging.LoggerFactory;
 import co.elastic.apm.agent.util.IOUtils;
+import co.elastic.apm.agent.util.LoggerUtils;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHost;
 import org.elasticsearch.client.Response;
@@ -36,7 +37,6 @@ import org.elasticsearch.client.ResponseException;
 import org.elasticsearch.client.ResponseListener;
 
 import javax.annotation.Nullable;
-import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.CancellationException;
@@ -44,6 +44,8 @@ import java.util.concurrent.CancellationException;
 public class ElasticsearchRestClientInstrumentationHelper {
 
     private static final Logger logger = LoggerFactory.getLogger(ElasticsearchRestClientInstrumentationHelper.class);
+
+    private static final Logger queryExtractionErrorLogger = LoggerUtils.logOnce(logger);
     private static final ElasticsearchRestClientInstrumentationHelper INSTANCE = new ElasticsearchRestClientInstrumentationHelper(GlobalTracer.requireTracerImpl());
 
     public static final List<WildcardMatcher> QUERY_WILDCARD_MATCHERS = Arrays.asList(
@@ -104,8 +106,8 @@ public class ElasticsearchRestClientInstrumentationHelper {
                 if (httpEntity != null && httpEntity.isRepeatable()) {
                     try {
                         IOUtils.readUtf8Stream(httpEntity.getContent(), span.getContext().getDb().withStatementBuffer());
-                    } catch (IOException e) {
-                        logger.error("Failed to read Elasticsearch client query from request body", e);
+                    } catch (Exception e) {
+                        queryExtractionErrorLogger.error("Failed to read Elasticsearch client query from request body", e);
                     }
                 }
             }


apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/esrestclient/ElasticsearchRestClientInstrumentationHelper.java
```diff
@@ -104,8 +106,8 @@ public Span createClientSpan(String method, String endpoint, @Nullable HttpEntit
                 if (httpEntity != null && httpEntity.isRepeatable()) {
                     try {
                         IOUtils.readUtf8Stream(httpEntity.getContent(), span.getContext().getDb().withStatementBuffer());
-                    } catch (IOException e) {
-                        logger.error("Failed to read Elasticsearch client query from request body", e);
+                    } catch (Exception e) {
+                        queryExtractionErrorLogger.error("Failed to read Elasticsearch client query from request body", e);
```
eyalkoren, line 110: [minor] Keep the former exception handling and apply this log-once handling only for `UnsupportedOperationException`, where the message can say that this may be a know hibernatesearch issue.

eyalkoren (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
```diff
@@ -441,6 +451,7 @@ private static ConstantCallSite internalBootstrap(MethodHandles.Lookup lookup, S
             MethodHandles.Lookup indyLookup = (MethodHandles.Lookup) lookupExposer.getMethod("getLookup").invoke(null);
             // When calling findStatic now, the lookup class will be one that is loaded by the plugin class loader
             MethodHandle methodHandle = indyLookup.findStatic(adviceInPluginCL, adviceMethodName, adviceMethodType);
+            methodHandle = wrapWithExceptionCatcher(methodHandle);
```
JonasKunz, line 454: [For Reviewer]
In the related user issue I noticed that uncaught exceptions from advices are ONLY logged to stdout.
I don't know if this bit us already at the past, but I feared it might in the future.
When the exception is only logged to stdout it (a) can confuse users and (b) means that the agent logs don't show us the full story. We would most likely see messages about orphaned spans with no idea where they come from without also having the stdout of the application.

I therefore thought it might be a good idea to improve this behaviour. With this change:
 * Exceptions are caught (they were caught before aswell) and logged to the agent logs
 * Exceptions are not logged to stdout anymore
 
 Essentially this change is equivalent to adding a `try {} catch(Throwable t){ log.error(t);}` around all advices. The performance might be slightly worse on Java 7 (additional method call), on Java 8+ it should be the same as directly inling the try-catch ([source](https://www.youtube.com/watch?v=rhBfvWQRe0w) making me think this).
 
 If you feel like this change is to risky performance wise, don't hesitate to speak up. I'm fine with removing this from this PR.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -511,10 +512,11 @@ public boolean matches(MethodDescription target) {
                 }
             }
         };
+        StackManipulation exceptionHandler = MethodInvocation.invoke(new MethodDescription.ForLoadedMethod(IndyBootstrap.getExceptionHandlerMethod(logger)));
         return new AgentBuilder.Transformer.ForAdvice(withCustomMapping)
             .advice(instrumentationStats.shouldMeasureMatching() ? statsCollectingMatcher : matcher, instrumentation.getAdviceClassName())
             .include(ClassLoader.getSystemClassLoader(), PrivilegedActionUtils.getClassLoader(instrumentation.getClass()))
-            .withExceptionHandler(PRINTING);
+            .withExceptionHandler(new Advice.ExceptionHandler.Simple(exceptionHandler));
```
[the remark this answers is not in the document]
felixbarny: Awesome! We were scratching our heads how to get advice exceptions into the agent logs without manually writing byte code for the stack manipulation. 👏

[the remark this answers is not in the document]
JonasKunz: Ah that's nice to hear! I initially tried a different approach (wrapping the `MethodHandles`, see https://github.com/elastic/apm-agent-java/pull/2993/commits/326db67e549b6c56921d1e6a729461d3bbcdfc27#diff-2ec63f545539ff7d65f52ef01087a18c7cdd9eb7130de156fb412ae6fe9e580eR465), but then noticed that some Advices might intentionally throw exceptions, at least according to the tests.

apm-agent-bootstrap/src/main/java/bootstrap/dispatcher/IndyBootstrapDispatcher.java
```diff
@@ -82,6 +83,18 @@ public static CallSite bootstrap(MethodHandles.Lookup lookup,
...
 
+    public static void logAdviceException(Throwable exception) {
+        try {
+            if (logAdviceException != null) {
+                logAdviceException.invoke(null, exception);
+            } else {
+                exception.printStackTrace();
+            }
+        } catch (Throwable t) {
+            t.printStackTrace();
```
eyalkoren, line 94: That's great, I missed the `Throwable#printStackTrace` usage in the exception handler we used so far when adding [`SystemStandardOutputLogger`](https://github.com/elastic/apm-agent-java/blob/main/apm-agent-common/src/main/java/co/elastic/apm/agent/common/util/SystemStandardOutputLogger.java#L58)!
Now we can avoid printing advice-based exceptions to std streams as well 🙂. 
You cannot use `SystemStandardOutputLogger` here, but please either add the same logic (look for the same system property or env variable that disables it, with consideration in security manager) or avoid using `printStackTrace`.
I am personally OK with not having a fallback at all as we expect our exception handler to always work.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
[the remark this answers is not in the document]
eyalkoren: Nice!! 👏

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
```diff
@@ -296,6 +315,10 @@ static void setJavaBaseModule(Class<?> targetClass) throws Throwable {
             .invoke(targetClass);
     }
 
+    public static void logExceptionThrownByAdvice(Throwable exception) {
+        LoggerFactory.getLogger(IndyBootstrap.class).error("Advice threw an exception, this should never happen!", exception);
```
eyalkoren, line 319: If there's no reason not to, please use a static `logger` field.

eyalkoren (commented):
Beautiful addition, well done!
Two minor comments.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/IndyBootstrap.java
JonasKunz: I think we can't use a static logger field here for the same reason we can't use it in the `bootstrap()`-method.
However, i'll add a static `logger()` method to keep the comment about it in a single place.

eyalkoren: That's very true, thanks for reminding me to look at my own comment 🤦‍♂️

eyalkoren (approved):

PR #2998, opened 2023-01-31
Skip deploying of java 17 subprojects

JonasKunz:
## What does this PR do?
During the last release I noticed that our release job attempts to deploy the new java 17-specific subprojects aswell, which is not intentional to my knowledge.

### DIFF
--- a/apm-agent-java-17/pom.xml
+++ b/apm-agent-java-17/pom.xml
@@ -17,6 +17,7 @@
     </modules>
 
     <properties>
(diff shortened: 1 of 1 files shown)


JonasKunz:
> Ahh, because it is a top level subproject we need to say that explicitly? LGTM

Exactly. The plugin projects for example do not need to specify it because they inherit the settings from their `apm-agent-plugins` (potentially transitive) parent, where the `maven-deploy-plugin.skip` is set to `true`.

PR #3001, opened 2023-02-02
Fix used instrumentations shown on shutdown

### LINKED ISSUE #2970: Agent prints invalid (too small) set of enable_isntrumentations at the end
Our current algorithm in `InstrumentationStats` for determining the options to put into `enable_instrumentations` for enabling all used instrumentations is flawed.

For example, running the agent on the following program
```
  @CaptureTransaction
  public static void main(String[] args) throws Exception {
      System.out.println(ElasticApm.currentSpan());
  }
```
yields the following output:
```
co.elastic.apm.agent.bci.InstrumentationStatsLifecycleListener - Used instrumentation groups: [annotations-capture-transaction, executor, executor-collection, fork-join, jul-error, slf4j-error, ssl-context, urlconnection]
```

This set is invalid, because `public-api` would be needed in order to enable the instrumentation for `ElasticApm.currentSpan()`.

Fortunately, the problem we are trying to solve here is a known problem in computer science: [the set cover problem](https://en.wikipedia.org/wiki/Set_cover_problem).
In our case the "universe" is formed by all Instrumentations. Every instrumentation-group-name forms a subset which contains all instrumentations which are enabled by adding this name to `enable_instrumentations`.

I suggest we use two phases of the greedy algorithm for solving the set-cover-problem:
 1. Find a set of instrumentation-group-names which enables all used instrumentations but minimized the number of enabled unused instrumentations
 2. Minimize this set of names while still keeping all used instrumentations enabled

JonasKunz:
## What does this PR do?
Closes #2970.
 * Made the output deterministic
 * Added correct handling for `experimental`
 * Ensured that no used instrumentations are missing

## Checklist

- [x] This is a bugfix


apm-agent-core/src/main/java/co/elastic/apm/agent/bci/InstrumentationStats.java
```diff
@@ -58,7 +62,9 @@ public Collection<String> getUsedInstrumentationGroups() {
         for (ElasticApmInstrumentation instrumentation : usedInstrumentations.keySet()) {
             usedInstrumentationGroups.addAll(instrumentation.getInstrumentationGroupNames());
         }
-        for (ElasticApmInstrumentation instrumentation : allInstrumentations) {
+        List<ElasticApmInstrumentation> allDeterministic = getAllInstrumentationsSorted();
+        for (ElasticApmInstrumentation instrumentation : allDeterministic) {
+            System.out.println("Instrumentation in order: " + instrumentation.getClass().getName());
```
eyalkoren, line 67: No more direct prints to std output please, only through [`SystemStandardOutputLogger`](https://github.com/elastic/apm-agent-java/blob/0e033282f2537c91e83e078c1506b1f763b5f956/apm-agent-common/src/main/java/co/elastic/apm/agent/common/util/SystemStandardOutputLogger.java#L29) please.
You may want to add a `DEBUG` level logging for this one.

eyalkoren (changes requested):

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/InstrumentationStats.java
JonasKunz: Yes, this is just for testing at the CI currently, I'll remove it when this PR is ready for review.
The test seems to be indeterminstic as in locally everything works and on the CI it doesn't, I'm trying to understand why

eyalkoren (approved):

PR #3002, opened 2023-02-06
Add configuration to suppress Async Profiler logging

### LINKED ISSUE #1789: WARNING: Install JVM debug symbols to improve profile accuracy
## Describe the bug
java agent is logging too many times "WARNING: Install JVM debug symbols to improve profile accuracy"
this affects logs readability

## Steps to reproduce
I'm using the apm-agent-attach-1.22.0-standalone agent 
Application is in container environment.
Logs are sent to stdout/stderr

```
 java -version
openjdk version "11.0.2" 2019-01-15 LTS
OpenJDK Runtime Environment 18.9 (build 11.0.2+7-LTS)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+7-LTS, mixed mode, sharing)
```

I tried to change the agent log_level to ERROR or to turn it OFF but warning messages are still there.

I cannot install JVM debug symbols 

I think those messages are coming from the async profiler. Is there a way to disable them?

eyalkoren:
## What does this PR do?
Closes #1789 

## Checklist
  - [x] I have added unit tests that proves the new configuration option is being used where required


apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/ProfilingConfiguration.java
```diff
@@ -54,6 +54,15 @@ public class ProfilingConfiguration extends ConfigurationOptionProvider {
...
         .buildWithDefault(false);
 
+    private final ConfigurationOption<Boolean> disableProfilingLogging = ConfigurationOption.<Boolean>booleanOption()
+        .key("profiling_logging_disabled")
+        .configurationCategory(PROFILING_CATEGORY)
+        .description("By default, async profiler prints warning messages about missing JVM symbols to standard output. \n" +
+            "Set this option to `true` to suppress such messages")
+        .dynamic(true)
+        .tags("added[1.37.0]")
+        .buildWithDefault(false);
```
SylvainJuge, line 64: [minor] maybe better to keep a positive wording here, so `profiling_logging_enabled` and make the default value `false`.

SylvainJuge (approved):
Thanks for taking care of this long-standing issue, I only have one minor suggestion on the option naming.

Have you been able to validate this manually with a JVM that makes async-profiler issue such messages ? I remember some "light" or "small" docker images where those debug symbols were not present, but I can't remember which ones.

apm-agent-plugins/apm-profiling-plugin/src/main/java/co/elastic/apm/agent/profiler/ProfilingConfiguration.java
eyalkoren: I started with `enabled`, but I figured it's better to reflect the opt-in-for-special-cases nature of this config with `disabled`. I am not sure...

SylvainJuge: Almost all the other configuration options that we have are `_enabled` suffix, thus that would be at least consistent with others:
- `circuit_breaker_enabled`
- `span_compression_enabled`
- `profiling_inferred_spans_enabled`

PR #3010, opened 2023-02-07
Switch to fully automated release based on nexus-staging-maven-plugin

### LINKED ISSUE #2855: Broken LATEST download link for snapshot version
The link we currently use in our readme does not return the expected artifact:

`https://oss.sonatype.org/service/local/artifact/maven/redirect?r=snapshots&g=co.elastic.apm&a=elastic-apm-agent&v=LATEST`

Which returns a `1.26.1-SNAPSHOT` version from ~1year ago when it should now return `1.34.2-SNAPSHOT`.

JonasKunz:
## What does this PR do?

Switches from the `maven-deploy-plugin` to the `nexus-staging-maven-plugin` for nexus deployment.
This makes the manual steps in the Nexus UI from our release process obsolete: The plugin automatically closed, releases and drops the repository in nexus. It should automatically perform a cleanup in case of a failure aswell.

I have the hope that his maybe fixes #2855 aswell.

Unfortunately this change can hardly be tested manually, I'll trigger a SNAPSHOT build after this is on stable to see if at least the snapshot deployment works properly.

### DIFF
--- a/apm-agent-plugins/apm-kafka-plugin/apm-kafka-spring-plugin/pom.xml
+++ b/apm-agent-plugins/apm-kafka-plugin/apm-kafka-spring-plugin/pom.xml
@@ -13,7 +13,7 @@
     </properties>
 
     <artifactId>apm-kafka-spring-plugin</artifactId>
-
+    <name>${project.groupId}:${project.artifactId}</name>
 
     <dependencies>
         <dependency>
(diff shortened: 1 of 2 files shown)


eyalkoren:
This seems very appealing indeed and testing it through snapshot is a great idea.

> The plugin automatically closed, releases and drops the repository in nexus

Does maven block and returns only after all these steps are completed?

PR #2686, opened 2022-06-24
Fix incorrect parent/child relationship in inferred span

felixbarny:
## Checklist

- [ ] This is an enhancement of existing features, or a new feature in existing plugins
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] I have made corresponding changes to the documentation
- [x] This is a bugfix
  - [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)
  - [x] I have added tests that would fail without this fix
- [ ] This is a new plugin
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)
  - [ ] My code follows the [style guidelines of this project](https://github.com/elastic/apm-agent-java/blob/main/CONTRIBUTING.md#java-language-formatting-guidelines)
  - [ ] I have made corresponding changes to the documentation
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/main/CONTRIBUTING.md#testing) pass locally with my changes
  - [ ] I have updated [supported-technologies.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/docs/supported-technologies.asciidoc)
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] Added an instrumentation plugin? Describe how you made sure that old, non-supported versions are not instrumented by accident.
- [ ] This is something else
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)

PR #3011, opened 2023-02-07
Using surefire `rerunFailingTestsCount` for flaky tests

eyalkoren:
Closes #2554

### DIFF
--- a/apm-agent-plugins/apm-profiling-plugin/src/test/java/co/elastic/apm/agent/profiler/CallTreeTest.java
+++ b/apm-agent-plugins/apm-profiling-plugin/src/test/java/co/elastic/apm/agent/profiler/CallTreeTest.java
@@ -59,6 +59,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.doReturn;
 
 @DisabledOnOs(OS.WINDOWS)
+@DisabledOnAppleSilicon
 class CallTreeTest {
 
     private MockReporter reporter;
--- a/pom.xml
+++ b/pom.xml
@@ -414,6 +414,7 @@
                 <configuration>
                     <enableAssertions>true</enableAssertions>
                     <trimStackTrace>false</trimStackTrace>
+                    <rerunFailingTestsCount>3</rerunFailingTestsCount>
                 </configuration>
             </plugin>
             <plugin>
@@ -502,11 +503,11 @@
                 </plugin>
                 <plugin>
                     <artifactId>maven-surefire-plugin</artifactId>
-                    <version>2.22.2</version>
+                    <version>3.0.0-M8</version>
                 </plugin>
                 <plugin>
                     <artifactId>maven-failsafe-plugin</artifactId>
-                    <version>2.22.2</version>
+                    <version>3.0.0-M8</version>
                 </plugin>
                 <plugin>
                     <artifactId>maven-jar-plugin</artifactId>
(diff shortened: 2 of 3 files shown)


apm-agent-plugins/apm-profiling-plugin/src/test/java/co/elastic/apm/agent/profiler/CallTreeTest.java
```diff
@@ -59,6 +59,7 @@
 import static org.mockito.Mockito.doReturn;
 
 @DisabledOnOs(OS.WINDOWS)
+@DisabledOnAppleSilicon
```
eyalkoren, line 62: Unrelated- was removed accidentally in #2686

pom.xml
```diff
@@ -414,6 +414,7 @@
                 <configuration>
                     <enableAssertions>true</enableAssertions>
                     <trimStackTrace>false</trimStackTrace>
+                    <rerunFailingTestsCount>3</rerunFailingTestsCount>
```
JonasKunz, line 417: We should add the same configuration for the integration-tests below in the `maven-failsafe-plugin`, shouldn't we?
Link to docs: https://maven.apache.org/surefire/maven-failsafe-plugin/examples/rerun-failing-tests.html

eyalkoren:
Since we are in the midst of the transition to using GH actions for reporting as well, the aggregation and report of such flaky tests results is out of scope in this PR
The only place I currently know this info is available is within the surefire output xml within a `<flakyFailure>` tag, but this means parsing all test output files. 
Example output for the `FlakyTest` added within this PR:
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuite ...>
  ...
  <testcase name="passOnSecondRun" classname="co.elastic.apm.agent.testutils.FlakyTest" time="0">
    <flakyFailure message="[This test is expected to fail in its first run] &#10;Expecting actual not to be null" type="java.lang.AssertionError">
      <stackTrace><![CDATA[java.lang.AssertionError: 
      ...
      </stackTrace>
    </flakyFailure>
  </testcase>
</testsuite>
```
Another option is using something like [maven-surefire-report-plugin](https://maven.apache.org/surefire/maven-surefire-report-plugin/usage.html), but as far as I can see this only creates HTML reports. If that is the case, we can probably create it, archive it and add a link to it, but we will still want to have some kind of indication within the GitHub PR page if there were flaky tests.

JonasKunz (approved):
So build wise it seems to work as intended:
```
[2023-02-08T07:29:15.204Z] [INFO] Results:
[2023-02-08T07:29:15.204Z] [INFO] 
[2023-02-08T07:29:15.204Z] [WARNING] Flakes: 
[2023-02-08T07:29:15.204Z] [WARNING] co.elastic.apm.agent.testutils.FlakyTest.passOnSecondRun
[2023-02-08T07:29:15.204Z] [ERROR]   Run 1: FlakyTest.passOnSecondRun:38 [This test is expected to fail in its first run] 
[2023-02-08T07:29:15.204Z] Expecting actual not to be null
[2023-02-08T07:29:15.204Z] [INFO]   Run 2: PASS
[2023-02-08T07:29:15.204Z] [INFO] 
[2023-02-08T07:29:15.204Z] [INFO] 
[2023-02-08T07:29:15.204Z] [WARNING] Tests run: 970, Failures: 0, Errors: 0, Skipped: 1, Flakes: 1
```
We'll just need to figure out a way to make it show up in CI-report and/or GH comment.

EDIT: Looks like we responded at the same time, aggree that fixing the reporting is out of scope for this PR.

PR #3015, opened 2023-02-08
Elasticsearch rest async context propagation

SylvainJuge:
## What does this PR do?

Initially reported [in our forum](https://discuss.elastic.co/t/elastic-java-client-async-apm-transaction/322986/3).

When using the new ES Java client (8.x), the asynchronous call is delegated to a `CompletableFuture` (CF) using code similar to the following:

```java
  RestClient restClient = RestClient.builder(ES_HOST).setDefaultHeaders(DEFAULT_HEADERS).build();
        ElasticsearchTransport transport = new RestClientTransport(restClient, new JacksonJsonpMapper());
        ElasticsearchAsyncClient client = new ElasticsearchAsyncClient(transport);

        CompletableFuture<SearchResponse<Product>> search = client.search(s -> s.index("products")
                        .query(q -> q.term(t -> t.field("name")
                                .value(v -> v.stringValue("product")))),
                Product.class);

        CompletableFuture<SearchResponse<Product>> result = search
                .thenApplyAsync(response -> {
                    
                    // We don't see the active transaction or span, even if there is one
                    System.out.println("span = "+ ElasticApm.currentSpan()); 
                    System.out.println("span = "+ ElasticApm.currentTransaction());

                    return response;
                }, _executorService);
```

In practice, the context propagation works already for the `CompletableFuture`, and I just had to extend `co.elastic.apm.agent.esrestclient.ResponseListenerWrapper` to ensure this entry-point in the ES driver is properly covered.

While the example above relies on the new 8.x ES client, the underlying method that is instrumented is available in recent 7.17 versions, thus we just provide it in a new7.x ES instrumentation plugin (previously it was only just for testing).

## Checklist

- [x] This is a bugfix


apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/esrestclient/ResponseListenerWrapper.java
```diff
@@ -28,24 +30,69 @@
 public class ResponseListenerWrapper implements ResponseListener, Recyclable {
 
     private final ElasticsearchRestClientInstrumentationHelper helper;
+    private final Tracer tracer;
+
     @Nullable
     private ResponseListener delegate;
+
```
SylvainJuge, line 37: [for reviewer] Here we can't easily sub-class due to the class instance pooling, thus there is effectively two implementations that are triggered depending on the use-case.

SylvainJuge:
Bugfix has been validated by the end-user (who initially reported it in the forum).

jackshirazi (approved):

PR #3024, opened 2023-02-18
HttpURLConnection (#2976) - fix getResponseCode handled exceptions

### LINKED ISSUE #2976: urlconnection plugin: HttpURLConnection exceptions are captured also handled by the application.
## Describe the bug

Also the application handles all exceptions, especially expected ones the agent is reporting them to APM.

## Steps to reproduce

Simple Spring Boot Web application

```
@RequestMapping(method = RequestMethod.GET, path = "/ping")
public ResponseEntity<Object> ping() throws IOException {
    HttpURLConnection conn = (HttpURLConnection) new URL("https://google.de/404").openConnection();
    conn.setInstanceFollowRedirects(true);
    conn.setRequestMethod("GET");
    int code = conn.getResponseCode();
    System.out.println("Response Code: " + code);
    System.out.println("I reached the end without an exception!");
}
```
(Using Servlet with `RequestMapping` is just used to trigger the code inside a Spring Boot application).

gives following console output

```
Response Code: 404
I reached the end without an exception!
```

But the agent runs into https://github.com/elastic/apm-agent-java/blob/main/apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java#L132 and reports the exception to APM

Screenshot 2023-01-18 at 20 22 00

----

Different example using an url which response with 401, results into an `java.io.IOException` to be reported to APM, also 

```
Response Code: 401
I reached the end without an exception!
```

Screenshot 2023-01-18 at 20 37 26

## Expected behavior

Since everything runs as expected and exceptions are handled, there shouldn't be any exception reported towards APM (which as follow up results in an alert, if someone is creating alerts based on occurring exceptions)

## Debug logs

Log for the 404 example.

[reproducer.log](https://github.com/elastic/apm-agent-java/files/10449961/reproducer.log)

## Other remarks

We have observed same behaviour with AWS SDK S3 instrumentation, also there exceptions are captured also handled later on. Which means we have detected errors in APM, also the code handled and just works fine.

markush81:
## What does this PR do?
Fixes #2976 same way OTel has implemented its instrumentation of `HttpURLConnection`

## Checklist

- [x] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -36,6 +36,7 @@ communication - {pull}2996[#2996]
 * Prevent potential connection leak on network failure - {pull}2869[#2869]
 * Fix for inferred spans where the parent id was also a child id - {pull}2686[#2686]
 * Fix context propagation for async 7.x and 8.x Elasticsearch clients - {pull}3015[#3015]
+* Fix for `HttpUrlConnection` throwing exception for responseCode >= 400 (OTel inspired way to fix) - {pull}3024[#3024]
 
 [[release-notes-1.x]]
 === Java Agent version 1.x
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -38,7 +38,7 @@ communication - {pull}2996[#2996]
 * Fix for inferred spans where the parent id was also a child id - {pull}2686[#2686]
 * Fix context propagation for async 7.x and 8.x Elasticsearch clients - {pull}3015[#3015]
 * Fix exceptions filtering based on <<config-ignore-exceptions>> when those are <<config-unnest-exceptions, nested>> - {pull}3025[#3025]
-* Fix for `HttpUrlConnection.getResponseCode` capturing an exception, also internally already handled - {pull}3024[#3024]
+* Fix usage of `HttpUrlConnection.getResponseCode()` causing an error event due to exception capturing, even when it is internally handled - {pull}3024[#3024]
 
 [[release-notes-1.x]]
 === Java Agent version 1.x
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -36,7 +36,7 @@ communication - {pull}2996[#2996]
 * Prevent potential connection leak on network failure - {pull}2869[#2869]
 * Fix for inferred spans where the parent id was also a child id - {pull}2686[#2686]
 * Fix context propagation for async 7.x and 8.x Elasticsearch clients - {pull}3015[#3015]
-* Fix for `HttpUrlConnection` throwing exception for responseCode >= 400 (OTel inspired way to fix) - {pull}3024[#3024]
+* Fix for `HttpUrlConnection.getResponseCode` capturing an exception, also internally already handled - {pull}3024[#3024]
 
 [[release-notes-1.x]]
 === Java Agent version 1.x
--- a/apm-agent-plugins/apm-urlconnection-plugin/src/test/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentationTest.java
+++ b/apm-agent-plugins/apm-urlconnection-plugin/src/test/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentationTest.java
@@ -122,7 +122,7 @@ public class HttpUrlConnectionInstrumentationTest extends AbstractHttpClientInst
     }
 
     @Test
-    public void testGetResponseCodeWithUnknownHost() {
+    public void testGetResponseCodeWithUnhandledException() {
         try {
             performGet("http://unknown");
         } catch (Exception e) {
--- a/apm-agent-plugins/apm-urlconnection-plugin/src/test/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentationTest.java
+++ b/apm-agent-plugins/apm-urlconnection-plugin/src/test/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentationTest.java
@@ -124,7 +124,9 @@ public class HttpUrlConnectionInstrumentationTest extends AbstractHttpClientInst
     @Test
     public void testGetResponseCodeWithUnhandledException() {
         try {
-            performGet("http://unknown");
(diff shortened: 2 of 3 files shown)


eyalkoren:
@markush81 thanks for taking a shot on this! 🙏 
Please see my last [comment](https://github.com/elastic/apm-agent-java/issues/2976#issuecomment-1437227531) on the original issue. It explains how we think it would be best to approach this.
If you want to implement this yourself, we will be happy to assist and review. Otherwise, if you don't mind and agree that this approach is preferable, we can implement that.
Please let us know.

apm-agent-plugins/apm-urlconnection-plugin/src/test/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentationTest.java
```diff
@@ -121,4 +121,27 @@ public void testFakeReuse() throws Exception {
...
+
+    @Test
+    public void testGetResponseWithHandledException() throws Exception {
+        final HttpURLConnection urlConnection = (HttpURLConnection) new URL(getBaseUrl() + "/non-existent").openConnection();
+        urlConnection.getResponseCode();
+        urlConnection.disconnect();
+
+        verifyHttpSpan("localhost", "/non-existent", 404);
+        assertThat(reporter.getErrors()).isEmpty();
+    }
```
eyalkoren, line 145: Please add a test for the same URL only using `getInputStream()` (or even better- use `performGet()`) instead of `getResponseCode()` and verifying the expected behavior - `assertThat(reporter.getErrors()).hasSize(1)`

CHANGELOG.asciidoc
```diff
@@ -38,6 +38,7 @@ communication - {pull}2996[#2996]
 * Fix for inferred spans where the parent id was also a child id - {pull}2686[#2686]
 * Fix context propagation for async 7.x and 8.x Elasticsearch clients - {pull}3015[#3015]
 * Fix exceptions filtering based on <<config-ignore-exceptions>> when those are <<config-unnest-exceptions, nested>> - {pull}3025[#3025]
+* Fix for `HttpUrlConnection.getResponseCode` capturing an exception, also internally already handled - {pull}3024[#3024]
```
eyalkoren, line 41:
```suggestion
* Fix usage of `HttpUrlConnection.getResponseCode()` causing an error event due to exception capturing, even when it is internally handled - {pull}3024[#3024]
```

eyalkoren (commented):
Looks good! Thanks!
Please add the requested test to fully verify our intentions are addressed and maintained.

apm-agent-plugins/apm-urlconnection-plugin/src/test/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentationTest.java
```diff
@@ -121,4 +121,27 @@ public void testFakeReuse() throws Exception {
...
+    public void testGetResponseCodeWithUnhandledException() {
+        try {
+            final HttpURLConnection urlConnection = (HttpURLConnection) new URL("http://unknown").openConnection();
+            urlConnection.getResponseCode();
+            urlConnection.disconnect();
+        } catch (Exception e) {
+            // intentionally ignored
+        }
+
+        assertThat(reporter.getErrors()).hasSize(1);
```
eyalkoren, line 134:
```suggestion
        assertThat(reporter.getErrors()).hasSize(1);
        
        Assertions.assertThat(reporter.getFirstSpan(500)).isNotNull();
        assertThat(reporter.getSpans()).hasSize(1);
        Span span = reporter.getSpans().get(0);
        assertThat(span.getNameAsString()).isEqualTo("GET unknown");
        Http http = span.getContext().getHttp();
        assertThat(http.getStatusCode()).isEqualTo(0);
        assertThat(http.getUrl().toString()).isEqualTo("http://unknown");
```

markush81: adjusted

apm-agent-plugins/apm-urlconnection-plugin/src/test/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentationTest.java
markush81: added, but kind like redundant to `co.elastic.apm.agent.httpclient.AbstractHttpClientInstrumentationTest#testNonExistingHttpCall` except for the additional assert. So i am not sure if changing the mentioned test, would also be an option.

eyalkoren (approved):
👍 🙏

apm-agent-plugins/apm-urlconnection-plugin/src/test/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentationTest.java
eyalkoren: The additional assertion (for an error being captured) is exactly what makes it non-redundant, it is what verifies this codes does what we intend it does.
And changing the mentioned test is not preferable as this is a very specific behavior for `HttpUrlConnection` instrumentation and only when using `getInputStream`. Other clients do not report errors in such 404 requests.

markush81: > And changing the mentioned test is not preferable as this is a very specific behavior for `HttpUrlConnection` instrumentation and only when using `getInputStream`. Other clients do not report errors in such 404 requests.

True, that makes sense then.

markush81:
run elasticsearch-ci/docs

jackshirazi (approved):
Great work, thanks!

PR #3026, opened 2023-02-20
Create GH actions workflow for agent release

JonasKunz:
## What does this PR do?

Add GH actions workflow for our existing release process.
Most of the process is unchanged, main difference is that all commiting / tagging is done in the very first job within the workflow and not during the actual build and deploy anymore.

I added a new automation for moving eveything in the `unreleased` section in the changelog to a new section for the target release.

As a result, we should now be able to retry the individual steps in this workflow, because each one is atomic and retriable in case of failure.

I testes the provided jobs as far as possible without authorization on my own fork.
The following things still need to be done:
 * Implement the actual buildkite integration for the `mvn deploy`
 * Implement the AWS Lambda build
 * Setup the `apmmachine` authorization (or whichever user we use)

I have marked all the places with `# TODO` comments in the workflow file where work is needed.

### DIFF
--- a/.ci/ReleaseChangelog.java
+++ b/.ci/ReleaseChangelog.java
@@ -1,0 +1,126 @@
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.time.Instant;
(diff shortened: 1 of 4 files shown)


JonasKunz:
Closed in favor of #3137

PR #3027, opened 2023-02-20
Implementing `long_field_max_length` according to spec

eyalkoren:
## What does this PR do?
Implementing the [`long_field_max_length` spec](https://github.com/elastic/apm/blob/main/specs/agents/field-limits.md#long_field_max_length-configuration).

## Checklist

  - [ ] ~I have added tests that prove my fix is effective or that my feature works~ - this is a static field that is globally used, which makes it hard to test configuration changes. We can change that only to allow testing, but it requires quite a bit of refactoring of all places that use this field, so I think it would be an overkill in this case.


Jenkinsfile
```diff
@@ -162,7 +162,7 @@ pipeline {
                   unstashV2(name: 'build', bucket: "${JOB_GCS_BUCKET_STASH}", credentialsId: "${JOB_GCS_CREDENTIALS}")
                   dir("${BASE_DIR}") {
                     withOtelEnv() {
-                      sh label: 'mvn test', script: './mvnw test'
+                      sh label: 'mvn test', script: './mvnw test -DargLine=\"-Delastic.apm.overwrite.config.docs=false\"'
```
JonasKunz, line 165: I'm assuming you intentionally sneaked in the changes to fix our documentation tests here? I'm fine with that, just asking to make sure it was intentional.

Depending on which PR gets merged first, we need to make sure to also add this to GH actions (#2870).

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Db.java
```diff
@@ -40,7 +40,7 @@ public class Db implements Recyclable {
         new Allocator<CharBuffer>() {
             @Override
             public CharBuffer createInstance() {
-                return CharBuffer.allocate(DslJsonSerializer.MAX_LONG_STRING_VALUE_LENGTH);
+                return CharBuffer.allocate(SerializationConstants.getMaxLongStringValueLength());
```
JonasKunz, line 43: This is not 100% spec conformant because some unicode characters might require two UTF-16 chars to be represented ([wiki link](https://en.wikipedia.org/wiki/UTF-16#Code_points_from_U+010000_to_U+10FFFF)), but this is such an edge case that I think this is acceptable.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/SerializationConstants.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    private static SerializationConstants get() {
+        if (INSTANCE != null) {
+            return INSTANCE;
+        }
+
+        ElasticApmTracer tracer = GlobalTracer.getTracerImpl();
+        if (tracer == null) {
+            // should only happen before tracer is initialized, mostly in test
```
JonasKunz, line 62: [Minor] How many tests are affected by this?
I would prefer moving the initialization of `SerializationConstants` without a `GlobalTracer` being set to the tests themselves if it is only a handful of tests. Otherwise it "risks" that the `tracer==null` branch is accidentally executed even in production.

If it is a lot effort it's not worth it IMO, feel free to ignore this comment.

apm-agent-plugins/apm-spring-webmvc-plugin/pom.xml
```diff
@@ -107,4 +107,16 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <!-- Required for testing -->
+                    <argLine>--add-opens java.base/java.lang=ALL-UNNAMED</argLine>
```
JonasKunz, line 116: [Question] why is this required for testing?

JonasKunz (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/Db.java
eyalkoren: Yeah, we can definitely live with that. 
Since this is a configuration to limit field size for transport and storage, it even makes more sense this way, assuming that it will enforce the maximum size we want in all encoded forms.

apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/SerializationConstants.java
eyalkoren: This is a question for @SylvainJuge

SylvainJuge: Only a few test classes actually require this, so I'll try pushing another version without this ugly lazy init.

SylvainJuge: Should be better with 5d8b06e7fec0ce78ed5f6775b8b3c19bf9f5d4e0

eyalkoren: `MockReporter` required this initialization as well - [`c9bd607`](https://github.com/elastic/apm-agent-java/pull/3027/commits/c9bd6073edfafa6b9804a10110966b0fde9ca35e).
Not the prettiest, but I agree it's preferable over leaking this concern into production code, so we'll go with that and rethink if we find this biting us again.

Issue #3030, opened 2023-02-20
Favour Tracer above ElasticApmTracer in plugins


raphw:
## Is your feature request related to a problem?
I would like to piggyback on Elastic's agent for non-APM-tracing, but at various points, the agent is bound to implementation details of Elastic's agent. This is mainly with regards to the use of `ElasticApmTracer` being requested from various places. However, this does not seem necessary for the most parts, the method missing from the `Tracer` API are `getConfig()` and `getObjectPoolFactory()` with few exceptions.

## Describe the solution you'd like
By adding only a few methods to the tracer API, the direct bounds to Elastic's APM can be removed from almost all plugins. This way it becomes possible to implement a custom agent whilst still using the comprehensive infrastructure that is offered by the Elastic agent and its plugins.

## Describe alternatives you've considered
Forking the repository and implementing a copy with individual customization, but that is of course not a good long-term strategy and hinders collaboration on the main repository.

## Additional context
Open Telementry offers a generic agent for implementing a blue print agent. Unfortunately, the agent does not support as many plugins as the Elastic agent and is not equally well maintained. It would be great if the Elastic agent could evolve in a direction where it would be possible to use it for other tracing purposes. As an ideal result, this would trigger additional contributions on the long run and help the Elastic agent to grow its standing in the field. To begin with, I would want to make some binary-compatible changes that improve the pluggability, but at some point one could even better differentiate the plugin SPI from the core agent API.

raphw:
I created a repo where I tried a POV to see if this is feasible and it seems like a small change which already separates the APIs much more. https://github.com/raphw/apm-agent-java/commit/53e70e0b5f50ada526930962bbbb006a18ba6918

PR #3031, opened 2023-02-20
3030: Reduce dependency on elastic APM tracer in plugins but rather depend on a more general API.

raphw:
POC of https://github.com/elastic/apm-agent-java/issues/3030


raphw:
This change set introduces a `Tracer` API that is independent of the *apm-agent-core* module to separate plugins from the elastic agent's implementation. Instead, the suggested tracer API is located in an independent *apm-agent-tracer-api* module which is declared a dependency of the current *apm-agent-plugin-sdk* module. As a result, it becomes possible to implement a `Tracer` without considering the API needed for instrumentation plugins. Furthermore, it becomes possible to implement custom agents that use the custom tracer while still using the, now generalized, plugins. The change set was created largely through automatic IDE refactoring in the following steps:

- *Interface extraction* for the `Tracer` class and its method parameter types such as `Span` or `Transaction` to create a parallel hierarchy. Manually, I generified the new `Span` and `Transaction` interfaces to allow for simple subtyping with correct return values.
- Moved the generated interface APIs to a new module without its own dependencies.
- Introduces a new `GlobalTracer` that is API compatible with the current `GlobalTracer` where the latter now delegates to the former.
- *Type substitution* only within the plugin modules only for subprojects of the *apm-agent-plugins*. As these plugins never instantiate any classes of the tracer API but only act as consumers, so replacing classes with interfaces was with little friction. Also, test code was exempt from substitution as all previous code is subtyping new interfaces of the same name. To satisfy some generic typing, additional code was substituted to a minor extend. 
- *Rollback* of some (sub-)plugins, namely the *apm-api-plugin*, *apm-awslambda-plugin*, *apm-logging-plugin*, *apm-micrometer-plugin*, *apm-opentelementry*, *apm-opentracing-plugin* and *apm-profiling-plugin*. Those plugin use a very different subset of methods of the `Tracer` API and I took them out of scope to keep this change set small, for a very contextual definition of small.
- *Removal of unused methods* of the extracted interfaces as they were not needed by the plugins that I extracted. I did some minimal clean up and was able to remove one method manually which I found was only used by a single plugin, but that was recreatable by using other methods of the interface. I did not explore this further, but discovered multiple methods that were only used few or even single places and that would likely be removable.
- Introduces a polymorphism mechanism on the `Tracer` API (`require`/`probe`) where a tracer can be checked for a subclass. As an experiment, I introduced a `ServiceAwareTracer` subtype of `Tracer`. This way, plugins can require extended tracer capabilities and only use them if available. This way, the plugins that were previously excluded could be moved to the new API, too, without lifting this burden upon other plugins. Also, external plugins can simply hook into the current tracer API with extended use without code customizations.
- Adjustment of the plugin instantiation mechanism. The code now checks for the most-specific constructor `Tracer` type. This way, plugins can deactivate themselves by requiring a given subtype of `Tracer` or initialize for limited capability. Without special requirements, they can also only declare a default constructor.
- *Interface extraction* of the configuration classes and offering of `MinimalConfiguration` interface and an implementation of all relevant configuration types. Those can be used if one is not interested in configuring extended capabilities without breaking the current plugin configuration mechanism. The interfaces were then moved to the *apm-agent-plugin-sdk* module.
- *Dependency substitution* of the plugins that were not set aside. Instead of depending on *apm-agent-core* they now depend on *apm-agent-plugin-sdk*. 
- Copying of some types that were used by the plugins to the *apm-agent-plugin-sdk* and *automatic import correction* on those plugins. For example, many plugins use `*Util` classes. Also, they rely on the `SimpleMethodSignature` annotation and now subclass a new `TracerAwareInstrumentation`. I did not delete the original classes to avoid binary incompatibilities.
- Introduced `WeakConcurrent.weakSpanMap()` which exposes `WeakConcurrentProviderImpl.weakSpanMapImpl()` which is needed in multiple plugins.
- Fix reminding compilation issues manually, typically with regards to generic types not being compatible when a `Span` class meets a `Span<?>` interface.

The change set is supposedly fully binary compatible, but might require some minor source code changes with regards to generic typing. I realize it is a large change, but it is mainly reorganizing the code, hopefully without any behavioral changes. One can think of it as introducing the JPA to Hibernate. While it does not add any direct value or feature to the Elastic agent, it offers a much smaller surface to plugin developers, plus allows plugin developers and their consumers, to use the implementations that you already offer. 

Possible improvements on this further would be:
- Add additional interfaces similar to the `ServiceAwareTracer` to the *apm-agent-tracer-api* module which allow supporting the skipped plugins. I already experimented with it by adding layers of a `SpanAwareTracer` and a `MetricsAwareTracer` to better understand the needs of the plugins I wanted to skip. This seems straight forward and would again mainly involve automated refactorings.
- Extract a base agent template, from *apm-agent-core* that contains modules like code for locating and instantiating plugins or setting up the invoke dynamic dispatch for them. Again, this would mainly involve code relocation which can be automated.

raphw:
Closed in favor of more focused PRs.

PR #3034, opened 2023-02-23
Update year 2023 in the NOTICE

v1v:
## What does this PR do?

Update year in the NOTICE file to reflect our current year


eyalkoren (approved):
I am not sure it is required anymore...

Issue #2982, opened 2023-01-23
OpenJDK CRaC support


ArtyomGabeev:
## Is your feature request related to a problem?

AWS announced support of snapstart in AWS Lambda based on Java runtime. 
It looks like this support based on OpenJDK CRaC, with ability to specify before/after context restore hooks, reference:
https://aws.amazon.com/blogs/compute/reducing-java-cold-starts-on-aws-lambda-functions-with-snapstart/
https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html#snapstart-compatibility

However if java agent attached before snapshot created, I think it will lead to incorrect results (as least sending APM data with same instance metadata).

## Describe the solution you'd like

Ideally, we would like to bytebuddy instrumentation happen before snapshot is took.
In afterRestore method, java agent should be able re update only necessary instance specific information

## Describe alternatives you've considered

Attaching agent in afterRestore hook, however it returns back initialisation latency

## Additional context

Implementing and providing CRaC restore hook may be useful not only for AWS Lambda application, but for any java application using CRaC project.

ArtyomGabeev:
I completed some testing with enabled SnapStart and want to provide some information, which should help with implementation.

Setup: 
- use official configuration instructions and add two layers to enable java agent.
- enable snapstart and publish java version

I see, after invoking function, trace data in Elasticsearch.
However, service.node.name field was not reported and in Kibana instance was shown as "\<unknown\>".

There is a difference in environment variables when function started from SnapStart, AWS_LAMBDA_LOG_STREAM_NAME is not populated, and ServiceFactory does not fill node name in method augmentServiceForAWSLambda.

I've opened an issue on AWS java lib, let see if its expected:
https://github.com/aws/aws-lambda-java-libs/issues/402

Note:
If AWS_LAMBDA_LOG_STREAM_NAME won't be available for SnapStart functions, we can provide a fallback implementation based on ephemeralId, e.g.:
```java
private static String generateNodeName(String ephemeralId) {
    DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy/MM/dd");
    LocalDate localDate = LocalDate.now();

    String version = "[" + PrivilegedActionUtils.getEnv("AWS_LAMBDA_FUNCTION_VERSION") + "]";

    return dtf.format(localDate) + "/" + version + ephemeralId.replaceAll("-", "");
  }
```

[...]

PR #3036, opened 2023-02-26
Refresh tracer metadata on CRaC snapshot restore

ArtyomGabeev:
## Checklist

Initial proposal for https://github.com/elastic/apm-agent-java/issues/2982

- [ ] This is a bugfix
- [ ] This is a new plugin
- [x] This is something else


apm-agent-plugins/apm-awslambda-plugin/src/main/java/co/elastic/apm/agent/awslambda/helper/AbstractLambdaTransactionHelper.java
```diff
@@ -143,6 +143,8 @@ private void completeMetaData(Context lambdaContext) {
             String region = arnSegments[3];
             String accountId = arnSegments[4];
 
+            String logStreamName = lambdaContext.getLogStreamName();
```
ArtyomGabeev, line 146: unfortunately, after snapshot restore, this method returns null, so lambda instance has \<uknown\> node name instead of CloudWatch log stream name.

AWS team replied, that this environment variable won't be available, since it's impossible to change env variables after snapshot it taken.

We need to decide for a fallback strategy, how to generate lambda instance identifier.

ArtyomGabeev: I've got a reply from AWS that neither AWS_LAMBDA_LOG_STREAM_NAME, neither context.getLogStreamName gonna be populated for SnapStart enabled functions - https://github.com/aws/aws-lambda-java-libs/issues/402 .

I do not know, if Kibana product have some special handling of AWS lambda APM data and if it relies on the format of configured_name from service.node field.

In initial proposal, I suggest to simulate and generate similar to log group node name, but right now I think we do not need to tie (if possible) instance name to log group name at all.

If we need just to distinguish different lambda instances, may be pair of lambda version + ephemeral id is enough, but this is a breaking change, if we are going to change it for all AWS Lambdas.

Another option I'm considering is to introduce a property, which will be a strategy for lambda instance naming.
By default, we should still try to use AWS_LAMBDA_LOG_STREAM_NAME env variable, but we may allow another one, which will use auto generated name.

ArtyomGabeev:
+1

jackshirazi:
Hi, thank you for this contribution it's really interesting! I'll be looking at this, it may take a little while as I'm also on other things. First thing though, can you please sign the [contributor agreement](https://www.elastic.co/contributor-agreement)

elastic-apm-agent-premain/src/main/java/co/elastic/apm/agent/premain/AgentMain.java
```diff
@@ -34,9 +37,14 @@
  */
 public class AgentMain {
 
+    private static final AgentCracResource agentCracResource = new AgentCracResource();
```
jackshirazi, line 40: Is there a reason to hold a reference to the object? It seems like it only needs to be registered, it could be created then?

ArtyomGabeev: @jackshirazi , Hi.
According to AWS documentation we need to have a strong reference for Crac resource:
https://docs.aws.amazon.com/lambda/latest/dg/snapstart-runtime-hooks.html

Thanks

jackshirazi:
You seem to have 2 email addresses, one is signed up but the one associated with this PR isn't. I've tried to add the latter email address, did you get an email?

elastic-apm-agent-premain/src/main/java/co/elastic/apm/agent/premain/AgentMain.java
jackshirazi: Thanks, I missed that

ArtyomGabeev:
> You seem to have 2 email addresses, one is signed up but the one associated with this PR isn't. I've tried to add the latter email address, did you get an email?

Thank you, looks like CLA is signed off.

ArtyomGabeev:
+1

NicklasWallgren:
+1

jackshirazi:
still pending :+1

v1v:
run docs-build

PR #3037, opened 2023-02-27
add agent config access to the plugin sdk

jackshirazi:
## What does this PR do?
Adds an ElasticApm.getConfig() method that allows external plugins and manual instrumentation to access the elastic config

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else


apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
```diff
@@ -249,4 +249,44 @@ public static void captureException(@Nullable Throwable e) {
...
+     * <p>
+     * Note the value returned can be null if an invalid key is used, and also
+     * if the "public-api" set of instrumentations has not been applied
+     * </p>
+     *
+     * @param key the string option
+     * @return The current value for the option, per the option type (String, Long, Integer, etc)
+     * @since 1.37.0
+     */
+    public static Object getConfig(String key) {
```
SylvainJuge, line 271:
```suggestion
    public static <T> T getConfig(String key, Class<T> type) {
```
[minor] this avoids casts in the calling code, and it allows to check the actual type at invocation time.

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/ElasticApmApiInstrumentation.java
```diff
@@ -162,4 +163,25 @@ public static void setServiceInfoForClassLoader(@Advice.Argument(0) @Nullable Cl
...
+            @Advice.AssignReturned.ToReturned
+            @Advice.OnMethodExit(suppress = Throwable.class, inline = false)
+            public static Object getConfig(@Advice.Argument(0) @Nullable String key) {
+                try {
+                    Object value= GlobalTracer.getTracerImpl().getConfigurationRegistry().getConfigurationOptionByKey(key).getValue();
+                    // should we return a special value for null - which means non-existent entry?
+                    return value;
+                } catch (NullPointerException e) {
+                    return null;
+                }
```
SylvainJuge, line 183: While we usually try to be very kind to our end-users, I would say that here we should throw an `IllegalArgumentException` if there is no matching config option. If we add the expected returned type as an argument, then we can also check the returned type and also throw an error if it's invalid.
Both cases would be bugs on the calling code, so we'd better throw exceptions early to prevent bugs from compounding later.

apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/agent/pluginapi/ConfigInstrumentationTest.java
```diff
@@ -0,0 +1,106 @@
...
+        tracer = MockTracer.createRealTracer();
+        configurationRegistry = tracer.getConfigurationRegistry();
+        coreConfig = configurationRegistry.getConfig(CoreConfiguration.class);
+    }
+
+    @AfterEach
+    void reset() {
+        ElasticApmAgent.reset();
+    }
+
```
SylvainJuge, line 65: [minor] we should add a test case for when calling `getConfig` with an unknown configuration key.

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
jackshirazi: excellent suggestion, thanks! changed now

apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/agent/pluginapi/ConfigInstrumentationTest.java
jackshirazi: good point, tests added

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
```diff
@@ -260,24 +260,42 @@ public static void setServiceInfoForClassLoader(@Nullable ClassLoader classLoade
...
+        Object value = doGetConfig(key);
+        if (value == null) {
+            throw new IllegalArgumentException("There is no such option: " + key);
+        } else if (value instanceof IllegalStateException) {
+            return null;
+        } else if (value.getClass().equals(type)) {
+            return (T) value;
+        } else {
+            throw new IllegalArgumentException("The option: '"+key+"' is not of type "+type.getName());
+        }
```
SylvainJuge, line 286: [minor] I think we could simplify this further by keeping the exception creation in the advice code, then only check if the returned value is an exception (and if yes, then throw it), also the type matching is not exact, and we could for example pass `Number` to get a `Long` configuration option, but that's definitely nit-picking here :-).

SylvainJuge (approved):

apm-agent-api/src/main/java/co/elastic/apm/api/ElasticApm.java
jackshirazi: it's similar branching just in the private method than the advice - I prefer keeping the advice as simple as possible.

I did think about looking for inheritance, but decided to keep it exact matching because I think that way the customer reads the option more thoroughly - and we seem to have a lot of scanning the options rather than reading them. But I'm not stuck on this opinion

jackshirazi: actually I decided you are right and changed it to an inheritance match

PR #3039, opened 2023-02-28
add comments to clarify it is already aligned to the compressed span spec

jackshirazi:
Just additional comments


jackshirazi:
for the reviewer, [this is the spec change](https://github.com/elastic/apm/pull/632/files)

JonasKunz (approved):
LGTM

My reading as well that we were already compliant with the spec.

PR #3040, opened 2023-03-01
Added documentation for Otel metrics

JonasKunz:
## What does this PR do?
Adds the documentation for #2968 and #3014.

### DIFF
--- a/docs/api-opentelemetry.asciidoc
+++ b/docs/api-opentelemetry.asciidoc
@@ -7,7 +7,7 @@ endif::[]
 === OpenTelemetry bridge
 
 The Elastic APM OpenTelemetry bridge allows creating Elastic APM `Transactions` and `Spans`,
-using the OpenTelemetry API. OpenTelemetry metrics are collected and shipped to elastic as well.
+using the OpenTelemetry API. OpenTelemetry metrics are also collected.
 In other words,
 it translates the calls to the OpenTelemetry API to Elastic APM and thus allows for reusing existing instrumentation.
 
@@ -169,7 +169,7 @@ counter.add(42);
 
 We don't require you to setup an OpenTelemetry `MeterProvider` within `GlobalOpenTelemetry` yourself.
 The Elastic APM Java Agent will detect if no `MeterProvider` was configured and will provide its own automatically in this case.
-If you provide your own `MeterProvider` (see <<otel-metrics-sdk>>), it will not be overridden by the agent.
+If you provide your own `MeterProvider` (see <<otel-metrics-sdk>>), the agent will use the provided instance.
 
 [float]
 [[otel-metrics-sdk]]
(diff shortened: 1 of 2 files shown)


docs/api-opentelemetry.asciidoc
```diff
@@ -7,7 +7,7 @@ endif::[]
 === OpenTelemetry bridge
 
 The Elastic APM OpenTelemetry bridge allows creating Elastic APM `Transactions` and `Spans`,
-using the OpenTelemetry API.
+using the OpenTelemetry API. OpenTelemetry metrics are collected and shipped to elastic as well.
```
SylvainJuge, line 10:
```suggestion
using the OpenTelemetry API. OpenTelemetry metrics are also collected.
```

`elastic` in itself is a bit vague destination, and we'd better avoid any confusion where the metrics are sent.

docs/api-opentelemetry.asciidoc
```diff
@@ -136,17 +136,106 @@ try (Scope scope = custom.makeCurrent()) {
...
+import io.opentelemetry.api.metrics.Meter;
+
+Meter myMeter = GlobalOpenTelemetry.getMeter("my_meter");
+LongCounter counter = meter.counterBuilder("my_counter").build();
+counter.add(42);
+----
+
+We don't require you to setup an OpenTelemetry `MeterProvider` within `GlobalOpenTelemetry` yourself.
+The Elastic APM Java Agent will detect if no `MeterProvider` was configured and will provide its own automatically in this case.
+If you provide your own `MeterProvider` (see <<otel-metrics-sdk>>), it will not be overridden by the agent.
```
SylvainJuge, line 172:
```suggestion
If you provide your own `MeterProvider` (see <<otel-metrics-sdk>>), the agent will use the provided instance.
```

docs/api-opentelemetry.asciidoc
```diff
@@ -136,17 +136,106 @@ try (Scope scope = custom.makeCurrent()) {
...
+
+[float]
+[[otel-metrics-sdk]]
+==== Using a customized MeterProvider
+
+In some cases using just the <<otel-metrics-api, OpenTelemetry API for metrics>> might not be flexible enough.
+Some example use cases are:
+
+ * You want to use OpenTelemetry Views (e.g. to customize histogram buckets on a per-metric basis)
+ * You want to export metrics to other tools in addition to Elastic APM (e.g. prometheus)
```
SylvainJuge, line 182:
```suggestion
 * Using OpenTelemetry Views (e.g. to customize histogram buckets on a per-metric basis)
 * Exporting metrics to other tools in addition to Elastic APM (e.g. prometheus)
```

docs/api-opentelemetry.asciidoc
```diff
@@ -136,17 +136,106 @@ try (Scope scope = custom.makeCurrent()) {
...
+
+For these use cases you can just setup you OpenTelemetry SDK `MeterProvider`.
+The Elastic APM Agent will take care of installing an additional `MetricExporter` via instrumentation,
+which will ship the metric data to Elastic APM.
+This requires you to use OpenTelemetry version `1.16.0` or newer.
```
SylvainJuge, line 187:
```suggestion
This requires using OpenTelemetry version `1.16.0` or newer.
```
[question] : is the version constraint from our agent instrumentation or something that is not supported by older OTel versions ? If it is the latter, then I don't think it's worth mentioning it here (as it is already implicitly covered in the Otel docs).

docs/api-opentelemetry.asciidoc
```diff
@@ -136,17 +136,106 @@ try (Scope scope = custom.makeCurrent()) {
...
 [[otel-caveats]]
 ==== Caveats
 Not all features of the OpenTelemetry API are supported.
 
-[float]
-[[otel-metrics]]
-===== Metrics
-This bridge only supports the tracing API.
-The Metrics API is currently not supported.
-
```
SylvainJuge, line 149: :heart_eyes:

SylvainJuge (approved):
LGTM, I mostly have a few suggestions for wording.

There is also a lot of "you", so I would suggest simplifying a bit to avoid duplication, also that's probably bothering me only because doing the same in French would imply that we try to be friends with the end-user, so I'm definitely biased.

docs/api-opentelemetry.asciidoc
JonasKunz: Otel 1.16.0 added the capability for exporters to define the default aggregations to use (and temporality if I remember correctly). These features are used by our Otel Metric Exporter. For this reason the instrumentation ignores otel-SDKs which are older than 1.16.
For the API-based usage we support all opentelemetry versions (metrics were added in 1.10 )

JonasKunz:
Great suggestions, thanks!

PR #3046, opened 2023-03-06
Capture exit spans with annotations

### LINKED ISSUE #2696: [Feature] Extend @CreateSpan or create new annotation @CreateExitSpan
If we want a custom span to appear as a dependency on the APM server, we need to manually create it and manage it to the end.
This would be important if we want to show external services or other microservices as a dependency and have them on the dashboard showing the impact that they have.

Extend the existing @CaptureScan flag with a boolean attribute called `exit` defaulting to false. If we send them as true, it would create an exit span instead of a regular one. The other option would be creating a new annotation called @CaptureExitScan.

SylvainJuge:
## What does this PR do?

Fixes #2696

## Checklist

### DIFF
--- a/apm-agent-api/src/main/java/co/elastic/apm/api/CaptureSpan.java
+++ b/apm-agent-api/src/main/java/co/elastic/apm/api/CaptureSpan.java
@@ -87,4 +87,16 @@ public @interface CaptureSpan {
      * </p>
      */
     boolean discardable() default true;
+
+    /**
+     * <p>
+     * Makes the captured span an exit span. Exit spans are usually meant to represent outgoing calls to external systems, like databases.
+     * When an exit span is created no nested span is created. For example when creating exit spans
+     * for a database, it can hide the fact that calls to this database are made through an HTTP client.
+     * </p>
+     * <p>
+     * When set to {@literal true}, it is recommended to also set values for {@link #type()} and {@link #subtype()}
+     * </p>
+     */
+    boolean exit() default false;
 }
--- a/apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/api/SpanExitTest.java
+++ b/apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/api/SpanExitTest.java
@@ -1,0 +1,76 @@
+package co.elastic.apm.api;
+
+import co.elastic.apm.AbstractApiTest;
+import co.elastic.apm.agent.impl.transaction.Span;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static co.elastic.apm.agent.testutils.assertions.Assertions.assertThat;
+
+
+public class SpanExitTest extends AbstractApiTest {
+
+    @BeforeEach
+    void before(){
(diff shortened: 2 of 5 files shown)


apm-agent-plugins/apm-asynchttpclient-plugin/src/test/java/co/elastic/apm/agent/asynchttpclient/AsyncHttpClientInstrumentationTest.java
```diff
@@ -30,8 +30,8 @@
 import java.io.IOException;
 import java.util.Arrays;
 
+import static co.elastic.apm.agent.testutils.assertions.Assertions.assertThat;
 import static org.asynchttpclient.Dsl.asyncHttpClient;
-import static org.assertj.core.api.Assertions.assertThat;
```
SylvainJuge, line 34: [for reviewer] this is just test code refactoring allowed by the addition of `AbstractSpanAssert#isExit`.

apm-agent-api/src/main/java/co/elastic/apm/api/CaptureSpan.java
```diff
@@ -87,4 +87,16 @@
...
+     * <p>
+     * Makes the captured span an exit span. Exit spans are usually meant to represent outgoing calls to external systems, like databases.
+     * When an exit span is created no nested span is created. For example when creating exit spans
+     * for a database, it can hide the fact that calls to this database are made through an HTTP client.
+     * </p>
+     * <p>
+     * When set to {@literal true}, it is recommended to also set values for {@link #type()} and {@link #subtype()}
+     * </p>
+     */
+    boolean exit() default false;
```
jackshirazi, line 101: I think this should be something like `asExit`. `discardable` is something that is clearly true or false, but `exit` is a command to do something rather than a boolean state

jackshirazi: or maybe `makeExit`

apm-agent-plugins/apm-api-plugin/src/test/java/co/elastic/apm/api/SpanExitTest.java
```diff
@@ -0,0 +1,76 @@
...
+    @CaptureSpan(exit = true)
+    void captureExitSpan1(){
+
+    }
+
+    @CaptureSpan(exit = true, subtype = "my-database")
+    void captureExitSpan2(){
+
+    }
+
```
jackshirazi, line 75: what about if `@CaptureSpan` (so default of false is used) but the span from that method is an exit span anyway? Maybe that's handled in another already existing test?

jackshirazi: and definitely worth the similar `@CaptureSpan(exit = false)` but also where the span is an exit span

apm-agent-api/src/main/java/co/elastic/apm/api/CaptureSpan.java
SylvainJuge: Let's pick `asExit`, I've fixed that in latest commit.

PR #3053, opened 2023-03-09
Migrate specs update to updatecli

reakaleek:


.ci/update-json-specs.yml
```diff
@@ -0,0 +1,119 @@
...
+actions:
+  pr:
+    kind: "github/pullrequest"
+    scmid: default
+    spec:
+      automerge: false
+      draft: false
+      labels:
+        - "automation"
+      description: |-
```
v1v, line 63: What do you think adding the `title` field so the [name of the PR](https://www.updatecli.io/docs/plugins/actions/github/#_parameters) is something handled also here? See

.ci/update-gherkin-specs.yml
```diff
@@ -0,0 +1,115 @@
+name: update-gherkin-specs
```
v1v, line 1: It seems we could use the proposal for `updatecli.d` to benefit from the vscode integration:
- https://github.com/updatecli/website/blob/83c62785e57908d19db621574b74e1b1aebaf8ad/layouts/shortcodes/ide_jsonschema.html#L7

.github/workflows/updatecli.yml
```diff
@@ -11,19 +11,34 @@ permissions:
 jobs:
   bump:
     runs-on: ubuntu-latest
+    strategy:
```
v1v, line 14: We could use a folder instead:
- https://github.com/updatecli/updatecli/blob/f2bd316dec2460456a780428b6cab1c8f745bfc3/.github/workflows/updatecli.yaml#L36

Then there is no need for a matrix job.

Though, if one of them failed I don't know how this will be behave

reakaleek: I tried it with `diff`

```
Run Summary
===========
Pipeline(s) run:
  * Changed:    1
  * Failed:     0
  * Skipped:    0
  * Succeeded:  2
  * Total:      3
```

looking at the summary, I assume succesful pipeline won't be affected by failing ones.

reakaleek: https://github.com/elastic/apm-agent-java/pull/3053/commits/8b4db4c12c3d8d191836ce682938caf7f3e0ee51

.ci/update-gherkin-specs.yml
reakaleek: https://github.com/elastic/apm-agent-java/pull/3053/commits/8b4db4c12c3d8d191836ce682938caf7f3e0ee51

v1v (approved):

.github/workflows/updatecli.yml
v1v: Thanks for double checking this!

v1v (approved):

JonasKunz (approved):
Regarding the failing tests: Those were already fixed on `main`, so you should be good to merge after rebasing/merging the current `main` into this PR.

PR #3063, opened 2023-03-21
add the sources from apm-agent in to the final sources jar

jackshirazi:
## What does this PR do?
In 1.27.0 we restructured the subprojects so that all the main agent dependency management was moved from `elastic-apm-agent` to `apm-agent`, with no direct dependency on `apm-agent` from `elastic-apm-agent`. Instead it was made into a shaded dependency. This was to make the agent more isolated to reduce conflicts in classloading related to agent classes. An unnoticed side-effect of this was that the sources jar in `elastic-apm-agent` no longer contained the full agent sources. 

To fix this, I tried a number of different configurations. Any configuration that directly added the `apm-agent` as a dependency caused the class jar file to contain the classes as well as the shaded classes, so those configurations failed. The shading plugin should have been able to create all the sources of the shaded dependencies, but although I tried really hard I couldn't find the magic combo that produced that. So I've gone with the brute force approach of just merging the sources from the `apm-agent` sources jar into the `elastic-apm-agent` sources jar. This has the advantage of being very simple, but the disadvantage of directly depending on a source jar file existing in another project.

Additionally this PR fixes the missing `co.elastic.apm.agent.configuration` -> `co.elastic.apm.agent.premain.configuration` that @JonasKunz noticed

## Checklist
- [ ] This is a bugfix
- [ ] This is a new plugin
- [x] This is something else

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -42,6 +42,7 @@ communication - {pull}2996[#2996]
 * Fix context propagation for async 7.x and 8.x Elasticsearch clients - {pull}3015[#3015]
 * Fix exceptions filtering based on <<config-ignore-exceptions>> when those are <<config-unnest-exceptions, nested>> - {pull}3025[#3025]
 * Fix usage of `HttpUrlConnection.getResponseCode()` causing an error event due to exception capturing, even when it is internally handled - {pull}3024[#3024]
+* Fix source code jar to contain apm-agent sources - {pull}3063[#3063]
 
 [[release-notes-1.x]]
 === Java Agent version 1.x


elastic-apm-agent/pom.xml
```diff
@@ -158,6 +158,23 @@
...
                         </goals>
                     </execution>
+                    <execution>
+                        <id>combine-sources</id>
+                        <phase>package</phase>
+                        <configuration>
+
+                            <target name="merge-sources" description="Merge source jars">
+                                <zip destfile="${project.basedir}/target/${project.artifactId}-${project.version}-sources.jar">
+                                    <zipgroupfileset dir="${project.basedir}/../apm-agent/target/" includes="apm-agent-*-sources.jar"/>
```
JonasKunz, line 168: Looks good to me, I do however have an idea on how to get rid of the direct `apm-agent` file based reference:
* use the `maven-dependency-plugin` execution in this pom above to explicitly copy the apm-agent sources JAR into the target folder (using the [copy goal](https://maven.apache.org/plugins/maven-dependency-plugin/copy-mojo.html) )
* use the copied sources-jar from the target folder instead of the file path `${project.basedir}/../apm-agent/target/`

What do you think?

jackshirazi: I tried that a slightly different config, it requires the dependency to be declared as such which then would be included in the class files. Then I tried to exclude the dependencies after including them, then they don't show in the sources. So if I then do that and then copy the source jar, then do the merge it might work - but sounds like a very complex addition

jackshirazi: Actually I think I even tried it and couldn't get the sources if the classes were excluded, and couldn't prevent the classes from being the final class jar if they weren't even when I tried to exclude them also using the shade target. It was a really painful week

jackshirazi: nice change, thank you! Much better

PR #3071, opened 2023-03-23
ci: migrate release to stable branch

amannocci:
## What does this PR do?

* Migration of the release stable branch to GitHub Actions.
* The sync between main and stable branch will be handled by the GitHub Actions.

## Validation

Ref: https://github.com/elastic/apm-agent-java/actions/runs/4502205643/jobs/7923696347

### DIFF
--- a/.github/workflows/sync-branches.yml
+++ b/.github/workflows/sync-branches.yml
@@ -1,0 +1,26 @@
+name: Sync branches
+
+on:
+  push:
+    branches:
+      - feature/sync-branches
+
+jobs:
+  sync:
+    name: Sync branches
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Synchronize source to target branch
+        run: |
+          git checkout -f "${SOURCE_BRANCH}"
+          git checkout -f -b "${TARGET_BRANCH}"
+          git rebase "${SOURCE_BRANCH}"
+          git --no-pager log -n1 --pretty=oneline
+          git rev-parse --abbrev-ref HEAD
+          git push origin "${TARGET_BRANCH}"
+        env:
+          SOURCE_BRANCH: feature/sync-branches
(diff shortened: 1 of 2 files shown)


.github/workflows/sync-branches.yml
```diff
@@ -0,0 +1,26 @@
...
+  sync:
+    name: Sync branches
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Synchronize source to target branch
+        run: |
+          git checkout -f "${SOURCE_BRANCH}"
```
v1v, line 18: I somehow feel nervous to use env variables, even though they were already used in Jenkins.

v1v (commented):
I have not tested it out, but wonder if this works with the current GitHub token permissions, IIRC, it's read only based.

amannocci:
> I have not tested it out, but wonder if this works with the current GitHub token permissions, IIRC, it's read only based.

By default, a workflow has read/write permissions scope in permissive mode.
It should be working, though it could be worth using the restricted mode as default and granting permissions on demand.

.github/workflows/sync-branches.yml
v1v: That's fine.

Can you confirm whether this step worked in GH action with the [current default read permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)?

amannocci: AFAIS, we use the default permissive mode, which grants read/write permissions for all scopes.
I tested the workflow using a different source and target branch.
https://github.com/elastic/apm-agent-java/actions/runs/4502205643/jobs/7923696347

v1v: Cool, I'd say to put the explicit permissions just in case those are changed in the future.

v1v (approved):
Thanks

PR #3075, opened 2023-03-24
ci: correct UI trigger and remove useless workflow_call trigger

amannocci:
## What does this PR do?

* Remove useless [workflow_call](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_call) trigger.
* Use [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) instead of `workflow_call`.
* It allows triggering a workflow from the GitHub Actions UI.

## Screenshoot

Screenshot 2023-03-24 at 12 21 44

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,7 @@
 name: main
 
 on:
-  workflow_call:
+  workflow_dispatch:
     inputs:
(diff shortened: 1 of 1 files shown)


v1v (approved):
Thanks

PR #3076, opened 2023-03-24
Added dedicated java 8 build with latest log4j2 dependency.

### LINKED ISSUE #3058: Publish a Java 8 agent artifact with updated log4J
In order to fix common issues of users with security scanners complaining about our log4j version we should ship an additional agent artifact for Java 8+ which comes with an updated log4j version.
We will only provide this artifact via maven central, all other ways of adding the agent (e.g. CloudFoundry buildpack, docker base images) will still use the Java 7 artifact.

The easiest way of implementing this is most likely to add a new `elastic-apm-agent-java8` subproject. It might make sense to introduce a common parent project for `elastic-apm-agent-java8` and `elastic-apm-agent` to simplify dependency management.
In addition, it might make sense to add a manifest-entry via the corresponding pom.xml in order to correctly log the agent artifact on startup.

JonasKunz:
## What does this PR do?

Closes #3058 .

I decided to introduce a common parent project for `apm-agent` and `apm-agent-java8`, so that we can keep the plugin dependency management still nicely in one place.

For `elastic-apm-agent` and `elastic-apm-agent-java8` I instead decided to go the copy paste route (after extracting all premain code into a separate project), because a common parent project wouldn't have yielded too much benefit and we don't touch those poms often anyway. This entire PR is supposed to be reverted when we finally drop Java 7 support.

As usual, I'm terrible at naming things, so feel free to suggest better project names.

## Checklist
- [x] This is something else


elastic-apm-agent-java8/README.md
```diff
@@ -0,0 +1,4 @@
+# elastic-apm-agent-java8
+
+This project is a secondary build of `elastic-apm-agent` which ships the latest log4j2 version and therefore
```
jackshirazi, line 3:
```suggestion
This project is a secondary build of `elastic-apm-agent` which ships a log4j2 version that```

jackshirazi (approved):

PR #3080, opened 2023-03-28
Extract configuration from core module to API

raphw:
This change extracts a minimal configuration API from the *agent-core* module which is required by all plugins. It furthermore avoids that the matcher implementation is used by


raphw:
One minor issue I find with this change is the need to explicitly wrap `WildcardMatcher` to represent the more general `Matcher` type as the former cannot depend on the latter. This could be addressed by adding the *common* module as a dependency to the tracer-api module. This way, configuration interfaces could depend on `WildcardMatcher` directly.

PR #3086, opened 2023-03-30
Update broken agent download link

JonasKunz:
## What does this PR do?

Replace the broken agent download link on `1.x` branch with two links:
 * One direct download link for the latest version
 * A link to `MvnRepository` with an overview of older releases
 
PR to update `main` will follow.

### DIFF
--- a/docs/setup/get-agent.asciidoc
+++ b/docs/setup/get-agent.asciidoc
@@ -2,7 +2,8 @@
 
 Java agent releases are published to https://repo.maven.apache.org/maven2/[Maven central], in order to get a copy you can either:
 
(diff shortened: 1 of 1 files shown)


SylvainJuge:
I think you need to update on `main` first, then cherry-pick on `1.x` branch, as the `1.x` branch is force-pushed on release so merging it directly to `1.x` means it will be overwritten on release.

JonasKunz:
I'm planning on adding the same changes to main with a separate PR (will open later).
This way the force push on release won't revert these changes.

SylvainJuge (approved):

Issue #2942, opened 2023-01-01
Add support for Spring 6.x and Spring Boot 3.x


eyalkoren:
Spring 6 and Spring Boot 3 are GA, requiring us to do some changes in our instrumentations and create tests.
Following are major changes in those frameworks and the corresponding required changes in the agent:
- **Using Java 17 as their baseline** - #2936 proposes a framework to facilitate the work with Java 17 while still maintaining compatibility with Java 11 wherever applicable
- **Relying on Jakarta EE** - our Spring MVC instrumentations and much of our tests will have to adapt
- **API changes** - currently, the only known such change is the one exposed through our WebFlux instrumentation, as reported in #2935 

This issue is currently blocked on #2936. Once we know how we want our structure to facilitate Java 17 modules, we can unblock and prioritize this one.

Checklist:
- [ ] Remove [project-global version definitions](https://github.com/elastic/apm-agent-java/blob/main/pom.xml#L118-L119) 
- [ ] Adapt Spring MVC instrumentations - if all Spring class names and APIs remained the same, we may not be required to have a Java-17+ instrumentation module like we had to do with WebFlux, only adapt to use Jakarta EE APIs, for which we can hopefully reuse the [Servlet API adapters](https://github.com/elastic/apm-agent-java/tree/main/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/adapter). 
- [ ] Duplicate/adapt WebFlux plugin test (will have to run only on Java 17+). With #2935 I was able to upgrade dependencies locally and run the tests.
- [ ] Duplicate/adapt Spring MVC plugin tests (will have to run only on Java 17+). I was able to upgrade dependencies locally and with some `javax`-->`jakarta` changes successfully run `SpringTransactionNameInstrumentationTest`.
- [ ] Adjust Spring RestTemplate tests to test multiple versions, including Spring 6.x (may require adding proper testing for `org.apache.httpcomponents.client5.httpclient5` as well). Tests mostly pass by upgrading to Spring 6 and `httpclient` to `httpclient5` 
- [ ] Add a Spring Boot 3.x integration test module
- [ ] Update supported technologies page
- [ ] Once all the above are merged, open an issue to upgrade [opbeans](https://github.com/elastic/opbeans-java/blob/main/opbeans/pom.xml#L24) to Spring Boot 3.x

JonasKunz:
[...]

PR #3094, opened 2023-04-03
Added spring 6 webmvc support

JonasKunz:
## What does this PR do?

Adds support and tests for spring 6 webmvc (Part of #2942).
We continue to test with both spring 5 and 6. The instrumentation itself did not require changes except for the `javax` -> `jakarta` package name change. The instrumentation has been adapted to rely on our existing Servlet-API adapters.

## Checklist


apm-agent-plugins/apm-spring-webmvc/apm-spring-webmvc-spring5/src/main/java/co/elastic/apm/agent/springwebmvc/Spring5ExceptionHandlerInstrumentation.java
```diff
@@ -0,0 +1,44 @@
...
+    public Constants.ServletImpl servletImpl() {
+        return Constants.ServletImpl.JAVAX;
+    }
+
+    public static class AdviceClass extends AbstractSpringExceptionHandlerInstrumentation.AdviceClass {
+
+        @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
+        public static void captureException(@Advice.Argument(0) @Nullable HttpServletRequest request,
+                                            @Advice.Argument(3) @Nullable Exception e) {
+            onEnter(JavaxServletApiAdapter.get(), request, e);
```
SylvainJuge, line 40: [minor] given the adapter is only used to set an attribute which is only a single line, we could probably make this simpler by directly calling the adapter and avoid the advice inheritance and indirection.

apm-agent-plugins/apm-spring-webmvc/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/springwebmvc/Spring6ServiceNameInstrumentation.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.web.context.WebApplicationContext;
+
+public class Spring6ServiceNameInstrumentation extends AbstractSpringServiceNameInstrumentation {
+
+    @Override
+    public Constants.ServletImpl servletImpl() {
+        return Constants.ServletImpl.JAKARTA;
+    }
+
+    public static class AdviceClass extends AbstractSpringServiceNameInstrumentation.AdviceClass {
```
SylvainJuge, line 33: [minor] if we attempt to not use advice class inheritance, then maybe moving the `onExit` implementation directly to the adapter could be an option (or use an helper class). While the advice inheritance is definitely possible and I have nothing against it, it's not a pattern widely used, also we have to open the actual `onExit` implementation to know what it does, whereas a call to an helper class method usually have a more descriptive name.

apm-agent-plugins/apm-spring-webmvc/apm-spring-webmvc-plugin/src/test/java/co/elastic/apm/agent/springwebmvc/template/Spring6ThymeleafTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.web.servlet.config.annotation.EnableWebMvc;
+import org.thymeleaf.spring6.SpringTemplateEngine;
+import org.thymeleaf.spring6.templateresolver.SpringResourceTemplateResolver;
+import org.thymeleaf.spring6.view.ThymeleafViewResolver;
+
+public class Spring6ThymeleafTest extends Java17OnlyTest {
+    public Spring6ThymeleafTest() {
+        super(AbstractThymeleafTest.class);
```
SylvainJuge, line 33: Why not `Impl.class` here ?

apm-agent-plugins/apm-spring-webmvc/apm-spring-webmvc-spring5/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+        <artifactId>apm-spring-webmvc</artifactId>
+        <groupId>co.elastic.apm</groupId>
+        <version>1.37.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>apm-spring-webmvc-spring5</artifactId>
+    <name>${project.groupId}:${project.artifactId}</name>
+
+    <properties>
+        <animal.sniffer.skip>false</animal.sniffer.skip>
```
SylvainJuge, line 16: [minor] I think it's enabled by default, and most plugins that need to disable opt-out, thus setting it to `false` is probably not needed.

apm-agent-plugins/apm-spring-webmvc/apm-spring-webmvc-spring5/src/main/java/co/elastic/apm/agent/springwebmvc/ViewRenderInstrumentation.java
```diff
@@ -134,10 +133,16 @@ public ElementMatcher<? super TypeDescription> getTypeMatcher() {
 
     @Override
     public ElementMatcher<? super MethodDescription> getMethodMatcher() {
+
+        ElementMatcher.Junction<MethodDescription> javaxMatcher = takesArgument(1, named("javax.servlet.http.HttpServletRequest"))
+            .and(takesArgument(2, named("javax.servlet.http.HttpServletResponse")));
+        ElementMatcher.Junction<MethodDescription> jakartaMatcher =
+            takesArgument(1, named("jakarta.servlet.http.HttpServletRequest"))
+                .and(takesArgument(2, named("jakarta.servlet.http.HttpServletResponse")));
+
```
SylvainJuge, line 142: [minor] can we use both implementations of the servlet adapter here ?

apm-agent-plugins/apm-spring-webmvc/apm-spring-webmvc-spring5/src/main/java/co/elastic/apm/agent/springwebmvc/Spring5ExceptionHandlerInstrumentation.java
SylvainJuge: Good for me with the helper + not inline the implementation, we get the helper+method name to provide relevant naming which helps to understand once I'll have forgotten about this conversation :-).

apm-agent-plugins/apm-spring-webmvc/apm-spring-webmvc-plugin/src/main/java/co/elastic/apm/agent/springwebmvc/Spring6ServiceNameInstrumentation.java
JonasKunz: > then maybe moving the onExit implementation directly to the adapter could be an option

I would prefer to not that do that because it contains spring-specific code (the `application.name` lookup).

> While the advice inheritance is definitely possible and I have nothing against it, it's not a pattern widely used

That's true, I used our servlet-api as template here.

> whereas a call to an helper class method usually have a more descriptive name

I will switch to a static helper method here aswell.

apm-agent-plugins/apm-spring-webmvc/apm-spring-webmvc-plugin/src/test/java/co/elastic/apm/agent/springwebmvc/template/Spring6ThymeleafTest.java
JonasKunz: Good catch, that was a bug. The test didn't actually execute anything here.

apm-agent-plugins/apm-spring-webmvc/apm-spring-webmvc-spring5/pom.xml
JonasKunz: It was skipped via the parent-pom (`spring-webmc`), that's why it was reenabled here. I removed the setting from the parent and moved it to the leaf projects to make it less confusing.

SylvainJuge (approved):

JonasKunz:
> Our team are watching this PR with hawk eyes. After the merge, I assume it is now safe to ask when the artifact will be available to use.
>
> PS. We were about to go to production with our system's Spring / Java upgrade with or without Elastic APM.

@arnoe we don't have a fixed release schedule, but we still do have a few more changes in our pipeline which we want to include in the next release. Therefore, it might take a few weeks until we publish the next release.

If it is this urgent for you, you can already use the [snapshot form the PR build](https://github.com/elastic/apm-agent-java/suites/12287683160/artifacts/651056857) and upgrade to the next release as soon as it is available. This would also help us identify potential issues not uncovered by our own test prior to a release.

Normally you can also use the [snapshot from the main branch](https://github.com/elastic/apm-agent-java#snapshots) in such cases. However, the daily pipeline run failed the last time, therefore that snapshot is not up to date yet.

arnoe:
Okay, thanks @JonasKunz for clarifying the expectations on the release.

We're taking a look at the snapshot for now. Would feedback here be helpful?

arnoe:
Okay, thanks for clarifying the expectations on the release.

We'll take a look at the snapshot. Would it help if we provide feedback about the snapshot version - if so, where would you prefer to get this feedback.

JonasKunz:
> Would feedback here be helpful?

Feedback is always welcome! If you discover any issues with the spring 6 support the best place to report it would be via comments in the corresponding issue: #2942.

PR #3105, opened 2023-04-11
Generate sources for tracer project, force agent javadoc

JonasKunz:
## What does this PR do?
 * Enables generation of the sources-jar for the `tracer` project (which we publish on maven central)
 * Force the generation of a javadoc-jar for the `elastic-apm-agent` / `elastic-apm-agent-java8` projects.
   * Javadoc was not generated because the projects contained no code anymore
   * To workaround this, a package-info.java and a dummy class were added to both projects

### DIFF
--- a/elastic-apm-agent-java8/src/main/java/apm/package-info.java
+++ b/elastic-apm-agent-java8/src/main/java/apm/package-info.java
@@ -1,0 +1,22 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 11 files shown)


elastic-apm-agent-java8/src/main/java/apm/package-info.java
```diff
@@ -0,0 +1,22 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/**
+ * This project contains the Java 8 build of the elastic apm java agent.
+ */
+package apm;
```
SylvainJuge, line 22: [question] why not use a more "traditional" package like `co.elastic.apm.java8` so it does not conflicts with the other `JavadocPlaceHolder.java` and still fits a bit the other packages even if we don't really publish it ?

SylvainJuge (approved):

elastic-apm-agent-java8/src/main/java/apm/package-info.java
JonasKunz: I decided to go with `co.elastic.apm.premain.java7/java8` to be a subpackage of the other non-shaded premain classes.

PR #3118, opened 2023-04-20
Refactor testcontainer for easier integration tests

SylvainJuge:
## What does this PR do?
- bypass javadoc for integration tests: not strictly required but saves a bit of build time
- make agent test container reusable
- remove testcontainers version, provided by bom

## Checklist

- [x] This is something else

### DIFF
--- a/apm-agent-common/src/test/java/co/elastic/apm/agent/test/AgentTestContainer.java
+++ b/apm-agent-common/src/test/java/co/elastic/apm/agent/test/AgentTestContainer.java
@@ -143,7 +143,7 @@ public abstract class AgentTestContainer<SELF extends GenericContainer<SELF>> ex
     }
 
     /**
-     * Sets the environment variable that will be used to set the '-javaagent' or other JVM arguments
+     * Sets the environment variable that will be used to set the '-javaagent', JVM System properties and remote debug
      *
      * @param name environment variable name
      * @return this


integration-tests/main-app-test/src/test/java/co/elastic/apm/test/TestAppContainer.java
```diff
@@ -1,214 +0,0 @@
...
-        return this;
-    }
-
-    /**
-     * Configures remote debugging automatically for the JVM running in the container.
-     * On the IDE side, all is required is to add debugger listening for incoming connections on port 5005
-     *
-     * @return this
-     */
-    public TestAppContainer withRemoteDebug() {
```
JonasKunz, line 118: Could you maybe extract this functionality to a superclass, which delegates the actual addition of the remote debug args to the subclass?
I'm thinking of the lambda integration test I work on. We could enable remote debug there via `JAVA_TOOL_OPTS` I think, but this can't use this class.

SylvainJuge: Why a superclass ? We could very probably move this implementation to an independant class so it's easier to reuse, even without testcontainers. Would that work for the use-case you are mentioning with lambda integration test ?

JonasKunz: > Why a superclass?
I was just thinking because the method currently invokes `withExtraHost`. But a static utility method / class would do the trick aswell.

SylvainJuge: discussed this directly, using `withJvmArgumentsVariable` like in `co.elastic.apm.agent.tomcatlogging.TomcatLoggingIT#testLogShading` with `"JAVA_TOOL_OPTIONS"` argument should work with AWS lambda.

apm-agent-plugins/apm-jdbc-plugin/pom.xml
```diff
@@ -27,44 +27,37 @@
         <dependency>
             <groupId>org.testcontainers</groupId>
             <artifactId>jdbc</artifactId>
-            <version>${version.testcontainers}</version>
```
SylvainJuge, line 30: [for reviewer] this is not required as we are using a bom import

integration-tests/main-app-test/src/test/java/co/elastic/apm/test/AgentSetupIT.java
```diff
@@ -61,9 +62,9 @@ void testServiceNameAndVersionFromManifest(String image, AgentFileAccessor.Varia
     void testSecurityManagerWarning() {
         String expectedMsg = "Security manager without agent grant-all permission";
 
-        try (TestAppContainer app = testAppWithJavaAgent("openjdk:17", AgentFileAccessor.Variant.STANDARD)) {
+        try (AgentTestContainer.Generic app = testAppWithJavaAgent("openjdk:17", AgentFileAccessor.Variant.STANDARD)) {
 
-            app.withSecurityManager()
+            app.withSecurityManager(null)
```
SylvainJuge, line 67: [for reviewer] using `null` here is a mean to indicate that we just enable the SM with the default policy.

PR #3120, opened 2023-04-24
Added AWS Lambda integration test

JonasKunz:
## What does this PR do?

Adds a basic integration test for AWS Lambda based on the [AWS Lambda Runtime Interface Emulator](https://github.com/aws/aws-lambda-runtime-interface-emulator).
The test uses the real aws lambda extension layer and connects to a fake apm server, which records the received events.

## Checklist

- [x] This is something else


integration-tests/aws-lambda-test/src/main/java/co/elastic/test/TestLambda.java
```diff
@@ -0,0 +1,46 @@
...
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+
+public class TestLambda implements RequestHandler<String, String> {
+    @Override
+    public String handleRequest(String command, Context context) {
+        ElasticApm.currentTransaction().addCustomContext("command", command);
+        if (command.startsWith("sleep ")) {
+            long millis = Long.parseLong(command.substring("sleep ".length()));
+            doSleep(millis);
```
jackshirazi, line 31: feels like this parse needs error handling for misconfigured comands

jackshirazi (approved):
I love that you've added this, excellently done!

integration-tests/aws-lambda-test/src/main/java/co/elastic/test/TestLambda.java
JonasKunz: What would an ideal solution look like for you?
E.g. a lambda with exactly the following behaviour regarding commands:
 * `greet XYZ` -> returns "Hello XYZ".
 * `sleep A` -> sleeps "A" amount of millis, returns empty string
 * `die` -> kills the Lambda JVM
 * everything else -> throws an error

jackshirazi: yes, why not tie it down tight, it's testing so we want to see anything odd

PR #3147, opened 2023-05-23
Enhance JMX metrics

SylvainJuge:
## What does this PR do?

- Adds capture of `non_heap` memory pool details: `used`, `commited` and `max` (so far only the `heap` ones were included).
- Adds capture of open file descriptors with `jvm.fd.used` and `jvm.fd.max`
- Include minor refactorings from #2712 

## Checklist

  - [x] ~~Added an API method or config option? Document in which version this will be introduced~~ N/A


apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/JvmFdMetrics.java
```diff
@@ -0,0 +1,94 @@
...
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.lang.management.ManagementFactory;
+import java.lang.management.OperatingSystemMXBean;
+
+public class JvmFdMetrics extends AbstractLifecycleListener {
+
+    private static final MethodHandle NOOP = MethodHandles.constant(String.class, "no-op");
+
+    private static final MethodHandle getOpenFileDescriptorCount = getMethodHandle("getOpenFileDescriptorCount");
```
SylvainJuge, line 38: [for reviewer] we have to use method handle to avoid direct reference to `com.sun.*` classes.

jackshirazi: worth adding a comment on that

jackshirazi:
Is there are backward compatibility break from this? Or are we saying that metrics (other than the semconv ones that will come soon) don't have to be there. But is the UI expecting any specific ones?

SylvainJuge:
> Is there are backward compatibility break from this? Or are we saying that metrics (other than the semconv ones that will come soon) don't have to be there. But is the UI expecting any specific ones?

The UI is not expecting those new metrics, I just added them because that's quite easy to do (and it was in the PoC PR), also there isn't any compatibility issue here as we just add new ones.

Those would definitely be replaced by the semconv ones in the future.

apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/builtin/JvmFdMetrics.java
SylvainJuge: fixed in cbc57c572e576a16dee1f50cd60f6379f895ffc2

PR #3148, opened 2023-05-24
Spring cleaning

SylvainJuge:
## What does this PR do?

- initial intent is to remove global spring versions as described in #2942 (see below)
- cleanup quartz plugin: rename modules for consistency + remove common quartz module dependency on quartz2.
- remove spring-boot version property (only used for packaging a test app).
- remove unused `version.logback`

The `spring.version` in the main `pom.xml` is used for two things:
- in test applications like benchmarks, where it's meant to be the current latest version of spring framework, in this case using any version will usually do and keeping it to the latest is fine
- as a "secondary" test dependency, for example in the quartz plugin, there is a spring-based integration that we need to test. In theory we should test for all spring-versions, but from experience the latest is usually fine.

There are a few other version properties that are used in a similar way:
- `version.jetty-server`
- `version.okhttp` (this one is also used outside test code, defining it in instrumentation plugin might be better than globally).
- `version.h2`

Having a single property here makes updating and reviewing dependabot PRs easier, so it might be relevant to keep that.
Maybe it could be enough to add a small comment in the `pom.xml` to make it more obvious that those are not actual dependencies and only used for testing outside of production code. 

## Checklist

- [ ] Decide on which approach to take to remove `spring.version` property and similar ones (see above)
- [x] This is something else
  - [x] ~~I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)~~ N/A


apm-agent-builds/pom.xml
```diff
@@ -284,12 +284,12 @@
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
-            <artifactId>apm-quartz-1-plugin</artifactId>
+            <artifactId>apm-quartz-plugin-1</artifactId>
```
SylvainJuge, line 287: [for reviewer] using same naming convention as grpc

apm-agent-plugins/apm-quartz/apm-quartz-common/pom.xml
```diff
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <parent>
-        <artifactId>apm-quartz-job-plugin</artifactId>
+        <artifactId>apm-quartz</artifactId>
```
SylvainJuge, line 4: [for reviewer] using `apm-quartz` as module name for all the other sub-modules, the `-plugin` sufix is kept for the instrumentation plugins (with an extra per-version suffix to fit the naming conventions).

apm-agent-plugins/apm-quartz/apm-quartz-common/pom.xml
```diff
@@ -1,17 +1,17 @@
...
         <version>1.38.1-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
-    <artifactId>quartz-common</artifactId>
+    <artifactId>apm-quartz-common</artifactId>
     <name>${project.groupId}:${project.artifactId}</name>
 
     <properties>
-        <version.quartz>2.3.1</version.quartz>
```
SylvainJuge, line 14: [for reviewer] the common part was compiled with quartz 2.x and contained some code that was only used with 2.x plugin, however making sure the common part is compiled with 1.x only is safer.

apm-agent-plugins/apm-quartz/apm-quartz-common/pom.xml
```diff
@@ -22,19 +22,6 @@
...
-            <version>${version.spring}</version>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>org.quartz-scheduler</groupId>
-            <artifactId>quartz-jobs</artifactId>
-            <version>${version.quartz}</version>
-            <scope>test</scope>
-        </dependency>
```
SylvainJuge, line 37: [for reviewer] those were only required for quartz 2.x

apm-agent-plugins/apm-quartz/apm-quartz-common/src/test/java/co/elastic/apm/agent/quartzjob/AbstractJobTransactionNameInstrumentationTest.java
```diff
@@ -124,24 +107,6 @@ void testJobWithResult() throws SchedulerException {
         assertThat(transaction.getResult()).isEqualTo("this is the result");
     }
 
-    @Test
-    void testDirectoryScan() throws SchedulerException, IOException {
```
SylvainJuge, line 128: [for reviewer] this was only relevant for quartz 2.x (hence the `Assumptions` call at start of test method.

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-testapp/pom.xml
```diff
@@ -96,7 +96,7 @@
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-maven-plugin</artifactId>
                 <!-- explicit version required as bom will not cover plugins, only dependencies -->
-                <version>${version.spring-boot}</version>
```
SylvainJuge, line 99: [for reviewer] the global `spring-boot` version property was removed because only used here in test application, thus in test code.

pom.xml
```diff
@@ -118,7 +118,6 @@
         <version.jackson>2.14.2</version.jackson>
         <version.junit>5.9.1</version.junit>
         <version.junit.vintage>4.13.2</version.junit.vintage>
-        <version.logback>1.2.3</version.logback>
```
SylvainJuge, line 121: [for reviewer] this wasn't used anymore.

JonasKunz (approved):

PR #3151, opened 2023-05-26
Prevent overhead from excessive span creation

### LINKED ISSUE #2788: Restrict the number of spans created through `trace_methods` misconfiguration
We saw lots of cases where the `trace_methods` configuration was abused to capture way-too-many methods using wildcards on full packages. In other cases, the problem is not with the amount of methods being captured, but with capturing a low-latency method that is very frequently being executed.
The only safety measure we currently employ to deal with that is the limitation of spans per transaction, however this limitation still doesn't prevent additional spans from being created (only they will be dropped), because we still need them for breakdown metrics calculations and for context propagation. 

So what we can do is simply use the `co.elastic.apm.agent.impl.transaction.Transaction#isSpanLimitReached()` within the `TraceMethodInstrumentation.TraceMethodAdvice` and if the limit is reached, simply return `null` instead of creating a span.

Possibly even better, we can stop creating `trace_methods`-related spans before the limit is reached, for example when 90% of the limit is reached, so that we won't drop the really "interesting" spans.

SylvainJuge:
## What does this PR do?

Fixes #2788 by skipping span creation when the max span limit is already reached within the active transaction.

This is applied to the following instrumentations:
- `trace_methods` set in configuration
- `@Traced` annotated methods
- `@CaptureSpan` annotated methods

While it does not prevent users from mis-using configuration or annotations, it removes most of the agent overhead by skipping the span creation.

The global number of dropped spans is captured in the transaction, however the dropped span metrics are not updated because they rely on the span being created (and are updated when the span ends).

As a consequence the dropped spans metrics might appear inconsistent, but I think it's safe to assume that inconsistency is a small price to pay here given it helps to lower the overhead in case of an agent mis-configuration.

## Checklist

  - [x] ~~Added an API method or config option? Document in which version this will be introduced~~ N/A
  - [x] ~~I have made corresponding changes to the documentation~~ N/A

### Performance improvements

10 millions very short spans created within a single transaction:
- the bottom half transactions with current version of the agent: transaction takes about ~7seconds
- the top half is with this change applied: transaction takes about 200~400ms

Screenshot from 2023-05-26 16-23-32

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/tracemethods/TraceMethodInstrumentation.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/tracemethods/TraceMethodInstrumentation.java
@@ -22,6 +22,8 @@ import co.elastic.apm.agent.bci.TracerAwareInstrumentation;
 import co.elastic.apm.agent.bci.bytebuddy.SimpleMethodSignatureOffsetMappingFactory;
 import co.elastic.apm.agent.configuration.CoreConfiguration;
 import co.elastic.apm.agent.impl.ElasticApmTracer;
+import co.elastic.apm.agent.sdk.logging.Logger;
+import co.elastic.apm.agent.sdk.logging.LoggerFactory;
 import co.elastic.apm.agent.tracer.GlobalTracer;
 import co.elastic.apm.agent.impl.transaction.AbstractSpan;
 import co.elastic.apm.agent.impl.transaction.Span;
@@ -126,6 +128,8 @@ public class TraceMethodInstrumentation extends TracerAwareInstrumentation {
 
     public static class TraceMethodAdvice {
 
+        public static final Logger logger = LoggerFactory.getLogger(TraceMethodAdvice.class);
+
         private static final ElasticApmTracer tracer = GlobalTracer.get().require(ElasticApmTracer.class);
         private static final long traceMethodThresholdMicros;
 
@@ -148,6 +152,7 @@ public class TraceMethodInstrumentation extends TracerAwareInstrumentation {
             } else if (parent.isSampled()) {
                 if (parent.shouldSkipChildSpanCreation()) {
                     // span limit reached means span will not be reported, thus we can optimize and skip creating one
+                    logger.debug("Not creating span for {} because span limit is reached.", signature);
                     return null;
                 }
                 span = parent.createSpan()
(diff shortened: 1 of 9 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/tracemethods/TraceMethodInstrumentation.java
```diff
@@ -146,6 +146,10 @@ public static Object onMethodEnter(@Advice.Origin Class<?> clazz,
                     span.withName(signature).activate();
                 }
             } else if (parent.isSampled()) {
+                if (parent.shouldSkipChildSpanCreation()) {
+                    // span limit reached means span will not be reported, thus we can optimize and skip creating one
+                    return null;
```
JonasKunz, line 151: You have a debug-log for this case in the other instrumentation, I think it would make sense to add a debug log here aswell.

apm-agent-plugins/apm-api-plugin/src/main/java/co/elastic/apm/agent/pluginapi/TracedInstrumentation.java
```diff
@@ -90,6 +90,10 @@ public static Object onMethodEnter(
 
             final AbstractSpan<?> parent = tracer.getActive();
             if (parent != null) {
+                if (parent.shouldSkipChildSpanCreation()) {
+                    // span limit reached means span will not be reported, thus we can optimize and skip creating one
```
JonasKunz, line 94: Add a debug log here aswell.

JonasKunz (approved):

apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/AbstractSpan.java
```diff
@@ -184,4 +184,17 @@
...
+     * is expected to call this method before every span creation operation for proper dropped spans accounting. If not
+     * called before attempting span creation, a span will be created and dropped before reporting.
+     * <br/>
+     * Expected caller behavior depends on the returned value:
+     * <ul>
+     *     <li>{@literal true} returned means the caller is expected to NOT call {@link #createSpan()} or {@link #createExitSpan()}</li>
+     *     <li>{@literal false} returned means the caller MAY call {@link #createSpan()} or {@link #createExitSpan()}</li>
+     * </ul>
+     */
+    boolean shouldSkipChildSpanCreation();
```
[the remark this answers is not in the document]
SylvainJuge: That's exactly why I took this simple approach here, and it would have been quite consistent with `createExitSpan` that is expected to return `null` in case there is already an active exit span.

Another alternative would be to have a way to return a no-op span, which could avoid most of the refactoring but could still have some unexpected side-effects.

I think it could be worth investing a bit of time for a refactoring/tech-debt PR.

PR #3154, opened 2023-05-31
Fix external plugin OpenTelemetry instrumentation

### LINKED ISSUE #3145: OpenTelemetry regression in v1.37.0 for external plugins
## Describe the bug

This was initially reported [in the forum](https://discuss.elastic.co/t/after-updating-java-apm-agent-from-1-36-to-1-37-context-propagation-in-reactor-is-no-longer-working-as-expected/332810/5).

The external plugins can use the OpenTelemetry API for tracing, when doing so, they have a shaded copy of the OpenTelemetry API within the agent jar.
The agent OpenTelemetry plugin then instruments this OpenTelemetry API copy to provide its own implementation.

However, there was a change between v1.36.0 and v1.37.0 that introduced a regression here: the otel plugin does not seem to instrument the otel api within an external plugin.

The commit that introduced the regression is `be93978c1`, which is ~4k lines and does relates to OpenTelemetry as it adds support for OpenTelemetry metrics (#3014).

## How to reproduce

- use a sample application using r2dbc (I don't think this is specific with r2dbc, but I did not manage to reproduce with OTel only).
- https://github.com/bezkoder/spring-data-r2dbc-postgresql-example
- configure the agent with minimal external plugin + minimal instrumentation for easier debug:
```bash
-Delastic.apm.enable_instrumentations=opentelemetry,r2dbc
-Delastic.apm.plugins_dir=<path-to-external-plugin-jar-folder>
```
- use a local postgres container for database
```bash
docker run --name postgresql \
-p 5432:5432 \
-e POSTGRES_DB=testdb \
-e POSTGRES_PASSWORD=123 \
-e POSTGRES_USER=postgres \
--rm \
postgres
```
- create an external plugin that have the following instrumentation:

```java
public class R2dbcPostgresqlInstrumentation extends ElasticApmInstrumentation {

    @Override
        return named("io.r2dbc.postgresql.PostgresqlStatement");
    }

    @Override
        return named("execute").and(takesArgument(0, named("java.lang.String")));
    }

    @Override
    public Collection<String> getInstrumentationGroupNames() {
        return Collections.singletonList("r2dbc");
    }

    public static class AdviceClass {

        @Advice.OnMethodExit(inline = false)
        public static void onExitExecute() {
            String implName = GlobalOpenTelemetry.get().getTracerProvider().getClass().getName();
            if (implName.equals("io.opentelemetry.api.trace.DefaultTracerProvider")) {
                throw new IllegalStateException("unexpected tracer implementation");
            }

        }

    }
}
```

With v1.36.0, the application starts, with v1.37.0, the exception is thrown and the application fails to start.

JonasKunz:
## What does this PR do?

Fixes #3145 .

The GlobalOpenTelemetry instrumentation was broken for external plugins due to a sideeffect of hiding the internal Otel-API shipped within the agent classloader from the `IndyPluginClassloader`:
 * External Plugin `IndyPluginClassloader`s contain a copy of *all* classes of the external plugin jar. This means also `GlobalOpenTelemetry` is copied to the `IndyPluginClassloader`.
 * Our instrumentation for `GlobalOpenTelemetry` has a classloader check ensuring that the target classloader can actually load the Otel-API. This is done by trying to load the `io.opentelemetry.context.propagation.TextMapSetter.class` resource
 * `IndyPluginClassloaders` cannot locate the resources for their corresponding classes and therefore delegate this search to their parent classloader.
 * The parent classloader filters out classes and resources which are supposed to be hidden from the plugin. With #3014  otel classes were added to this list.
 * For this reason the instrumentation thinks that the copied `GlobalOpenTelemetry` is not instrumentable.

## Checklist

- [x] This is a bugfix


integration-tests/external-plugin-otel-test/external-plugin-otel-test-app/pom.xml
```diff
@@ -44,10 +44,20 @@
     <build>
         <finalName>${project.artifactId}</finalName>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
SylvainJuge, line 48: [nit] default maven plugin `groupId` can be omitted.

Is there any particular reason the jar plugin is required here ? For example to run it in the IDE ?
The packaged version is provided with dependencies with `maven-assembly-plugin` below.

integration-tests/external-plugin-otel-test/external-plugin-otel-test-app/src/main/java/testapp/Main.java
```diff
@@ -48,11 +50,27 @@ private static void span() {
...
+                @Override
+                public void run() {
+                    SpanContext spanContext = Span.current().getSpanContext();
+                    if (!spanContext.isValid()) {
+                        System.out.println("no active OTel Span");
+                    } else {
+                        System.out.printf("active span ID = %s, trace ID = %s%n", spanContext.getSpanId(), spanContext.getTraceId());
+                    }
+                }
+            }.run();
```
SylvainJuge, line 64: Does wrapping into an anonymous class allows to not load it directly from the parent class ? If so then maybe it would be worth adding a small comment for the futures ourselves that might wonder why.

SylvainJuge (approved):

integration-tests/external-plugin-otel-test/external-plugin-otel-test-app/pom.xml
JonasKunz: The integration tests have been adapted to use two versions of this app: One containing the Otel-dependencies (`jar-with-dependencies`) and one without (the "standard" jar built by the maven-jar-plugin).

The entry for the `maven-jar-plugin` is required to add the `Main-Class` entry to the Jar without the dependencies.

integration-tests/external-plugin-otel-test/external-plugin-otel-test-app/src/main/java/testapp/Main.java
JonasKunz: It is not possible to load a class which contains references to classes which are not on the classpath.
So if we would run this class without the `Runnable`-wrapping and without Otel on the classpath, the class would fail to load. This is prevent by moving the Otel-calls to another class (anonymous in this case). This anonymous class is never loaded when we run without Otel because we first make sure that Otel is actually on the classpath. This way, it is possible to run the App without problems with and without Otel on the classpath.

I'll add a comment to explain this behaviour.

Issue #3152, opened 2023-05-29
agent 1.38.0 having issue with cloud gateway project having spring boot 3.1.0 and spring cloud 2022.0.3


clockrun:
## Describe the bug
By upgrading to agent version 1.38.0, I have connection issue on our gateway (Using spring cloud gateway). From logs, I can see it caused by stackoverflow during request processing. After downgrade back to 1.36.0, issue is gone.

BTW, I have several other spring cloud gateway, and they don't have stackoverflow issue after upgrade to 1.38.0. So I guess this bug require some special configuration to reproduce. Unfortunately, I cannot narrow down to specific confiugration which result in this issue.

## Steps to reproduce
using spring-boot 3.1.0 and spring cloud 2022.0.3, with spring cloud gateway

## Expected behavior
connection can be built from client to backend service through cloud gateway

## Debug logs
see attachment

[stackoverflow.txt](https://github.com/elastic/apm-agent-java/files/11590358/stackoverflow.txt)

JonasKunz:
Hi @clockrun, thanks for reporting this!

To me this seems like a side effect of #2993. Before this change we would catch and print exceptions occuring within our instrumentation to stdout. This has been altered to redirect the exception printing into the logging instead.

It looks like in your case the instrumentation was running into a `StackOverflowError`. We then try to print this `StackOverflowError` to our logs, which again caused the stack to overflow, but this time the exception was not caught.

I would ask you to try out two things please:
 * Please verify that the same problem occurs with agent version 1.37.0. This helps us exclude that the root cause of the StackOverflow is related to our spring changes in 1.38.0
 * Please try out the [snapshot build](https://github.com/elastic/apm-agent-java/suites/13290862742/artifacts/725555905) from this [PR](https://github.com/elastic/apm-agent-java/pull/3159). This should ensure that the exception from the agent code does not propagate into your code. Nonetheless, it would be great if you could watch the logs to see whether there are any exceptions, even though they should not cause your requests to fail anymore.

clockrun:
Hi JonasKunz,
Yes, same issue can be reproduced with version 1.37.0
Unfortunately I have similar issue after using the snapshot build. Here is lastest error message:
[errorlog.txt](https://github.com/elastic/apm-agent-java/files/11637452/errorlog.txt)

clockrun:
sorry for the late reply.
After upgrade to apm agent 1.42.0 and spring boot 3.1.2, this issue is gone.

PR #3159, opened 2023-06-01
Protect against stackoverflow when printing advice exceptions

JonasKunz:
## What does this PR do?

Tries to mitigate #3152 .

## Checklist

- [x] This is a bugfix
  - [ ] ~I have added tests that would fail without this fix~ not feasible to produce a StackOverflow in a test while keeping the system in a known state.


apm-agent-bootstrap/src/main/java/bootstrap/dispatcher/IndyBootstrapDispatcher.java
```diff
@@ -91,7 +91,11 @@ public static void logAdviceException(Throwable exception) {
                 printStackTrace(exception);
             }
         } catch (Throwable t) {
-            printStackTrace(t);
+            try {
+                printStackTrace(t);
+            } catch (Exception e) {
```
jackshirazi, line 96:
```suggestion
            } catch (Throwable e) {
```
`StackOverflowError` is an error not exception, so needs to be `Throwable'

JonasKunz: Good catch! (pun intended)

jackshirazi (approved):
Fine without a test but, if you feel like it, you could add a test with a non-tail recursive call that logs on each call until it hits actual stack overflow and confirm the expected "should never happen"

PR #3172, opened 2023-06-09
Prevent NoSuchMethodErrors on pre Servlet-api 3.0 versions in spring instrumentation.

JonasKunz:
## What does this PR do?

We had a user reporting that they are seeing the following error caused by our spring transaction-naming instrumentation:

```
[main] ERROR co.elastic.apm.agent.bci.IndyBootstrap - Advice threw an exception, this should never happen!
java.lang.NoSuchMethodError: javax.servlet.ServletContext.getClassLoader()Ljava/lang/ClassLoader;
```

While we only support Servlet-API versions 3.0+, the spring instrumentation does currently not guard against the presence older versions. This PR adds such a guard for this method to our servlet-api adapters.

## Checklist

- [x] This is a bugfix

### DIFF
--- a/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/adapter/JavaxServletApiAdapter.java
+++ b/apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/adapter/JavaxServletApiAdapter.java
@@ -43,7 +43,17 @@ public class JavaxServletApiAdapter implements ServletApiAdapter<HttpServletRequ
 
     private static final JavaxServletApiAdapter INSTANCE = new JavaxServletApiAdapter();
 
+    private final boolean ServletContext_getClassLoader_available;
+
     private JavaxServletApiAdapter() {
+        boolean probeResult;
+        try {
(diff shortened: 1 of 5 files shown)


apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/adapter/JavaxServletApiAdapter.java
```diff
@@ -43,7 +43,17 @@ public class JavaxServletApiAdapter implements ServletApiAdapter<HttpServletRequ
...
+
     private JavaxServletApiAdapter() {
+        boolean probeResult;
+        try {
+            ServletContext.class.getMethod("getClassLoader");
+            probeResult = true;
+        } catch (NoSuchMethodException e) {
+            probeResult = false;
+        }
+        ServletContext_getClassLoader_available = probeResult;
```
jackshirazi, line 56: I think it's cleaner to have
`ServletContext_getClassLoader_available = canExecuteServletContextgetClassLoader()`
as the constructor, but that's my preference, will approve as is

jackshirazi (approved):

jackshirazi (approved):

PR #3185, opened 2023-06-13
Fix lambda layer publish script

reakaleek:
## What does this PR do?

The lambda layer publish script did not work as expected locally and presumably in GH actions and created 20+ lambda layer versions in eu-west-1 instead of 1 version in every region. (Assumption: maybe due to Jenkins compatibility)

The script added by this PR proved to work in a GH Actions workflow in https://github.com/elastic/apm-agent-python and https://github.com/elastic/apm-aws-lambda.

### DIFF
--- a/.ci/publish-aws.sh
+++ b/.ci/publish-aws.sh
@@ -21,20 +21,6 @@ ALL_AWS_REGIONS=$(aws ec2 describe-regions --output json --no-cli-pager | jq -r 
 rm -rf "${AWS_FOLDER}"
 mkdir -p "${AWS_FOLDER}"
 
-# Delete previous layers
-for region in $ALL_AWS_REGIONS; do
-  layer_versions=$(aws lambda list-layer-versions --region="${region}" --layer-name="${ELASTIC_LAYER_NAME}" | jq '.LayerVersions[].Version')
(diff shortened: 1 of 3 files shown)


.ci/publish-aws.sh
```diff
@@ -0,0 +1,66 @@
...
+export AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION:-eu-west-1}
+
+export AWS_FOLDER=${AWS_FOLDER:-.ci/.aws}
+FULL_LAYER_NAME=${ELASTIC_LAYER_NAME:?layer name not provided}
+ALL_AWS_REGIONS=$(aws ec2 describe-regions --output json --no-cli-pager | jq -r '.Regions[].RegionName')
+
+rm -rf "${AWS_FOLDER}"
+mkdir -p "${AWS_FOLDER}"
+
+# Delete previous layers
```
JonasKunz, line 24: Isn't automatically deleting stuff a bit dangerous?

reakaleek: this was a copy-paste from another project.

But we can remove it.

https://github.com/elastic/apm-agent-java/pull/3185/commits/335305adc652d20259c2c8cab2f492393b7b1611

PR #3187, opened 2023-06-14
Upgrade micrometer, workaround for histogram bug in tests.

JonasKunz:
## What does this PR do?
 Upgrades micrometer in the tests to the latest version. Includes a workaround in the tests for https://github.com/micrometer-metrics/micrometer/issues/3904.

## Checklist

- [x] This is something else


jackshirazi (approved):
But I think we should put the `publishPercentileHistogram` back when it's fixed in micrometer. If you agree, we should raise an issue for that and target to the next milestone, I see that micrometer have targeted your bug to be fixed in their next milestone

PR #3190, opened 2023-06-15
Wildlfy test upgrades

JonasKunz:
## What does this PR do?

Upgrades the wildfly integration tests and the docs regarding supported versions.
In addition this contains an experiment for how to receive dependabot updates for our integration tests: A dummy dockerfile was added, which should be scanned by dependabot. Our integration tests read the image to use from that Dockerfile.

## Checklist

- [x] This is something else

### DIFF
--- a/integration-tests/application-server-integration-tests/src/test/resources/Dockerfile-wildfly-latest
+++ b/integration-tests/application-server-integration-tests/src/test/resources/Dockerfile-wildfly-latest
@@ -1,0 +1,4 @@
+FROM quay.io/wildfly/wildfly:27.0.0.Final-jdk17
+# This is a dummy dockerfile which is just there to receive dependabot updates
+# The FROM clause at the top is parsed by our integration tests to use the latest but fixed version
+# This has the benefit over using docker :latest tags in tests that the tests should remain stable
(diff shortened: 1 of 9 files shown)


integration-tests/application-server-integration-tests/src/test/resources/Dockerfile-wildfly-latest
```diff
@@ -0,0 +1,4 @@
+FROM quay.io/wildfly/wildfly:27.0.0.Final-jdk17
```
JonasKunz, line 1: [For Reviewer]
This is intentionally not the very latest version, which is `28`. This is to check whether the dependabot approach works as desired.

integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/WildFlyIT.java
```diff
@@ -50,13 +50,10 @@ public static Iterable<Object[]> data() {
...
             {"8.2.1.Final"},
             {"9.0.0.Final"},
-            {"10.0.0.Final"},
             {"11.0.0.Final"},
-            {"12.0.0.Final"},
             {"13.0.0.Final"},
-            {"14.0.0.Final"},
             {"15.0.0.Final"},
-            {"16.0.0.Final"}
+            {"25.0.0.Final"}
```
JonasKunz, line 56: [For Reviewer]
I removed a few inbetween versions as they are IMO not that useful to get confidence. I decided to remove them to not further increase the testing time.

jackshirazi (approved):

PR #3207, opened 2023-06-26
[ci] enhance CI build artifacts

SylvainJuge:
## What does this PR do?

Publish other agent artifacts in github workflow to have a ready-to-download link for each PR build.
For now it's one zip file per artifact, but we may switch to a single archive later.

## Checklist

- [x] decide on single zip or many zip files
- [x] This is something else

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -82,6 +82,21 @@ jobs:
         with:
           name: elastic-apm-agent-java8
           path: ./elastic-apm-agent-java8/target/elastic-apm-agent-java8-*.jar
+      - name: Upload agent attach CLI artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: apm-agent-attach-cli
+          path: ./apm-agent-attach-cli/target/apm-agent-attach-cli-*.jar
+      - name: Upload agent attach artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: apm-agent-attach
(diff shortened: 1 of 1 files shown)


jackshirazi (approved):
There are other artifacts we publish, but I think these are enough, or you could add apm-agent-plugin-sdk?

Others:
apm-agent-parent
apm-agent-plugin-sdk
apm-agent-tracer
apm-opentracing

SylvainJuge:
I'm in to add the 4 other artifacts, however that means we'll have 9 in total, thus maybe having a single zip file isn't a bad idea after all for simplicity, this would only require 1 line per artifact in the workflow file, which seems easier to maintain over time.

The only downside however would be the possibility for end-users to pick the wrong artifact, but given the audience here is mostly support and "experienced" users, I would be in to bet that it should not happen that often (and the agent already self disables when used as a library, thus harm is limited).

Having zip files for every artifact is a limitation of artifacts that are created during GH actions build, releases can however have other types of artifacts.

what do you think @jackshirazi ?

jackshirazi:
I think just adding the one plugin is sufficient. In fact we could go with the current even without that - my thinking is who wants snapshot builds of these various artifacts, and essentially it's us providing a bugfix for someone to test, and that's the set of of artifacts you've done and maybe the plugin - though we've never needed to provide a snapshot for that last.

So either merge this, or add that one plugin artifact, I don't think we need the rest nor a zip of all of them

PR #3215, opened 2023-06-30
Don't shutdown the logging configuration when running from test

jackshirazi:
## What does this PR do?
Bugfix, the logging configuration was being shutdown in tests as part of the reset between tests, but this meant that all the logging was switched to OFF after that reset, which meant subsequent tests in a class had no log output

## Checklist

- [x] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
@@ -647,12 +647,7 @@ public class ElasticApmAgent {
         if (instrumentation == null) {
             return;
         }
-        Tracer tracer = GlobalTracer.get().require(Tracer.class);
-        if (tracer instanceof ElasticApmTracer) {
-            ((ElasticApmTracer) tracer).stop();
-        } else {
-            tracer.stop();
-        }
+        GlobalTracer.get().require(Tracer.class).stop();
         GlobalTracer.setNoop();
         Exception exception = null;
         if (resettableClassFileTransformer != null) {
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
@@ -649,7 +649,7 @@ public class ElasticApmAgent {
         }
         Tracer tracer = GlobalTracer.get().require(Tracer.class);
         if (tracer instanceof ElasticApmTracer) {
-            ((ElasticApmTracer) tracer).stopForTest();
+            ((ElasticApmTracer) tracer).stop();
         } else {
             tracer.stop();
         }
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
@@ -647,7 +647,12 @@ public class ElasticApmAgent {
         if (instrumentation == null) {
             return;
         }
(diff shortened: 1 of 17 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -565,7 +565,15 @@ public void recycle(TraceContext traceContext) {
...
-    public synchronized void stop() {
+    public void stop() {
+        stop(false);
+    }
+
+    public void stopForTest() {
+        stop(true);
+    }
+
+    private synchronized void stop(boolean fromTest) {
```
SylvainJuge, line 576:
```suggestion
    public synchronized void stop(boolean shutdownLogging) {
```

There is only a few cases where we actually shutdown the logging, so we'd rather make `stop(){...}` variant NOT stop logging as it's mostly used for tests.

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
```diff
@@ -647,7 +647,12 @@ public static synchronized void reset() {
         if (instrumentation == null) {
             return;
         }
-        GlobalTracer.get().require(Tracer.class).stop();
+        Tracer tracer = GlobalTracer.get().require(Tracer.class);
```
SylvainJuge, line 650: we can use `assertionsEnabled` as a proxy for "this is running in tests" without having to do a cast, or even better do this transparently within `ElasticApmTracer#stop` without having the caller to deal with "is this test code or not".

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
jackshirazi: I did that originally because fewer changes like you say, but it looked clearer to explicitly say in the test that it's a test usage of stop. I'm not convinced my choice was better, but I don't think that just because there are fewer changes makes it better to do it the other way?

apm-agent-core/src/main/java/co/elastic/apm/agent/bci/ElasticApmAgent.java
jackshirazi: yes, you're right that's definitely better

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
jackshirazi: lol, doesn't matter given the other suggestion

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
```diff
@@ -590,7 +590,9 @@ public synchronized void stop() {
         } catch (Exception e) {
             logger.warn("Suppressed exception while calling stop()", e);
         }
-        LoggingConfiguration.shutdown();
+        if (!assertionsEnabled) {
```
SylvainJuge, line 593: [minor] maybe add a comment to explain why we do this when assertions are enabled so one of our future selves would not wonder why we do this.

SylvainJuge (approved):

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java
jackshirazi, line 593:
```suggestion
        //Shutting down logging resets the log level to OFF - subsequent tests in the class will get no log output, hence the guard
        if (!assertionsEnabled) {
```

PR #3231, opened 2023-07-10
Make most plugins depend on only the SDK and tracer dependencies.

raphw:
Avoids that most plugins depend on apm-agent-core. To do so, some remaining classes are moved to more appropriate locations.


JonasKunz (approved):

JonasKunz:
run elasticsearch-ci/docs

Issue #3004, opened 2023-02-06
[META 757] Baggage in APM Agents


elastic-apm-tech:
See meta issue and spec for the description and details:
- Meta issue: https://github.com/elastic/apm/issues/757
- Spec issue: https://github.com/elastic/apm/issues/758

JonasKunz:
Implementation plan:
 - [ ] move capability of starting spans from `AbstractSpan` to `ElasticContext` (otherwise baggage updates might be lost)
 - [ ] move header context propagation responibility from `AbstractSpan` to `ElasticContext`, context propagation needs to happen even if no span is active (e.g. propagating just baggage)
 - [ ] Implement baggage in `ElasticContext`, implement header propagation for baggage
 - [ ] Adjust our in-process context propagation to propagate contexts instead of spans
 - [ ] Bridge OpenTelemetry baggage to our internal baggage implementation

JonasKunz:
The implementation will have the following limitations:
 * When creating spans via `ElasticAPM.currentSpan().createSpan()` baggage updates which happened after the activation of the current span will not be visible on the child span. This should not be a problem though, as baggage can only be updated via the OpenTelemetry API. In that case, users should be using the OpenTelemetry API for creating spans aswell.
 * The same limitation applies to our OpenTracing bridge
 * Baggage propagation support has *not* been implemented for Kafka. This will be done via #3017

PR #3236, opened 2023-07-13
W3C Baggage propagation and tracer APIs

JonasKunz:
## What does this PR do?

Part of #3004. Implements W3C baggage propagation and parsing.
Adds tracer-API s for reading and updating baggage.

## Checklist

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/baggage/Baggage.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/baggage/Baggage.java
@@ -1,3 +1,21 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
 package co.elastic.apm.agent.impl.baggage;
 
 import javax.annotation.Nullable;
@@ -9,7 +27,7 @@ import java.util.Set;
 
 public class Baggage implements co.elastic.apm.agent.tracer.Baggage {
 
-    public static final Baggage EMPTY = new Baggage(Collections.emptyMap(), Collections.emptyMap());
+    public static final Baggage EMPTY = new Baggage(Collections.<String, String>emptyMap(), Collections.<String, String>emptyMap());
 
     private final Map<String, String> baggage;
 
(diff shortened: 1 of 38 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/baggage/Baggage.java
```diff
@@ -0,0 +1,166 @@
...
+    public String getMetadata(String key) {
+        return baggageMetadata.get(key);
+    }
+
+    public boolean isEmpty() {
+        return baggage.isEmpty();
+    }
+
+    public static Builder builder() {
+        return EMPTY.toBuilder();
```
jackshirazi, line 71: I don't understand why you would want a builder that holds immutable empty maps so cannot be changed?

jackshirazi: I see, you're replacing them if they are from EMPTY. It's not intuitive

jackshirazi: It does minimize the object creation, I see that's why. Okay, ignore these comments

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/baggage/Baggage.java
```diff
@@ -0,0 +1,166 @@
...
+                setBaggageValue(key, value);
+                setBaggageMetadata(key, metadata);
+            }
+            return this;
+        }
+
+        /**
+         * Builds the resulting baggage.
+         * If no modifications were done, the returned instance might be the same as the parent baggage.
+         * The builder must not be used anymore after calling build.
```
jackshirazi, line 119: Worth enforcing this with a boolean check?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/baggage/RecyclableCharArray.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package co.elastic.apm.agent.impl.baggage;
+
```
jackshirazi, line 20: Why is this class in this package?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/baggage/W3CBaggagePropagation.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+
+    private static String encodeToHeader(Baggage baggage) {
+        StringBuilder header = new StringBuilder();
+        for (String key : baggage.keys()) {
+            String value = baggage.get(key);
+            String metadata = baggage.getMetadata(key);
+            if (isEncodeableBaggage(key, value)) {
```
jackshirazi, line 76: should there be log output for non-encodable baggage?

jackshirazi (approved):
Few minor suggestions, otherwise very nice!

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/baggage/RecyclableCharArray.java
JonasKunz: It was added as a replacement for the `ThreadLocal` usage within `PercentEncoder`. However, after this PR is merged I'll rebase #3239 and replace this class with the easier pooling added in that PR.

jackshirazi: yes, I mean why THIS package rather than the one with PercentEncoder?

Issue #3059, opened 2023-03-20
Virtual Thread Support


JonasKunz:
Virtual Threads are likely to be available as non-preview in Java 21.
We should ensure that our agent properly supports virtual threads:
 * We need to check our `ThreadLocal` usages and make sure that we don't use too much heap in case there are millions of threads
 * Check that context propagation still works properly with virtual threads (check if new APIs were introduced which our instrumentation needs to cover)
 * Ensure that virtual threads are excluded from the inferred spans feature (for start, we might add support later)
 * Add tests spawning millions of virtual threads and see what happens

NicklasWallgren:
Glad that this is prioritised.

Will 8.10 support JDK 21?

JonasKunz:
We already continuously test with JDK 21-EA. Therefore it should already be supported by current versions with the exception of potential high memory consumption for virtual threads which will be addressed by this issue.

PR #3239, opened 2023-07-17
Replace IOUtils thread local usage with loom friendly pooling

JonasKunz:
## What does this PR do?
Part of #3059.
Our `IOUtils` uses thread-locals for reusing `ByteBuffer` (2kb size) and `CharsetDecoders`.
This usage is not feasible with virtual threads, because 1 million threads for example would equate to 2gb memory usage due to the bytebuffers.

This PR replaces the thread-local based pooling with a queue based pool which we already use for other recycling use-cases. A new `ObjectHandle`-variant has been added to make the usage safe and easy via try-with-resources.

Resource pooling via a queue is expected to be more expensive than thread-local based pooling, as it involves at least two atomic CAS operations. Those should however not be to costly, as we expect low contention.

An alternative approach would be to use thread-locals for platform threads and queue based pooling for virtual threads, but I would expect the complexity here to not be worth it.

## Checklist
- [x] This is something else


apm-agent-core/src/main/java/co/elastic/apm/agent/util/IOUtilsProviderImpl.java
```diff
@@ -0,0 +1,161 @@
...
+            public DecoderWithBuffer createInstance() {
+                return new DecoderWithBuffer();
+            }
+        });
+    }
+
+    @Override
+    public boolean readUtf8Stream(final InputStream is, final CharBuffer charBuffer) throws IOException {
+        // to be compatible with Java 8, we have to cast to buffer because of different return types
+        try (ObjectHandle<DecoderWithBuffer> pooled = pool.createInstance()) {
```
JonasKunz, line 59: [For Reviewer] Unfortunately the diff does not recognize it, but the actual implementations have not been touched. The only change was to replace the `ThreadLocal` usage with the resource pool.

jackshirazi (approved):

PR #3240, opened 2023-07-17
Removed ThreadLocal usage for tracestate serialization

JonasKunz:
## What does this PR do?

Part of #3059 .
Prior to this PR, we used a `ThreadLocal<StringBuilder>` for generating tracestate headers.
This has been replaced by
 * caching the header within the `TraceState`-instance once it is generated
 * use the existing `StringBuilder` within the `TraceState` for generating the headers with a lock for thread safety
 
TBH I'm not sure why the `ThreadLocal` was there in the first place. I'm afraid that I'm missing something here, therefore I would like to wait for a Review from @SylvainJuge because he is the original author of this code.

## Checklist
- [x] This is something else

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceState.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceState.java
@@ -243,11 +243,11 @@ public class TraceState implements Recyclable {
     }
 
     /**
-     * NOT THREAD SAFE! Uses a shared string builder.
+     * Needs to be synchronized because of the use of the instance field StringBuilder.
      *
      * @return the text header generated from {@link #tracestate}.
      */
-    private String generateTracestateHeader() {
+    private synchronized String generateTracestateHeader() {
         String singleEntry = tracestate.size() != 1 ? null : tracestate.get(0);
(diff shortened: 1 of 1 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceState.java
```diff
@@ -230,82 +243,63 @@ public void setSizeLimit(int limit) {
...
+     *
+     * @return the text header generated from {@link #tracestate}.
      */
-    private static class TextTracestateAppender {
-
-        private static final TextTracestateAppender INSTANCE = new TextTracestateAppender();
-        private final ThreadLocal<StringBuilder> tracestateBuffer = new ThreadLocal<StringBuilder>();
-
-        private TextTracestateAppender() {
+    private String generateTracestateHeader() {
```
jackshirazi, line 250: for maintainability, why not make this a synchronized method? It won't impact the sync block it's called in, but will ensure that if someone later adds a call to it elsewhere forgetting to synchronize it, it will be thread-safe

jackshirazi (approved):

SylvainJuge (approved):

PR #3244, opened 2023-07-18
Exclude virtual threads from inferred spans feature

JonasKunz:
## What does this PR do?

Excludes virtual threads from the inferred spans feature. Before doing this I did a short test, showing that our async-profiler version can't deal with virtual threads switching their carrier thread.

## Checklist

- [x] This is something else


apm-agent-plugin-sdk/src/main/java/co/elastic/apm/agent/sdk/internal/ThreadUtil.java
```diff
@@ -0,0 +1,79 @@
...
+     * <p>
+     * Otherwise we runtime generate an implementation which invokes Thread.isVirtual().
+     * We use runtime proxy generation because Thread.isVirtual() has been added in Java 19 as preview and Java 21 as non preview.
+     * Therefore we would require a compilation with Java 19 (non-LTS), because Java 20+ does not allow targeting Java 7.
+     * <p>
+     * Alternatively we could simply invoke Thread.isVirtual via reflection.
+     * However, because this check can be used very frequently we want to avoid the penalty / missing inline capability of reflection.
+     *
+     * @return the implementation for {@link VirtualChecker}.
+     */
```
SylvainJuge, line 54: Nice ! I think that's definitely a good approach we could use for other deprecated or soon-to-be-removed APIs like the security manager.

SylvainJuge (approved):

Issue #3186, opened 2023-06-13
Improve GH actions release process


JonasKunz:
Version `1.39.0` was the first attempt of releasing via GH actions. It uncovered a few problems.

1st attempt:
 * There was a bug in the job which calls build-kite to perform the release, which meant that `mvn deploy was not triggered`
 * The release preparation step already had committed the version bumps and the new tag.
 * Resolution:
     * Fix the GH-action via a PR
     * Revert the version bump commits via a PR
     * Manual deletion of the tag (required admin permissions!)
     * Retry in 2nd attempt.
2nd attempt:
  * The `Deploy to maven central` "failed", however the actual artifact was actually published to maven central correctly. We think that the final "publish" of the release returned a strange response / timed out, but actually went through.
  * As a result, all of the follow-up tasks were aborted due to "failed" maven central publish step
  * Resolution:
     * We added boolean inputs to skip the `prepare release` and `Deploy to maven central` steps
     * This made everything work except for the AWS-lambda upload, as it depends on the uploaded lambda-zip from the buildkite build. Subsequently, also the creation of the GH release failed.
     * Those had to be done manually (required admin permissions and AWS credentials).
     
The GH actions release workflow was designed with atomicity of the individual steps in mind, so that in theory a failure could be mitigate by just using the `Rerun failed Jobs` feature. Unfortunately, this does not take the case where a job fails but silently succeeds (e.g. our maven deploy) into account, which is where things went wrong.
In addition, the `Rerun failed jobs` does not work correctly if there is a bug in the workflow or a related script. To my knowledge, the `Rerun failed jobs` would always use the same revision, therefore no fixes to the scripts can be applied.

We'll most likely need to split the process up into individual workflows, which ideally are idempotent.
We especially should adjust the lambda-task to not depend on a previous uploaded build artifact, but to use the apm-agent from maven central.

SylvainJuge:
Note to my future self so my present self can forget the details:

- We have to delegate some tasks to buildkite for artifacts signature
- We should probably split the release workflow in independent workflows
- Ideas for splitting: 
  - first part that creates the tag and pushes change to `main`, should end with triggering the artifact publication with buildkite
  - second part executes in buildkite
  - third part waiting on the publication in maven central (or better triggered by buildkite job when it completes)
  - if the buildkite part fails, then we should be able to investigate it, then trigger the third part manually if needed

jackshirazi:
[...]

PR #3251, opened 2023-07-20
Decouple lambda layer release from buildkite build

JonasKunz:
## What does this PR do?

Makes the AWS lambda layer build optionally download the agent-jar from maven central instead of using the locally built one. This allows us to remove the buildkite download-artifact step from the release workflow.

This achieved by using the maven-dependency-plugin for fetching the elastic-apm-agent.jar . For this to work, the lambda layer build had to be moved to a separate module.

It makes use of the dependency use resolution of the dependency plugin:
> Artifacts are resolved from the following sources in order:
> 
> * the current reactor,
> * the local repository,
> * the configured remote repositories.

When running the standard `./mvnw package`, the aws lambda layer will be built using the locally built agent jar, which is exactly what we want for the lambda integration test.
Running the following command will instead force the jar to be fetched from maven central:
 ```
 ./mvnw dependency:purge-local-repository package -pl apm-agent-lambda-layer
 ```
`dependency:purge-local-repository` ensures that no artifact from the local repository is used, `-pl apm-agent-lambda-layer` ensures that only the lambda layer is built and not the agent jar itself.

Part of #3186 .

## Checklist

- [x] This is something else


jackshirazi (approved):
CI magic!

jackshirazi (approved):

reakaleek (approved):

PR #3254, opened 2023-07-25
Fix invalid SQS propagation introduced with #3206

JonasKunz:
## What does this PR do?

Fixes a `ClassDefNotFoundError` reported in this [discussion forums thread](https://discuss.elastic.co/t/bug-apm-java-springboot-v1-39-0-no-jvm-metrics-found/339104).

The invocation of the `SQSHelper` was moved behind the span activation to ensure that even if no span is created, baggage propagation still occurs.

This however caused the `SQSHelper` to be invoked for all aws services, including non-sqs ones.

## Checklist

- [x] This is a bugfix
  - [x] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)
  - [ ] I have added tests that would fail without this fix

PR #3266, opened 2023-08-03
Fix invalid SQSHelper usage in AWS async client instrumentation

JonasKunz:
## What does this PR do?

Fixes the same issue as #3254, but for the async-client. Looks like I forgot about the async client in the previous bugfix.
This time I double-checked: The Async and Sync clients are the only references to the problematic `SQSHelper`-class.

## Checklist

- [x] This is a bugfix


jackshirazi (approved):
good catch

PR #3273, opened 2023-08-07
Add error reporting API to tracer-api.

raphw:
This change adds an representation of `ErrorCapture` to the tracer API to avoid a dependency on the core module. To reduce complexity, errors are always reported on the current span. No plugin derives from this practice what makes this sensible.


v1v:
@mergifyio update

raphw:
Force pushed with new hash to trigger new build which failed due to low memory on testcontainers.

JonasKunz (approved):
LGTM, I've done the rebase and just added two missing `@Override` annotations.

JonasKunz:
run elasticsearch-ci/docs

Issue #3259, opened 2023-07-31
Open Liberty JMS message beans doesn't contain elastic apm logging 


marinkton:
I'm referring to issue [3192](https://github.com/elastic/apm-agent-java/issues/3192). 

We've tested the new 1.40.0 version and it works well with Spring Boot, but unfortunately, we're encountering issues with open liberty. You can download the project from [this link](https://drive.google.com/file/d/135_hFFXyXAoyo1Q059SnaOuvoPvsHCbr/view?usp=sharing) .

By the way, you can ignore the dockerfile in the project.

To build the project, simply run mvn clean install, and to run it, use mvn liberty:dev. If needed, you can modify the property values in liberty/config/server.env.

Thanks in advance!

JonasKunz:
Thanks for providing the sample application!
After looking at the code I suspect the root-cause is that the [jms_listener_packages](https://www.elastic.co/guide/en/apm/agent/java/current/config-messaging.html#config-jms-listener-packages) is not set.

Could you configure that option to your application package? I assume all your `MessageListener`s are in that package (or a sub package).

The documentation of the `jms_listener_packages` is unfortunately misleading: If the configuration is not set, it does not mean that all `MessageListeners` are instrumented. Instead, for performance reasons a heuristic for the classname is used: Only classes whose name contains "Message" or "Listener" will be checked.

We will:
 * Clarify the documentation about the behavior
 * Ensure that the `application_packages` configuration value is also always scanned for `MessageListeners`: This would already have avoided the problem in your case
 
 Could you verify that configuring `jms_listener_packages` solves the issue for you?

marinkton:
@JonasKunz thanks for the reply. 

I've tested by adding `jms_listener_packages ` but it didn't work. As you can see `application_packages ` is already in the server.env file aswell. 

I renamed the classes also to "Listener" which didn't work also.

SylvainJuge:
Hi @marinkton,

The current implementation requires to pick one of the following strategies:
- set `jms_listener_packages` to `com.demo` (following the provided your example), even if that duplicates the `application_packages` configuration (you will have to set both) (1)
- rename `ConsumerBean` to `ConsumerListener` or `MessageConsumer` (2)

(1) For now, the `jms_listener_packages` do not "inherit" from the value set in `application_packages`.

(2) Here, renaming without changing the agent configuration should work, thus that's definitely a bit surprising. Could you re-test this configuration just to be sure ?

[...]

PR #3299, opened 2023-08-23
 Enhance/document msg listener heuristic

### LINKED ISSUE #3135: Allow opt-out for MessageListener matcher performance optimization
#2930 added a performance optimization by pre-filtering instrumentation candidates for our `MessageListener` instrumentation.
Unfortunately, there is currently no way of restoring the old behaviour: It is not possible to make the agent scan all packages for `MessageListeners`. If a user has no knowledge of where the `MessageListeners` are placed, he has no chance of instrumenting them.

A simple, backwards-compatible fix would be to enhance the `elastic.apm.jms_listener_packages` config option to accept wildcard matchers instead of strings. This way, a user can set `jms_listener_packages=*` to make the agent scan all packages.

A downside of this approach is a bit of inconsistency in the configuration: Other wildcard options such as `classes_excluded_from_instrumentation` match the wildcards against the entire class name, whereas `jms_listener_packages` only checks that the classname begins with the provided pattern / string.

SylvainJuge:
## What does this PR do?

Fixes #3135 
Relates to #3259 

I had to move the test JMS classes to a package outside of `co.elastic.apm` as `co.elastic.apm` is already covered by the test configuration which is automatically loaded from classpath.

## Checklist

### DIFF
--- a/apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/configuration/MessagingConfiguration.java
+++ b/apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/configuration/MessagingConfiguration.java
@@ -100,7 +100,7 @@ public class MessagingConfiguration extends ConfigurationOptionProvider {
         .configurationCategory(MESSAGING_CATEGORY)
         .description("Defines which packages contain JMS MessageListener implementations for instrumentation." +
             "\n" +
-            "When empty (default), only the inner-classes or that have 'Listener' or 'Message' in their names are considered.\n"+
+            "When empty (default), all inner-classes or any classes that have 'Listener' or 'Message' in their names are considered.\n"+
             "\n" +
             "This configuration option helps to make MessageListener type matching faster and improve application startup performance.\n" +
             "\n" +
--- a/apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/configuration/MessagingConfiguration.java
+++ b/apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/configuration/MessagingConfiguration.java
@@ -100,8 +100,11 @@ public class MessagingConfiguration extends ConfigurationOptionProvider {
         .configurationCategory(MESSAGING_CATEGORY)
         .description("Defines which packages contain JMS MessageListener implementations for instrumentation." +
             "\n" +
-            "When set to a non-empty value, only the classes matching configuration will be instrumented.\n" +
-            "This configuration option helps to make MessageListener type matching faster and improve application startup performance."
+            "When empty (default), only the inner-classes or that have 'Listener' or 'Message' in their names are considered.\n"+
+            "\n" +
+            "This configuration option helps to make MessageListener type matching faster and improve application startup performance.\n" +
+            "\n" +
+            "Starting from version 1.43.0, the classes that are part of the 'application_packages' option are also included."
         )
         .dynamic(false)
         .buildWithDefault(Collections.<String>emptyList());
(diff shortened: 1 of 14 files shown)


docs/configuration.asciidoc
```diff
@@ -2430,9 +2430,12 @@ Prepending an element with `(?-i)` makes the matching case sensitive.
 ==== `jms_listener_packages` (performance added[1.36.0])
 
 Defines which packages contain JMS MessageListener implementations for instrumentation.
-When set to a non-empty value, only the classes matching configuration will be instrumented.
+When empty (default), only the inner-classes or that have 'Listener' or 'Message' in their names are considered.
```
jackshirazi, line 2433: only which inner classes? any in the whole application?

docs/configuration.asciidoc
```diff
@@ -2430,9 +2430,12 @@ Prepending an element with `(?-i)` makes the matching case sensitive.
...
+
 This configuration option helps to make MessageListener type matching faster and improve application startup performance.
 
+Starting from version 1.43.0, the classes that are part of the 'application_packages' option are also included.
```
jackshirazi, line 2437: how does that affect the default vs non-default?

SylvainJuge: It does not affect it, it's just that classes that are part of the `application_packages` are also evaluated during class matching. In a sense it's just making the exclusion pre-filtering less restrictive than previously.

apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/configuration/MessagingConfiguration.java
```diff
@@ -100,8 +100,11 @@ public class MessagingConfiguration extends ConfigurationOptionProvider {
         .configurationCategory(MESSAGING_CATEGORY)
         .description("Defines which packages contain JMS MessageListener implementations for instrumentation." +
             "\n" +
-            "When set to a non-empty value, only the classes matching configuration will be instrumented.\n" +
-            "This configuration option helps to make MessageListener type matching faster and improve application startup performance."
+            "When empty (default), only the inner-classes or that have 'Listener' or 'Message' in their names are considered.\n"+
```
jackshirazi, line 103: only which inner classes? any in the whole application?

jackshirazi, line 103:
```suggestion
            "When empty (default), all inner-classes or any classes that have 'Listener' or 'Message' in their names are considered.\n"+
```

docs/configuration.asciidoc
jackshirazi: so the default option will be inner-classes, classes with 'Listener' or 'Message' in the name, and all classes in application_packages?

apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/configuration/MessagingConfiguration.java
```diff
@@ -100,8 +100,11 @@ public class MessagingConfiguration extends ConfigurationOptionProvider {
         .configurationCategory(MESSAGING_CATEGORY)
         .description("Defines which packages contain JMS MessageListener implementations for instrumentation." +
             "\n" +
-            "When set to a non-empty value, only the classes matching configuration will be instrumented.\n" +
-            "This configuration option helps to make MessageListener type matching faster and improve application startup performance."
+            "When empty (default), all inner-classes or any classes that have 'Listener' or 'Message' in their names are considered.\n"+
+            "\n" +
+            "This configuration option helps to make MessageListener type matching faster and improve application startup performance.\n" +
+            "\n" +
+            "Starting from version 1.43.0, the classes that are part of the 'application_packages' option are also included."
```
jackshirazi, line 107:
```suggestion
            "Starting from version 1.43.0, the classes that are part of the 'application_packages' option are also included in the list of classes considered."
```

jackshirazi (approved):

docs/configuration.asciidoc
SylvainJuge: yes

PR #3300, opened 2023-08-25
Replace binary propagation with UTF8 encoded text propagation

### LINKED ISSUE #3017: [META 764] Replace binary context propagation with UTF8 encoded textual headers for Otel compatibility
See meta issue for the description and details:
- Meta issue: https://github.com/elastic/apm/issues/764

JonasKunz:
## What does this PR do?

Closes #3017.
This change has been tested manually for sending / receiving kafka interoperability with both older elastic APM agents and the opentelemetry agent.

## Checklist


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java
```diff
@@ -56,23 +58,6 @@
  *       Header name     Version           Trace-Id                Span-Id     Flags
  * </pre>
  * <p>
- * Binary representation (e.g. 0.11.0.0+ Kafka record header), based on
```
jackshirazi, line 59: This comment is still useful since we're supporting the backward compatible header

jackshirazi (approved):
OMG thank you, this leaves it so much cleaner! The only thing I suggest is to move the deleted binary header structure comment over to the KafkaRecordHeaderAccessor class, since we're still supporting that header

PR #3303, opened 2023-08-29
Disable instrumentation for ES client versions with native instrumentation

### LINKED ISSUE #3235: Disable ES instrumentation when a language client with included OTel instrumentation is used
https://github.com/elastic/elasticsearch-java/pull/588 will bake instrumentation into the elasticsearch java client.
As soon as this PR is merged, we should make sure that our instrumentation detects versions of the client which already are instrumented and disable our instrumentation in that case.
Otherwise we might end up double recording the data.

AlexanderWert:
Fixes #3235

https://github.com/elastic/elasticsearch-java/pull/652 introduced native instrumentation into the ES client.
We need to disable agent instrumentation for versions with native instrumentation.

The native instrumentation introduced the class  `co.elastic.clients.transport.instrumentation.Instrumentation` in the ES client. This PR adds a check for the presence of that class.

### DIFF
--- a/apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/esrestclient/ElasticsearchRestClientInstrumentation.java
+++ b/apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/esrestclient/ElasticsearchRestClientInstrumentation.java
@@ -23,11 +23,11 @@ import co.elastic.apm.agent.tracer.GlobalTracer;
 import co.elastic.apm.agent.tracer.Tracer;
 import net.bytebuddy.matcher.ElementMatcher;
 
-import static co.elastic.apm.agent.sdk.bytebuddy.CustomElementMatchers.classLoaderCanLoadClass;
-import static net.bytebuddy.matcher.ElementMatchers.not;
-
 import java.util.Collection;
 import java.util.Collections;
+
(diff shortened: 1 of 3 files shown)


AlexanderWert:
**TODO**: we need to make dependabot stop bumping the ES client versions from 8.10+ on

apm-agent-plugins/apm-es-restclient-plugin/apm-es-restclient-plugin-common/src/main/java/co/elastic/apm/agent/esrestclient/ElasticsearchRestClientInstrumentation.java
```diff
@@ -35,4 +39,10 @@ public Collection<String> getInstrumentationGroupNames() {
         return Collections.singleton("elasticsearch-restclient");
     }
 
+    @Override
+    public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
+        // ensure only ES client versions without native instrumentation are instrumented
```
SylvainJuge, line 44: [minor] adding the ES client version number in the comment here might be relevant for future reference, alternatively we could have also used the `co.elastic.apm.agent.sdk.bytebuddy.CustomElementMatchers#implementationVersionLte` method that makes it self-documented, however doing so here would have required to already have the new client released, which is not the case yet.

JonasKunz: I've added the cutoff version to the comment

PR #3313, opened 2023-09-07
Fix DB2 jdbc connection instrumentation

JonasKunz:
## What does this PR do?

It seems like DB2 JDBC-connection classes are sometimes obfuscated and therefor do not contain "Connection" in their name. Our JDBC instrumentation does not trace statemented if the SQL is not available, which is only available if the connection was also instrumented.

## Checklist

- [x] This is a bugfix
  - [ ] ~I have added tests that would fail without this fix~


JonasKunz:
Fix has been tested and verified by the user.

PR #3316, opened 2023-09-08
Change 'cloud.project.id' for GCP metadata to be the 'project-id'

trentm:
Closes: https://github.com/elastic/apm-agent-java/issues/3311

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -44,6 +44,7 @@ Use subheadings with the "=====" level for adding notes for unreleased changes:
 * Added `baggage_to_attach` config option to allow automatic lifting of baggage into transaction, span and error attributes - {pull}3288[#3288], {pull}3289[#3289]
 * Exclude elasticsearch 8.10 and newer clients from instrumentation because they natively support OpenTelemetry  - {pull}3303[#3303]
 * Switched to OpenTelemetry compatible context propagation for Kafka - {pull}3300[#3300]
+* Changed `cloud.project.id` collected in Google Cloud (GCP) to be the `project-id` - {issues}3311[#3311]
 
 [[release-notes-1.x]]
 === Java Agent version 1.x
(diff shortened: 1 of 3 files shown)


CHANGELOG.asciidoc
```diff
@@ -44,6 +44,7 @@ Use subheadings with the "=====" level for adding notes for unreleased changes:
 * Added `baggage_to_attach` config option to allow automatic lifting of baggage into transaction, span and error attributes - {pull}3288[#3288], {pull}3289[#3289]
 * Exclude elasticsearch 8.10 and newer clients from instrumentation because they natively support OpenTelemetry  - {pull}3303[#3303]
 * Switched to OpenTelemetry compatible context propagation for Kafka - {pull}3300[#3300]
+* Changed `cloud.project.id` collected in Google Cloud (GCP) to be the `project-id` - {issues}3311[#3311]
```
JonasKunz, line 47: Isn't this rather a `Potentially Breaking Change` and should be put in a section with the corresponding heading? How did you treat this for Node?

trentm: I see now that the Java agent's changelog sometimes has a "Potentially breaking changes" section. I can definitely move the entry to that section.

This change means that a query using hardcoded `project-id` values could now return empty results. Is that a breaking change? https://github.com/elastic/apm/blob/main/specs/agents/breaking-changes.md#what-is-not-a-breaking-change suggests we don't consider it one. I think it is helpful to users to have a "Potentially breaking changes" section that can highlight these things.

JonasKunz: I didn't know about that spec doc, thanks for pointing it out! I'd say we follow the spec then and leave it as a feature as you suggested.

JonasKunz (approved):

JonasKunz (approved):
Thanks for implementing this!

JonasKunz (approved):

jackshirazi (approved):

PR #3327, opened 2023-09-20
Fixes ByteBuddy AbstractMethodError which could occur when running tests from IDE

JonasKunz:
We have now had for a time the strange behaviour that certain tests would fail when run from the IDE with the following error:

```
java.lang.AbstractMethodError: Receiver class co.elastic.apm.agent.bci.bytebuddy.MinimumClassFileVersionValidator does not define or inherit an implementation of the resolved method 'abstract net.bytebuddy.jar.asm.ClassVisitor wrap(net.bytebuddy.description.type.TypeDescription, net.bytebuddy.jar.asm.ClassVisitor, net.bytebuddy.implementation.Implementation$Context, net.bytebuddy.pool.TypePool, net.bytebuddy.description.field.FieldList, net.bytebuddy.description.method.MethodList, int, int)' of interface net.bytebuddy.asm.AsmVisitorWrapper.
	at net.bytebuddy.asm.AsmVisitorWrapper$Compound.wrap(AsmVisitorWrapper.java:746) ~[byte-buddy-1.14.8.jar:?]
```

I've been able to find the root cause of the issue and fix it with this PR.

Our `MinimumClassFileVersionValidator` implements `AsmVisitorWrapper` from `byte-buddy-dep`, which means it uses the **unshaded** `org.objectweb.asm.ClassVisitor` as parameter type.
The error message shows that at runtime the `AsmVisitorWrapper` interface is picked up from `byte-buddy` instead, which expects the shaded `net.bytebuddy.jar.asm.ClassVisitor`.

I assume that when the tests are run from maven the classpath order is slightly different, so that `byte-buddy-dep` wins over `byte-buddy`. In the IDE the order seems to be the other way around, causing the failure.

I've fixed it by excluding `byte-buddy` from the dependencies which pull it in transitively: `assertJ` and `mockito`.

## Checklist

- [x] This is something else

### DIFF
--- a/apm-agent-attach-cli/pom.xml
+++ b/apm-agent-attach-cli/pom.xml
@@ -56,6 +56,13 @@
             <version>${version.bouncy-castle.bc-fips}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <!-- required for the inherited assertJ and Mockito dependencies-->
+            <groupId>net.bytebuddy</groupId>
+            <artifactId>byte-buddy-dep</artifactId>
+            <version>${version.byte-buddy}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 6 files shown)


jackshirazi (dismissed):

JonasKunz:
> I think before merging, it's worth doing a manual test using a snapshot and just run through to see all the normal things expected are displayed in the UI

The shanges should (in theory) only have altered the testing classpath. However, I agree that it makes still sense to do a manual test.

PR #3336, opened 2023-09-28
Offer direct span creation API for use within explicit tracers.

raphw:
Adds a `DirectTracer` abstraction that allows for recording data explicitly, for example when using tracer APIs such as OpenTelementry.


SylvainJuge:
Hi @raphw , if I am not mistaken most of the changes here are also included in #3340 , is that correct ?

For now the only concern that I have is the exact meaning we want to give to `direct` implementations, as it's not straightforward to know which one to pick apart from relying on what is available in the classpath, do you have a few other naming strategies that would be less abstract ?

raphw:
Direct is maybe not the best term. Maybe explicit is a better prefix?

raphw:
I am creating a new, independant PR.

PR #3348, opened 2023-10-06
[mergify] fix mergify config again

SylvainJuge:
## What does this PR do?

The default merge queue strategy is `merge` and that is not compatible with our branch protection rules (we require a linear history), as a consequence we need to explicitly set `merge_method` on the queue (and not the pull-request like I did previously).

## Checklist

- [x] This is something else

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,7 +1,8 @@
 queue_rules:
   - name: default
-  # branch protection rules are automatically included, no extra rule needed for now
+  # merge method needs to be set to 'squash' to match the repository config
     merge_method: squash
+  # branch protection rules are automatically included, no extra rule needed for now
 
 pull_request_rules:
   - name: Automatic squash and merge on approval with success checks and ready-to-merge label
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -11,3 +11,4 @@ pull_request_rules:
       - base=main
     actions:
       queue:
+        name: default
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,6 +1,7 @@
 queue_rules:
   - name: default
   # branch protection rules are automatically included, no extra rule needed for now
+    merge_method: squash
 
 pull_request_rules:
   - name: Automatic squash and merge on approval with success checks and ready-to-merge label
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -9,5 +11,3 @@ pull_request_rules:
       - base=main
     actions:
       queue:
-        merge_method: squash
-        name: default
```
v1v, line 13: What's the reason for the name being removed?

The APM Server uses `merge` instead `merge_method`:
- https://github.com/elastic/apm-server/blob/547b7c8bcd6a86006f1470a859dd757fa947e342/.mergify.yml#L149-L152

IIUC, `merge` is deprecated so `merge_method` is the new approach, surprisingly it works in the APM Server, see https://github.com/elastic/apm-server/pull/11784 that was automatically merged as soon as it got an approval.

SylvainJuge: good catch for the `name`, I've re-added it.

The "pull_request_rules" is applied to make the PRs enter the merge queue.

When a PR is added manually using the "queue" command, then the `pull_request_rules` is not applied as the user added the PR to the merge queue. But that also means that the PR will be merged with the default strategy `merge`.

Setting this by default on the merge queue means that by default all PRs will be merged using this strategy (which is consistent with our "linear history" requirement and default repo config for merging PRs), and that should be applied to the PRs that are added manually or automatically to the queue.

v1v: I'll be finishing writing down the proposal with the release automation but I see something that we might need to figure out.

If there will be a PR with the version bump then the `linear history` will be required hence the merge method will need to be forced to 'fast-forward". How do you envision this particular settings if we use Mergify?

v1v (approved):

.mergify.yml
SylvainJuge: For the release PRs, once they are approved we can make mergify `rebase` them instead of a `squash`, this would be handled by a dedicated `pull_request_rules` that will use the `merge_method: 'rebase'` instead of the queue default.

v1v: Brilliant, thanks Sylvain!

PR #3353, opened 2023-10-10
Fix multiple spans being created for HTTPUrlConnection HEAD requests

JonasKunz:
## What does this PR do?

While testing with a local spring-boot app I noticed that out instrumentation created 14 spans per single RestTemplate request.

This issue occured because I instructed the `RestTemplate` to ignore the response body by setting the expected type to `Void.class`. This causes spring to send `HEAD` requests instead. And as it turns out, our `HTTPUrlConnection` instrumentation doesn't deal well with those:

The internal `connected` field is not switched to `true` for `HEAD` requests, causing multiple spans to be created when invoking `getResponseCode` multiple times.

This PR fixes this by also checking the `responseCode` field to infer whether the `HTTPUrlConnection` has already connected or not.

## Checklist

- [x] This is a bugfix

### DIFF
--- a/apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
+++ b/apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
@@ -79,7 +79,11 @@ public abstract class HttpUrlConnectionInstrumentation extends ElasticApmInstrum
             @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
             public static Object enter(@Advice.This HttpURLConnection thiz,
                                        @Advice.FieldValue("connected") boolean connected,
+                                       @Advice.FieldValue("responseCode") int responseCode,
                                        @Advice.Origin String signature) {
+
+                //With HEAD requests the connected stays false
+                boolean actuallyConnected = connected || responseCode != -1;
 
                 boolean isNestedCall = callDepth.isNestedCallAndIncrement();
                 ElasticContext<?> activeContext = tracer.currentContext();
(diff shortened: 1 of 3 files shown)


apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
```diff
@@ -79,15 +79,19 @@ public static class AdviceClass {
             @Advice.OnMethodEnter(suppress = Throwable.class, inline = false)
             public static Object enter(@Advice.This HttpURLConnection thiz,
                                        @Advice.FieldValue("connected") boolean connected,
+                                       @Advice.FieldValue("responseCode") int responseCode,
                                        @Advice.Origin String signature) {
 
+                //With HEAD requests the connected stays false
```
SylvainJuge, line 85: [minor] that's really an implementation detail here, could we also check the HTTP verb to be sure ?

SylvainJuge (approved):
Good catch !

apm-agent-plugins/apm-urlconnection-plugin/src/main/java/co/elastic/apm/agent/urlconnection/HttpUrlConnectionInstrumentation.java
JonasKunz: I'd argue that even for non-HEAD requests this approach is safer: If a responseCode is available, this implies that the connection has already been made. Therefore I think an additional check would not be good here.

SylvainJuge: Good for me, let's keep it as-is then !

SylvainJuge:
@Mergifyio refresh

Issue #3356, opened 2023-10-11
[META 834] Implement context_propagation_only setting


JonasKunz:
Implement the setting described in the [spec](https://github.com/elastic/apm/blob/a3510c0484d1eeefc26ae424fdb2ebd151e3bdcd/specs/agents/transport.md#disable_send-configuration).

PR #3357, opened 2023-10-12
Fix off-by-one error in tests which effectively disabled recycling validation

JonasKunz:
## What does this PR do?

While working on #3356 I discovered that our tests did not catch leaked objects in all cases.
Turns out that this was caused by an off-by-one error in `TestObjectPoolFactory`.

After fixing the error, it uncovered a few missing recyclings in our tests, but the production code is fine.
For that reason I won't add a changelog, as it is tests-only.

## Checklist

- [x] This is something else

### DIFF
--- a/apm-agent-core/src/test/java/co/elastic/apm/agent/objectpool/TestObjectPoolFactory.java
+++ b/apm-agent-core/src/test/java/co/elastic/apm/agent/objectpool/TestObjectPoolFactory.java
@@ -79,7 +79,7 @@ public class TestObjectPoolFactory extends ObjectPoolFactory {
                     // silently ignored
                 }
             }
-        } while (retry-- > 0 && hasSomethingLeft);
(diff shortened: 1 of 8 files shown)


apm-agent-core/src/test/java/co/elastic/apm/agent/objectpool/TestObjectPoolFactory.java
```diff
@@ -79,7 +79,7 @@ public void checkAllPooledObjectsHaveBeenRecycled() {
                     // silently ignored
                 }
             }
-        } while (retry-- > 0 && hasSomethingLeft);
+        } while (--retry > 0 && hasSomethingLeft);
```
JonasKunz, line 82: The postfix decrement caused `retry` to become `-1` at the end of the loop, effectively causing the `if` condition below to always evaluate to false.

jackshirazi (approved):

PR #3410, opened 2023-11-06
jdk update reminder

SylvainJuge:
Whenever there is a new JDK release GA, we need to update our JDK integration tests.
The main problem is that it's quite easy to forget about this over time as they are expected only twice a year.

## What does this PR do?
- provides a simple release schedule that will fail in CI whenever a new JDK GA is available:
```
This test fails to remind you that JDK 21 is about or already released,

please update the following:

- .github/workflows/main.yml in 'jdk-compatibility-tests' : replace early-access with released GA version
- in this test: remove released version in the release schedule
- in this test: update release schedule if needed
```
- disabled by default, enabled only in CI through system property to prevent noise outside of CI and in unrelated pull-requests.

While it does not ensures that we properly update the tests in CI, which could require parsing the github workflow YAML, at least it allows us to avoid forgetting about doing so, which should be fine given it only happens twice a year.

## Checklist

- [x] This is something else
  - [x] manually tested with settings for JDK 21: `Arguments.of(21, LocalDate.parse("2023-09-01"))`, result when it fails: https://github.com/elastic/apm-agent-java/actions/runs/6773390429

### DIFF
--- a/apm-agent-core/src/test/java/co/elastic/apm/agent/util/JdkVersionTest.java
+++ b/apm-agent-core/src/test/java/co/elastic/apm/agent/util/JdkVersionTest.java
@@ -19,6 +19,7 @@
 package co.elastic.apm.agent.util;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfSystemProperty;
 import org.junit.jupiter.params.provider.Arguments;
 
 import java.time.LocalDate;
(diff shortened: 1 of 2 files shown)


jackshirazi:
The releases happen after the middle of the month, and we don't need a reminder immediately, so I'd make it the 1st of Oct and Apr

JonasKunz (dismissed):

apm-agent-core/src/test/java/co/elastic/apm/agent/util/JdkVersionTest.java
```diff
@@ -0,0 +1,92 @@
...
+        if (now.isBefore(gaDate)) {
+            return;
+        }
+
+        assertThat(false)
+            .describedAs(
+                "This test fails to remind you that JDK %d is about or already released,\n\n" +
+                    "please update the following:\n" +
+                    "\n" +
+                    "- .github/workflows/main.yml in 'jdk-compatibility-tests' : replace early-access with released GA version\n" +
```
jackshirazi, line 67:
```suggestion
                    "- .github/workflows/main.yml in 'jdk-compatibility-tests' : replace early-access with released GA version and add the new early-access version\n" +
```

PR #3422, opened 2023-11-09
feat: new release process using PRs

amannocci:
## What is the change being made?

* Use PRs to release a new version of the apm-agent-java.
* The release process has been split into two workflows.
* The first one will create a PR with pre/post-release changes labeled with `release`.
* It will push artifacts in staging/snapshot maven repositories.
* It will also block any further change happening in the `main` branch.
* The new PR should be reviewed by a code owner.
* If everything looks good then the release process will continue after the PR has been approved.
* The second workflow will perform the remaining tasks.
* It will also create a new PR to update the `1.x` major branch.
* This PR could be merged after review.

## Why is the change being made?

* The new branch protection doesn't allow direct push in the `main` branch even for the automation.

### DIFF
--- a/.github/workflows/release-from-pr.yml
+++ b/.github/workflows/release-from-pr.yml
@@ -1,0 +1,303 @@
+---
+
+# Releases the agent
+name: Release On PR
+
+on:
+  pull_request_review:
+    types: [ submitted ]
+    branches:
+      - main
+    paths:
+      - 'pom.xml'
+
+env:
+  JAVA_VERSION: 17
+  JAVA_DIST: temurin
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}
+
+jobs:
+  check:
+    name: "Check release from PR"
+    runs-on: ubuntu-latest
+    if: github.event.review.state == 'approved'
+    outputs:
+      is_release: ${{ steps.check.outputs.is_release }}
+    steps:
+      - id: check
+        name: Check for specific release label
+        run: |
+          LABEL="release"
+          echo "Checking for label: $LABEL"
+          LABEL_EXISTS=$(echo '${{ toJSON(github.event.pull_request.labels.*.name) }}' | jq '.[] | select(. == "'$LABEL'")')
+          if [ -z "$LABEL_EXISTS" ]; then
+            echo "Label $LABEL does not exist on the merged PR."
+            echo "::set-output name=is_release::false"
+          else
+            echo "::set-output name=is_release::true"
+          fi
+
+  merge_ff:
+    name: "Unlock & Merge FF"
+    runs-on: ubuntu-latest
+    needs:
+      - check
+    if: needs.check.outputs.is_release == 'true'
+    steps:
+      - uses: elastic/apm-pipeline-library/.github/actions/github-token@current
+        with:
+          url: ${{ secrets.VAULT_ADDR }}
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+      - uses: elastic/apm-pipeline-library/.github/actions/setup-git@current
+        with:
+          username: ${{ env.GIT_USER }}
+          email: ${{ env.GIT_EMAIL }}
+          token: ${{ env.GITHUB_TOKEN }}
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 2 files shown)


amannocci:
Given the significant alterations to the release process introduced by this PR, I'm seeking feedback on its implementation.
What do you though @v1v @reakaleek @SylvainJuge?

SylvainJuge:
Is there any particular reason to use a separate PR for the version bump ? This is part of the usual release process, and merging non-snapshot version directly into main will make the CI attempt to build non-snapshot artifacts when merged which is not necessary.

Ideally if everything is in the same PR, it means we only have to review the changes once, however in this case the commit that will be used to generate the release would be the HEAD~1 from the release branch (and not the state that we see in the PR review).

Alternatively, we could have one PR with two reviews needed: one for the commit that generates the release binaries, then another one for the next snapshot version (which should only contain version changes in the `pom.xml` files, thus tedious but not complicated to review).

amannocci:
> Is there any particular reason to use a separate PR for the version bump ? This is part of the usual release process, and merging non-snapshot version directly into main will make the CI attempt to build non-snapshot artifacts when merged which is not necessary.

Are talking about the PR to push changes in the major 1.x branch?

.github/workflows/release-from-pr.yml
```diff
@@ -0,0 +1,303 @@
...
+          ref: ${{ env.TAG_NAME }}
+          fetch-depth: 0 # Load entire history as it is required for the push-script
+      - uses: elastic/apm-pipeline-library/.github/actions/docker-login@current
+        with:
+          registry: docker.elastic.co
+          secret: secret/apm-team/ci/docker-registry/prod
+          url: ${{ secrets.VAULT_ADDR }}
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+      - name: "Build docker image"
```
v1v, line 146: I'd envision the `release-from-pr.yml` to consume the artifacts generated in the `release.yml`. This can help with ensuring that no artifacts need to be generated in two different places the ones in `release.yml` and the ones in `release-from-pr.yml`. But making the `release-from-pr.yml` to be the orchestrator.

v1v (commented):
I cannot find the changes related to using the [nexus staging](https://help.sonatype.com/repomanager2/staging-releases) rather than the production one. IIRC, that's configured in the settings.xml or pom.xml

.github/workflows/release.yml
```diff
@@ -12,28 +12,6 @@ on:
       version:
         description: 'The version to release (e.g. 1.2.3). This workflow will automatically perform the required version bumps'
         required: true
-      update_changelog:
```
v1v, line 15: What's the proposal for this?

amannocci: That's a good point. Let me try to reintroduce some of them where possible or relevant.

.github/workflows/release-from-pr.yml
amannocci: AFAIU, the docker build script will consume artifacts from the Sonatype repository since they aren't present locally.

v1v: I'm not talking about the `docker build image` but a more generic aspect, I'd say the `release.yml` workflow shall be the one creating all the required artifacts, and the `release-from-pr.yml` workflow being the one promoting them - while `promotion` means interacting with Nexus, Docker, AWS or similar artifact managers.

v1v: On side note regarding the `docker build script`, will it work when using the staging nexus repository? As I mentioned https://github.com/elastic/apm-agent-java/pull/3422#pullrequestreview-1727110547, this proposal will need to use the staging nexus repository, so releases won't be public available, only after the PR with the bumps is merged (that's the task in `release-from-pr.yml`). 

The promotion in Nexus from staging to production could be done in a follow up. The process for the Release Manager, aka RM, could be something like:

1) RM runs the `release.yml` workflow with the relevant input parameters
2) RM waits for some notification regarding the PR with the bumps to be approved.
  a) Once approved the `release-from-pr.yml` workflow will do the remaining steps.
3) RM promotes from stating to production in Nexus (manually if no automated).

v1v:
run docs-build

amannocci:
Superseded by https://github.com/elastic/apm-agent-java/pull/3567

PR #3436, opened 2023-11-22
extend InstrumentationTest timeout to reduce flakeyness

jackshirazi:
## What does this PR do?
lots of tests are hitting a timeout on InstrumentationTest but it's passing fine when run manually.

In the end I commented out the test. It's odd that it's started happening but there is no guarantee that the class will be unloaded in any time period, I think I'd have to set metaspace to some small value and it's easier to just stop doing the test

### DIFF
--- a/apm-agent-core/src/test/java/co/elastic/apm/agent/bci/InstrumentationTest.java
+++ b/apm-agent-core/src/test/java/co/elastic/apm/agent/bci/InstrumentationTest.java
@@ -447,6 +447,8 @@ class InstrumentationTest {
 
         System.gc();
         System.gc();
+        Runtime.getRuntime().runFinalization();
+        System.gc();
         await().untilAsserted(() -> assertThat(applicationCLRef.get()).isNull());
     }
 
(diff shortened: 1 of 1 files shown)


jackshirazi:

apm-agent-core/src/test/java/co/elastic/apm/agent/bci/InstrumentationTest.java
```diff
@@ -421,7 +421,8 @@ void testPluginClassLoaderGCdAfterUndoingInstrumentation() {
         await().untilAsserted(() -> assertThat(pluginClassLoader.get()).isNull());
     }
 
-    @Test
+    //this is failing from unspecified - likely metaspace specific GC not happening - in CI
+//    @Test
```
SylvainJuge, line 425: [minor] maybe better to use `@Ignore` here, and keep the comment so we have a clue about the reason it's disabled.

apm-agent-core/src/test/java/co/elastic/apm/agent/bci/InstrumentationTest.java
```diff
@@ -441,7 +442,10 @@ void testNoClassLoaderLeakWhenInstrumentedApplicationIsUndeployed() throws Excep
         applicationCL = null;
         instrumentedClass = null;
 
-        System.gc();
+        long start = System.currentTimeMillis();
+        while(System.currentTimeMillis()-start < 10_000) {
+
+        }
```
SylvainJuge, line 448: [minor] let's remove this if we disable the test for now.

SylvainJuge (approved):

PR #3453, opened 2023-12-04
Disable SpringAmqpBatchIT due to flakyness

JonasKunz:
## What does this PR do?

Unfortunately is looks like SpringAmqpBatchIT still causes flakyness on the CI, see [this run](https://github.com/elastic/apm-agent-java/actions/runs/7063241319/job/19228982570?pr=3450) for exmaple.
I assume it is due to the RabbitMQ containers not being properly shutdown, because it can be reproduced by running the test in repeated mode locally.


JonasKunz:

PR #3462, opened 2023-12-12
[wip] work-around for JVM crash with exceptions

SylvainJuge:
## Checklist

- [x] This is a bugfix

### DIFF
--- a/apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/EagerThrowable.java
+++ b/apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/EagerThrowable.java
@@ -20,20 +20,24 @@ package co.elastic.apm.agent.tracer;
 
 public class EagerThrowable extends Throwable {
 
-    private final Class<? extends Throwable> originalType;
+    private final Class<? extends Throwable> originalClass;
 
     public EagerThrowable(Throwable t) {
         super(t.getMessage(), t.getCause(), true, false);
         setStackTrace(t.getStackTrace());
-        this.originalType = t.getClass();
+        this.originalClass = t.getClass();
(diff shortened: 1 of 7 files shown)


apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/EagerThrowable.java
```diff
@@ -0,0 +1,43 @@
...
+ * under the License.
+ */
+package co.elastic.apm.agent.tracer;
+
+public class EagerThrowable extends Throwable {
+
+    private final Class<? extends Throwable> originalClass;
+
+    public EagerThrowable(Throwable t) {
+        super(t.getMessage(), t.getCause(), true, false);
```
JonasKunz, line 26: The cause should also be wrapped:
```suggestion
        super(t.getMessage(), t.getCause() == null ? null : new EagerThrowable(t.getCause()), true, false);
```

apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/EagerThrowable.java
```diff
@@ -0,0 +1,43 @@
...
+        setStackTrace(t.getStackTrace());
+        this.originalClass = t.getClass();
+        Throwable[] suppressed = t.getSuppressed();
+        for (int i = 0; i < suppressed.length; i++) {
+            addSuppressed(suppressed[i]);
```
JonasKunz, line 31: same here, I think we should wrap those too?
```suggestion
            addSuppressed(new EagerThrowable(suppressed[i]));
```

v1v:
run docs-build

PR #3474, opened 2023-12-21
fix span stacktrace capture

SylvainJuge:
## What does this PR do?

Span stack trace is captured when the span ends, currently when the span is reported to the tracer. This implementation worked well until the introduction of compressed spans because the call to `ElasticApmTracer.report(Span)` was synchronous.

However, with the addition of compressed spans now the calls to `ElasticApmTracer.report(...)` can happen while the compression algorithm is executed, thus after the span has already been ended. As a consequence, the captured stack trace needs to be captured when the span ends and not when it's reported.

## Checklist

- [x] This is a bugfix

### DIFF
--- a/apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ElasticApmTracerTest.java
+++ b/apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ElasticApmTracerTest.java
@@ -159,11 +159,20 @@ class ElasticApmTracerTest {
             Span span = tracerImpl.getActive().createSpan();
             try (Scope spanScope = span.activateInScope()) {
                 Thread.sleep(10);
-                span.end();
+                stackTraceEndSpan(span);
             }
             transaction.end();
         }
-        assertThat(reporter.getFirstSpan().getStacktrace()).isNotNull();
+        Throwable stackTrace = reporter.getFirstSpan().getStacktrace();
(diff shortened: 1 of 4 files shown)


apm-agent-core/src/test/java/co/elastic/apm/agent/impl/ElasticApmTracerTest.java
```diff
@@ -159,11 +160,20 @@ void testEnableStacktraces() throws InterruptedException {
...
             }
             transaction.end();
         }
-        assertThat(reporter.getFirstSpan().getStacktrace()).isNotNull();
+        Throwable stackTrace = reporter.getFirstSpan().getStacktrace();
+        assertThat(stackTrace).isNotNull();
+        assertThat(Arrays.stream(stackTrace.getStackTrace()).filter(stackTraceElement ->
+            stackTraceElement.getMethodName().equals("stackTraceEndSpan")
+                && stackTraceElement.getClassName().equals(ElasticApmTracerTest.class.getName()))).hasSize(1);
+    }
```
SylvainJuge, line 172: [for reviewer] this test does not actually check that the bugfix is effective, but just improves a bit the heuristic because until now there was none.

JonasKunz (dismissed):

JonasKunz (approved):

PR #3492, opened 2024-01-15
Use LRU cache for statement parser

JonasKunz:
## What does this PR do?

Replaces the caching for our Statement parsing with a simple LRU cache.
Previously we, used a bounded Map and made some exclusions to caching:
 * Non-prepared statements were not cached
 * Too long or too short statements were not cached
 
This could lead to edge cases where the caching is not effective, which we encountered with a user.
This PR switches to the same caching approach also used by the OTel-agent: Simply cache everything, but use a bounded LRU cache.

## Checklist
- [x] This is something else


v1v:
run docs-build

PR #3498, opened 2024-01-18
Added a note on centralized agent config support for AWS Lambda

someshwaranM:
## What does this PR do?

The centralized Agent configuration does not work well with the AWS Lambda because it is not supported. The problem is you will be able to add the Lambda services in the central config without an understanding of its support. So, this PR will keep an alert when creating Lambda services.


v1v:
run docs-build

SylvainJuge (approved):

PR #3513, opened 2024-02-02
Fix agent downloader with new signing key for 1.46.0

SylvainJuge:
## What does this PR do?

The signing key for deployed artifacts in maven central has changed, so we need to update the Key ID and public key used for validation.

Apart from breaking unit tests, this only impacts the agent downloader that attemps to validate the downloaded agent with an embedded copy of the public key and key ID.
So the 1.46.0 agent won't be able to download itself.

This fix adds support for multiple keys and embeds the new key.

## Checklist

- [x] This is something else
  - [x] tests added/modified to verify it works as expected
  - [x] update PR with new public key


apm-agent-attach-cli/src/main/java/co/elastic/apm/attach/AgentDownloader.java
```diff
@@ -250,20 +284,31 @@ static TreeSet<Version> parseMavenMetadataXml(InputStream htmlInputStream) throw
...
+            }
+            return toByteArray(inputStream, size);
+        } catch (IOException e) {
+            throw new IllegalStateException(e);
+        }
     }
+
+    private static byte[] toByteArray(InputStream inputStream, int size) throws IOException {
+        byte[] result = new byte[size];
+        if (size == 0 || size != inputStream.read(result) || inputStream.read() >= 0) {
```
SylvainJuge, line 308: [for reviewer] we can get away without a loop here as the content has known size and is rather small (less than 1kb).

apm-agent-attach-cli/src/test/java/co/elastic/apm/attach/AgentDownloaderTest.java
```diff
@@ -36,6 +41,7 @@
 
 class AgentDownloaderTest {
 
+    // valid key stored in valid_key.asc, but not the one used to sign agent artifacts
```
SylvainJuge, line 44: [for reviewer] here we need to have a valid GPG key.

JonasKunz (approved):

PR #3527, opened 2024-02-14
cleanup servlet request attribute

SylvainJuge:
- remove duplicated method
- remove attribute after use
- add test assertion for elastic attribute leak

## What does this PR do?

Removes the servlet request attributes used to capture spring MVC exceptions after they have been used. This helps prevent undesirable side effects that could be related to the presence of extra request attributes.

## Checklist

- [x] This is a bugfix

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -31,12 +31,12 @@ Use subheadings with the "=====" level for adding notes for unreleased changes:
 
 === Unreleased
 
-[[release-notes-1.x]]
-=== Java Agent version 1.x
-
 [float]
 ===== Bug fixes
 * Cleanup extra servlet request attribute used for Spring exception handler - {pull}3527[#3527]
+
+[[release-notes-1.x]]
+=== Java Agent version 1.x
 
 [[release-notes-1.47.0]]
 ==== 1.47.0 - 2024/02/13
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -34,6 +34,10 @@ Use subheadings with the "=====" level for adding notes for unreleased changes:
 [[release-notes-1.x]]
 === Java Agent version 1.x
 
+[float]
+===== Bug fixes
+* Cleanup extra servlet request attribute used for Spring exception handler - {pull}3527[#3527]
+
 [[release-notes-1.47.0]]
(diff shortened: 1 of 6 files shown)


jackshirazi (dismissed):

apm-agent-plugins/apm-servlet-plugin/src/main/java/co/elastic/apm/agent/servlet/ServletApiAdvice.java
```diff
@@ -230,10 +238,23 @@ public static <HttpServletRequest, HttpServletResponse, ServletContext, ServletC
...
                         if (throwable instanceof Throwable) {
                             t2 = (Throwable) throwable;
-                            if (!attributeName.equals("javax.servlet.error.exception") && !attributeName.equals("jakarta.servlet.error.exception")) {
+
+                            // elastic exception can be removed as it's not needed after transaction end
+                            if (attributeName.equals(ELASTIC_EXCEPTION)) {
+                                try {
+                                    // extra safety: try to remove hard reference from the map entry to the exception
+                                    // but map implementation might throw if it does not support null values
+                                    adapter.setAttribute(httpServletRequest, attributeName, null);
```
JonasKunz, line 250: So if we know that `ConcurrentHashMap` will always throw here and we are trying to circumvent the problem exactly for `ConcurrentHashMap` I wonder if we want this cost of the extra safety:

How about we go with your initial suggest and just do the remove without trying to set it to `null`?

SylvainJuge: Yes, I think that's a better option here, the removal is expected to remove the complete map entry.

I'll revert to only remove.

CHANGELOG.asciidoc
```diff
@@ -34,6 +34,10 @@ Use subheadings with the "=====" level for adding notes for unreleased changes:
 [[release-notes-1.x]]
 === Java Agent version 1.x
 
+[float]
```
JonasKunz, line 37: This is the wrong place, it must be above `[[release-notes-1.x]]` but under `===Unreleased`

JonasKunz (changes requested):

JonasKunz (approved):

PR #3538, opened 2024-02-23
Fix labeler workflow

JonasKunz:
The `MY_GITHUB_TOKEN` env var which was removed in #3536 is actually required as an input to the project assignment step.


v1v (approved):
ups 🙏

PR #3568, opened 2024-03-21
Add exclusion list option for calling DatabaseMetaData.getUserName

jackshirazi:
## What does this PR do?
Add an option to exclude calls to DatabaseMetaData.getUserName dependent on whether the DatabaseMetaData contains a string specified in the option


SylvainJuge (approved):
LGTM, I'm not sure about the naming but as long as it's internal it should be fine to provide a workaround.

PR #3589, opened 2024-04-18
Add thread dump debug option

SylvainJuge:
## What does this PR do?

Adds an internal option that triggers a thread dump at regular interval.
minimal frequency is limited at 100ms

One of the expected usages is when the application code is not known and relies on unsupported technologies, in this case `trace_methods` has to be configured for transactions, the expected usage scenario is:
- configure the agent with `log_level=debug` and `thread_dump_interval=1000ms`
- start the application and trigger a few transactions manually
- analyze the agent logs with the thread dumps to find relevant classes and methods to instrument.
- thread dump interval needs to be adjusted relative to the usual transaction duration, for very fast transactions less than 100ms triggering more transactions will be required to get at least a few thread dumps that contain relevant application code.

## Checklist

- [x] This is something else

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfiguration.java
@@ -488,6 +488,13 @@ public class CoreConfiguration extends ConfigurationOptionProvider implements co
             "exist and use it to dump bytecode of instrumented classes.")
         .buildWithDefault("");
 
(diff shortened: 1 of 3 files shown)


jackshirazi (approved):

JonasKunz:
Unfortunately this option seems less useful than expected: The `ThreadInfo.toString()` of the returned values from `ThreadMxBean.dumpAllThreads` only prints the top 8 stack frames for each. If we want this feature to be useful, I think we'd need to print the stacktraces ourselves.

PR #3598, opened 2024-04-25
Universal profiling integration: write shared memory

JonasKunz:
## What does this PR do?

Implements the first part of the universal profiling correlation: Writing the shared memory to be read by the profiler.

This PR also already adds all the config options, but marks them as "internal" so that the feature is basically hidden.
The full feature will likely be split across three pull requests. Until the last one is merged, the feature should not be used.

Therefore I also won't add it to the changelog. This PR also should not prevent us from doing patch-releases.

## Checklist

- [x] This is an enhancement of existing features, or a new feature in existing plugins
  - [ ] ~I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)~
  - [x] I have added tests that prove my fix is effective or that my feature works
  - [x] Added an API method or config option? Document in which version this will be introduced
  - [ ] ~I have made corresponding changes to the documentation~

PR #3602, opened 2024-04-26
Universal Profiling integration: open socket for communication

JonasKunz:
## What does this PR do?

Follow up of #3598. Like that PR, this PR won't have a changelog because the feature should not be used yet and is disabled by a internal config option.

This PR doesn't add tests for the consumption of messages yet, this will be tested indirectly in the next PR when we actually add the `profiler_stacktrace_ids` attribute to spans.

## Checklist

### DIFF
--- a/apm-agent-core/src/main/java/co/elastic/apm/agent/universalprofiling/UniversalProfilingIntegration.java
+++ b/apm-agent-core/src/main/java/co/elastic/apm/agent/universalprofiling/UniversalProfilingIntegration.java
@@ -33,6 +33,10 @@ import co.elastic.otel.UniversalProfilingCorrelation;
 
 import javax.annotation.Nullable;
 import java.nio.ByteBuffer;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Random;
 
 public class UniversalProfilingIntegration {
 
@@ -42,6 +46,9 @@ public class UniversalProfilingIntegration {
 
     // Visible for testing
     volatile boolean isActive = false;
+
+    // Visible for testing
+    String socketPath = null;
 
     private ActivationListener activationListener = new ActivationListener() {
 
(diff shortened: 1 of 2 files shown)


apm-agent-core/src/main/java/co/elastic/apm/agent/universalprofiling/UniversalProfilingIntegration.java
```diff
@@ -70,21 +95,49 @@ public void start(ElasticApmTracer tracer) {
...
+
     public void stop() {
         try {
+            if (executor != null) {
+                executor.shutdown();
+                executor.awaitTermination(10, TimeUnit.SECONDS);
+                executor = null;
+            }
             if (isActive) {
+                UniversalProfilingCorrelation.stopProfilerReturnChannel();
```
jackshirazi, line 140: do you need to set isActive false at scope exit?

apm-agent-core/src/main/java/co/elastic/apm/agent/universalprofiling/UniversalProfilingIntegration.java
```diff
@@ -114,4 +167,65 @@ public void correlateAndReport(Transaction endedTransaction) {
...
+        for (int i = 0; i < 8; i++) {
+            int idx = rnd.nextInt(allowedChars.length());
+            name.append(allowedChars.charAt(idx));
+        }
+        return name.toString();
+    }
+
+    private void consumeProfilerMessages() {
+        try {
+            while (true) {
```
jackshirazi, line 201: should this be while(isActive)

jackshirazi (dismissed):

apm-agent-core/src/main/java/co/elastic/apm/agent/universalprofiling/UniversalProfilingIntegration.java
JonasKunz: IIRC, the object won't be reused after a call to `stop()`. Nonetheless, it looks more clean to me to reset it.

apm-agent-core/src/main/java/co/elastic/apm/agent/universalprofiling/UniversalProfilingIntegration.java
JonasKunz: No, the intention of the loop is to drain all available messages. The loop can be aborted by either of the following two conditions:
 * No more messages are available (the returned message is `null`)
 * There is some problem (except for a single message being non decodeable, represented by the `DecodeException`). The loop will be aborted by an exception.

jackshirazi (approved):

PR #3600, opened 2024-04-26
release: use tag version for docker build/push

v1v:
## What does this PR do?

Pass explicit version when generating the docker images rather than inspecting the environment variables.

No more Jenkins support

Wait for the release to do the docker specific images, to be able to push the latest docker image

Fixes #3604 

## Checklist

- [ ] This is an enhancement of existing features, or a new feature in existing plugins
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] I have made corresponding changes to the documentation
- [ ] This is a bugfix
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)
  - [ ] I have added tests that would fail without this fix
- [ ] This is a new plugin
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)
  - [ ] My code follows the [style guidelines of this project](https://github.com/elastic/apm-agent-java/blob/main/CONTRIBUTING.md#java-language-formatting-guidelines)
  - [ ] I have made corresponding changes to the documentation
  - [ ] I have added tests that prove my fix is effective or that my feature works
  - [ ] New and existing [**unit** tests](https://github.com/elastic/apm-agent-java/blob/main/CONTRIBUTING.md#testing) pass locally with my changes
  - [ ] I have updated [supported-technologies.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/docs/supported-technologies.asciidoc)
  - [ ] Added an API method or config option? Document in which version this will be introduced
  - [ ] Added an instrumentation plugin? Describe how you made sure that old, non-supported versions are not instrumented by accident.
- [ ] This is something else
  - [ ] I have updated [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.asciidoc)

PR #3605, opened 2024-04-29
fix docker image publication

### LINKED ISSUE #3604: Fix docker image publication version
OK
```
docker pull docker.elastic.co/observability/apm-agent-java:1.48.0
```

KO
```
docker pull docker.elastic.co/observability/apm-agent-java:1.49.0
```

Impacts https://github.com/elastic/opbeans-java/pull/259 as the opbeans update relies on it.

From the [release execution tasks](https://github.com/elastic/apm-agent-java/actions/runs/8524059306/job/23384899243) the previous tag was selected, so the image was pushed with `1.48.1` tag and not `1.49.0`.

So in practice it's likely that we have been pushing the N-1 release in docker when the release N is executed for a while.

Update: should be fixed by #3600

SylvainJuge:
## What does this PR do?

Fixes #3604 

Will require executing docker image publication outside of a release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -128,6 +128,7 @@ jobs:
       - await-maven-central-artifact
     env:
       SONATYPE_FALLBACK: 1
+      TAG_NAME: ${{ env.RELEASE_VERSION_TAG }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


JonasKunz:
Doesn't this PR have the same goal as #3600? I also think @v1v wanted to manually add the missing 1.49.0 image after that PR is merged

PR #3607, opened 2024-04-29
Universal profiling integration: Added serialization of stacktrace IDs as profiler_stack_trace_ids otel attributes

JonasKunz:
## What does this PR do?

Follow up of https://github.com/elastic/apm-agent-java/pull/3598. Like that PR, this PR won't have a changelog because the feature should not be used yet and is disabled by a internal config option.

This PR adds
 * storage of profiling stacktrace IDs on transactions with pooling
 * Serialization as otel `profiler_stack_trace_ids` attribute

## Checklist


apm-agent-core/src/main/java/co/elastic/apm/agent/report/serialize/Base64SerializationUtils.java
```diff
@@ -0,0 +1,43 @@
...
+        'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X',
+        'Y', 'Z', 'a', 'b', 'c', 'd', 'e', 'f',
+        'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n',
+        'o', 'p', 'q', 'r', 's', 't', 'u', 'v',
+        'w', 'x', 'y', 'z', '0', '1', '2', '3',
+        '4', '5', '6', '7', '8', '9', '-', '_',
+    };
+
+
+    public static void writeBytesAsBase64UrlSafe(byte[] data, JsonWriter jw) {
```
JonasKunz, line 19: The java `Base64` class used in the otel-distro is only available in Java 8+, that's why we need to add a custom implementation here. Also this implementation is allocation free, keeping it in line with the rest of the agent design.

jackshirazi (approved):

PR #3623, opened 2024-05-09
ci: build and push Docker image based on Chainguard base image

v1v:

Release two flavours of Docker images:
- the one we always do
- the one based on [Wolfi](https://www.chainguard.dev/unchained/introducing-wolfi-the-first-linux-un-distro) from [Chainguard](https://edu.chainguard.dev/chainguard/chainguard-images/overview/) 

Please note that we are going to preserve the current `Dockerfile`, so that users will still be able to build their custom images based on `Alpine`: this is needed because `docker.elastic.co/wolfi/chainguard-base` is not a public base image, so docker build would fail for unauthenticated users.

## Tests

I created a feature branch `test/docker-images-wolfi` and ran the snapshot workflow:

- https://github.com/elastic/apm-agent-java/actions/runs/9019062649/job/24781080512

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/Dockerfile.wolfi
+++ b/Dockerfile.wolfi
@@ -1,0 +1,7 @@
+FROM docker.elastic.co/wolfi/chainguard-base@sha256:9f940409f96296ef56140bcc4665c204dd499af4c32c96cc00e792558097c3f1
+RUN mkdir /usr/agent
+ARG JAR_FILE
+ARG HANDLER_FILE
+COPY ${JAR_FILE} /usr/agent/elastic-apm-agent.jar
+COPY ${HANDLER_FILE} /usr/agent/elastic-apm-handler
+RUN chmod +x /usr/agent/elastic-apm-handler
--- a/scripts/docker-release/build_docker.sh
+++ b/scripts/docker-release/build_docker.sh
@@ -18,11 +18,13 @@ readonly SCRIPT_PATH="$( cd "$(dirname "$0")" ; pwd -P )"
 readonly PROJECT_ROOT=$SCRIPT_PATH/../../
 readonly NAMESPACE="observability"
 
-if [ "$(ls -A  ${PROJECT_ROOT}elastic-apm-agent/target/*.jar)" ]
+FILE=$(ls -A ${PROJECT_ROOT}elastic-apm-agent/target/*.jar | grep -E "elastic-apm-agent-[0-9]+.[0-9]+.[0-9]+(-SNAPSHOT)?.jar" )
+
+if [ -n "${FILE}" ]
 then
   # We have build files to use
   echo "INFO: Found local build artifact. Using locally built for Docker build"
-  find -E ${PROJECT_ROOT}elastic-apm-agent/target -regex '.*/elastic-apm-agent-[0-9]+.[0-9]+.[0-9]+(-SNAPSHOT)?.jar' -exec cp {} ${PROJECT_ROOT}apm-agent-java.jar \; || echo "INFO: No locally built image found"
+  cp "${FILE}" "${PROJECT_ROOT}apm-agent-java.jar" || echo "INFO: No locally built image found"
 elif [ ! -z ${SONATYPE_FALLBACK+x} ]
 then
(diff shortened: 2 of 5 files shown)


scripts/docker-release/build_docker.sh
```diff
@@ -18,11 +18,13 @@ readonly SCRIPT_PATH="$( cd "$(dirname "$0")" ; pwd -P )"
...
 readonly NAMESPACE="observability"
 
-if [ "$(ls -A  ${PROJECT_ROOT}elastic-apm-agent/target/*.jar)" ]
+FILE=$(ls -A ${PROJECT_ROOT}elastic-apm-agent/target/*.jar | grep -E "elastic-apm-agent-[0-9]+.[0-9]+.[0-9]+(-SNAPSHOT)?.jar" )
+
+if [ -n "${FILE}" ]
 then
   # We have build files to use
   echo "INFO: Found local build artifact. Using locally built for Docker build"
-  find -E ${PROJECT_ROOT}elastic-apm-agent/target -regex '.*/elastic-apm-agent-[0-9]+.[0-9]+.[0-9]+(-SNAPSHOT)?.jar' -exec cp {} ${PROJECT_ROOT}apm-agent-java.jar \; || echo "INFO: No locally built image found"
```
v1v, line 25: I'm afraid `-E` is MacOS based, using ls with grep instead

JonasKunz:
> Please note that we are going to preserve the current Dockerfile, so that users will still be able to build their custom images based on Alpine: this is needed because docker.elastic.co/wolfi/chainguard-base is not a public base image, so docker build would fail for unauthenticated users.

If users cannot use this new image, what is the purpose of adding it?

v1v:
> If users cannot use this new image, what is the purpose of adding it?

IIUC, users can use the docker image we generate based on Wolfi, but they cannot use the base docker image `docker.elastic.co/wolfi/chainguard-base`.

reakaleek (approved):

JonasKunz:
> IIUC, users can use the docker image we generate based on Wolfi, but they cannot use the base docker image docker.elastic.co/wolfi/chainguard-base.

I see. I suppose the purpose of this PR is to enhance security so that we don't ship any outdated libs in our docker image, correct?

I'd argue then why not simple not use any base-image (`FROM scratch` in the Dockerfile)?

Our image is not intended to be used as a base image, the docs state that it should be used in a `COPY`-statement:
`COPY --from=docker.elastic.co/observability/apm-agent-java:latest /usr/agent/elastic-apm-agent.jar /elastic-apm-agent.jar`

I feel like this would be a much simpler and easier to maintain approach (if I'm not misunderstanding the goal here).
I'd need to double check with them that we really don't want to use our published images as base images anywhere.

v1v:
> I'd argue then why not simply not use any base-image at all (FROM scratch in the Dockerfile)?

For completeness in this thread, I just added you to the loop on an internal GitHub issue.

Dockerfile.wolfi
```diff
@@ -0,0 +1,7 @@
+FROM docker.elastic.co/wolfi/chainguard-base@sha256:9f940409f96296ef56140bcc4665c204dd499af4c32c96cc00e792558097c3f1
```
JonasKunz, line 1: We'll eventually switch to just publishing the wolfi images, right?
When that happens I'd suggest to use an `ARG` for the base image:

```
ARG BASE_IMAGE=alpine@sha256:c5b1261d6d3e43071626931fc004f70149baeba2c8ec672bd4f27761f8e1ad6b
FROM $BASE_IMAGE
```

This way we can have a single Dockerfile which locally builts on top of Alpine, but from CI we can pass in the wolfi image as base image.

Nothing to do for now, just something to maybe keep in mind for when we drop publishing the alpine based image.

JonasKunz (approved):
LGTM

JonasKunz:
> I'd argue then why not simply not use any base-image at all (FROM scratch in the Dockerfile)?

For reference: We found out that the K8s-attacher requires executing `cp` within the container, so using `FROM scratch` is not an option.

Dockerfile.wolfi
v1v: > We'll eventually switch to just publishing the wolfi images, right?

I think so, that's something we are still discussing internally

PR #3684, opened 2024-06-15
Uses MockWebServer instead of a docker container for servlet ITs

codefromthecrypt:
## What does this PR do?

This uses MockWebServer instead of a docker container for servlet ITs. In doing so, this simplifies the docker containers explicitly managed in our tests to one. That allows a simpler migration to the container jupiter extension in the future.

As this is test fixture only, I didn't check any checkbox as it isn't user-affecting enhancement or bug.

I chose okhttp MockWebServer..
* because we already have most classes in the classpath and a single version to manage via okhttp
* instead of `com.sun.net.httpserver.HttpServer`: as this is a dodgy type that only gets worse with new JREs
* instead of wiremock as this is doing simple dispatch, and MWS is more straightforward

Thoughts on the future enabled by this:

If we were to simplify to extensions, a better devex is possible, notably being able to run a single test, such as a method in `SoapTestApp` without running all other tests in a loop. This would be via other jupiter things such as `@Nested`, and in consideration of some limitations (in my experience, it is best to have a top-level class per docker image under test, vs a matrix, as it is easier to manage the sub-tests that way).

## Checklist
- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/integration-tests/application-server-integration-tests/pom.xml
+++ b/integration-tests/application-server-integration-tests/pom.xml
@@ -119,6 +119,12 @@
             <version>${version.okhttp}</version>
         </dependency>
         <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>mockwebserver</artifactId>
+            <version>${version.okhttp}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.testcontainers</groupId>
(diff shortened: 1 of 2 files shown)


jackshirazi (approved):

codefromthecrypt:
Thanks for the support @jackshirazi!

PR #3692, opened 2024-06-20
Capture request body for Apache HttpClient 4.x

JonasKunz:
## What does this PR do?

Adds an experimental, internal configuration option to enable capturing of HTTP client request bodies.
Adds support for this option in the apache http client v4 (both sync and async)

 * The new option `elastic.apm.capture_http_client_request_body_size` can be set to the number of bytes to be captured (default is `0`, meaning the feature is disabled).
 * The feature respects the content type filtering using the existing `capture_body_content_types` config option
 * The body bytes will be converted to a string and output as otel-attribute: `http.request.body.content`. As a result, this will currently end up as `labels.http_request_body_content` in ES. We'll have to decide what we want to do here
 * The feature is designed as a two-step process:
    * A span is marked explicitly "eligible" to have the HTTP body captured
    * An HTTP instrumentation should capture the body for *current* span if it marked as eligible, even if the instrumentation itself did not create that span: This way we can for example capture request bodies for spring `RestTemplate` via existingm, lower-level instrumentations such as the apache HTTP-client instrumentation.

We also sniff the content charset from the HTTP `Content-Type` header. This will be later used to attempt to convert the body bytes back to a string. This is left as a TODO in this PR, at the moment we simply cast the `bytes` to `chars`, which is equivalent to the HTTP default `ISO-8859-1` encoding.

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else


apm-agent-plugins/apm-apache-httpclient/apm-apache-httpclient4-plugin/src/test/java/co/elastic/apm/agent/httpclient/v4/ApacheHttpAsyncClientInstrumentationTest.java
```diff
@@ -125,5 +134,42 @@ public void testSpanWithIllegalProtocol() throws Exception {
...
+            public void cancelled() {
+                responseFuture.cancel(true);
+            }
+        });
+        responseFuture.get();
+
 
+        expectSpan("/")
+            .withRequestBodySatisfying(body -> {
+                // assertThat(body).endsWith("line  101\nline");
```
codefromthecrypt, line 172: hehe

JonasKunz: whoopsie, fixed in https://github.com/elastic/apm-agent-java/pull/3692/commits/1181ced5180594e4db715d86bdc1105486131a10

apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/metadata/BodyCapture.java
```diff
@@ -0,0 +1,47 @@
+package co.elastic.apm.agent.tracer.metadata;
+
+import javax.annotation.Nullable;
+
+public interface BodyCapture {
```
jackshirazi, line 5: Why is there no access to the captured body from the interface?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/BodyCaptureImpl.java
```diff
@@ -0,0 +1,136 @@
...
+            }
+        });
+
+    private enum CaptureState {
+        NOT_ELIGIBLE,
+        ELIGIBLE,
+        STARTED
+    }
+
+    private CaptureState state;
```
jackshirazi, line 36: Given the description of changing state in one place but capturing in another, I think the `state` access and updates need to made thread-safe?

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/BodyCaptureImpl.java
```diff
@@ -0,0 +1,136 @@
...
+
+    @Override
+    public boolean isEligibleForCapturing() {
+        return state != CaptureState.NOT_ELIGIBLE;
+    }
+
+    @Override
+    public boolean startCapture(@Nullable String charset, int numBytesToCapture) {
+        if (numBytesToCapture > WebConfiguration.MAX_BODY_CAPTURE_BYTES) {
+            throw new IllegalArgumentException("Capturing " + numBytesToCapture + " bytes is not supported");
```
jackshirazi, line 77:
```suggestion
            throw new IllegalArgumentException("Capturing " + numBytesToCapture + " bytes is not supported, maximum is " + WebConfiguration.MAX_BODY_CAPTURE_BYTES + " bytes");
```

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/BodyCaptureImpl.java
```diff
@@ -0,0 +1,136 @@
...
+        }
+        if (state == CaptureState.ELIGIBLE) {
+            if (charset != null) {
```
jackshirazi, line 80: how can `charset` be null?

apm-agent-tracer/src/main/java/co/elastic/apm/agent/tracer/metadata/BodyCapture.java
JonasKunz: This is intentional: Instrumentations only need to write it, it is not intended to be read back by them.

apm-agent-core/src/main/java/co/elastic/apm/agent/impl/context/BodyCaptureImpl.java
JonasKunz: `charset` is the parameter here, not the field (I'll rename the parameter to avoid confusion).

The charset may not always be available, e.g. if the `Content-Type` header was not present or did not contain a charset. In that case we'll later default to simply casting from `byte` to `char` (assuming the HTTP default charset).

jackshirazi (approved):

PR #3708, opened 2024-07-08
Run opentelemetry benchmarks for individual features

JonasKunz:
## What does this PR do?

Runs the opentelemetry benchmarks for our distro snapshot with just individual features enabled.
This should allow us to more easily spot regressions in those.

### DIFF
--- a/.buildkite/scripts/opentelemetry-benchmark.sh
+++ b/.buildkite/scripts/opentelemetry-benchmark.sh
@@ -61,21 +61,23 @@ ELASTIC_OTEL_DISTRO_FEATURE_ENTRIES=""
 for FEATURE in "infspan" "cloudres" "up-integ" "spanstack" ; do
   FEATURE_ARGS=""
   if [ "$FEATURE" = "infspan" ]; then
-    FEATURE_ARGS="$FEATURE_ARGS -Delastic.otel.inferred.spans.enabled=true"
+    FEATURE_ARGS="$FEATURE_ARGS, \\\"-Delastic.otel.inferred.spans.enabled=true\\\""
   fi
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/opentelemetry-benchmark.sh
```diff
@@ -57,8 +57,29 @@ ELASTIC_LATEST_ENTRY="new Agent(\\\"elastic-latest\\\",\\\"latest available rele
...
+    FEATURE_ARGS="$FEATURE_ARGS, \\\"-Delastic.otel.inferred.spans.enabled=true\\\""
+  fi
+  if [ "$FEATURE" != "cloudres" ]; then
+    FEATURE_ARGS="$FEATURE_ARGS, \\\"-Dotel.resource.providers.aws.enabled=false\\\", \\\"-Dotel.resource.providers.gcp.enabled=false\\\""
+  fi
+
+  if [ "$FEATURE" = "up-integ" ]; then
+    FEATURE_ARGS="$FEATURE_ARGS, \\\"-Delastic.otel.universal.profiling.integration.enabled=true\\\""
+  else
+    FEATURE_ARGS="$FEATURE_ARGS, \\\"-Delastic.otel.universal.profiling.integration.enabled=false\\\""
```
jackshirazi, line 73: won't this get added every time except once?

JonasKunz: Yes, that is the intention: Fully disable the feature (e.g. not even load the native library).
By default the setting is on "auto", listening for a profiler to startup.

jackshirazi (approved):

PR #3724, opened 2024-07-18
Capture request body for HttpUrlConnection

JonasKunz:
## What does this PR do?

Adds support for capturing request bodies for `HttpUrlConnection`, also cleans up some shared code and adds test cases.

Not adding changelog and docs yet because this feature is hidden behind an internal feature flag.

## Checklist


apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/SpanImpl.java
```diff
@@ -174,6 +181,16 @@ public SpanImpl withAction(@Nullable String action) {
         return this;
     }
 
+    @Override
+    public void addEndListener(SpanEndListener<? super SpanImpl> listener) {
+        endListeners.add(listener);
```
jackshirazi, line 186: We should probably add a warning here for the set size exceeding some threshold in case we generate a leak somehow. And drop adding if so

jackshirazi (dismissed):

jackshirazi (approved):

PR #3799, opened 2024-09-13
[DOCS] Clarify the frequency of config polling

lucabelluccini:
## What does this PR do?

Clarify the frequency of config polling

https://github.com/elastic/apm-agent-java/blob/9a12b98ee959abb4a596c823325d9c78d827724b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ApmServerConfigurationSource.java#L59

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -4062,6 +4062,7 @@ Example: `5ms`.
 # trace_methods_duration_threshold=0ms
 
 # When enabled, the agent will make periodic requests to the APM Server to fetch updated configuration.
+# The frequency of the periodic request is driven by the `Cache-Control` header returned from APM Server/Integration, falling back to 5 minutes if not defined.
 #
 # This setting can be changed at runtime
 # Type: Boolean
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -1269,7 +1269,8 @@ Example: `0ms`.
 [[config-central-config]]
 ==== `central_config` (added[1.8.0])
 
-When enabled, the agent will make periodic requests to the APM Server (every 5 minutes) to fetch updated configuration.
+When enabled, the agent will make periodic requests to the APM Server to fetch updated configuration.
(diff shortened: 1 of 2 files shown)


docs/configuration.asciidoc
```diff
@@ -1269,7 +1269,7 @@ Example: `0ms`.
 [[config-central-config]]
 ==== `central_config` (added[1.8.0])
 
-When enabled, the agent will make periodic requests to the APM Server to fetch updated configuration.
+When enabled, the agent will make periodic requests to the APM Server (every 5 minutes) to fetch updated configuration.
```
jackshirazi, line 1272: It's not 5 minutes. The logic is to check the "Cache-Control" header from the central config, and use the "max-age" value if set. If not set, then the 5 minute default is used. The cache control header max age from the central config looks like it is set to 30 seconds

lucabelluccini: 🤦 Right - it was on the same file... https://github.com/elastic/apm-agent-java/blob/9a12b98ee959abb4a596c823325d9c78d827724b/apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/ApmServerConfigurationSource.java#L96

lucabelluccini, line 1272:
```suggestion
When enabled, the agent will make periodic requests to the APM Server to fetch updated configuration.
The frequency of the periodic request is driven by the `Cache-Control` header returned from APM Server/Integration, falling back to 5 minutes if not defined.
```

It would be good to link to https://www.elastic.co/guide/en/observability/current/apm-configure-agent-config.html#_apm_agent_configuration_options

lucabelluccini: I've proposed an update to the statement.

JonasKunz (commented):
LGTM, if @jackshirazi agrees I'll take care of this PR.
The config file is actually auto-generated, we'll need to add the paragraph in the Java sources instead.

jackshirazi:
@JonasKunz yes please

jackshirazi (approved):

PR #3808, opened 2024-09-19
ci: use GitHub app for ephemeral tokens

v1v:
## What does this PR do?

Use the GitHub app to generate the required ephemeral tokens with the least permissive principle.

Token expires after 55 minutes since it was requested. It's asked for once per step in the BK pipeline. Hence, if one step takes longer than 60 minutes and the token is required afterwards, then it will expire.

I cannot see this is the case in this project. See below

## Tasks

- [x] Confirm CI builds don't take longer than 55 minutes.

`apm-agent-java-snapshot` never took longer than 45 minutes and average is 15 minutes

image

`apm-agent-java-release` never took longer than 45 minutes and average is 18 minutes

image

`apm-agent-java-benchmarks` took only once over 6k minutes and average is 24 minutes:

image

## Why is it important?

* Finer-grained tokens with Service Machine accounts are required to rotate the secrets manually.
* GitHub app to generate temporary tokens is the advanced approach to avoid the above
* Document what the CI requires to run in terms of access

## Checklist

- [ ] This is a bugfix
- [ ] This is a new plugin
- [ ] This is something else

### DIFF
--- a/.buildkite/hooks/prepare-common.sh
+++ b/.buildkite/hooks/prepare-common.sh
@@ -37,9 +37,11 @@ fi
 java -version
 
 echo "--- Prepare github secrets :vault:"
-VAULT_SECRET_PATH=kv/ci-shared/observability-ci/github-bot-user
-GITHUB_SECRET=$(vault kv get -field token "${VAULT_SECRET_PATH}")
-GIT_USER=$(vault kv get -field username "${VAULT_SECRET_PATH}")
-GIT_EMAIL=$(vault kv get -field email "${VAULT_SECRET_PATH}")
+# Only accessible by Elastic employees
+# The GitHub Permission Set can be found at:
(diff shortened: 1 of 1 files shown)


SylvainJuge (approved):
LGTM, in case the token expires, does the failure mode of the build process provides a relevant error message ? Or should we remember that a long build time combined with an authentication error might be related to this ?

reakaleek (approved):

v1v:
> LGTM, in case the token expires, does the failure mode of the build process provides a relevant error message ? Or should we remember that a long build time combined with an authentication error might be related to this ?

It will fail when using the token. I don't know what the build error is; it depends on each specific build system or automation. I bet it will say something about the token having expired or 404 when pushing changes to GitHub.

Then, correlating the build time versus the error might help us understand the reason. Unfortunately, I cannot think of any other way to detect those issues natively.

PR #3940, opened 2025-01-15
remove obsolete tomcat docker image from integration tests

SylvainJuge:
## What does this PR do?

Removes the tomcat 8.5.0 docker image.

When trying to pull this docker image from CLI with `docker pull tomcat:8.5.0-jre8` we get the following error:
```
8.5.0-jre8: Pulling from library/tomcat
[DEPRECATION NOTICE] Docker Image Format v1 and Docker Image manifest version 2, schema 1 support is disabled by default and will be removed in an upcoming release. Suggest the author of docker.io/library/tomcat:8.5.0-jre8 to upgrade the image to the OCI Format or Docker Image manifest v2, schema 2. More information at https://docs.docker.com/go/deprecated-image-specs/
```

This docker image is being used by the `co.elastic.apm.servlet.TomcatIT` integration tests, which fails when run locally and in CI, but does not trigger a failure in CI, which will need proper investigation.

This 9 year old docker image is not maintained anymore, thus the simplest way to move forward is to remove it.

### DIFF
--- a/integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/TomcatIT.java
+++ b/integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/TomcatIT.java
@@ -40,7 +40,6 @@ public class TomcatIT extends AbstractTomcatIT {
     @Parameterized.Parameters(name = "Tomcat {0}")
     public static Iterable<Object[]> data() {
         return Arrays.asList(new Object[][]{
-            {"8.5.0-jre8"},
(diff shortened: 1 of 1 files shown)


jackshirazi (approved):

JonasKunz:
We'll wait on merging this one, because this PR removes our way to reproduce the CI issue where the CI is green even though some tests (=the tomcat test) failed.

JonasKunz:
Merging now, because the root cause of undetected CI failures has been found and will be fixed with #3952.

PR #3950, opened 2025-01-22
Partition tests on CI

JonasKunz:
## What does this PR do?

Attempts to automatically partition our tests on CI so that we get smaller logs and lower total runtime.


JonasKunz:
Unfortunately this partitioning approach doesn't work for the application-server integration tests, because those live in a single project. We'd need to split them into multiple sub-projects to make it work correctly.

For the non-application server integration tests it works, but has a slight imbalance: The jdbc-integration-tests take much longer than the rest, therefore the split is a bit uneven.

PR #3952, opened 2025-01-23
Investigate incorrect CI failure reporting

JonasKunz:
## What does this PR do?

Fixes the problems with CI sometimes not correctly reporting test failures as actually failed.
The root cause for this problem was a old `maven-surefire-plugin` combined with parameterized tests.

E.g. our `TomcatIT` test is parameterized, the parameter is the name of the tomcat docker image to test.
#3940 removes an image which doesn't exist anymore, thus the tests should have been failing but were not.

The reason was that the used version `maven-surefire-plugin` was naming all runs of parametrized tests with the same name - even when executed with different parameters. E.g. All test runs for different tomcat images were named just `TomcatIT.testAllScenarios` in the resulting test report XML file.
Subsequently, if at least one of the runs was successful (aka at least one image works) the tests would not count as failure, but as a single test being flaky. Flaky tests do not cause maven to report the tests as failed.

The update of the `maven-surefire-plugin` fixes this, because it assigns unique names to parametrized tests based on the parameters.

### DIFF
--- a/apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/JdbcDbIT.java
+++ b/apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/JdbcDbIT.java
@@ -43,7 +43,8 @@ public class JdbcDbIT extends AbstractJdbcInstrumentationTest {
             {"jdbc:tc:postgresql:9://hostname/databasename", "postgresql", "databasename", true},
             {"jdbc:tc:postgresql:10://hostname/databasename", "postgresql", "databasename", true},
             {"jdbc:tc:mariadb:10://hostname/databasename", "mariadb", "databasename", true},
-            {"jdbc:tc:sqlserver:2017-CU12://hostname/databasename", "mssql", "master", false}, // for mssql the 'master' name comes from the runtime catalog fallback
+            // SQL Server image seems to be broken with recent kernel versions: https://github.com/microsoft/mssql-docker/issues/868
+            //{"jdbc:tc:sqlserver:2017-CU12://hostname/databasename", "mssql", "master", false}, // for mssql the 'master' name comes from the runtime catalog fallback
             {"jdbc:tc:db2:11.5.0.0a://hostname/databasename", "db2", "test", true},
             {"jdbc:tc:oracle://hostname/databasename", "oracle", "xepdb1", true},
         });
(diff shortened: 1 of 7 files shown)


integration-tests/application-server-integration-tests/src/test/java/co/elastic/apm/servlet/JakartaeeTomcatWithSecurityManagerIT.java
```diff
@@ -38,8 +37,9 @@ public JakartaeeTomcatWithSecurityManagerIT(final String tomcatVersion) {
     @Parameterized.Parameters(name = "Tomcat {0}")
     public static Iterable<Object[]> data() {
         return Arrays.asList(new Object[][]{
-            {"10.0.10-jdk11-adoptopenjdk-hotspot"},
-            {"10.1.0-jdk17-temurin"}, // Servlet 6.x
+            {"10.0.10-jdk11-adoptopenjdk-hotspot"}
+            // TODO: investigate why on 10.1.0-jdk17-temurin it returns 500 for the status page
+            //{"10.1.0-jdk17-temurin"}, // Servlet 6.x
```
JonasKunz, line 42: I've created an issue for this one:
https://github.com/elastic/apm-agent-java/issues/3953

apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/JdbcDbIT.java
```diff
@@ -43,7 +43,8 @@ public static Iterable<Object[]> data() {
             {"jdbc:tc:postgresql:9://hostname/databasename", "postgresql", "databasename", true},
             {"jdbc:tc:postgresql:10://hostname/databasename", "postgresql", "databasename", true},
             {"jdbc:tc:mariadb:10://hostname/databasename", "mariadb", "databasename", true},
-            {"jdbc:tc:sqlserver:2017-CU12://hostname/databasename", "mssql", "master", false}, // for mssql the 'master' name comes from the runtime catalog fallback
+            // SQL Server image seems to be broken with recent kernel versions: https://github.com/microsoft/mssql-docker/issues/868
+            //{"jdbc:tc:sqlserver:2017-CU12://hostname/databasename", "mssql", "master", false}, // for mssql the 'master' name comes from the runtime catalog fallback
```
SylvainJuge, line 47: I don't think there is a significant risk to disable this test case, maybe add a TODO prefix to make it clear that it might be something we are looking to fix later.

SylvainJuge (dismissed):

apm-agent-plugins/apm-jdbc-plugin/src/test/java/co/elastic/apm/agent/jdbc/JdbcDbIT.java
JonasKunz: Added in [b9e4b85](https://github.com/elastic/apm-agent-java/pull/3952/commits/b9e4b85d89b5846a8157027201dfe79617418cff)

SylvainJuge (approved):

PR #3959, opened 2025-01-28
github-action: Add AsciiDoc freeze warning

reakaleek:
## Details

⚠️ This PR was created by an automated tool. Please review the changes carefully. ⚠️ 

Add a workflow that will comment on PRs with AsciiDoc changes.

## Why

During the migration to Elastic Docs v3, the Docs team will focus exclusively on migrating content.
To maintain consistency, prevent conflicts, and ensure a smoother transition we will freeze all AsciiDoc changes.

This means you will get a warning when you create AsciiDoc changes in your PRs.

   See https://github.com/elastic/docs-builder/issues/281 for details

If there are any questions, please reach out to the @elastic/docs-engineering

### DIFF
--- a/.github/workflows/release-step-3.yml
+++ b/.github/workflows/release-step-3.yml
@@ -151,7 +151,6 @@ jobs:
         run: ./scripts/docker-release/push_docker.sh "${{ env.RELEASE_VERSION }}"
 
   publish-aws-lambda:
-    if: false
     name: "Publish AWS Lambda"
     permissions:
       contents: read
--- a/.github/workflows/release-step-3.yml
+++ b/.github/workflows/release-step-3.yml
@@ -151,6 +151,7 @@ jobs:
         run: ./scripts/docker-release/push_docker.sh "${{ env.RELEASE_VERSION }}"
 
   publish-aws-lambda:
+    if: false
     name: "Publish AWS Lambda"
     permissions:
       contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/release-step-3.yml
```diff
@@ -151,6 +151,7 @@ jobs:
         run: ./scripts/docker-release/push_docker.sh "${{ env.RELEASE_VERSION }}"
 
   publish-aws-lambda:
+    if: false
```
bmorelli25, line 154: @reakaleek is this expected?

reakaleek: No it's not. Thank you. Probably some leftover state of my fork.

```suggestion
```

JonasKunz:
Our changelog is stored in an aciidoc file, too. That one cannot be frozen, as it would prevent us from releasing IINM?

Also how long will the migration take, as it will essentially block us from updating docs, right? Or is there an alternate process?

reakaleek:
> Our changelog is stored in an aciidoc file, too. That one cannot be frozen, as it would prevent us from releasing IINM?
> 
> Also how long will the migration take, as it will essentially block us from updating docs, right? Or is there an alternate process?

Generally, the check is not a required status check. Hence, you can ignore it if necessary.

For the timeline, I'm gonna share on slack.

Where is the changelog asciidoc, maybe we can ignore it.

JonasKunz (approved):

SylvainJuge (approved):

PR #3975, opened 2025-02-13
update struts to latest version

jackshirazi:


jackshirazi:
We decided not to do this

PR #4014, opened 2025-03-24
add internal config option to capture thread info on span/transaction start

SylvainJuge:
## What does this PR do?

Adds internal `capture_thread_on_start` boolean option to capture thread id and name as label.

option is dynamic and can be changed at runtime if set through external java properties, it's not supported through central configuration backend.


apm-agent-core/src/main/java/co/elastic/apm/agent/configuration/CoreConfigurationImpl.java
```diff
@@ -883,6 +883,14 @@ public String toSafeString(List<WildcardMatcher> value) {
         .dynamic(true)
         .buildWithDefault(Arrays.asList(WildcardMatcher.valueOf("*")));
 
+    private final ConfigurationOption<Boolean> captureThreadOnStart = ConfigurationOption.booleanOption()
+        .key("capture_thread_on_start")
+        .configurationCategory(CORE_CATEGORY)
+        .description("Whether to capture tread name and ID as labels.")
```
hegerchr, line 889:
```suggestion
        .description("Whether to capture thread name and ID as labels.")
```

PR #4037, opened 2025-04-03
fix labeler + automaticaly add to GH project

SylvainJuge:
- fixes the `agent-java` label 
- automatically adds internal PRs to the project board in `In Progress` status and in current iteration with something reusable across projects.

Depends on https://github.com/elastic/oblt-actions/pull/263 to be merged

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -64,7 +64,7 @@ jobs:
       with:
         github-token: ${{ steps.get_token.outputs.token }}
         project-id: 1829
-        item-url: "https://github.com/elastic/apm-agent-java/pull/${{ github.event.pull_request.number }}"
+        item-url: "${{ github.event.pull_request.url }}"
 
     - name: set status in project
       id: set-project-status-field
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -107,5 +107,5 @@ jobs:
         project-id: 1829
         item-id: ${{ steps.add-to-project.outputs.item-id }}
         field-name: 'Iteration'
-        field-value: '@next'
+        field-value: 'it-110'
         field-type: 'iteration'
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -64,7 +64,8 @@ jobs:
         enable-versioned-regex: 0
 
     - name: Assign new internal pull requests to project
-      if: (steps.checkUserMember.outputs.isTeamMember == 'true' || steps.checkUserMember.outputs.isExcluded == 'true') && github.event_name == 'pull_request'
+      if: (steps.checkUserMember.outputs.isTeamMember == 'true' || steps.checkUserMember.outputs.isExcluded == 'true')
+      # && github.event_name == 'pull_request'
       run: |
         item_id=$(gh project item-add ${PROJECT_NUMBER_ID} --url ${PULL_REQUEST_URL} --owner elastic --format json --jq .id)
         gh project item-edit --id "${item_id}" --project-id "${PROJECT_ID}" --field-id "${STATUS_FIELD_ID}" --single-select-option-id "${STATUS_FIELD_VALUE_ID}"
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -45,6 +45,22 @@ jobs:
         github-user: ${{ github.actor }}
         github-token: ${{ steps.get_token.outputs.token }}
 
+    - name: Check team membership for user
+      uses: elastic/get-user-teams-membership@1.1.0
+      id: checkUserMember
+      with:
+        username: ${{ github.actor }}
+        team: 'apm'
+        usernamesToExclude: |
+          apmmachine
+          dependabot
+          dependabot[bot]
+        GITHUB_TOKEN: ${{ steps.get_token.outputs.token }}
+    - name: Show team membership
+      run: |
+        echo "::debug::isTeamMember: ${{ steps.checkUserMember.outputs.isTeamMember }}"
+        echo "::debug::isExcluded: ${{ steps.checkUserMember.outputs.isExcluded }}"
+
     - name: Add community and triage labels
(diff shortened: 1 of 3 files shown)


.github/workflows/labeler.yml
```diff
@@ -14,16 +15,13 @@ jobs:
...
-        script: |
-          github.rest.issues.addLabels({
-            issue_number: context.issue.number,
-            owner: context.repo.owner,
-            repo: context.repo.repo,
-            labels: ["agent-java"]
-          })
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"
+        configuration-path: .github/java-label.yml
+        enable-versioned-regex: 0
```
SylvainJuge, line 24: [for reviewer] this part failed, so switching to the same pattern we have elsewhere for example in the android agent seem to work. Also, I wonder if we could/should replace the current version as it relies on Alex repository.

.github/workflows/labeler.yml
```diff
@@ -38,41 +36,54 @@ jobs:
...
-          
-          It may take some time before we review a PR, so even if you don’t see activity for some time, it **does not** mean that we have forgotten about it. 
-          
-          Every once in a while we go through a process of prioritization, after which we are focussing on the tasks that were planned for the upcoming [milestone](https://github.com/elastic/apm-agent-java/milestones). The prioritization status is typically reflected through the PR labels. It could be pending triage, a candidate for a future milestone, or have a target milestone set to it.
-    - name: Assign new internal pull requests to project
-      uses: elastic/assign-one-project-github-action@1.2.2
-      if: contains(steps.is_elastic_member.outputs.result, 'true') && github.event.pull_request
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"
+        configuration-path: .github/community-label.yml
+        enable-versioned-regex: 0
```
SylvainJuge, line 62: [for reviewer] this part is a copy-paste from android agent repo.

.github/workflows/labeler.yml
```diff
@@ -38,41 +36,54 @@ jobs:
...
+        agent_field_id=$(echo "$agent_field_json" | jq -r '.id' )
+        agent_field_value_id=$(echo "$agent_field_json" | jq -r ".options[]|select(.name==\"${AGENT}\")|.id" )
+        gh project item-edit --id "${item_id}" --project-id "${project_id}" --field-id "${agent_field_id}" --single-select-option-id "${agent_field_value_id}"
+
+        iteration_field_json=$(gh api graphql -f query='query($project_id:ID!){ node(id:$project_id) { ... on ProjectV2 { fields(first: 20) { nodes { ... on ProjectV2IterationField { id name configuration { iterations { startDate id } } } } } } } }' --jq '.data.node.fields.nodes[] | select( .name == "Iteration" )' -F project_id=${project_id})
+        iteration_field_id=$(echo "$iteration_field_json" | jq -r '.id' )
+        iteration_field_value_id=$(echo "$iteration_field_json" | jq -r 'last(.configuration.iterations[]|select(.startDate<=(now|strftime("%Y-%m-%d"))))|.id')
+        gh project item-edit --id "${item_id}" --project-id "${project_id}" --field-id "${iteration_field_id}" --iteration-id "${iteration_field_value_id}"
```
SylvainJuge, line 83: [for reviewer] this masterpiece of inlined graphql, jq and shell allows to avoid having to hardcode cryptic IDs of the project fields and their values. It should be reusable as-is in other repositories by changing the environment variables below.

SylvainJuge:
draft PR to add those to global github actions and make it easier to reuse https://github.com/elastic/oblt-actions/pull/260

.github/workflows/labeler.yml
```diff
@@ -38,41 +36,97 @@ jobs:
...
```
SylvainJuge, line 126: @v1v for now setting the field values does not work with the "project-field-set" action, however here this part is working as expected, removing the `if` condition allows to make it run with the expected result. I'm giving up on this for today, let's continue on this next week when you have time.

.github/workflows/labeler.yml
```diff
@@ -14,16 +14,9 @@ jobs:
...
-      uses: actions/github-script@v7
-      with:
-        script: |
-          github.rest.issues.addLabels({
-            issue_number: context.issue.number,
-            owner: context.repo.owner,
-            repo: context.repo.repo,
-            labels: ["agent-java"]
-          })
+      run: gh issue edit "${NUMBER}" --add-label "agent-java" --repo "${{ github.repository }}"
```
v1v, line 19: you need the support for the `NUMBER` env variable and the `GH_TOKEN: github.secret` or the one with the GITHUB_TOKEN secret value:

IIUC, it supports GH issues and PRs, so the github context might be different, see https://github.com/elastic/apm-agent-python/blob/0473d9b5592861ed4ca5dea43ecf6b66d4fa35ff/.github/workflows/labeler.yml#L17-L19

.github/workflows/labeler.yml
```diff
@@ -38,41 +31,66 @@ jobs:
...
+      uses: elastic/get-user-teams-membership@1.1.0
+      id: checkUserMember
       with:
-        github-org: "elastic"
-        github-user: ${{ github.actor }}
-        github-token: ${{ steps.get_token.outputs.token }}
+        username: ${{ github.actor }}
+        team: 'apm'
+        usernamesToExclude: |
+          apmmachine
```
v1v, line 41: we don't use `apmmachine` anymore

.github/workflows/labeler.yml
```diff
@@ -38,41 +31,66 @@ jobs:
...
+    - name: Show team membership
+      run: |
+        echo "::debug::isTeamMember: ${{ steps.checkUserMember.outputs.isTeamMember }}"
+        echo "::debug::isExcluded: ${{ steps.checkUserMember.outputs.isExcluded }}"
 
     - name: Add community and triage labels
-      if: contains(steps.is_elastic_member.outputs.result, 'false') && github.actor != 'dependabot[bot]' && github.actor != 'elastic-renovate-prod[bot]' && github.actor != 'elastic-observability-automation[bot]'
-      uses: actions/github-script@v7
+      if: steps.checkUserMember.outputs.isTeamMember != 'true' && steps.checkUserMember.outputs.isExcluded != 'true'
+      run: gh issue edit "${NUMBER}" --add-label "community,triage" --repo "${{ github.repository }}"
```
v1v, line 53: ditto

v1v (approved):

PR #4228, opened 2025-09-26
fix reference config properties build

SylvainJuge:
The generated reference config properties fails in CI, this is probably due to moving back the reference documentation in this repository, the `{applies_to}` marker is not very relevant so I've fixed it by removing it.


theletterf:
@SylvainJuge I never touched the APM Java agent docs except for little maintenance, so I don't know if there was any pre-existing automation in place. I, too, would focus automation efforts on EDOT docs only from now on. +CC @colleenmcginnis 

Bear in mind that the docs preview in a PR only provides links to the affected (altered) docs, but you should be able to navigate anywhere else within the repo's docs.

PR #4274, opened 2025-10-23
rollback #4270

jackshirazi:
We would have missed a regression without the integration tests

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,8 +13,6 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 3
-    exclude-paths:
-      - "integration-tests/**"
     allow:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -13,8 +13,6 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 3
-    exclude-paths:
-      - "integration-tests/**"
```
LikeTheSalad: Was this causing an issue? @jackshirazi

jackshirazi: The byte buddy update would have passed if the integration tests are ignored, but it's a real problem and only the integration tests are showing that. So best we keep those tests in for upgrades

LikeTheSalad (approved):

PR #4284, opened 2025-10-27
relax instrumentation matcher

jackshirazi:
trying out things for spring web mvc


apm-agent-plugins/apm-spring-webmvc/apm-spring-webmvc-spring5/src/main/java/co/elastic/apm/agent/springwebmvc/AbstractSpringTransactionNameInstrumentation.java
```diff
@@ -75,8 +75,7 @@ public abstract class AbstractSpringTransactionNameInstrumentation extends Elast
      */
     @Override
     public ElementMatcher<? super TypeDescription> getTypeMatcher() {
-        return nameStartsWith("org.springframework.web.servlet")
-            .and(hasSuperType(named("org.springframework.web.servlet.HandlerAdapter")))
+        return hasSuperType(named("org.springframework.web.servlet.HandlerAdapter"))
```
SylvainJuge: Doing that will make the matching expensive as all potential instrumentation candidates will need to have all their super type chain resolved. We could however maybe use `getTypeMatcherPreFilter` to keep an heuristic on name while not impacting performance too much.

The total matching time is shown in agent debug logs when the agent stops, so that helps to give an idea about the efficiency of the heuristic.

PR #4288, opened 2025-10-29
disable HttpClient3InstrumentationTest and enable JDK 26-ea

### LINKED ISSUE #4256: investigate and fix test failures with JDK 26
`HttpClient3InstrumentationTest` currently fails with JDK 26, for example [here in CI](https://github.com/elastic/apm-agent-java/actions/runs/18459647946/job/52587970756?pr=4255).

This can be reproduced locally, which seems directly related to JDK 26.

jackshirazi:
closes #4256


SylvainJuge:
> @SylvainJuge it looks like at least some of the issues are resolved by bumping byte buddy (1.17.8 supports java 26). The changes left in this PR are reasonable but might not be needed at all, shall I test without them, or just merge this?

If that's not too time consuming, I would prefer we try without them and keep the required changes to minimal.

PR #4515, opened 2026-06-17
fix(webflux): fix NoSuchMethodError when using spring framework 7 with webflux

jihad-bouhaiji:
## What does this PR do?
This PR fixes an issue with the experimental support for Spring framework 7 that was added [PR](https://github.com/elastic/apm-agent-java/pull/4489). When running the agent, it will throw a NoSuchMethodError exception due to the `HttpHeaders.entrySet` method. This is due to HttpHeaders [no longer implementing the MultiValueMap contract](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpHeaders.html).\
I have changed the copying of header to only use API's that are still available across version 5, 6 and 7.
The test `co.elastic.apm.agent.springwebflux.AbstractServerInstrumentationTest#dispatchHello` covers this fix.

## Checklist

  - [ ] I have updated [CHANGELOG.next-release.md](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.next-release.md)
- [x] This is a bugfix
  - [x] I have updated [CHANGELOG.next-release.md](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.next-release.md)
- [ ] This is a new plugin
  - [ ] I have updated [CHANGELOG.next-release.md](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.next-release.md)
  - [ ] I have updated [supported-technologies.md](https://github.com/elastic/apm-agent-java/blob/main/docs/reference/supported-technologies.md)
- [ ] This is something else
  - [ ] I have updated [CHANGELOG.next-release.md](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.next-release.md)

### DIFF
--- a/apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-spring5/src/main/java/co/elastic/apm/agent/springwebflux/WebfluxHelper.java
+++ b/apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-spring5/src/main/java/co/elastic/apm/agent/springwebflux/WebfluxHelper.java
@@ -58,7 +58,9 @@ import java.lang.reflect.Type;
 import java.net.InetSocketAddress;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.BiFunction;
+import java.util.stream.Collectors;
 
 import static co.elastic.apm.agent.tracer.AbstractSpan.PRIORITY_HIGH_LEVEL_FRAMEWORK;
 import static co.elastic.apm.agent.tracer.AbstractSpan.PRIORITY_LOW_LEVEL_FRAMEWORK;
@@ -313,7 +315,11 @@ public class WebfluxHelper {
     }
 
     private static void copyHeaders(HttpHeaders source, PotentiallyMultiValuedMap destination) {
-        for (Map.Entry<String, List<String>> header : source.entrySet()) {
+        // This is the only way I could find to make a multiValueMap without having access
+        // to the api from spring framework 7 that will work across versions 5, 6 and 7
+        Map<String, List<String>> multiValueHeaderMap = source.toSingleValueMap().keySet()
+            .parallelStream().collect(Collectors.toMap(e -> e, key -> Objects.requireNonNull(source.get(key))));
+        for (Map.Entry<String, List<String>> header : multiValueHeaderMap.entrySet()) {
             for (String value : header.getValue()) {
                 destination.add(header.getKey(), value);
             }
--- a/apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-spring5/src/main/java/co/elastic/apm/agent/springwebflux/WebfluxHelper.java
+++ b/apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-spring5/src/main/java/co/elastic/apm/agent/springwebflux/WebfluxHelper.java
@@ -60,6 +60,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.function.BiFunction;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static co.elastic.apm.agent.tracer.AbstractSpan.PRIORITY_HIGH_LEVEL_FRAMEWORK;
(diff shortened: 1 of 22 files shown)


apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-spring5/src/main/java/co/elastic/apm/agent/springwebflux/WebfluxHelper.java
```diff
@@ -313,7 +316,16 @@ private static void fillResponse(Transaction<?> transaction, ServerWebExchange e
     }
 
     private static void copyHeaders(HttpHeaders source, PotentiallyMultiValuedMap destination) {
-        for (Map.Entry<String, List<String>> header : source.entrySet()) {
+        // This is the only way I could find to make a multiValueMap without having access
+        // to the api from spring framework 7 that will work across versions 5, 6 and 7
```
SylvainJuge: [nit] reword the comment to explain the "why".

```suggestion
        // This allows to to make a MultiValueMap without having access
        // to the api from spring framework 7 that will work across versions 5, 6 and 7
```

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-spring5/src/main/java/co/elastic/apm/agent/springwebflux/WebfluxHelper.java
```diff
@@ -313,7 +316,16 @@ private static void fillResponse(Transaction<?> transaction, ServerWebExchange e
...
+        Map<String, List<String>> multiValueHeaderMap = source.toSingleValueMap().keySet()
+            .parallelStream().collect(Collectors.toMap(Function.identity(), new Function<>() {
```
SylvainJuge: [minor] we don't need a parallelStream here as only a few values are expected, so it is probably better to use the non parallel one.

SylvainJuge (commented):
Can you try to add an explicit test that allows to reproduce it ? Is the issue happening with the version that is being used for compilation ? if yes, then a simple unit test should be enough. If not, then do you have a sample application that allows to reproduce the issue ?

jihad-bouhaiji:
I added a module `apm-spring-webflux-plugin-spring7` and `apm-spring-webflux-testapp-spring7` that tests the same flows that we have for Spring 6 in `apm-spring-webflux-plugin`. That will catch any breaking api mismatch across 5, 6, and 7.\

I do have a question.\
It turns out that the api for `HttpHeaders.get()` has also changed in springframework 7. So I changed it to use `getValuesAsList`. However for the test `co.elastic.apm.agent.springwebflux.AbstractServerInstrumentationTest#hello`, it expects a comma seperated string for the header "Accept" instead of a List of Strings.\
I have changed the test to reflect that the headers are copied as a list instead of one String, but i want to double check with you if that was okay, or if i should parse all multi value headers into one comma seperated string?

SylvainJuge:
> I do have a question. It turns out that the api for `HttpHeaders.get()` has also changed in springframework 7. So I changed it to use `getValuesAsList`. However for the test `co.elastic.apm.agent.springwebflux.AbstractServerInstrumentationTest#hello`, it expects a comma seperated string for the header "Accept" instead of a List of Strings. I have changed the test to reflect that the headers are copied as a list instead of one String, but i want to double check with you if that was okay, or if i should parse all multi value headers into one comma seperated string?

I think this is OK here, having distinct implementations for versions 5,6 and 7 would probably be over-complication here.

jihad-bouhaiji:
@SylvainJuge Could you have one last look at this PR to see if it is ready to merge?

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-testapp-spring7/src/main/java/co/elastic/apm/agent/springwebflux/testapp/GreetingAnnotated.java
```diff
@@ -0,0 +1,197 @@
...
+
+        return greetingHandler.childSpans(count, delayMillis, durationMillis);
+    }
+
+    @GetMapping(path = "/child-flux/sse")
+    public Flux<ServerSentEvent<String>> getChildSpansSSE(@RequestParam(value = "count", required = false, defaultValue = "3") int count,
+                                                          @RequestParam(value = "duration", required = false, defaultValue = "5") long durationMillis,
+                                                          @RequestParam(value = "delay", required = false, defaultValue = "5") long delayMillis) {
+
+        return greetingHandler.childSpans(count, durationMillis, delayMillis)
```
jackshirazi, line 164:
```suggestion
        return greetingHandler.childSpans(count, delayMillis, durationMillis)
```

jackshirazi:
Spring Boot 4/Spring Framework 7 require Java 17, but these modules will still run in the JDK 11 compatibility job. Please add the testing-jdk-11 profile with <skipTests>true</skipTests> to both new POMs, matching apm-spring-webmvc-spring7-test. Also reuse Java17OnlyTest for Spring7Test and add the Java 17 guard to the servlet and WebSocket wrapper tests.

jackshirazi:
closed in favour of https://github.com/elastic/apm-agent-java/pull/4556

PR #4556, opened 2026-08-21
Fix: Spring 7 Webflux NoSuchElementError on HttpHeaders#entrySet() 

mtomik:
## What does this PR do?

Since the original [PR](https://github.com/elastic/apm-agent-java/pull/4515 ) is stuck for some time,
I created another one with the same commits + a bit better solution for the core of this issue.

Also when I was running these newer tests I found one more issue:

`Spring7ServerFunctionalInstrumentationTest#dispatchError` test was failing due to TracedSubscriber not removing the reference of the subscription on `cancel()` from `contextMap`. so the solution might be the `CancellationAwareSubscription` that is just the wrapper calling the `discardIf()` after cancel.

Is it possible that the issue was there even with older versions of reactor (from older Spring version), but now in that Spring 7 ( reactor 3.8.6 ) it was always failing.

## Checklist

  - [ ] I have updated [CHANGELOG.next-release.md](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.next-release.md)
- [x] This is a bugfix
  - [x] I have updated [CHANGELOG.next-release.md](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.next-release.md)
- [ ] This is a new plugin
  - [ ] I have updated [CHANGELOG.next-release.md](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.next-release.md)
  - [ ] I have updated [supported-technologies.md](https://github.com/elastic/apm-agent-java/blob/main/docs/reference/supported-technologies.md)
- [ ] This is something else
  - [ ] I have updated [CHANGELOG.next-release.md](https://github.com/elastic/apm-agent-java/blob/main/CHANGELOG.next-release.md)


CHANGELOG.next-release.md
```diff
@@ -11,6 +11,7 @@ This file contains all changes which are not released yet.
 <!--FIXES-START-->
 
 - avoid caching when reading version from jar to prevent side effects - [#4543](https://github.com/elastic/apm-agent-java/pull/4543)
+- fix Spring Webflux 7 NoSuchMethodError on HttpHeaders#entrySet()
```
jackshirazi:
```suggestion
- fix Spring Webflux 7 NoSuchMethodError on HttpHeaders#entrySet() - [#4556](https://github.com/elastic/apm-agent-java/pull/4556)

```

apm-agent-core/src/test/java/co/elastic/apm/agent/testutils/Java17OnlyTest.java
```diff
@@ -19,7 +19,7 @@
 package co.elastic.apm.agent.testutils;
 
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.EnabledForJreRange;
+import org.junit.jupiter.api.condition.EnabledOnJre;
```
jackshirazi:
```suggestion
import org.junit.jupiter.api.condition.EnabledForJreRange;
```

apm-agent-core/src/test/java/co/elastic/apm/agent/testutils/Java17OnlyTest.java
```diff
@@ -32,24 +32,24 @@
...
-    private Class<?> actualTestClass;
+    private final Class<?> actualTestClass;
 
     public Java17OnlyTest(Class<?> testClazz) {
         this.actualTestClass = testClazz;
     }
 
-    @EnabledForJreRange(min = JRE.JAVA_17)
     @Test
+    @EnabledOnJre(JRE.JAVA_17)
```
jackshirazi:
```suggestion
    @EnabledForJreRange(min = JRE.JAVA_17)
```

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-spring5/src/main/java/co/elastic/apm/agent/springwebflux/WebfluxHelper.java
```diff
@@ -58,7 +58,11 @@
 import java.net.InetSocketAddress;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
```
jackshirazi:
```suggestion
```

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-spring5/src/main/java/co/elastic/apm/agent/springwebflux/WebfluxHelper.java
```diff
@@ -58,7 +58,11 @@
...
+import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.stream.Collectors;
```
jackshirazi:
```suggestion
```

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-spring5/src/test/java/co/elastic/apm/agent/springwebflux/AbstractServerInstrumentationTest.java
```diff
@@ -39,19 +39,24 @@
...
 import reactor.test.StepVerifier;
 
+import java.lang.reflect.Field;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.function.Predicate;
 
+import static io.netty.handler.codec.http.HttpHeaders.Values.APPLICATION_JSON;
```
jackshirazi:
```suggestion
```

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-spring5/src/test/java/co/elastic/apm/agent/springwebflux/AbstractServerInstrumentationTest.java
```diff
@@ -39,19 +39,24 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.doReturn;
+import static org.springframework.http.MediaType.TEXT_PLAIN;
```
jackshirazi:
```suggestion
```

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-plugin-spring7/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+    </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+
```
jackshirazi, line 95: Need to exclude Java 11 testing
```suggestion

    <profiles>
        <profile>
            <!-- Spring Boot 4 / Spring Framework 7 requires junit 6 which requires java 17 -->
            <id>testing-jdk-11</id>
            <activation>
                <property>
                    <name>test_java_version</name>
                    <value>11</value>
                </property>
            </activation>
            <properties>
                <skipTests>true</skipTests>
            </properties>
        </profile>
    </profiles>

```

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-testapp/pom.xml
```diff
@@ -119,4 +119,20 @@
...
+                    <name>test_java_version</name>
+                    <value>11</value>
+                </property>
+            </activation>
+            <properties>
+                <skipTests>true</skipTests>
+            </properties>
+        </profile>
+    </profiles>
+
```
jackshirazi: I think this legacy testapp tests remain Java-11-compatible
```suggestion
```

apm-agent-core/src/test/java/co/elastic/apm/agent/testutils/Java17OnlyTest.java
mtomik: It seemed to me that based on the name of the class, that it should be only for java17, but ok :+1:  also thanks to junit6 ( from spring 7 ) I have to also define the max due to this condition to make it different to default. 

DEFAULT_MINIMUM_JRE=17
DEFAULT_MAXIMUM_JRE=OTHER

image

mtomik:
thanks for the review :+1: 
 
the failed build was due to incompatibility of that sub cancel with java 7. I reworked that using Instrumentation -`SubscriptionCancelInstrumentation` that will call the discardIf on cancelled subscription

apm-agent-plugins/apm-reactor-plugin/src/test/java/co/elastic/apm/agent/reactor/TracedSubscriberTest.java
```diff
@@ -223,6 +223,18 @@ void contextPropagation_Flux_error() {
             .verifyErrorMatches(t -> t == error);
     }
 
+    @Test
+    void cancelledSubscription_releasedContext() {
+        transaction = startTestRootTransaction("root");
+        int initialReferenceCount = transaction.getReferenceCount();
+
+        StepVerifier.create(Flux.never())
+            .thenCancel()
```
jackshirazi, line 238: I think the failure was async, so this should be
```suggestion
            .subscribeOn(SUBSCRIBE_SCHEDULER))
            .thenCancel()
```

mtomik: yeah right.. fixed. I verified again that without that new instrumentation it will fail on that referenceCount assertion

jackshirazi (dismissed):

apm-agent-plugins/apm-reactor-plugin/src/main/java/co/elastic/apm/agent/reactor/SubscriptionCancelInstrumentation.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class SubscriptionCancelInstrumentation extends ElasticApmInstrumentation {
+
+    @Override
+    public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
+        return classLoaderCanLoadClass("reactor.core.CoreSubscriber");
+    }
+
+    @Override
+    public ElementMatcher<? super TypeDescription> getTypeMatcher() {
```
SylvainJuge, line 54: Is there a common name to all the implementations we need to instrument here ? the `hasSuperType` is very expensive without `getTypeMatcherPreFilter`, so if possible adding `getTypeMatcherPreFilter` would be relevant here.

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-testapp-spring7/src/test/resources/application.properties
```diff
@@ -0,0 +1,6 @@
+# set to DEBUG for easier test application debugging
+logging.level.root=DEBUG
```
SylvainJuge: we probably don't need to have DEBUG by default here, this will make the test execution very verbose.

mtomik: it seems like it is just copy from the existing "testapp" module. so should I remove it also from there?

SylvainJuge: yes, please do that.

apm-agent-plugins/apm-reactor-plugin/src/main/java/co/elastic/apm/agent/reactor/SubscriptionCancelInstrumentation.java
mtomik: only I can think of is something like this, but that probably will not narrow it much, right?

```java
    @Override
        return nameContains("Subscription").or(nameContains("Subscriber"));
    }
```

or maybe have multiple these instrumentations by package/library that implements the CoreSubscriber ? or have just one for the implementations from `reactor.core` and the rest will be cleaned potentially by the GC later?

jackshirazi:
run docs-build

apm-agent-plugins/apm-reactor-plugin/src/main/java/co/elastic/apm/agent/reactor/SubscriptionCancelInstrumentation.java
SylvainJuge: I think this should be fine as a first step, do you have a list of all the classes that are being currently instrumented here to verify all of them match ? If those are spring classes I would expect them to stay consistently named. If those are user-provided then it might be too narrow and we need another approach.

mtomik: the problem is, that there can come any implementation of `CoreSubscriber`. if we miss some thanks to this filter, the `TracedSubscriber` will not remove it immediately from its `contextMap` / `subscriptionMap`. 

if I understand it correctly, the memory leak caused by this was always handled by GC. in this newer spring 7 came newer reactor or other library that allowed us to spot it more easily? that was at least my thought...

I had here another solution to solve it - the wrapper around that subscription (ecde26e454f932ab06e504f62bf6e30c6622ddb0) , but I had there compile issue with the older java versions. so maybe just try to solve that to avoid using this instrumentation?

SylvainJuge: I am definitely not familiar with the implementation details of webflux here, so I'll rely on your expertise to know if it's fine or not to add such filter.

What I know though is that the matching can become time-expensive, when the agent is started with debug logging there is an ascii-art table displayed at the JVM shutdown that prints the matching time per instrumentation. With that, you should be able to evaluate if the matching is actually expensive or not relative to other instrumentations. With a bit of luck the classes that are being instrumented here do not have very deep class hierarchies (unlike servlets for example), so the matching might not be as-expensive as it could be with such matcher.

In order to evaluate the impact in practice I would thus suggest to:
- run the webflux test with this instrumentation disabled and log_level=debug, store the output as baseline
- run the webflux test with this instrumentation enabled and log_level=debug, compare the output and report if there is any significant difference.

jackshirazi:
@mtomik are you able to complete the remaining suggested feedback, I was waiting for this to do a release but if it's not going to be completed soon I'll push this to the next release

apm-agent-plugins/apm-spring-webflux/apm-spring-webflux-testapp-spring7/src/test/resources/application.properties
mtomik: done

apm-agent-plugins/apm-reactor-plugin/src/main/java/co/elastic/apm/agent/reactor/SubscriptionCancelInstrumentation.java
mtomik: run the benchmarks as you mentioned - 

**without instrumentation**
```
2026-09-03 00:17:05,567 [ForkJoinPool-2-worker-1] DEBUG co.elastic.apm.agent.bci.InstrumentationStatsLifecycleListener - Total time spent matching: 239,064,579ns
| Advice name                                        | Type ns         | Method ns       |
| ExecutorInstrumentation$ExecutorRunnableInstrumentation |      49,737,495 |       5,401,685 |
| JakartaAsyncInstrumentation$JakartaStartAsyncInstrumentation |      26,006,870 |               0 |
| JakartaFilterInstrumentation                       |      25,307,667 |               0 |
```

**instrumentation without name filter**
```
2026-09-02 23:56:42,555 [ForkJoinPool-2-worker-1] DEBUG co.elastic.apm.agent.bci.InstrumentationStatsLifecycleListener - Total time spent matching: 716,530,294ns
| Advice name                                        | Type ns         | Method ns       |
| SubscriptionCancelInstrumentation                  |     501,739,464 |      18,140,762 |
| ExecutorInstrumentation$ExecutorRunnableInstrumentation |      33,892,075 |       5,670,217 |
| JakartaAsyncInstrumentation$JakartaStartAsyncInstrumentation |      16,321,812 |               0 |
| JakartaFilterInstrumentation                       |      15,601,878 |               0 |
```

**instrumentation with name filter**
```
2026-09-02 23:58:28,541 [ForkJoinPool-2-worker-1] DEBUG co.elastic.apm.agent.bci.InstrumentationStatsLifecycleListener - Total time spent matching: 280,546,991ns
| Advice name                                        | Type ns         | Method ns       |
| ExecutorInstrumentation$ExecutorRunnableInstrumentation |      54,391,782 |       7,226,246 |
| SubscriptionCancelInstrumentation                  |      27,428,938 |      11,004,509 |
| JakartaFilterInstrumentation                       |      30,907,346 |               0 |
```

**instrumentation with name filter + small type matcher optimization** 
```
2026-09-03 00:00:57,965 [ForkJoinPool-2-worker-1] DEBUG co.elastic.apm.agent.bci.InstrumentationStatsLifecycleListener - Total time spent matching: 266,680,947ns
| Advice name                                        | Type ns         | Method ns       |
| ExecutorInstrumentation$ExecutorRunnableInstrumentation |      52,177,027 |       5,465,724 |
| SubscriptionCancelInstrumentation                  |      24,501,340 |       8,916,960 |
```
so the name matcher made huge difference.

I also solved the java7 compile issue with the "non-instrumentation" solution, but I couldn't find cleaner solution than that - you can find it in last commits. 

so if that 10% increase in time spent is not much I would pick that

mtomik:
I found there one more issue. cancel of the subscription was called before collecting all the headers could be finished and that could cause this exception:
```
java.lang.NullPointerException: null
at java.util.Objects.requireNonNull(Objects.java:220)
at org.springframework.http.HttpHeaders$CaseInsensitiveEntrySet$CaseInsensitiveEntry.getValue(HttpHeaders.java:2334)
at org.springframework.http.HttpHeaders$CaseInsensitiveEntrySet$CaseInsensitiveEntry.getValue(HttpHeaders.java:2319)
at org.springframework.http.HttpHeaders.lambda$forEach$0(HttpHeaders.java:2030)
at java.lang.Iterable.forEach(Iterable.java:75)
at org.springframework.http.HttpHeaders.forEach(HttpHeaders.java:2030)
at co.elastic.apm.agent.springwebflux.WebfluxHelper.copyHeaders(WebfluxHelper.java:320)
at co.elastic.apm.agent.springwebflux.WebfluxHelper.fillResponse(WebfluxHelper.java:310)
at co.elastic.apm.agent.springwebflux.WebfluxHelper.endTransaction(WebfluxHelper.java:182)
at co.elastic.apm.agent.springwebflux.TransactionAwareSubscriber.cancelTransaction(TransactionAwareSubscriber.java:214)
at co.elastic.apm.agent.springwebflux.TransactionAwareSubscriber.cancel(TransactionAwareSubscriber.java:91)
at reactor.core.publisher.FluxOnAssembly$OnAssemblySubscriber.cancel(FluxOnAssembly.java:658)
```

so I fixed it with the correct order + added the same error handling that we got for status code there.

 @jackshirazi that should be all from me. if the review of my last changes might block your release, no problem. lets move it to another one then.

SylvainJuge:
I just checked the matching performance in the test standard output, and it's totally acceptable:

```
   Advice name                                                                                    Type ns            Method ns         
   ExecutorInstrumentation$ExecutorRunnableInstrumentation                                             91,301,814         13,214,102   
   JavaxFilterInstrumentation                                                                          46,153,889            616,531   
   InitServiceNameInstrumentation$JavaxInitServiceNameInstrumentation                                  32,963,871          1,479,452   
   SubscriptionCancelInstrumentation                                                                   25,883,477          8,483,966   
   JavaxServletInstrumentation                                                                         30,973,489            656,024   
   JavaxAsyncInstrumentation$JavaxStartAsyncInstrumentation                                            27,429,023            660,029   
   ExecutorInstrumentation$ExecutorInvokeAnyAllInstrumentation                                         14,844,983         11,137,511   
   RunnableCallableForkJoinTaskInstrumentation                                                         16,824,272          1,212,259   
   ExecutorInstrumentation$ExecutorCallableInstrumentation                                              9,833,095          3,811,010   
   ServletContainerInitializerServiceNameInstrumentation$JavaxInitServiceNameInstrumentation           12,067,723                  0   
   ExecutorInstrumentation$ForkJoinPoolInstrumentation                                                  9,219,421          2,371,271   
   ForkJoinTaskInstrumentation                                                                          8,114,142            558,088   
   JavaxFilterChainInstrumentation                                                                      6,956,189            147,314   
   JavaxRequestStreamRecordingInstrumentation                                                           6,463,642            466,558   
   DispatcherHandlerInstrumentation                                                                     6,393,830            468,880   
   HandlerAdapterInstrumentation                                                                        4,913,156          1,504,565   
   ReactorInstrumentation                                                                               5,121,863          1,232,776   
   JakartaServletVersionInstrumentation$JakartaInit                                                     6,340,990                  0   
   JavaxServletVersionInstrumentation$JavaxInit                                                         4,625,790            579,637   
   JavaxAsyncInstrumentation$JavaxAsyncContextInstrumentation                                           4,688,562            190,258   
   JakartaServletVersionInstrumentation$JakartaService                                                  4,754,681                  0   
   JavaxServletVersionInstrumentation$JavaxService                                                      4,353,296            351,680   
   JakartaServletInstrumentation                                                                        3,074,576                  0   
   WarmupInstrumentation                                                                                2,764,028                  0   
   BodyInsertersInstrumentation                                                                         1,995,783            722,432   
   JakartaFilterChainInstrumentation                                                                    2,679,905                  0   
   SystemSingleEnvVariablesInstrumentation                                                              2,247,976            323,788   
   JakartaAsyncInstrumentation$JakartaAsyncContextInstrumentation                                       2,485,043                  0   
   JakartaAsyncInstrumentation$JakartaStartAsyncInstrumentation                                         2,409,624                  0   
   InitServiceNameInstrumentation$JakartaInitServiceNameInstrumentation                                 2,360,988                  0   
   JakartaFilterInstrumentation                                                                         2,317,672                  0   
   ServletContainerInitializerServiceNameInstrumentation$JakartaInitServiceNameInstrumentation          2,218,153                  0   
   JakartaRequestStreamRecordingInstrumentation                                                         2,167,849                  0   
   SystemAllEnvVariablesInstrumentation                                                                 1,637,550            243,327   

```

This instrumentation uses 3 classes, and none of them is an outlier in the table above.
- `HandlerAdapterInstrumentation`
- `DispatcherHandlerInstrumentation`
- `BodyInsertersInstrumentation`

apm-agent-plugins/apm-reactor-plugin/src/main/java/co/elastic/apm/agent/reactor/SubscriptionCancelInstrumentation.java
SylvainJuge: Sorry I did not saw your reply here before commenting, in the current state of the PR this is fine for me performance-wise. What you observed here with SubscriptionCancelInstrumentation being at the top of table is the kind of performance regression that we try to avoid.

So, I think we can consider this aspect as solved.

apm-agent-plugins/apm-reactor-plugin/src/main/java/co/elastic/apm/agent/reactor/SubscriptionCancelInstrumentation.java
```diff
@@ -0,0 +1,82 @@
...
+public class SubscriptionCancelInstrumentation extends ElasticApmInstrumentation {
+
+    @Override
+    public ElementMatcher.Junction<ClassLoader> getClassLoaderMatcher() {
+        return classLoaderCanLoadClass("reactor.core.CoreSubscriber");
+    }
+
+    @Override
+    public ElementMatcher<? super NamedElement> getTypeMatcherPreFilter() {
+        return nameContains("Subscriber").or(nameContains("Subscription"));
```
jackshirazi, line 50: I think there are a bunch in reactor that don't use those in the names. Maybe add that as a special case?
```suggestion
        return nameStartsWith("reactor.")
            .or(nameContains("Subscriber"))
            .or(nameContains("Subscription"));
```

SylvainJuge: Asking an LLM to find known implementations of `Subscription` yields the following list, so `reactor.` is a good prefix, we might also add `io.reactivex.` and `io.smallrye.mutiny.` but I'm definitely not familiar with any of those reactor frameworks. Here it would not hurt to have a few false positives rather than lots of false positives creating lots of useless overhead, so this should be fine as a first step, and we can refine further if needed.

```
reactor.core.publisher.FluxInterval$IntervalRunnable
reactor.core.publisher.FluxSwitchMap$SwitchMapInner
reactor.core.publisher.SinkManyUnicastNoBackpressure
io.reactivex.internal.operators.flowable.FlowableAmb$AmbCoordinator
io.reactivex.internal.operators.flowable.FlowablePublishMulticast$OutputCanceller
io.smallrye.mutiny.operators.multi.MultiFlatMapOp$FlatMapInner
io.smallrye.mutiny.operators.multi.MultiCombineLatestOp$CombineLatestCoordinator
io.smallrye.mutiny.operators.multi.MultiGroupByOp$State
io.smallrye.mutiny.operators.multi.MultiZipOp$ZipCoordinator
io.smallrye.mutiny.operators.multi.processors.UnicastProcessor
io.smallrye.mutiny.operators.multi.MultiOperatorProcessor
io.smallrye.mutiny.operators.multi.builders.IntervalMulti$IntervalRunnable
org.springframework.http.server.reactive.ChannelSendOperator$WriteBarrier
org.springframework.http.server.reactive.ChannelSendOperator$WriteCompletionBarrier
```

mtomik: one drawback of this additional name matching is increased instruction time spent. So I run the numbers again:

**current filter**
```
2026-09-04 00:03:18,209 [ForkJoinPool-2-worker-1] DEBUG co.elastic.apm.agent.bci.InstrumentationStatsLifecycleListener - Total time spent matching: 257,422,186ns
| Advice name                                        | Type ns         | Method ns       |
| ExecutorInstrumentation$ExecutorRunnableInstrumentation |      46,278,675 |       5,203,226 |
| SubscriptionCancelInstrumentation                  |      17,952,170 |      11,488,096 |
| JakartaFilterInstrumentation                       |      24,123,333 |               0 |
```

**OR starts with "reactor."**
```
2026-09-03 23:59:18,358 [ForkJoinPool-2-worker-1] DEBUG co.elastic.apm.agent.bci.InstrumentationStatsLifecycleListener - Total time spent matching: 282,551,157ns
| Advice name                                        | Type ns         | Method ns       |
| ExecutorInstrumentation$ExecutorRunnableInstrumentation |      48,814,752 |       6,779,428 |
| SubscriptionCancelInstrumentation                  |      42,561,353 |       8,489,875 |
| JakartaFilterInstrumentation                       |      23,363,543 |               0 |
| JakartaAsyncInstrumentation$JakartaStartAsyncInstrumentation |      20,613,139 |               0 |
```

**OR all 3 starts with you mentioned**
```
2026-09-04 00:02:17,662 [ForkJoinPool-2-worker-1] DEBUG co.elastic.apm.agent.bci.InstrumentationStatsLifecycleListener - Total time spent matching: 289,889,601ns
| Advice name                                        | Type ns         | Method ns       |
| SubscriptionCancelInstrumentation                  |      47,053,567 |      14,306,418 |
| ExecutorInstrumentation$ExecutorRunnableInstrumentation |      48,076,345 |       5,605,395 |
| JakartaFilterInstrumentation                       |      24,638,337 |               0 |
| JakartaAsyncInstrumentation$JakartaStartAsyncInstrumentation |      22,838,079 |               0 |
```

if we want to optimize the filter even further - with something like just `nameEndsWith("Subscriber")` we cover most of them with just this overhead. 
```
2026-09-04 00:12:48,681 [ForkJoinPool-2-worker-1] DEBUG co.elastic.apm.agent.bci.InstrumentationStatsLifecycleListener - Total time spent matching: 259,733,930ns
| Advice name                                        | Type ns         | Method ns       |
| ExecutorInstrumentation$ExecutorRunnableInstrumentation |      62,550,712 |       5,650,035 |
| JakartaFilterInstrumentation                       |      24,774,004 |               0 |
| SubscriptionCancelInstrumentation                  |      14,498,788 |       9,413,708 |
| JakartaAsyncInstrumentation$JakartaStartAsyncInstrumentation |      20,837,849 |               0 |
``` 
the `nameContains("Subscription")` was there just to match it even when someone did his own implementation - bigger chance that we catch it


**nameEndsWith("Subscriber") OR all 3 your package prefixes** 
```
2026-09-04 00:09:40,953 [ForkJoinPool-2-worker-1] DEBUG co.elastic.apm.agent.bci.InstrumentationStatsLifecycleListener - Total time spent matching: 281,649,260ns
| Advice name                                        | Type ns         | Method ns       |
| SubscriptionCancelInstrumentation                  |      38,772,826 |      14,850,740 |
| ExecutorInstrumentation$ExecutorRunnableInstrumentation |      48,154,398 |       5,293,119 |
| JakartaAsyncInstrumentation$JakartaStartAsyncInstrumentation |      24,675,265 |               0 |
| JakartaFilterInstrumentation                       |      23,984,791 |               0 |
```

the thing is that if we miss some of them, we just rely on GC to clean it up later. so it just about finding that balance I guess :smile: but at the end, every one of these will be better than it was before, where we missed all of them

jackshirazi: Okay so I think it all looks good, let's go with 
```
@Override
    return nameEndsWith("Subscriber")
        .or(nameContains("Subscription"))
        .or(nameStartsWith("reactor."))
        .or(nameStartsWith("io.reactivex."))
        .or(nameStartsWith("io.smallrye.mutiny."))
        .or(nameStartsWith("org.springframework.http.server.reactive."));
}
```

mtomik: ok, done.

jackshirazi (approved):

SylvainJuge (approved):

